PR #44, opened 2015-01-22
Gigasecond exact time off

masters3d:
Found error on Daylight Savings


kytrinyx:
This is a good point, however not all locations observe daylight savings time.

masters3d:
that is right. I think the swift api may have a bug on this.

masters3d:
I was wrong about this and that is why I closed the pull request. See spread sheet. 
image

kytrinyx:
Thanks for following up.

PR #45, opened 2015-01-24
gigasecond.json: add test cases

soniakeys:
Add two test cases, one with non-zero seconds but no late-in-the-day roll over, and one just before midnight causing roll over to another day.  Discussion at Go issue https://github.com/exercism/xgo/issues/122.


kytrinyx:
Thanks!

PR #47, opened 2015-01-30
Improve sublist readme

### LINKED ISSUE #46: Definition of "sublist" in sublist exercise
I recently submitted a pull request to xpython regarding what I thought "sublist" should mean in the context of this problem, and @sjakobi redirected me here.

In my reading of the problem, a list is essentially a set that allows repetition - order doesn't matter, inherently. The way that the tests are written treats the lists more like words, where order does matter.

Given that there's this ambiguity in the concept of lists, can we change the problem specification to clarify this? I don't have an opinion on which way to change things, but it should go one of two ways:
* Clarify that order doesn't matter in a list, and change `test_spread_sublist` and its equivalents so that `multiples_of_15` is a sublist of `multiples_of_3`
* Clarify that order does matter in a list, and keep the tests as they are.

On a related note, I've only had any real experience with Python. Are lists handled in a different way in other languages such that it's less of an issue?

pminten:
Fixes #46

### DIFF
--- a/sublist.md
+++ b/sublist.md
@@ -2,3 +2,16 @@ Write a function that given two lists determines if the first list is
 contained within the second list, if the second list is contained within
 the first list, if both lists are contained within each other or if none
 of these are true.
+
+Specifically, a list A is a sublist of list B if by dropping 0 or more elements
+from the front of B and 0 or more elements from the back of B you get a list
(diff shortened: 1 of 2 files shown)


kytrinyx:
Thank you!

PR #48, opened 2015-01-31
add custom-set.json

soniakeys:
It's big.  I'm finally ready to add this as I have a test program in Go ready to use it.  As usual when I assembled this I went through existing test programs and combined and consolidated things.  The section at the end with mixed types is still kind of strange.  I don't use it in Go.  Someone who can actually use the data might fix it up someday.


kytrinyx:
Thanks! I hope to start trying these out with the Ruby track.

PR #49, opened 2015-02-02
add bob.json test data

soniakeys:
Snow day today here in Boston!


soniakeys:
wait, i'll add another...

PR #50, opened 2015-02-02
add bob.json, leap.json test data

soniakeys:
A couple of common first exercises.

### DIFF
--- a/bob.json
+++ b/bob.json
@@ -1,0 +1,143 @@
+{
+   "#": [
+      "For 'gibberish', some languages generate a random strings.",
+      "Similarly for 'prolonged silence' and 'alternate silence' some",
+      "languages generate random length strings of a whitespace character.",
+	  "",
+      "The \\u escaped 'other' whitespace (in case you don't have all of",
+	  "these codes memorized) are b: vertical tab, a0: non-breaking space,",
+	  "and 2002: en-space."
+   ],
+   "cases": [
+      {
+         "description": "stating something",
+         "input": "Tom-ay-to, tom-aaaah-to.",
+         "expected": "Whatever."
+      },
+      {
+         "description": "shouting",
+         "input": "WATCH OUT!",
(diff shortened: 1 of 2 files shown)


kytrinyx:
Thanks, it's really nice to get these common problems into a common format. There's been a lot of drift over the past 18 months.

soniakeys:
Mm hm, thank you.  I didn't find much drift in leap, but a few languages had made interesting additions to bob.  

I've been including pretty much whatever I find but I've been thinking about how to handle cases where a language might want to use a subset of the test cases.  For example one addition for bob was some unusual Unicode space characters--a non-breaking space and an en-width space.  It's fine for the moment because I believe Go is the only language at this point attempting to generate code from the json and Go handles these well.  But what if Unicode was tricky to deal with?  You might not want to test it in a beginner-level exercise.  A recourse with the current json would be to generate code for all cases where "description" didn't equal "other whitespace".  It's a little kludgy but I'm not inclined to invent any new conventions until simple solutions like this become inadequate.

An example that Go didn't handle well was the mixed data type test cases in custom-set.  When I first collected the test cases, I tried to use the word "mixed" in the description, thinking my Go program could skip any test case with the word "mixed".  That turned out not to work.  Oh, I could have made it work, but it was going to be a extra effort.  I backtracked and put these mixed-type test cases in a separate section of the json, which is easily ignored in Go.  I think it's a Go thing that the first technique was awkward.  I'm guessing that with a dynamically typed language--or maybe just a different json decoder in Go--it wouldn't have been a problem.  Anyway, I did what was easy for the moment.  Tradeoffs of doing it differently can be considered if or when someone has trouble with a code generator in another language.

kytrinyx:
> I'm not inclined to invent any new conventions until simple solutions like this become inadequate.
> [...]
> Tradeoffs of doing it differently can be considered if or when someone has trouble with a code generator in another language.

Yes, I think this is the best approach for now.

Thanks again.

PR #51, opened 2015-02-07
add hamming.json

soniakeys:
Hope the comment is okay.  I know there's a conversation somewhere about unequal length strings but I didn't find it.  My opinion is that ignoring excess length is incorrect, although I know several languages do it that way currently.


kytrinyx:
I've gone back and forth on this one, but I've concluded that I believe the correct behavior is to have an error condition.

PR #53, opened 2015-02-26
correct T nucleotide name

bjmllr:
thymidine is the nucleoside that contains the nucleotide thymine

http://rosalind.info/glossary/dna-string/ (source of the exercise)

http://en.wikipedia.org/wiki/Thymidine

http://en.wikipedia.org/wiki/Thymine


kytrinyx:
Sheesh :) I have a degree in molecular biology and I still get all of these confused! Thank you :)

PR #55, opened 2015-03-06
Clarify how ties in ordering should be resolved

pminten:
This is to support the updated Go version of the exercise where this is part of the test suite.


kytrinyx:
Thanks.

PR #56, opened 2015-03-10
Rename problem to "Singly-Linked List"

masters3d:
According to the source_url the problem is called Singly-Linked List
https://github.com/exercism/x-common/blob/master/linked-list.md

We already have a https://github.com/exercism/x-common/blob/master/simple-linked-list.yml

Using this description would be more accurate for the problem name itself used in the JSON files in each track.


kytrinyx:
Yeah, that is less confusing. Thanks!

PR #59, opened 2015-03-15
Make it clear you're not supposed to use standard libraries to implement nth-prime

Carpetsmoker:
See: https://github.com/exercism/xruby/issues/95


kytrinyx:
Nice, thanks.

PR #61, opened 2015-03-19
Added acronym problem metadata

monkbroc:
Metadata for new problem acronym.

[Problem added to Ruby track](https://github.com/exercism/xruby/pull/97)


kytrinyx:
I like this. It sounds like a great exercise to put early on in a track.

PR #62, opened 2015-03-28
Add poker exercise

pminten:
This is a somewhat more involved exercise due to poker hand evaluation having various corner cases (such as flushes being based on suits and pretty much everything else being based on ranks).


kytrinyx:
I recently saw a talk that used poker hands as an example of how a good abstraction can change the entire approach to the solution. It was very interesting.

PR #67, opened 2015-04-05
Reword the blurb of accumulate to make the arguments clear

Quintasan:
Currently the blurb suggests that the `accumulate` function should have the following signature:
```haskell
accumulate [a] -> (a -> b) -> [b]
```
However the test suite assumes that `accumulate` has the following signature:
```haskell
accumulate (a -> b) -> [a] -> [b]
```

### DIFF
--- a/accumulate.yml
+++ b/accumulate.yml
@@ -1,4 +1,4 @@
 ---
-blurb: "Implement the `accumulate` operation, which, given a collection and an operation to perform on each element of the collection, returns a new collection containing the result of applying that operation to each element of the input collection."
+blurb: "Implement the `accumulate` operation, which, given an operation and a collection applies that operation to each element of the collection and returns a new collection containing the result of applying that operation to each element of the input collection."
 source: "Conversation with James Edward Gray II"
 source_url: "https://twitter.com/jeg2"


pminten:
I don't read the old text as specifying an order of arguments. It talks about a collection and an operation on that collection, but that doesn't necessarily imply that those two are passed in that order.

And you really can't specify the order globally since some languages have the convention that the function goes first (Haskell's `map`) and others that the list goes first (Elixir's `Enum.map`).

Quintasan:
It certainly does not explicitly mention that's the order but the first time I read it I was pretty much sure that the collection goes first just to be proved wrong by the test suite. Perhaps an adnotation saying that one should read the test suite beforehand would be the best solution.

pminten:
The important thing is to remember that the README is there to give a general idea about the exercise but every track has to adapt the exercises to the way its language works.

That said, I agree there's a potential for confusion but I see it more as a problem of not communicating clearly enough what the README is.

@kytrinyx A solution could be to have some way for track maintainers to add text that's appended to a README for an exercise in the same way that SETUP.md is already appended to each README.

kytrinyx:
@pminten I think that's probably the right approach. I'll open an issue in x-api.

kytrinyx:
I've added the annotation in 62f4ae775720e472fddad8132fa360c9e83ee884. Thanks!

PR #70, opened 2015-04-12
Added POV metadata

alexkalderimis:
A proposal to add a graph re-parenting exercise.

There isn't much in the way of graph orientated exercises, so I felt this would fill a gap (on the more challenging side maybe).

An implementation has been prepared in clojure (see xclojure for the pull-request).

Possible extensions: require users to implement `path-to`


kytrinyx:
This looks interesting, thank you!

PR #73, opened 2015-04-19
Make bowling readme less object-oriented specific.

pminten:
Please note that I don't think I've ever played a game of bowling in my life, so it could be I broke the description of the rules.

### DIFF
--- a/bowling.md
+++ b/bowling.md
@@ -1,3 +1,7 @@
+Bowling is game where players roll a heavy ball to knock down pins
+arranged in a triangle. Write code to keep track of the score
+of a game of bowling.
+
 ## Scoring Bowling
 
 The game consists of 10 frames. In each frame the player has two
@@ -6,12 +10,13 @@ total number of pins knocked down, plus bonuses for strikes and spares.
 
 A spare is when the player knocks down all 10 pins in two tries.  The
 bonus for that frame is the number of pins knocked down by the next
-roll.  So in frame 3 above, the score is 10 (the total number knocked
-down) plus a bonus of 5 (the number of pins knocked down on the next
-roll.)
+roll. For example a player rolls in the first three frames respectively
+4+2, 3+7 and 5+1. For frame 2, the score is 10 (the total number knocked
+down) plus a bonus of 5 (the number of pins knocked down on the next roll.)
 
 A strike is when the player knocks down all 10 pins on his first try.
-The bonus for that frame is the value of the next two balls rolled.
+The bonus for that frame is the value of the next two balls rolled
+(the value of the next frame, not counting the bonus).
 
 In the tenth frame a player who rolls a spare or strike is allowed to
 roll the extra balls to complete the frame.  However no more than three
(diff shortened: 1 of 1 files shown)


kytrinyx:
This looks correct to me (although I believe I've only played bowling once, and only because they guilted me into participating).

Dog:
When I was younger I bowled in a league. 

A frame is composed of one or two ball throws with ten pens standing at frame initialization. There are three cases for the tabulation of a frame.

* An open frame is where a score of less than 10 is recorded for the frame. In this case the score for the individual frame is the number of pins knocked down.

* A spare is where all ten pens are knocked down after the second throw is called a spare. The total value of a spare is 10 plus the value of their next immediate throw.

* A strike is where all ten pens are knowcked down after the firt throw. The total value of a strike is 10 plus the value of their next two immediate throws.

Here is a three frame example:
1  2  3
X 5/ 90

Frame 1 is (10 + 5 + 5) = 20
Frame 2 is (5 + 5 + 9) = 19
Frame 3 is (9 + 0) = 9

This means the current running total is 48.

The tenth frame in the game is a special case. If someone throws a strike or a spare then they get a fill ball. Fill balls exist to calculate the total of the 10th frame. Scoring a strike or spare on the fill ball does not give the player more fill balls. The total value of the 10th frame is the total number of pins knocked down.

Assume a tenth frame of:
X1/

Here the total value of the 10th frame is 20.

Assume a tenth frame of:
XXX

Here the total value of the 10th frame is 30.

pminten:
Thanks. Would you mind if we use your explanation I in the readme? It's clearer than what it currently says.

Considering the not too low not too high complexity of the notation it may actually make a good theme for a second exercise: parse a bowling scoreboard and determine if it's valid.

"I can not only fetch JSON, but parse it too." <notifications@github.com> schreef op 19 april 2015 19:52:50 CEST:
>When I was younger I bowled in a league. 
>
>A frame is composed of one or two ball throws with ten pens standing at
>frame initialization. There are three cases for the tabulation of a
>frame.
>
>* An open frame is where a score of less than 10 is recorded for the
>frame. In this case the score for the individual frame is the number of
>pins knocked down.
>
>* A spare is where all ten pens are knocked down after the second throw
>is called a spare. The total value of a spare is 10 plus the value of
>their next immediate throw.
>
>* A strike is where all ten pens are knowcked down after the firt
>throw. The total value of a strike is 10 plus the value of their next
>two immediate throws.
>
>Here is a three frame example:
>1  2  3
>X 5/ 90
>
>Frame 1 is (10 + 5 + 5) = 20
>Frame 2 is (5 + 5 + 9) = 19
>Frame 3 is (9 + 0) = 9
>
>This means the current running total is 48.
>
>The tenth frame in the game is a special case. If someone throws a
>strike or a spare then they get a fill ball. Fill balls exist to
>calculate the total of the 10th frame. Scoring a strike or spare on the
>fill ball does not give the player more fill balls. The total value of
>the 10th frame is the total number of pins knocked down.
>
>Assume a tenth frame of:
>X1/
>
>Here the total value of the 10th frame is 20.
>
>Assume a tenth frame of:
>XXX
>
>Here the total value of the 10th frame is 30.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/exercism/x-common/pull/73#issuecomment-94301248

-- 
Verzonden van mijn Android telefoon met K-9 Mail.

PR #76, opened 2015-04-25
Add error handling exercise

pminten:
This is a focus exercise, so it doesn't require figuring out an algorithm, only applying some standard error handling strategies.


kytrinyx:
This is a great idea, thanks!

PR #81, opened 2015-05-14
** in markdown will bold, changed to code blocks so they show up as i…

dalexj:
…ntended


kytrinyx:
Excellent catch, thank you!

In the future, would you mind making the subject of the commit a short summary sentence (under 50 characters)?

Then you can have a blank line and add more context. E.g.:

```
Avoid accidental formatting as bold in arithmetic equation

The power symbol, '**', gets translated to bold in markdown, causing some very weird formatting.
```

dalexj:
I'll be sure to do that with commits

PR #82, opened 2015-05-15
add diagram about exercism.io workflow

studiolo:
diagram
An illustration showing new users the steps in using exercism.io. Based on a conversation with @kytrinyx at rossconf.io. The Open Document Drawing is provided as the 'source'. It can be edited with Open/LibreOffice and used to generate various export formats.

This is a first draft, looking too boring and serious, it should look more like fun :-)


kytrinyx:
I love where this is going. I'm going to show it to a couple of people who are about product design, to see if it makes sense to them in terms of explaining what exercism is about.

Thanks!

studiolo:
Yes, I'd like to hear some opinions! Maybe we should have separate diagrams for the different aspects such as learn - teach? Or be more graphic, showing people?

kytrinyx:
> Maybe we should have separate diagrams for the different aspects 

Maybe. It's so hard to say at this point.

> such as learn - teach

I don't want to use the terms learn and teach, since I learn more from providing feedback than I do from solving a problem... but yes, aside from that, I think you might be on to something.

I'd love to get the more human aspect into this as well. We'll see!

kytrinyx:
I've moved this to the discussions repository. Thank you :heart:

https://github.com/exercism/discussions/issues/30

PR #83, opened 2015-05-16
Add react exercise

pminten:
This exercise is intended to make people play with the observer pattern.


pminten:
First implementation: exercism/xgo#170.

kytrinyx:
This looks interesting, thanks!

PR #84, opened 2015-05-19
Add comma in "Hello, world!" instruction blurb.

mandel01:
The comma is necessary for the test to authenticate in Python (and also in the other examples I checked, though I did not check all of them).


kytrinyx:
Yepp, this looks better. Would you rebase and squash this so that there is only one commit please?

mandel01:
Hmmm, I thought I did it right but it shows up as a new pull request. Please let me know if that is good or if I need to do it within this pull request.

kytrinyx:
Usually you can do a `git push -fu yourbranchname` and it will update the existing PR.

kytrinyx:
I found the other one, we're all good :)

Thank you!

PR #85, opened 2015-05-19
Add comma in "Hello, world!" instruction blurb

mandel01:
The comma is necessary for the test to authenticate in Python (and also in the other examples I checked, though I did not check all of them).


kytrinyx:
Thanks!

PR #90, opened 2015-06-05
Verbal, not written, word problem

kotp:
This informs about the type of problem this should be considered, as
mentioned in exercism/xruby#116


kytrinyx:
Excellent, thank you!

PR #100, opened 2015-08-12
changed description of problem hexadecimal to remove terms specific to Ruby

ekse:
The description mentioned ruby and gems, it now applies to all languages.


kytrinyx:
Thanks!

PR #103, opened 2015-08-28
Added core set of test cases for word-count

hankturowski:
word-count.json now has a base set of test cases for the word-count exercise.


kytrinyx:
Sweet, thank you for this!

PR #104, opened 2015-08-29
Slim down robot-simulator README

macintux:
* Eliminate redundancy in instructions.
* Eliminate confusing language about constrained rooms of different sizes,
  appears to have been legacy text irrelevant to this problem.
* Eliminate step headers (also legacy?).
* Clarify that coordinates increase when moving north and east.


kytrinyx:
This is much better, thank you.

PR #108, opened 2015-09-10
Make README match tests as best as I can determine

macintux:
The bank account challenge is fairly uncommon, perhaps due to the fact that most languages don't have trivial support for concurrency.

Of the 3 examples I've found thus far (go, erlang, clojure) 2 of them reinterpreted (or misinterpreted) the bank closure paragraph as account closure; clojure ignored it entirely.

So, let's make the theory reflect practice instead of the other way around.


kytrinyx:
This seems like a very reasonable re-imagining :) Thanks

PR #109, opened 2015-09-14
 poker: metadata

MartinDelille:


kytrinyx:
We're going to need an `md` file as well. This is what becomes the meat of the README that gets delivered with the problem.

I really need to do a write-up of adding new problems!

kytrinyx:
ping @MartinDelille did this get lost in the shuffle?

PR #111, opened 2015-10-11
Corrected case in leap y2k test json

mikeyjcat:


leap.json
```diff
@@ -26,7 +26,7 @@
          "expected": true
       },
       {
-         "description": "Y2K",
+         "description": "y2k",
```
kytrinyx: Let's just let each track downcase if they want. In Go it's common to use `CamelCase`, in Ruby, `snake_case`. I don't know about other languages.

mikeyjcat: There's a bit of a catch here. If I don't change this, the Robucop check fails as the Y2K transfers through to the method name of the test, which becomes test_Y2K. 
The current logic in the test generators appears to expect the case in the JSON files to be lower case, and in Ruby, intersperses with a "_" to make valid names. This fails in the case of Y2K.
It feels like there is a design decision to be made regarding what the generators can expect, and for them to code it up according. The current implicit design decision in the Ruby generators would appear to be to expect lower case descriptions, and create test method names accordingly.
By me changing the case to lower, the current Ruby generator pattern works. Looking at the Go generator, it appears as though it does not use the description for generating a test method name, so no issue.

kytrinyx: The test generators in Ruby should be fixed to lowercase everything.

mikeyjcat:
The README.md for the json test files needs a major rework, which I'll do next.

kytrinyx:
Would you mind squashing these commits, please?

mikeyjcat:
I've squished the commits. Lets push this one in, but when I looked at all the JSON files and their corresponding generated tests, it became clear that there are a couple of different patterns in place. While I've documented the simplest in the readme.md, I think we need to get a couple of agreed approaches that cover most of the scenarios, and refactor the existing tests and generators towards that. I'm quite happy to give it a shot if you'd like?

kytrinyx:
> I think we need to get a couple of agreed approaches that cover most of the scenarios

I am not sure yet. We don't have very many, and I don't want to leap to a conclusion about patterns too early. If you have specific suggestions for improvements, then go ahead and open an issue to discuss it.

In the future, please make the subject line of commits a short (>50 characters), high-level summary of what changed.

Now that I see it all in one commit, I would probably have split this into two unrelated pull requests: One for documenting the json files in the README, one for the leap json file.

If you need to add details, it's better to add those in the body of the commit message.
For more information, please see http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

PR #112, opened 2015-10-14
Added messages to JSON test file

mikeyjcat:
I am a little uncomfortable with the consistency of the JSON files and generators. I've added these messages and kept the structure the same, but feel as though some JSON refactoring across all the tests and generators may be required?

### DIFF
--- a/gigasecond.json
+++ b/gigasecond.json
@@ -20,25 +20,31 @@
       "cases": [
          {
             "input": "2011-04-25",
-            "expected": "2043-01-01T01:46:40"
+            "expected": "2043-01-01T01:46:40",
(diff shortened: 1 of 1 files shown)


kytrinyx:
I wouldn't worry about the consistency between json files yet. We don't really have enough information at the moment to see whether we've come across all the complexities or not. Until we have more information, I'm comfortable treating each individual JSON file as its own thing.

In the generators we've abstracted out the part that actually reads JSON files and creates test cases, both in Ruby and Go.

kytrinyx:
This looks good. In the future, please include the name of the exercise that is affected in the subject line of the commit. In these repositories we look at history a lot, and having the context about which problems are affected is very helpful when we do.

PR #115, opened 2015-10-18
Add more test files

mikeyjcat:
Added two json test files

### DIFF
--- a/hello-world.json
+++ b/hello-world.json
@@ -1,0 +1,33 @@
+{
+   "#": [
+     "Language implementations vary on the issue of missing input variables.",
+     "In this set of test cases, the input variable is left out of the definition,",
+     "and the generator would need to handle the language implementation."
+   ],
+   "cases": [
+      {
+         "description": "no name",
+         "expected": "Hello, World!",
+         "msg": ["When given no name, it should greet the world!"]
+      },
+      {
+         "description": "sample name",
+         "name": "Alice",
+         "expected": "Hello, Alice!",
+         "msg": ["When given Alice, it should greet Alice!"]
+      },
+      {
+         "description": "other sample name",
+         "name": "Bob",
+         "expected": "Hello, Bob!",
+         "msg": ["When given Bob, it should greet Bob!"]
+      },
+      {
+         "description": "no strange name",
+         "name": "",
+         "expected": "Hello, !",
+         "msg": ["When given an empty string it should have a space",
+                "and punctuation, though admittedly this is strange."]
+      }
+   ]
+}
--- a/raindrops.json
+++ b/raindrops.json
@@ -1,0 +1,106 @@
+{
+   "#": [
+     "Language implementations vary when one attempts a prime factor on a number with",
+     "and underscore in it.",
+     "In this set of test cases, we have one of these, and the solution would need to",
+     "handle the language implementation."
+   ],
+   "cases": [
+      {
+         "description": "1",
+         "number" : "1", 
+         "expected": "1",
+         "msg": "No, raindrop-speak of 1 is 1."
+      },
+      {
+         "description": "3",
+         "number" : "3", 
+         "expected": "Pling",
+         "msg": "No, raindrop-speak of 3 is Pling."
+      },
+      {
+         "description": "5",
+         "number" : "5", 
+         "expected": "Plang",
+         "msg": "No, raindrop-speak of 5 is Plang."
+      },
+      {
+         "description": "7",
+         "number" : "7", 
+         "expected": "Plong",
+         "msg": "No, raindrop-speak of 7 is Plong."
+      },
+      {
+         "description": "6",
+         "number" : "6", 
+         "expected": "Pling",
+         "msg": "No, raindrop-speak of 6 is Pling."
+      },
+      {
+         "description": "9",
+         "number" : "9", 
+         "expected": "Pling",
+         "msg": "No, raindrop-speak of 9 is Pling."
+      },
+      {
+         "description": "10",
+         "number" : "10", 
+         "expected": "Plang",
+         "msg": "No, raindrop-speak of 10 is Plang."
+      },
+      {
+         "description": "14",
+         "number" : "14", 
+         "expected": "Plong",
+         "msg": "No, raindrop-speak of 14 is Plong."
+      },
+      {
+         "description": "15",
+         "number" : "15", 
+         "expected": "PlingPlang",
+         "msg": "No, raindrop-speak of 15 is PlingPlang."
+      },
+      {
+         "description": "21",
+         "number" : "21", 
+         "expected": "PlingPlong",
+         "msg": "No, raindrop-speak of 21 is PlingPlong."
+      },
+      {
+         "description": "25",
+         "number" : "25", 
+         "expected": "Plang",
+         "msg": "No, raindrop-speak of 25 is Plang."
+      },
+      {
+         "description": "35",
+         "number" : "35", 
+         "expected": "PlangPlong",
+         "msg": "No, raindrop-speak of 35 is PlangPlong."
+      },
+      {
+         "description": "49",
+         "number" : "49", 
+         "expected": "Plong",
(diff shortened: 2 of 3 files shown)


kytrinyx:
Why did you close the other PR?

In the future, please submit individual PRs for different test definitions. If everything is bundled in a single PR, then any discussion on a single JSON file will prevent all the files that are fine from being merged.

hello-world.json
```diff
@@ -0,0 +1,33 @@
...
+      {
+         "description": "other sample name",
+         "name": "Bob",
+         "expected": "Hello, Bob!",
+         "msg": ["When given Bob, it should greet Bob!"]
+      },
+      {
+         "description": "no strange name",
+         "name": "",
+         "expected": "Hello, !",
```
kytrinyx: Where is this expectation defined? I'd rather remove it from whatever language it's implemented in, since it makes no sense.

hello-world.json
```diff
@@ -0,0 +1,33 @@
...
+   "#": [
+     "Language implementations vary on the issue of missing input variables.",
+     "In this set of test cases, the input variable is left out of the definition,",
+     "and the generator would need to handle the language implementation."
+   ],
+   "cases": [
+      {
+         "description": "no name",
+         "expected": "Hello, World!",
+         "msg": ["When given no name, it should greet the world!"]
```
kytrinyx: Why is the message an array?

raindrops.json
```diff
@@ -0,0 +1,106 @@
...
+   "#": [
+     "Language implementations vary when one attempts a prime factor on a number with",
+     "and underscore in it.",
+     "In this set of test cases, we have one of these, and the solution would need to",
+     "handle the language implementation."
+   ],
+   "cases": [
+      {
+         "description": "1",
+         "number" : "1",
```
kytrinyx: This should be an integer.

raindrops.json
```diff
@@ -0,0 +1,106 @@
...
+         "msg": "No, raindrop-speak of 52 is 52."
+      },
+      {
+         "description": "105",
+         "number" : "105", 
+         "expected": "PlingPlangPlong",
+         "msg": "No, raindrop-speak of 105 is PlingPlangPlong."
+      },
+      {
+         "description": "12_121",
```
kytrinyx: Ruby uses underscores for readability, but that's just an artifact of the development of the exercise. This shouldn't have underscores.

raindrops.json
```diff
@@ -0,0 +1,106 @@
+{
+   "#": [
+     "Language implementations vary when one attempts a prime factor on a number with",
+     "and underscore in it.",
+     "In this set of test cases, we have one of these, and the solution would need to",
+     "handle the language implementation."
```
kytrinyx: We should not have underscores; that's a rubyism that is used purely for readability, and should not leak into the common test inputs/outputs.

rna-transcription.json
```diff
@@ -0,0 +1,105 @@
+{
+   "#": [
+     "Language implementations vary on the issue of missing input variables.",
+     "In this set of test cases, the input variable is left out of the definition,",
+     "and the generator would need to handle the language implementation."
```
kytrinyx: Let's also mention that -1 represents an error case, each language must implement it in a way that makes sense for that language.

kytrinyx: Actually, let's not use -1 for expected. If expected is a string, then let's use an empty string.

rna-transcription.json
```diff
@@ -0,0 +1,105 @@
...
+         "msg": "No, the dna complement of adenine is thymine."
+      },
+      {
+         "description": "dna complement",
+         "rna": "UGAACCCGACAUG",
+         "expected": "ACTTGGGCTGTAC",
+         "msg": "No, the dna complement of UGAACCCGACAUG is ACTTGGGCTGTAC."
+      },
+      {
+         "description": "dna raises argument error",
```
kytrinyx: The message should not talk about exceptions or argument errors, that's an implementation detail.

The relevant detail is _invalid input_.

rna-transcription.json
```diff
@@ -0,0 +1,105 @@
...
+         "dna": "U",
+         "expected": -1,
+         "msg": "Dna must raise an exception on invalid input!"
```
kytrinyx: The message should not specify implementation details.

raindrops.json
```diff
@@ -0,0 +1,106 @@
...
+         "description": "1",
+         "number" : "1", 
+         "expected": "1",
+         "msg": "No, raindrop-speak of 1 is 1."
+      },
+      {
+         "description": "3",
+         "number" : "3", 
+         "expected": "Pling",
+         "msg": "No, raindrop-speak of 3 is Pling."
```
kytrinyx: These messages don't really add any value, they're just an echo of the implementation. Let's ditch it.

mikeyjcat:
Will do

hello-world.json
mikeyjcat: Expectation defined in existing tests. Happy to remove it.

hello-world.json
mikeyjcat: The existing tests contain multi-line commentary. It was simpler to implement this in the comments rather than introduce complexity trying to split long comments into multiple lines intelligently. In addition, since this was the first problem, I felt it was appropriate that the longer comments be preserved as is, rather than shortened.

raindrops.json
mikeyjcat: Hmm, good point

raindrops.json
mikeyjcat: Will correct

raindrops.json
mikeyjcat: Will correct, misunderstood the intent of the existing test.

rna-transcription.json
mikeyjcat: In the cold light of day, I see that this commentary is unclear. -1 refers to invalid input and for some reason my brain was not working with the commentary regarding this. I'll correct it and resubmit. This is as per the hamming.json file.

rna-transcription.json
mikeyjcat: Will do. Again was trying to keep existing tests the same, but I've missed the generalisation of them for other languages.

raindrops.json
mikeyjcat: All messages, or just for raindrops?

hello-world.json
kytrinyx: That was the weird case with an empty name, right? I think we should remove that one.

rna-transcription.json
kytrinyx: Right, but lets use a type that is consistent, then each language's generator can implement the test to expect whatever value they want.

rna-transcription.json
kytrinyx: Yeah, the cross-language stuff is tricky. It's often only when we start trying to generalize that we see where we've added implementation-specific language.

raindrops.json
kytrinyx: Just raindrops -- or any test suites where there's a pure comparison between two values, and no additional context. For the leap year problem, the extra context really helps. Here, not so much.

hello-world.json
mikeyjcat: Yep, consider it done.

kytrinyx: cool, thanks!

rna-transcription.json
mikeyjcat: If we use an empty string, will we not run into the scenario where we could really be wanting an empty string returned? Maybe a better way to flag an exception (acknowledge language dependant reference:) ) would be to not have an expected tag, and then the generator could take the absence of a return value as the indicator of an exception?

Basically the return value is undefined, so why have it in the test json?

kytrinyx: Right, the transcription of an empty string should be an empty string.

How about `null`?

mikeyjcat:
I believe all of the feedback in here is incorporated in the other individual PR's. I assume I should close this PR?

kytrinyx:
Good point. I'll close it now. Thanks!

PR #117, opened 2015-10-19
Added raindrops.json test definition file

mikeyjcat:
Removed the last test with the 12_121 number. I feel this test adds no value, and introduces unnecessary complexity into the test generation to create a test that passes style checks.

### DIFF
--- a/raindrops.json
+++ b/raindrops.json
@@ -1,0 +1,79 @@
+{
+   "cases": [
+      {
+         "description": "1",
+         "number" : 1, 
+         "expected": "1"
+      },
+      {
+         "description": "3",
+         "number" : 3, 
+         "expected": "Pling"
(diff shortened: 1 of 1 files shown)


kytrinyx:
Why not just use 12121?

mikeyjcat:
Can do, then rubocop gets shirty and wants the underscore (which is a language specific thing, again).

I didn’t see the value in adding yet another value without prime factors 3,5,7. The other tests have already proven the point of being able to return a number without the factors turned into strings, 12121 does not test any additional capability of a solution that I can see.

kytrinyx:
Yeah that's fair. I think we originally added it because people wrote solutions that hard-coded solutions, and wouldn't pass for larger numbers. This is fine, though.

PR #118, opened 2015-10-25
Removed test messages from hello-world.json

mikeyjcat:


kytrinyx:
Excellent. Thanks for taking the time to have that discussion so we could figure out the best approach to this.

mikeyjcat:
Its a pleasure

PR #120, opened 2015-10-26
Mention possibility of other strain function names

cayennes:
Also added just a little more exlanatory text.


kytrinyx:
Excellent, thank you.

PR #123, opened 2015-10-30
Added difference-of-squares.json test definition

mike-hewitson:
Some extra comments as this one brought in a few new challenges. Hope you like my approach.

### DIFF
--- a/difference-of-squares.json
+++ b/difference-of-squares.json
@@ -1,0 +1,59 @@
+{
+   "#": [
+      "Note that this test will need a bit more complexity in the test generator.",
+      "The generator will need to take into account that we have three method",
+      "calls, as well as to correctly format the long expected results number",
+      "dependant on the relevant language style eg 25_164_150 in Ruby.",
+      "In addition, some languages can include a test for consistent differences",
+      "which will need to be explictly articulated in the test generation template.",
+      "This test would have the form: assert_equal squares.difference, squares.difference",
+      "expressed in a lnaguage apprpriate syntax an style"
+   ],
+   "cases": [
+      {  
+         "description": "square of sums to 5",
+         "square_of_sums": 5,
+         "expected": 225
+      },
+      {
+         "description": "sum of squares to 5",
+         "sum_of_squares": 5,
+         "expected": 55 
+      },
+      {
+         "description": "difference of sums to 5",
+         "difference": 5,
+         "expected": 170
+      },
+      {
+         "description": "square of sums to 10",
+         "square_of_sums": 10,
+         "expected": 3025
+      },
+      {
+         "description": "sum of squares to 10",
+         "sum_of_squares": 10,
+         "expected": 385
+      },
+      {
+         "description": "difference of sums to 10",
+         "difference": 10,
+         "expected": 2640
+      },
+      {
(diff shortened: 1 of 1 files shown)


soniakeys:
It's workable, but I would prefer that test cases for different methods be grouped in separate json objects.  See for example clock.json and custom-set.json.  I do agree that the consistent difference checks don't need anything extra from the json.

mikeyjcat:
@soniakeys Let me have a look at it and see how it would work out with the generator.

kytrinyx:
@mikeyjcat the generators are written separately for each problem for exactly this reason.

mikeyjcat:
@soniakeys Wouldn't this change the order of the tests for all languages? Currently the tests fir a specific number are grouped together (e.g. sum of squares, square of sums, difference, all together for 5). If we were to group together by function, then we'd intersperse them, which I think would affect the understandability of the test suite? What do you think?

soniakeys:
Maybe just some languages, but I think I see your point.  A problem though is that in your PR here, the grouping by number is implicit, not structural.  This means test programs (or generators) must include some logic to see which fields are present or missing.  If objects have a more consistent structure, test programs and generators stay simpler.  So, for example, the three results could all be included in a single object, like

```
{
   "input": 5,
   "square-of-sums": 225,
   "sum-of-squares": 55,
   "difference": 170
}
```

Then each object could have the same structure.

mikeyjcat:
@soniakeys I think your suggestion is great, and I like it. I'll rework the generator and see how it goes, and let you know.

soniakeys:
@kytrinyx this grouping by number still leaves it easy for test program to order tests by method if they like.  They just run through the list of test cases multiple times, once for each method they are testing.

kytrinyx:
@soniakeys Agreed, your suggestion is simpler.

mikeyjcat:
Yep, I just need to work out how to do it elegantly. We've got the test names (n) , input data (1), and methods (n) (not sure what the language agnostic word is). The pattern we've been considering suggested by Sonia doesn't consider the test name, and I want to check the idea against the other test definitions to see if anything else emerges. This seems to be my experience for each one I've done so far:)

kytrinyx:
That sounds like a plan. We don't need test names in Go, since we use table tests rather than individual tests. That said, a description for each test case can be useful in some situations, and in languages like Ruby where we have individual tests, the description can be used as the name.

I have to board my flight. I'll think about how to name the thing that becomes method/function.

kytrinyx:
@mikeyjcat I never came up with any good word for method/function, but on reconsidering this discussion, I don' think it needs it. Also, I don't think that a test name is necessary. Each language can construct the name based on the test data however they like. It will depend on whether or not they put these in the same test or not.

kytrinyx:
@mike-hewitson Would you mind taking a look at this to see if we can take it across the finish line?

mike-hewitson:
Will do, apologies. I'll get it done this weekend.

petertseng:
Oof, realized too late that we had two people do this: https://github.com/exercism/x-common/pull/146

kytrinyx:
Aaah, in that case, apologies for the mixup! I'm going to go ahead and close this one. Thanks all!

PR #124, opened 2015-10-30
meetup common test data

soniakeys:
No test cases for fifth weeks.  See https://github.com/exercism/exercism.io/issues/2142 (and some discussion in individual languages, like https://github.com/exercism/xgo/issues/151.)

### DIFF
--- a/meetup.json
+++ b/meetup.json
@@ -1,0 +1,764 @@
+{
+   "cases": [
+      {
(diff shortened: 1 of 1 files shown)


kytrinyx:
Yeah, agreed. We can't have a monthly meetup only in months with five weeks. Thanks!

PR #126, opened 2015-11-06
Fixes "Cysteine" typo

mamenama:
Because cystine is something else entirely.

Minor, but I am a protein scientist and care about such things.


kytrinyx:
Oh my goodness, great catch. Thank you!

PR #128, opened 2015-11-12
word-count: tweak normalization test case

kytrinyx:
In exercism/xpython#234 we discussed how adding a second
word in the normalization case might allow
some languages to use sorted counts as a normalization-agnostic
assertion.

I don't know that this will work in practice as allowing people to normalize to a different case, since all the other tests have lowercase words.

### DIFF
--- a/word-count.json
+++ b/word-count.json
@@ -27,8 +27,8 @@
         },
         {
             "description": "normalize case",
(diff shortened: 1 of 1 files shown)


ryanplusplus:
Looks good to me.

PR #125, opened 2015-11-15
WIP! Create contributing guide specific to tracks and exercises

kytrinyx:
I'd like to add a CONTRIBUTING document to this repository with suggestions and guidelines for track maintainers, which we could then link people to when we ask if they'd like to be added as a maintainer on a track.

I don't know what should go into it, but here are some suggested starting points, based on a discussion I had recently with a new maintainer.

**PRs**

- If there are multiple maintainers, how do we avoid diffusion of responsibility?
- When should we ask for a second opinion when merging vs just merging right off the bat?
- How do we deal with typos and simple fixes? (just commit to master? Create a PR and then merge it yourself so the others get a notification?)

**When creating a problem**

- What behaviors should be explicitly specified in the readme?
- Should corner cases be explicitly called out or not?
- How hard should problems be?

I'd be interested in hearing other suggestions for topics that would be useful in this sort of document.

/cc @jtigger @petertseng @mixolidia (because you're fairly new maintainers)
/cc @exercism/lisp because you have really lively and interesting discussions
/cc @exercism/track-maintainers because you all have been doing this for a while!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,472 @@
+# Contributing
+
+It's pretty tricky to keep all the language tracks in good working order,
+and it's not uncommon that people discover typos in the README or the test
+suites, inconsistencies between the README and the tests, missing edge cases,
+factual errors, logical errors, and confusing ambiguities.
+
+We welcome contributions of all sorts and sizes, from reporting issues to
+submitting patches, to hanging out in the support chat to help people get
+unstuck.
+
+We are grateful for any help in making Exercism better!
+
+-----
+
+This guide covers several common scenarios pertaining to **improving the language tracks**
+themselves. There are other guides about contributing to other parts of the Exercism ecosystem.
+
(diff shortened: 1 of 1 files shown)


kytrinyx:
Update: I converted this issue to a PR, and will address the various things that were discussed here.

mikeyjcat:
Yep, very happy to do that. Been off the radar for the last few weeks, bad time of the year from a work perspective, budgets, rush to get things complete, regulator on-site inspections, but should get stick in again soon.

What I think think would make sense is gotchas and done patterns for different shapes of tests. I did feel like these were emerging, and I was starting to use them myself.

kytrinyx:
> bad time of the year from a work perspective

Yeah, I bet!

> What I think think would make sense is gotchas and done patterns for different shapes of tests.

That would be very useful. If they've not quite emerged yet, that's fine, we can always add things later. I just wanted to capture any of the existing knowledge so far.

kytrinyx:
The discussion seems to have wound down. I've tidied up what's there, and am going to merge it and create new issues for the missing bits and pieces.

Issue #130, opened 2015-11-17
Robot Name Requirements


tleen:
I'm finding the uniqueness requirements of the robot name task unclear. 

```text
Random names means a risk of collisions. In some exercism language tracks there are tests to ensure that the same name is never used twice.
```

Does this indicate that collisions should be avoided in every environment or just in those with tests specifically disallowing collisions?

kytrinyx:
Hm, yeah, I see what you mean. We should clarify that.

The original intention was to avoid collisions, but in tracks where there are no tests about that this might be more part of the discussion than part of the original requirement.

tleen:
Actually while we are looking at it, does it read to you that the randomness of the names is a requirement? Or would an acceptable solution (+1) iterate the names as a means to avoid collision?

kytrinyx:
The way I read it randomness is a requirement, but you're right--this could be reworded to provide a broader range of acceptable approaches.

kytrinyx:
Fixed in #131

neurogenesis:
@kytrinyx ... i took a quick look to see how other people implemented this. it looks like in many cases people just use either `sample` or `rand`, without explicitly preventing duplicates. Even with the committed change, it's not entirely clear what the intent is:

1. follow TDD and allow use of rand or sample as sufficient to meet the requirements for random number generation
2. write or use an algorithm that is less likely to generate predictable / repeatable random values (SecureRandom)
3. include additional logic to reject generated random names that have already been used (lightweight ephemeral registry via array/hash/set)

I'm curious if this might be better served with another test to verify that a given random name is already in use (implying a name registry of sorts):

- test for valid_name to verify a previously assigned name is rejected as a duplicate

kytrinyx:
I'm not sure. It seems a bit awkward for the robot API to expose a method for checking whether or not a name is valid, but I do think that you're right that the problem becomes more interesting if we are able to ensure that names are not duplicated.

tleen:
The test(s) would have to be reformulated to track for that. While the robot generator would have to guard against duplicates it would be weird to have the generator prove each name is unique. 

I read the problem to indicate the the name had to be in the format LLNNN, which ends up limiting the total namespace. Fundamentally having to enforce randomness and uniqueness is tricky given the finite namespace. It would definitely make the problem harder to solve.

[...]

PR #131, opened 2015-11-18
stricter definition of robot-name generator instructions regarding randomness and collisions

tleen:
Result of discussion from issue #130 creating more clarity in the expected program output.


kytrinyx:
Thank you!

PR #133, opened 2015-11-22
Add new exercise pangram

behrtam:
We have `anagrams` and `palindromes` so we can't leave out `pangrams`. I think it's a nice short and simple task for beginners.

### DIFF
--- a/pangram.json
+++ b/pangram.json
@@ -1,0 +1,29 @@
+{
+   "cases": [
+      {
+         "description": "sentence empty",
+         "input": "",
+         "expected": false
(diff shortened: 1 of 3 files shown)


matthewmorgan:
@rchavarria how about this?  I would be happy to implement this in the ES track.  I could do JS too at the same time I guess.  Let me know.  @behrtam thanks for doing this

rchavarria:
@matthewmorgan It would be awesome to have both!

kytrinyx:
This is great, thank you!

PR #135, opened 2015-11-28
corrects typo in crypto-square readme

jtrudell:


kytrinyx:
Good find, thank you.

PR #144, opened 2015-12-01
Add new exercise rail fence cipher

behrtam:
`simple-cipher` and `atbash-cipher` definitely need a companion. Meet `rail-fence-cipher`.


kytrinyx:
Very neat!

PR #147, opened 2015-12-02
Fix a minor grammatical error in forth.md

arnavroy:


kytrinyx:
Nice catch, thank you.

PR #151, opened 2015-12-04
Add shared inputs/outputs json for Pig Latin exercise

suzannehamilton:
This is a proposed fix for exercism/todo#125.

The tests all take an input string and compare the output to an expected string, so the test cases all have the same structure.

Some languages provide a named test for each case, and some group them together into parameterized tests. I've included descriptions at both levels so that the generators can use one or both types of description.

What do you think?

### DIFF
--- a/pig-latin.json
+++ b/pig-latin.json
@@ -1,0 +1,128 @@
+{
+  "#" : [
+    "In languages with parameterized tests, test cases are grouped together into similar cases with",
+    "test names or comments to describe the group. Other languages just name the individual tests."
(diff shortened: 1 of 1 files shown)


kytrinyx:
Thank you, @suzannehamilton, this looks great.

suzannehamilton:
Three minutes to merge. Amazing, thanks. :grin:

kytrinyx:
Haha, I was right in the middle of processing my GitHub notifications :) Just good timing is all!

PR #152, opened 2015-12-04
Fix typo in Hamming test definitions

suzannehamilton:
Just fixes a small typo that I noticed when building another file.


kytrinyx:
Nice catch, thank you.

PR #158, opened 2015-12-08
simple-linked-list: update Ruby example to include null object pattern

connorlay:
In the simple-linked-list README we have the following example Ruby code:
~~~ruby
Element.to_a(nil) #=> []
Element.from_a([]) #=> nil
Element.reverse(nil) #=> nil
one = Element.new(1, nil)
one #=> <Element @datum=1, @next=nil>
one.datum #=> 1
one.next #=> nil
two = Element.new(2, one)
two #=> <Element @datum=2, @next=<Element @datum=1 @next=nil>>
Element.to_a(two) #=> [2, 1]
Element.reverse(two) #=> <Element @datum=1 @next=<Element @datum=2 @next=nil>>
range = Element.from_a(1..10)
range.datum #=> 1
range.next.next.next.next.next.next.next.next.next.next.datum #=> 10
~~~

However in `simple_linked_list_test.rb` we have the following test:
~~~ruby
  def test_empty_list_operations
    skip
    assert_equal [], Element.from_a([]).to_a
    assert_equal [], Element.from_a([]).reverse.to_a
  end
~~~

The issue of the linked-list returning `nil` was addressed in [this commit](https://github.com/exercism/xruby/commit/fa6366e6f8a41da170205c0df6bd69fbf6d1e2ed), so I figured the Ruby code in the README should be updated to reflect that.

### DIFF
--- a/simple-linked-list.md
+++ b/simple-linked-list.md
@@ -20,20 +20,3 @@ linked list and convert to and from arrays.
 
 When implementing this in a language with built-in linked lists,
 implement your own abstract data type.
-
-Examples (Ruby):
-
-    Element.to_a(nil) #=> []
-    Element.from_a([]) #=> nil
-    Element.reverse(nil) #=> nil
-    one = Element.new(1, nil)
-    one #=> <Element @datum=1, @next=nil>
-    one.datum #=> 1
-    one.next #=> nil
-    two = Element.new(2, one)
-    two #=> <Element @datum=2, @next=<Element @datum=1 @next=nil>>
-    Element.to_a(two) #=> [2, 1]
-    Element.reverse(two) #=> <Element @datum=1 @next=<Element @datum=2 @next=nil>>
-    range = Element.from_a(1..10)
-    range.datum #=> 1
-    range.next.next.next.next.next.next.next.next.next.next.datum #=> 10


kytrinyx:
Ah. I didn't realize we still had ruby code in the README. I don't think we should have language-specific in the code, since the README gets delivered for all the languages.

@exercism/track-maintainers should we just remove the example code? Should we translate the code example into generic pseudocode?

rchavarria:
I don't mind if there is very little code, just two or three lines of code, to provide an example. 

The code shown in this issue looks like tests. Even translating it to pseudocode, it's too much for a README. It looks like specs describing the `Element` API, so I'd put that code in tests (it probably is), writing several test cases.

matthewmorgan:
I know zero Ruby, so the above example code is a bit cryptic for me.  Maybe pseudo code is better, consider mine the LCD opinion!

ryanplusplus:
I agree with @rchavarria -- that looks like tests more than pseudo-code.

kytrinyx:
Yeah, this is going to be confusing for all the other language tracks.

Should we remove the entire bit of code for now? People can always look at the test suites for details.

masters3d:
+1 for removal

The liked list example does not include sample code.

https://github.com/exercism/x-common/blob/master/linked-list.md

connorlay:
Removing the example code makes the most sense to me. That way all implementation details are left in the tests themselves. I've updated the README accordingly.

kytrinyx:
Excellent! Thanks for taking the time to discuss and work through this.

Issue #164, opened 2015-12-26
largest-series-product: add common data?


petertseng:
I feel like this problem is a pretty good candidate for having common test data.

Well, I'm sure a lot of other problems are too, but this one in particular caught my attention when reviewing code.

Things to think about:
* Any invalid inputs? `largest_series_product("1234a5", 2)`? `largest_series_product("12345", -1)`?
* I'd really like tests for `largest_series_product("123", 0)` (should return 1 for empty product) and `largest_series_product("", 1)` (should error). Currently most tracks only test `largest_series_product("", 0) == 1` which is reasonable since it's a boundary case, but in a two-input function isn't it right to test cases where only one input is a boundary as well as both?
* I'd really like a test for `largest_series_product("99099", 3)` - this should be 0 (but some people's code say 1 because they assume the minimum is 1... it's not!)
* Different languages handle error cases differently, be careful. Example: Ruby raise exceptions, Go return `(int, error)`, Clojure [just... doesn't](https://github.com/exercism/xclojure/blob/master/largest-series-product/largest_series_product_test.clj#L21)? (I feel like this should change...). But this is a problem for each track to handle individually.
* Obviously making each track actually use the common data is a bigger question, but that's a problem for each track to handle individually - let's at least get the process started by *having* common data.

tracks with this problem: http://x.exercism.io/problems/largest-series-product

I have an intent to do this myself, but I can't do it immediately so this issue will serve as a reminder for me when I can.

I'll survey the test data being used by the various tracks in a bit and see if I see any interesting features as well.

kytrinyx:
Yes, definitely! Eventually I'd like to have common data for all the problems that are used in multiple tracks. I made a bunch of individual issues for it over in the "todo" repository (which is used for hackdays and things like 24pullrequests where a bunch of people are new to exercism and new to open source).

- https://github.com/exercism/todo/labels/shared-test-data
- https://github.com/exercism/todo/issues/104

petertseng:
starting work on this.

PR #167, opened 2016-01-07
largest-series-product: Add JSON test data

petertseng:
These are a union of the existing tests among various language tracks.
Various error and boundary cases have been added as well.

Issues:
Issue #164
https://github.com/exercism/todo/issues/104

### DIFF
--- a/largest-series-product.json
+++ b/largest-series-product.json
@@ -1,0 +1,119 @@
+{
+    "#": [
+        "A negative expected value means the input is invalid.",
+        "Different languages may handle this differently.",
+        "e.g. raise exceptions, return (int, error), return Option<int>, etc.",
+
+        "Some languages specifically test the string->digits conversion",
+        "and the 'slices of size N' operation.",
+        "These cases *deliberately* do not cover those two operations.",
+        "Those are implementation details.",
+        "Testing them constrains implementations,",
+        "and not all implementations use these operations.",
+        "e.g. The implementation which makes a single pass through the digits."
+    ],
+    "cases": [
+        {
+            "description": "can find the largest product of 2 with numbers in order",
+            "digits": "0123456789",
+            "span": 2,
+            "expected": 72
+        },
+        {
+            "description": "can find the largest product of 2",
+            "digits": "576802143",
+            "span": 2,
+            "expected": 48
+        },
+        {
+            "description": "finds the largest product if span equals length",
+            "digits": "29",
+            "span": 2,
+            "expected": 18
+        },
+        {
+            "description": "can find the largest product of 3 with numbers in order",
+            "digits": "0123456789",
+            "span": 3,
+            "expected": 504
+        },
+        {
+            "description": "can find the largest product of 3",
+            "digits": "1027839564",
+            "span": 3,
+            "expected": 270
+        },
+        {
+            "description": "can find the largest product of 5 with numbers in order",
+            "digits": "0123456789",
+            "span": 5,
+            "expected": 15120
+        },
+        {
+            "description": "can get the largest product of a big number",
+            "digits": "73167176531330624919225119674426574742355349194934",
+            "span": 6,
+            "expected": 23520
+        },
+        {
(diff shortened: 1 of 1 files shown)


largest-series-product.json
```diff
@@ -0,0 +1,119 @@
...
+        "Those are implementation details.",
+        "Testing them constrains implementations,",
+        "and not all implementations use these operations.",
+        "e.g. The implementation which makes a single pass through the digits."
+    ],
+    "cases": [
+        {
+            "description": "can find the largest product of 2 with numbers in order",
+            "digits": "0123456789",
+            "span", 2,
```
petertseng: oops my bad. just a second

largest-series-product.json
```diff
@@ -0,0 +1,119 @@
...
+            "span": 6,
+            "expected": 28350
+        },
+        {
+            "description": "can get the largest product of a big number (Project Euler)",
+            "digits": "7316717653133062491922511967442657474235534919493496983520312774506326239578318016984801869478851843858615607891129494954595017379583319528532088055111254069874715852386305071569329096329522744304355766896648950445244523161731856403098711121722383113622298934233803081353362766142828064444866452387493035890729629049156044077239071381051585930796086670172427121883998797908792274921901699720888093776657273330010533678812202354218097512545405947522435258490771167055601360483958644670632441572215539753697817977846174064955149290862569321978468622482839722413756570560574902614079729686524145351004748216637048440319989000889524345065854122758866688116427171479924442928230863465674813919123162824586178664583591245665294765456828489128831426076900422421902267105562632111110937054421750694165896040807198403850962455444362981230987879927244284909188845801561660979191338754992005240636899125607176060588611646710940507754100225698315520005593572972571636269561882670428252483600823257530420752963450",
```
petertseng: slight concern here - the number is quite large and the answer exceeds a 32-bit int (but fits well within a 64-bit int). Go is the only track that has this test case

petertseng:
Going to test this against my implementations briefly to make sure I didn't include any bad data.

largest-series-product.json
```diff
@@ -0,0 +1,119 @@
...
+            "description": "reports zero if all spans include zero",
+            "digits": "99099",
+            "span": 3,
+            "expected": 0
+        },
+        {
+            "description": "rejects span longer than string length",
+            "digits": "123",
+            "span": 4,
+            "expected": 1
```
petertseng: -1, not 1.

petertseng:
```ruby
require_relative 'largest_series_product'
require 'json'

cases = JSON.parse(IO.read('largest-series-product.json'))
cases['cases'].each { |c|
  expected = c['expected']

  raised = nil
  begin
    observed = Series.new(c['digits']).largest_product(c['span'])
  rescue => e
    raised = e
  end

  prefix = "expected #{expected} for #{c['digits']} and #{c['span']}"

  if expected >= 0
    if observed != expected
      raise prefix + ", got #{observed}"
    elsif raised
      raise prefix + ", raised #{raised}"
    end
  elsif !raised
    raise prefix + ", got #{observed}"
  end
}

puts 'pass'
```

OK, looks good to me

petertseng:
Discussion in https://github.com/exercism/xclojure/issues/49 is interesting.

The argument goes that `LSP("", 0)` should be `max([])` which is an error. I can agree that if you get `max([])` that's an error. However, I disagree that `LSP("", 0)` is `max([])`.

If I ask for how many ways to make 1-character strings out of "123", there are three: "1", "2", "3". Of those three strings, we find the string that gets the max product, which is 3.
If I ask for how many ways to make 0-character strings out of "" (or "123" for that matter) there is one way: "". Of that one string, the string that gets the max product is whatever the product of "" is, which is 1. So `LSP("", 0)` is actually `max([product("")])` which is 1, not an error.

We could avoid the problem completely by not testing this at all, but I consider it reasonable to do so. Could probably add a comment to the test cases explaining that.

kytrinyx:
A comment sounds like a good idea.

PR #176, opened 2016-01-30
Updated the definition of a triangle

hankturowski:


hankturowski:
There is some question about the definition of a triangle, specifically about the degenerate triangle. If the sum of the lengths of any two sides equal the length of the third, then the triangle is degenerate, but still a triangle. It just has no area, and is isomorphic to a line segment.

If we don't want to include degenerate triangles in our definition of a triangle, then we should just update this file to make the rules clear.

hankturowski:
I'm closing my own pull request until we decide the correct solution. I don't want to confuse things.

PR #178, opened 2016-02-06
Clarify documentation

copiousfreetime:
Only mention singly linked list and that they may be used to implement
LIFO stacks. At this point there is no need to mention things like
persistent data structures and immutability.


copiousfreetime:
This pull request is to be paired with exercism/xruby#252

kytrinyx:
Yes, much better! Thanks :)

PR #180, opened 2016-02-07
Fix markdown table formatting for protein-translation

ryanplusplus:
Use the rich diff to see the difference when the markdown is rendered


kytrinyx:
Nice, thanks!

PR #181, opened 2016-02-07
Fix typos and improve consistency for POV overview

ryanplusplus:


pov.md
```diff
@@ -5,7 +5,7 @@ point of view. For example family trees are usually presented from the
 ancestor's perspective:
 
 ```
-          O
+          0
```
ryanplusplus: The letter `O` was used instead of the number zero.

pov.md
```diff
@@ -17,19 +17,18 @@ in the graph, by pulling it up to the root and dragging its relationships along
...
-          1-----+-----3
-          |           |
-        4-+-5       8-+-9
+        6
+        |
+  7-----+-----2
+              |
+        1-----0-----3
+        |           |
+      4-+-5       8-+-9
```
ryanplusplus: I aligned this with the above graph and used the same notation for a group of three children as used in the above graph.

pov.md
```diff
@@ -17,19 +17,18 @@ in the graph, by pulling it up to the root and dragging its relationships along
...
 ```
 
 This lets us more simply describe the paths between two nodes. So for example
 the path from 6-9 (which in the first graph goes up to the root and then down to
 a different leaf node) can be seen to follow the path 6-2-0-3-9
 
-This exercise involves taking an input graph and re-orientating it one one of
-the nodes. 
+This exercise involves taking an input graph and re-orientating it from the point
+of view of one of the nodes.
```
ryanplusplus: It wasn't 100% clear to me what the original author tried to type, but I think this captures the intent.

kytrinyx:
This looks great, thank you.

PR #183, opened 2016-02-16
Update hello-world.md workflow.

DeonHua:
The old link was broken, so I replaced it with an inline command as it's not too complex.


kytrinyx:
Good call!

PR #188, opened 2016-03-03
Rewrite of the linked-list problem.

cebial:
The description and blurb for this problem seemed very confusing and contained incorrect information.

1) The original blurb and description specifically mentioned *singly linked lists*, but the goal of the exercise was to write a *doubly linked list* (or alternatively a *circular linked list*). See [Wikipedia](https://en.wikipedia.org/wiki/Linked_list) for the definitions and explanations.

2) The text was filled with references to "deques", without any explanation of the term (it's short for **d**ouble-**e**nded **que**ue). These deques were compared with arrays, but as I understand it, an array, like a linked list, is a possible implementation of a deque, which is a more abstract concept. See [this explanation on Quora](https://www.quora.com/What-is-the-difference-between-a-deque-and-a-doubly-linked-list).

3) A "Proxy pattern" was mentioned in the blurb and text, but I don't think this exercise has anything to do with that pattern. From [Wikipedia](https://en.wikipedia.org/wiki/Proxy_pattern):
>A proxy, in its most general form, is a class functioning as an interface to something else.
>For the client, usage of a proxy object is similar to using the real object, because both implement the same interface.

I did a major rewrite, trying to stay as close to the original problem as possible, and also trying to educate the users by mentioning and explaining singly and doubly linked lists, nodes, and some common use cases for linked lists.

I copied the last paragraph of the original description almost verbatim, I know nothing about Elixir or Erlang, so I can't really judge if this information is accurate. Maybe someone can check this.

### DIFF
--- a/linked-list.md
+++ b/linked-list.md
@@ -1,29 +1,34 @@
-Deques have four fundamental operations (using Array terminology):
+Like an array, a linked list is a simple linear data structure. Several 
+common data types can be implemented using linked lists, like queues, 
+stacks, and associative arrays.
 
-* push (insert value at back)
-* unshift (insert value at front)
-* pop (remove value at back)
-* shift (remove value at front)
+A linked list is a collection of data elements called *nodes*. In a 
+*singly linked list* each node holds a value and a link to the next node. 
+In a *doubly linked list* each node also holds a link to the previous 
+node.
 
-The difference between deque and Array performance is that a deque
-implements all of these operations in constant time. Even clever
-implementations of Array will often have to copy the entire Array in
-order to unshift.
+You will write an implementation of a doubly linked list. Implement a 
+Node to hold a value and pointers to the next and previous nodes. Then 
+implement a List which holds references to the first and last node and 
+offers an array-like interface for adding and removing items:
 
-Under the hood we'll use an Element class with two fields, `next` and
-`prev`, that are both writable.
+* `push` (*insert value at back*);
+* `pop` (*remove value at back*);
+* `shift` (*remove value at front*).
+* `unshift` (*insert value at front*);
 
-To make the API usable, you'll use a Deque class as a proxy for this
-list. There two good ways to implement Deque: maintain separate
-references to the first and last Element, or maintain a reference to one
-of them and ensure that the list is circular.
+To keep your implementation simple, the tests will not cover error 
+conditions. Specifically: `pop` or `shift` will never be called on an 
+empty list.
 
-To keep your implementation simple, the tests will not cover error
-conditions. Specifically: pop or shift will never be called on an empty
-Deque.
+If you want to know more about linked lists, check [Wikipedia](https://en.wikipedia.org/wiki/Linked_list).
+
+---
 
 In languages that do not have good support for mutability (such as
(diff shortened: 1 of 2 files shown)


linked-list.md
```diff
@@ -1,29 +1,35 @@
...
+*singly linked list* each node holds a value and a link to the next node. 
+In a *doubly linked list* each node also holds a link to the previous 
+node.
 
-The difference between deque and Array performance is that a deque
-implements all of these operations in constant time. Even clever
-implementations of Array will often have to copy the entire Array in
-order to unshift.
+You will write an implementation of a doubly linked list. You will use a
+`Node`-class (or similar) to hold a value and pointers to the next and
```
kytrinyx: Can we avoid calling it a "class" by saying something like _you will implement a Node to hold a value and pointers..._?

linked-list.yml
```diff
@@ -1,4 +1,4 @@
 ---
-blurb: "Write a Singly-Linked List implementation that uses the Proxy pattern"
+blurb: "Implement a doubly linked list"
 source: "Inspired by 'Data Structures and Algorithms with Object-Oriented Design Patterns in Ruby', singly linked-lists."
```
kytrinyx: Let's also change the source to "Classic computer science topic" or something, so that it doesn't reference singly-linked list here.

kytrinyx:
Thanks so much for taking the time to rework this. We've had a huge amount of confusion between the simple-linked-list exercise (singly linked) and this one. I think we got the simple one sorted out a while back, but we clearly missed this in the confusion.

cebial:
My pleasure, I decided to rework this exercise because I got very confused myself when I encountered it first in the JavaScript track and it has been a nice learning experience already (also learning to use github).

I followed your suggestions, only the capitalization of Node and List suggests the use of a class now.

kytrinyx:
This is great, thanks! Would you mind squashing the commits? http://blog.steveklabnik.com/posts/2012-11-08-how-to-squash-commits-in-a-github-pull-request

cebial:
That was quite an adventure with git, thanks for the learning opportunity.

kytrinyx:
I never know if "adventure" is a good thing. I worked with someone once who said they were always getting themselves up "git creek".

Thanks :)

cebial:
It seemed simple enough at first, but then there was a lot of exploring and getting lost involved. Eventually I managed to succeed, but I'm not exactly sure how. 

All in all a solid adventure, will do again. :)

kytrinyx:
If you have time, there's a (long) video of a conference session that is really good for wrapping your head around the underlying model of Git: https://www.youtube.com/watch?v=1ffBJ4sVUb4 Git Ages 4 and Up

petertseng:
I liked [The Git Parable](http://tom.preston-werner.com/2009/05/19/the-git-parable.html) for learning the underlying concepts. The squash operation is covered in "Rewriting History" though it probably makes the most sense if reading the whole thing.

I have not watched the above video to see whether there's any overlap in information between our suggestions.

cebial:
@kytrinyx and @petertseng, thanks for the links, I will check them out when I have time.

PR #189, opened 2016-03-06
octal: Remove Ruby-specific terms

petertseng:
These descriptions need to be language-agnostic since they are shown for
each language. #100 did the same for hexadecimal.


kytrinyx:
Good catch, thanks.

Issue #193, opened 2016-03-08
linked-list: Suspicious alternative for functional languages


NobbZ:
During reading the changes made by PR #188, I encountered some paragraph in the `linked-list.md`, which I do find is giving a suspicious alternative for implementing the exercise in functional languages:

> In languages that do not have good support for mutability (such as Elixir or Erlang), you may choose to implement a functional list where shift and pop have amortized O(1) time. The simplest data structure for this will have a pair of linked lists for the front and back, where iteration order is front ++ reverse back.

In a commit of the afforementioned PR, I already did a quick comment:

https://github.com/cebial/x-common/commit/552006a4610f917233e604ace3e964158ad01d6b#diff-25093f3da30fc2fdff5ad6a3967a7dfbL29

> So this paragraph is describing an Okasaki-Functional-Double-Ended-Queue. This is not comparable with Doubly-Linked-Lists in any way.
> 
> Perhaps we should split this exercise up into this one and describe it in a imperativ way only while creating another exercise which does require you to implement a dequeue.
> 
> Another way to interprete this paragraph is very similar to a zipper as in [zipper.md](https://github.com/exercism/x-common/blob/master/zipper.md). If it is meant like a zipper we should probably don't do this exercise in languages which are immutable only and stick to that zipper-exercise.

Are there any further opinions on this?

kytrinyx:
I think the right move is to _not_ implement the linked-list in functional languages. We should remove the suspicious alternative from the README.

If functional languages have implemented linked-list with the alternative, we should consider deprecating it.

If the zipper problem is sufficiently close to the dequeue, I would not want to implement a new exercise specifically for the dequeue.

NobbZ:
The Zipper is actually far away from a dequeue, its just another way to read that paragraph.

The functional dequeue is something that holds two lists, one "front" and one "back" queue. One appends by "consing" to the back, and pops by taking the head from the front. If the front is empty when one tries to pop, the "back" is reversed and used as front.

[...]

PR #194, opened 2016-03-09
Update linked-list to remove outdated instructions

cebial:
This is an addition to pull request #188 which has already been merged. As mentioned by @NobbZ there and in issue #193, the track specific instructions for Elixir and Erlang mentioned an incorrect alternative solution. 

I left these instructions in during the rewrite because I wasn't sure how relevant or important they were for those tracks.

This PR removes the outdated/incorrect paragraph.


kytrinyx:
Thanks!

Issue #197, opened 2016-03-12
Fix confusion in Largest Series Product exercise


kytrinyx:
See exercism/ruby#262 and exercism/ruby#244

The README for the exercise uses the term _consecutive_ in a way that is ambiguous. In this context it means _consecutive by position_, not _consecutive by numerical sequence_.

To fix this we should:

- [x] update the README to clarify that the digits need to be consecutive by position, not necessarily in numeric order (e.g. '487' has the consecutive numbers '48', '487', and '87', even though 4 and 8 or 8 and 7 are not considered _consecutive numbers_.) It would help to not use _0123456789_ as the example.
- [x] create a set of shared test data (`largest-series-product.json`)
- [ ] survey the existing language tracks that implement this exercise to see that they're correctly implemented against the requirements as defined above.

At the moment the existing implementations are as follows:

- [ ] https://github.com/exercism/xclojure/tree/master/exercises/largest-series-product
- [ ] https://github.com/exercism/xcsharp/tree/master/exercises/largest-series-product
- [ ] https://github.com/exercism/xecmascript/tree/master/exercises/largest-series-product
- [ ] https://github.com/exercism/xelixir/tree/master/exercises/largest-series-product
- [ ] https://github.com/exercism/xerlang/tree/master/exercises/largest-series-product
- [ ] https://github.com/exercism/xgo/tree/master/exercises/largest-series-product
- [ ] https://github.com/exercism/xjavascript/tree/master/exercises/largest-series-product
- [ ] https://github.com/exercism/xlfe/tree/master/exercises/largest-series-product
- [ ] https://github.com/exercism/xlua/tree/master/exercises/largest-series-product
- [ ] https://github.com/exercism/xpython/tree/master/exercises/largest-series-product
- [ ] https://github.com/exercism/xruby/tree/master/exercises/largest-series-product
- [ ] https://github.com/exercism/xscala/tree/master/exercises/largest-series-product
- [ ] https://github.com/exercism/xswift/tree/master/exercises/largest-series-product
- [ ] https://github.com/exercism/xhaskell/tree/master/largest-series-product
- [ ] https://github.com/exercism/xperl5/tree/master/largest-series-product

neontapir:
Thanks. I've been traveling on business and couldn't file this right away.

petertseng:
I wil update READMEs, but have my hands full with other tasks right now so I can't get to the other two items, so they are free for others to take!

petertseng:
[...]

PR #200, opened 2016-03-13
series, largest-series-product: clarify "consecutive"

petertseng:
It was unclear whether "consecutive" meant consecutive in *numerical
value* or simply *positionally adjacent*. Due to this problem's origin
in Project Euler, it was the latter.

To avoid any confusion this commit does the following:
* Replace the word "consecutive" in .yml files with "contiguous"
* Clarifies in the .md files that the series of digits only need be
  positionally adjacent, not consecutive in numerical value.
* Uses examples in .md files that should make clear that numerical value
  is not important.

This should prevent any further confusion of the form expressed in the
two issues

* https://github.com/exercism/xruby/issues/244
* https://github.com/exercism/xruby/issues/262

This is the first subtask of #197.

### DIFF
--- a/largest-series-product.md
+++ b/largest-series-product.md
@@ -1,6 +1,9 @@
-For example, for the input `'0123456789'`, the largest product for a
-series of 3 digits is 504 (7 * 8 * 9), and the largest product for a
-series of 5 digits is 15120 (5 * 6 * 7 * 8 * 9).
+For example, for the input `'1027839564'`, the largest product for a
+series of 3 digits is 270 (9 * 5 * 6), and the largest product for a
+series of 5 digits is 7560 (7 * 8 * 3 * 9 * 5).
+
+Note that these series are only required to occupy *adjacent positions*
+in the input; the digits need not be *numerically consecutive*.
 
 For the input `'73167176531330624919225119674426574742355349194934'`,
 the largest product for a series of 6 digits is 23520.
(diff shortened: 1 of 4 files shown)


largest-series-product.md
```diff
@@ -1,6 +1,8 @@
-For example, for the input `'0123456789'`, the largest product for a
-series of 3 digits is 504 (7 * 8 * 9), and the largest product for a
-series of 5 digits is 15120 (5 * 6 * 7 * 8 * 9).
+For example, for the input `'1027839564'`, the largest product for a
+series of 3 digits is 270 (9 * 5 * 6), and the largest product for a
+series of 5 digits is 7560 (7 * 8 * 3 * 9 * 5).
+
+Note that these series are only required to occupy *adjacent positions* in the original string; the digits need not be *numerically consecutive*.
```
petertseng: wrap? Considering that the above lines are wrapped

series.md
```diff
@@ -1,13 +1,15 @@
...
 
-- 0123
-- 1234
+- 4914
+- 9142
 
 And if you ask for a 6-digit series from a 5-digit string, you deserve
 whatever you get.
+
+Note that these series are only required to occupy *adjacent positions* in the original string; the digits need not be *numerically consecutive*.
```
petertseng: wrap? Considering that the above lines are wrapped

largest-series-product.md
kytrinyx: Yeah, wrap.

series.md
kytrinyx: ditto.

petertseng:
wrapped to 72

PR #201, opened 2016-03-13
Fix expected output of an anagram test

tgecho:
It seems that the intent is for passing candidates to be returned unchanged.

### DIFF
--- a/anagram.json
+++ b/anagram.json
@@ -78,7 +78,7 @@
             "description": "detects anagrams using case-insensitve possible matches",
             "subject": "orchestra",
             "candidates": ["cashregister", "Carthorse", "radishes"],
-            "expected": ["carthorse"]
+            "expected": ["Carthorse"]
         },
         {
             "description": "does not detect a word as its own anagram",


petertseng:
I agree with this; seems in line with most languages' tests (at least the ones where I've done the exercise...)

Go is one exception I'm aware of, it always outputs lowercase (like how the json file is before this PR).

kytrinyx:
Yeah, agreed. Thanks.

petertseng:
> Go is one exception I'm aware of, it always outputs lowercase

It will no longer be if https://github.com/exercism/xgo/pull/694.

behrtam:
Okay, after that change. Do we still need "detects anagrams case-insensitively" and "detects anagrams using case-insensitve possible matches"?

```diff
         {
             "description": "detects anagrams case-insensitively",
             "subject": "Orchestra",
             "candidates": ["cashregister", "Carthorse", "radishes"],
             "expected": ["Carthorse"]
         },
         {
             "description": "detects anagrams using case-insensitive subject",
             "subject": "Orchestra",
             "candidates": ["cashregister", "carthorse", "radishes"],
             "expected": ["carthorse"]
         },
         {
              "description": "detects anagrams using case-insensitve possible matches",
              "subject": "orchestra",
              "candidates": ["cashregister", "Carthorse", "radishes"],
 -            "expected": ["carthorse"]
 +            "expected": ["Carthorse"]
          },
```

petertseng:
> Do we still need "detects anagrams case-insensitively" and "detects anagrams using case-insensitve possible matches"?

I think I saw a solution (I don't know when I saw it, so I can't find it to link to it) that only accepted capitalised candidates if the subject is also capitalised, so the latter two cases would find that. Don't know how common it is.

Issue #190, opened 2016-03-07
crypto-square implementations disagree with README in many tracks


matthewmorgan:
Taken from [xjava](https://github.com/exercism/xjava/issues/55#issuecomment-193389759):

Chunk sizes are wrong: The README gives several examples of how to deal with strings of text that can be made into perfect ( ie n X n ) squares, and those that can be made into imperfect ( r X c ) squares. In all the README examples, the imperfect squares are r X c with the additional constraint that r >= c.

As documented by @jtigger [here](https://github.com/exercism/exercism.io/issues/2432#issuecomment-118467422) most of the tracks take a 26-letter string and chunk it in to 5, 5, 4, 4, 4, 4. This is not the behavior specified in the README, and seems like a error from early on that got propagated to many tracks.

There's no json file for this in x-common, but it appears this needs to be changed in many places, not just on the Java track.

If I'm missing something please let me know, but it seems like this is a good case for common test data.

kytrinyx:
I remember having trouble with inconsistencies on this before, but I've never taken the time to go through everything and tidy it up.

I'd love to have a canonical list of test cases (`crypto-square.json` would be great).

kytrinyx:
Pinging @exercism/track-maintainers in case you're not watching the repo :)

ErikSchierboom:
Just to be clear, in the linked post it shows show variants of how tracks implement the algorithm for the strings "Madness, and then illumination." and "Vampires are people too!". What is the correct output for those strings?

petertseng:
So the question is whether the imperfect squares (which have r rows and c columns) should have `r >= c` or `c >= r`?

Well, I don't care how we decide as long as we all reach some consensus, but if I look at the [.yml file for the exercise](https://github.com/exercism/x-common/blob/master/crypto-square.yml#L4), it lists source of http://users.csc.calpoly.edu/~jdalbey/103/Projects/ProgrammingPractice.html, and the example on that page is a `c >= r` example, as it says:

> so it is written into a rectangle with 7 rows and 8 columns. 

This is supported by the following sentence in the README:

> A message between 5 and 8 characters long should use a rectangle 3 characters wide.

So the message "abcde" fills in this way (`c >= r`):

    abc
    de

Rather than this way (`r >= c`):

    ab
    cd
    e

The following portion also supports `c >= r`

> Which has length: 30
> And splits into 5 6-character rows: 

---

Finally, I found a contradictory sentence in the README:

[...]

PR #204, opened 2016-03-14
Edit README to clarify exercise instructions (crypto-square)

matthewmorgan:
This PR is to address #190 .  

@petertseng asked if the output should be chunked.  My opinion, shared with @jtigger,  is that it should be, as it was in the original exercise that inspired this problem, and chunking the output fits aesthetically with the idea of creating a visual 'square' which is what the original users of this method would have done.

We're hoping we can get some consensus on the content of the README, then as a next step add a `config.json` for this exercise.  Thanks all!

### DIFF
--- a/crypto-square.md
+++ b/crypto-square.md
@@ -1,4 +1,7 @@
-The input is first normalized: The spaces and punctuation are removed
+Write a program that, given an English text, outputs the encoded version
+of that text.
+
+First, the input is normalized: the spaces and punctuation are removed
 from the English text and the message is downcased.
 
 Then, the normalized characters are broken into rows.  These rows can be
@@ -6,13 +9,19 @@ regarded as forming a rectangle when printed with intervening newlines.
 
 For example, the sentence
 
-> If man was meant to stay on the ground god would have given us roots
+> If man was meant to stay on the ground, god would have given us roots.
 
-is 54 characters long.
+is normalized to:
 
-Broken into 8-character columns, it yields 7 rows.
+> ifmanwasmeanttostayonthegroundgodwouldhavegivenusroots
 
-Those 7 rows produce this rectangle when printed one per line:
+The plaintext should be organized in to a rectangle.  The size of the
+rectangle (`r x c`) should be decided by the length of the message,
+such that `c >= r` and `c - r <= 1`, where `c` is the number of columns
+and `r` is the number of rows.
+
+Our normalized text is 54 characters long, dictating a rectangle with
+`c = 8` and `r = 7`:
 
 ```plain
 ifmanwas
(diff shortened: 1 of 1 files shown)


petertseng:
I think these are great changes because they make clear the that the decision to chunk the output is a separate step from actually figuring out what the letters in the output should be. Further it explains the rationale for the chunk size that we chose. Thanks for taking this on.

Do you anticipate that removing the examples of 4 characters -> 2x2, 81 characters -> 9x9, 5-8 characters -> 3 columns 2 rows will cause students to have a hard time figuring out the best way to compute them? From the original text, it was quick to see that finding the smallest number whose square is >= the message size would do it (alternatively, `ceil(sqrt(message_size))`). Is the new challenge a good or bad thing?

It seems that "print out in chunks of `r` letters" would misalign some letters. If you look at the letters in the README, I can decode "If man was meant to..." and then after that the letters get mistaligned with "etay on tho ground gad would huve given srootss". The page at http://users.csc.calpoly.edu/~jdalbey/103/Projects/ProgrammingPractice.html seems to have a slightly different chunking of "imtgdvs fearwer mayoogo anouuio ntnnlvt wttddes aohghn  sseoau" - the difference being that the last two chunks are both of size 6, rather than 7 and 5. That would stack up to form:

```
imtgdvs
fearwer
mayoogo
anouuio
ntnnlvt
wttddes
aohghn 
sseoau
```

from which the letters do get aligned in a readable manner So we can either match their scheme (which would require adjusting our explanation slightly to match) or stick with this and clarify that some letters will get shifted around.

matthewmorgan:
@petertseng you're absolutely right about the letter alignment.  Good catch.

It appears the examples in the original (calpoly) exercise distribute the `n` unfilled spaces across the final `n` rows of the square.  Do I have that right?  If so, then we should change the README to reflect that.

As far as omitting the extra examples, I think that one well-written example is OK. We just have to get this one there!

jtigger:
Yup, and the .json will inform unit tests that should help convey these corner cases.

(sent from mobile)

petertseng:
> It appears the examples in the original (calpoly) exercise distribute the n unfilled spaces across the final n rows of the square. Do I have that right?

I agree with this.

crypto-square.md
```diff
@@ -27,43 +36,31 @@ sroots
...
-
-The size of the square (number of columns) should be decided by the
-length of the message.
-
-If the message is a length that creates a perfect square (e.g. 4, 9, 16,
-25, 36, etc), use that number of columns.
+Output the encoded text in chunks.  Phrases that fill perfect squares
+`(r X r)` should be output in `r`-length chunks separated by spaces.
+Imperfect squares should distribute the `n` empty spaces across the last
+`n` rows of the square by making the length of those final `n` rows `c-1`.
```
petertseng: `r - 1 ` here or `c - 1`? I *think* it's `r - 1` but check me on that

matthewmorgan:
Yep.  I just noticed that row and column counts are swapped in the 'stacked' chunks.

Thanks again @petertseng !

petertseng:
> Yep. I just noticed that row and column counts are swapped in the 'stacked' chunks.

Ah that's right, that's why I too got a bit mixed up. Probably best if we keep just one value of `r` throughout the readme (as in, what I see now is right, with `r - 1` there).

So personally I would be :+1: on this. If anyone else wishes to chime in?

kytrinyx:
This is much clearer, thanks so much for sorting through this! I'm happy with this as well.

I'm about to get on a plane for a ridiculously long trip. If nobody else has chimed in by the time I get to my destination (40 hours), then I'll go ahead and merge it then.

jtigger:
Hey @kytrinyx, do you have a process by which you stage up updates to existing exercises such that the README and the various track implementations deploy together?

Do we need to create issues in each track that will now have a delta from the README?

kytrinyx:
Sorry it took so long to get to this... I've been traveling and have barely checked github (or email) for the past 5 days.

I don't have a process; sounds like I probably should, though. See https://github.com/exercism/discussions/issues/2 for a brand new discussion about it.

If one of you comes up with a suggested text for the issue, I can cross-post it to all the tracks that have the crypto-square exercise in it.

matthewmorgan:
At this point I'm planning on creating a `crypto-square.json` to flesh out this PR.  @jtigger and I discussed yesterday and it seems like it would be a good idea.

kytrinyx:
That sounds excellent!

IanWhitney:
I'm going to start implementing crypto-square for Rust soon. Can this PR be merged? And @matthewmorgan, how is the json file coming along? Anything I can do to help?

kytrinyx:
Yeah, let's merge this and have a separate PR for the JSON file.

Issue #198, opened 2016-03-12
Improve Sum of Multiples exercise


kytrinyx:
See the discussions in https://github.com/exercism/exercism.io/issues/2654 and https://github.com/exercism/xgo/issues/256

There's been a bit of confusion due to the whole "defaults" thing (use 3 and 5 as defaults if no arguments are passed). I made that up sometime at the dawn of Exercism, and I regret it.

I would much rather have a cleaner problem statement with no defaults.

In order to fix this we would need to

- [x] clarify the README
- [x] create a canonical set of shared test inputs/outputs in sum-of-multiples.json
- [x] fix all of the existing implementations

alebaffa:
Hi, is this exercise still under revision?

kytrinyx:
Yeah, if you look at all the related links in this thread, the issue needs to stay open as long as there are green badges:

<img width="687" alt="screen shot 2016-08-07 at 9 11 02 am" src="https://cloud.githubusercontent.com/assets/276834/17463230/e56f95ca-5c7e-11e6-84ec-ffc782f97246.png">

NobbZ:
I will try to tackle xlfe this weekend.

NobbZ:
I took the freedom to tackle `xlfe` as well as `xclojure`.

But I am far from beeing an expert about the various lisp dialects and there different idioms, so building up the new testsuites was a very straightforward mechanical process, for sclojure, I even used 2 replace by regexp of my editor (against the lfe version) and I was ready to go in theory, but I fine tuned a little bit, because in the original suite there was used a `zero?` predicate instead of checking equality to `0`…

So both should be reviewed by track maintainers or track-VIPs at least.

NobbZ:
Both got merged so it should be done now.

petertseng:
Great job, all involved

PR #206, opened 2016-03-16
Add sum-of-multiples.json

petertseng:
This is a synthesis of the existing tests in current tracks.

A short attempt to discern the history of the various tests:

Almost all tracks have the [3, 5] cases in common, as well as the
[7, 13, 17] and [43, 47] case. Indeed, these cases were present since
the first known sighting of sum-of-multiples at
https://github.com/exercism/exercism.io/commit/607be68494d30487c001303b6dc56c2b7b975409.

The [3, 5] to 100 case was added after discussion in
https://github.com/exercism/todo/issues/20 and
https://github.com/exercism/exercism.io/issues/2486. Most tracks had
adopted them at the time of this PR.

The [4, 6] and [5, 6, 8] cases arose from discussion in
https://github.com/exercism/exercism.io/issues/2341 - some tracks, but
not all, had adopted these tests at the time of this PR.

The [5, 25] case was a Haskell-specific case added in
https://github.com/exercism/xhaskell/pull/50.

The [] and [1] cases were Go-specific cases that have been present ever
since the exercise was added to the Go track in
https://github.com/exercism/xgo/pull/53.

Cases found in various tracks that were not included:

The [1] case of Go was originally a [1, 1] case but this is a conscious
decision not to test duplicate factors.

There were [0] and [0, 1] cases added to the Python track in
https://github.com/exercism/xpython/pull/149 but this is a conscious
decision not to test 0 as a factor.

This is the second subtask of #198.

Closes https://github.com/exercism/todo/issues/154

### DIFF
--- a/sum-of-multiples.json
+++ b/sum-of-multiples.json
@@ -1,0 +1,64 @@
+{
+    "cases": [
+        {
+            "factors": [3, 5],
+            "limit": 1,
+            "expected": 0
+        },
+        {
+            "factors": [3, 5],
+            "limit": 4,
+            "expected": 3
+        },
+        {
+            "factors": [3, 5],
+            "limit": 10,
+            "expected": 23
+        },
+        {
+            "factors": [3, 5],
+            "limit": 100,
+            "expected": 2318
+        },
+        {
+            "factors": [3, 5],
+            "limit": 1000,
+            "expected": 233168
+        },
+        {
+            "factors": [7, 13, 17],
+            "limit": 20,
+            "expected": 51
+        },
+        {
+            "factors": [4, 6],
+            "limit": 15,
+            "expected": 30
+        },
+        {
+            "factors": [5, 6, 8],
+            "limit": 150,
+            "expected": 4419
+        },
+        {
+            "factors": [5, 25],
+            "limit": 51,
+            "expected": 275
+        },
+        {
+            "factors": [43, 47],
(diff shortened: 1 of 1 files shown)


petertseng:
Discussion is welcomed on whether to include tests for:

* Having duplicates in the list of factors
* Having 0 in the list of factors

petertseng:
Verified no mistakes in this file:

```ruby
require 'json'

cases = JSON.parse(IO.read('sum-of-multiples.json'))

cases['cases'].each { |c|
  expected = c['expected']
  factors = c['factors']
  limit = c['limit']
  observed = (1...limit).select { |n|
    factors.any? { |f| n % f == 0 }
  }.reduce(0, :+)

  raise "Nope on #{c}, got #{expected}" if observed != expected
}

puts 'pass'
```

kytrinyx:
I think that it's the right move to not test duplicate factors. Also, 0 is a really weird case, with the identify for multiplication being 1 and I'd rather not complicate this exercise with it.

I think that this PR is good as is, and would love a second (and third) opinion.

petertseng:
Thanks for review. Just FYI I'm changing the commit message (the json file contents haven't changed) so as to reference https://github.com/exercism/todo/issues/20, as I missed it in my history exploration

kytrinyx:
Excellent! Let me know if you run into any weirdness with blazon.

PR #207, opened 2016-03-18
Tree Building grammar and capitalization fixes

ryanplusplus:
- Use correct form of "its"
- Capitalize "id" as "ID"
- "an parent" => "a parent"


kytrinyx:
Thanks! I can't believe I keep missing these its/it's issues.

PR #209, opened 2016-03-23
sum-of-multiples: Mark as no longer under revision

petertseng:
This partially reverts 1f0ad90774e3922558c39f8aa3ccd81b83b1c99f,
which partially reverted 89308fd925b604e84fb652d8b72ee74ca231474b.

The default behavior was added in #157 as a result of confusion in
https://github.com/exercism/exercism.io/issues/2654. Since then, we have
decided the way forward is to remove the default behavior altogether, as
discussed in https://github.com/exercism/x-common/issues/198.

This commit should be held until all tracks have removed their
expectations of default behavior.

### DIFF
--- a/sum-of-multiples.md
+++ b/sum-of-multiples.md
@@ -5,15 +5,3 @@ The sum of these multiples is 78.
 
 Write a program that can find the sum of the multiples of a given set of
 numbers.
-
-This exercise is currently under revision.
-You are still free to do this exercise and submit solutions.
-Depending on whether this track has been updated recently,
-the tests may differ between this track and other tracks.
-
-In most tracks, the set of numbers must be explicitly provided.
-In some other tracks, if no set of numbers is given, it defaults to 3 and 5.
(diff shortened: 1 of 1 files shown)


petertseng:
Hello, it's been a month. I think that's been quite ample notice. In a few hours, I would like to:

* Submit PRs to the remaining tracks myself
* Not wait for them to merge those PRs, instead just merge this PR when I submit those

petertseng:
Updated plan: Instead, modify this PR to say "This exercise is currently undergoing change, some tracks have defaults, some don't, if you want to help look at #198", then I'll merge.

petertseng:
That was done in https://github.com/exercism/x-common/pull/236.

IanWhitney:
@petertseng If I'm reading your comments correctly, we can merge this. Yes?

petertseng:
No, I should have clarified. This should only be merged after all tracks complete - currently clojure, cpp, and lfe have yet to do

PR #210, opened 2016-03-26
clock: add more test cases to canonical test data

kytrinyx:
In the Go track I've seen a lot of people who only account for a clock
overflowing once in either direction. As the README describes the
problem, any overflow should be corrected, even if it accounts for
several days.

The second problem that I've seen a lot is people who adjust in the
display method, which means that the clocks would fail the equality
check. We weren't verifying equality against overflow, however, so we
didn't catch this.

I also removed two redundant test cases.

### DIFF
--- a/clock.json
+++ b/clock.json
@@ -18,22 +18,10 @@
             "expected": "08:00"
          },
          {
-            "description": "on the hour",
-            "hour": 9,
-            "minute": 0,
-            "expected": "09:00"
-         },
-         {
             "description": "past the hour",
             "hour": 11,
             "minute": 9,
             "expected": "11:09"
-         },
-         {
-            "description": "past the hour",
-            "hour": 11,
-            "minute": 19,
-            "expected": "11:19"
          },
          {
             "description": "midnight is zero hours",
@@ -48,6 +36,12 @@
             "expected": "01:00"
          },
          {
+            "description": "hour rolls over continuously",
+            "hour": 100,
+            "minute": 0,
+            "expected": "04:00"
+         },
+         {
             "description": "sixty minutes is next hour",
             "hour": 1,
             "minute": 60,
@@ -60,10 +54,28 @@
             "expected": "02:40"
          },
          {
+            "description": "minutes roll over continuously",
+            "hour": 0,
+            "minute": 1723,
+            "expected": "04:43"
+         },
+         {
             "description": "hour and minutes roll over",
             "hour": 25,
             "minute": 160,
             "expected": "03:40"
+         },
+         {
+            "description": "hour and minutes roll over continuously",
+            "hour": 201,
+            "minute": 3001,
+            "expected": "11:01"
+         },
+         {
+            "description": "hour and minutes roll over to exactly midnight",
+            "hour": 72,
+            "minute": 8640,
+            "expected": "00:00"
          },
          {
             "description": "negative hour",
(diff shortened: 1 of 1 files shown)


kytrinyx:
When this gets pulled in, I'll notify all of the tracks using blazon. Here's the suggested text for the issue:

```
clock: updated canonical test data

While providing feedback on this problem, it became clear that people were submitting solutions with bugs that our tests weren't catching. There were two types of bugs:

- insufficient handling of overflow. A clock wraps around, but our tests were only checking for a single day's worth of wrapping; the new test suite adds tests for wrapping multiple days.
- modifying the clock in the display method. Clocks should be equal, but instead of actually making the clocks equal, many solutions were adjusting the values in the display method. This adds new test cases to ensure a more robust equality.

If the test suite in this track is generated, then regenerating the exercise should be sufficient to close this issue. If not, then someone will need to manually update the tests to include the edge cases.
```

clock.json
```diff
@@ -259,6 +351,78 @@
...
+               "minute": 32
+            },
+            "clock2": {
+               "hour": -12,
+               "minute": -268
+            },
+            "expected": true
+         },
+         {
+            "description": "clocks with negative hours and minutes that wrap",
```
petertseng: this doesn't appear to be negative

clock.json
```diff
@@ -249,6 +293,54 @@
...
+               "minute": 40
+            },
+            "clock2": {
+               "hour": -2,
+               "minute": 40
+            },
+            "expected": true
+         },
+         {
+            "description": "clocks with negative hour that wraps",
```
petertseng: could you explain the difference between this and the previous case "clocks with negative hour"? It doesn't look like they are different enough to warrant a separate case since both negative values are < -24, unless this is testing something different

clock.json
```diff
@@ -60,12 +54,30 @@
             "expected": "02:40"
          },
          {
+            "description": "minutes roll over continuously",
+            "hour": 0,
+            "minute": 1723,
+            "expected": "05:13"
```
petertseng: 4:43 instead?

Since 1723 - 1440 = 283

clock.json
kytrinyx: Ah, no, I think I meant to do something to wrap exactly one day, so -31 instead of -7. Let me push up a fix.

clock.json
kytrinyx: Oops, both of those were supposed to be negative. It's harder to test these--I should have regenerated the Go tests using this; I'll go do that now.

clock.json
kytrinyx: Yepp, I've fixed that.

kytrinyx:
I generated the Go test suite locally to test the changes this time.

petertseng:
:+1:  from me, though you've thought about this exercise more than I have. That means that while I am able to reason about whether the tests added by this PR make sense, I may miss tests that are *not* added by this PR that maybe should be. Luckily that's not so terrible since they can just be added in another PR.

I did have one thought: Is there a test for adding 0 to a clock, and should there be?

kytrinyx:
There is no such test, and I think that it's a good case to have.

kytrinyx:
I've added the missing test. I regenerated the Go test suite and verified that it generates a passing set of tests.

petertseng:
:+1: , made a quick clock implementation and tested it against the json file, seems good

kytrinyx:
Ok, thanks for being a second pair of eyes on this!

PR #213, opened 2016-03-28
Update queen-attack test expectations

IanWhitney:
Fixing my copy & paste errors that @petertseng noticed in https://github.com/exercism/x-common/pull/211

### DIFF
--- a/queen-attack.json
+++ b/queen-attack.json
@@ -37,7 +37,7 @@
       "black_queen": {
         "position": "(2,5)"
       },
-      "expected": false
+      "expected": true
     },
     {
       "description": "can attack on first diagonal",
@@ -47,7 +47,7 @@
       "black_queen": {
         "position": "(0,4)"
       },
-      "expected": false
+      "expected": true
     },
     {
       "description": "can attack on second diagonal",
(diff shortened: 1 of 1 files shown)


kytrinyx:
This looks good to me. Would you mind updating the commit message to have `queen-attack` in it? That helps immeasurably when digging into the history.

queen-attack.json
```diff
@@ -67,7 +67,7 @@
       "black_queen": {
         "position": "(1,1)"
       },
-      "expected": false
+      "expected": true
```
petertseng: I believe the coordinates will need updating here if we want this to be true - (2, 4) can't attack (1, 1)

queen-attack.json
```diff
@@ -77,7 +77,7 @@
       "black_queen": {
         "position": "(5,5)"
       },
-      "expected": false
+      "expected": true
```
petertseng: Coordinates probably need to be changed here too - (2, 4) can't attack (5, 5)

IanWhitney:
You'd think it would be simple to write a dang json file. Apparently not. Fixing those two errors that @petertseng spotted.

IanWhitney:
Ok. All diagonal tests now look good to me.

petertseng:
Could you check real quick whether the change got pushed right? Still looks the same to me

IanWhitney:
Add Git to the things I can't get right today.

kytrinyx:
@IanWhitney -- would you mind tweaking the commit subject line? It sounds like you're having a chaotic day, and I'm happy to amend the commit message if you've got too much stuff going on. Just say the word.

petertseng:
OK from me on the json content, so it'll be ready to go when getting the commit message sorted.

IanWhitney:
Commit message changed. It's not that today was busy, but I'm traveling tomorrow. After that I'll be back at work and without these nice 3-4 hour stretches to think about test ordering, etc.

kytrinyx:
Thanks! Good luck with travel (I know how that can go!)

PR #214, opened 2016-03-30
Split queen-attack tests into creation and attacking

IanWhitney:
As discussed here:
https://github.com/exercism/x-common/pull/211#issuecomment-202596536


kytrinyx:
I think this makes sense, thanks!

PR #219, opened 2016-04-07
rna-transcription: remove rna->dna cases from canonical data

kytrinyx:
There is no biological basis for RNA->DNA transcription.

A couple of people suggested that we could rephrase this and make the
question _Given this RNA, what DNA sequence could produce it?_, but
it seemed simpler to simply remove it.

There is also the question of whether or not having the reverse
transcription helps guide people to a better/more generic solution.
I think that this is one place where we could have the discussion on the
website about hard-coding vs not.

See https://github.com/exercism/x-common/issues/148

### DIFF
--- a/rna-transcription.json
+++ b/rna-transcription.json
@@ -14,7 +14,7 @@
    "cases": [
       {
          "description": "rna complement of cytosine is guanine",
-         "dna" : "C", 
+         "dna" : "C",
          "expected": "G"
       },
       {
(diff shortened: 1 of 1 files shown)


petertseng:
Data change looks good (everything with "rna" key removed, nothing with "dna" key removed).

Commit message: you say "There is no biological basis for DNA->RNA transcription." but I think you meant to say "There is no biological basis for RNA->DNA transcription."

> There is also the question of whether or not having the reverse transcription helps guide people to a better/more generic solution.

Oh now this is interesting! Though I imagine there will be good discussion to be had for this problem anyway (did the student hard-code each pair, or are the pairs in some sort of dictionary that maps from DNA to RNA equivalent?). There are probably other problems that check two-way correspondences (bracket-pairs might be considered one, rail-fence-cipher as well)

kytrinyx:
> I think you meant to say

Yes, totally. I actually initially wrote that, and then my brain was all _no, wait, that's backwards_. D'oh.

I've fixed the commit message.

IanWhitney:
I guess we should follow biology here!

petertseng:
Commit message looks good now, and I've said what I needed to say on the discussion points

kytrinyx:
Thanks. Merging!

PR #220, opened 2016-04-07
Reorder Sieve tests

IanWhitney:
This comes from a discussion we had in the xrust track

https://github.com/exercism/xrust/pull/97

Typically tests are ordered from the simplest case to the most
complex/edge-case. Putting the simplest test first lets students get a
passing test faster.

I think this re-ordered set of Sieve tests follows
this standard. The simplest test -- finding no primes -- is now first.
Followed by finding just 1 prime, then a few primes, then a lot of
primes.

### DIFF
--- a/sieve.json
+++ b/sieve.json
@@ -1,9 +1,24 @@
 { 
     "cases": [
         {
+            "description": "no primes under two",
+            "limit": 1,
+            "expected": []
+        },
+        {
+            "description": "find first prime",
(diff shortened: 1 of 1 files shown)


petertseng:
Seems good to me, though I was a party to that discussion already

kytrinyx:
Yeah, this looks like a more reasonable ordering.

PR #227, opened 2016-04-14
sublist: Add JSON test data

callahat:
The test data is based off the python tests for the sublist example:
https://github.com/exercism/xpython/blob/master/exercises/sublist/sublist_test.py

### DIFF
--- a/sublist.json
+++ b/sublist.json
@@ -1,0 +1,92 @@
+{
+	"#": [
+		"Lists are ordered and sequential.",
+		"https://en.wikipedia.org/wiki/List_%28abstract_data_type%29",
+		"",
+		"Depending on your language, there may need to be some translation",
+		"to go from the JSON array to the list representation.",
+		"The expectation can be used to generate an expected value",
+		"based on your implementation (such as a constant 'EQUAL', 'SUBLIST', etc.)."
+	],
+	"cases": [{
+		"description": "empty lists",
+		"list_one": [],
+		"list_two": [],
+		"expectation": "equal"
+	}, {
+		"description": "empty list within non empty list",
+		"list_one": [],
+		"list_two": [1, 2, 3],
+		"expectation": "sublist"
+	}, {
+		"description": "non empty list contains empty list",
+		"list_one": [1, 2, 3],
+		"list_two": [],
+		"expectation": "superlist"
+	}, {
+		"description": "list equals itself",
+		"list_one": [1, 2, 3],
+		"list_two": [1, 2, 3],
+		"expectation": "equal"
+	}, {
+		"description": "different lists",
+		"list_one": [1, 2, 3],
+		"list_two": [2, 3, 4],
+		"expectation": "unequal"
+	}, {
+		"description": "false start",
+		"list_one": [1, 2, 5],
+		"list_two": [0, 1, 2, 3, 1, 2, 5, 6],
+		"expectation": "sublist"
+	}, {
+		"description": "consecutive",
+		"list_one": [1, 1, 2],
+		"list_two": [0, 1, 1, 1, 2, 1, 2],
+		"expectation": "sublist"
+	}, {
+		"description": "sublist at start",
+		"list_one": [0, 1, 2],
+		"list_two": [0, 1, 2, 3, 4, 5],
+		"expectation": "sublist"
+	}, {
+		"description": "sublist in middle",
+		"list_one": [2, 3, 4],
+		"list_two": [0, 1, 2, 3, 4, 5],
+		"expectation": "sublist"
+	}, {
+		"description": "sublist at end",
+		"list_one": [3, 4, 5],
+		"list_two": [0, 1, 2, 3, 4, 5],
+		"expectation": "sublist"
+	}, {
+		"description": "at start of superlist",
+		"list_one": [0, 1, 2, 3, 4, 5],
+		"list_two": [0, 1, 2],
+		"expectation": "superlist"
+	}, {
+		"description": "in middle of superlist",
+		"list_one": [0, 1, 2, 3, 4, 5],
+		"list_two": [2, 3],
+		"expectation": "superlist"
+	}, {
+		"description": "at end of superlist",
+		"list_one": [0, 1, 2, 3, 4, 5],
+		"list_two": [3, 4, 5],
+		"expectation": "superlist"
+	}, {
+		"description": "first list missing element from second list",
+		"list_one": [1, 3],
+		"list_two": [1, 2, 3],
+		"expectation": "unequal"
+	}, {
+		"description": "second list missing element from first list",
+		"list_one": [1, 2, 3],
+		"list_two": [1, 3],
+		"expectation": "unequal"
+	}, {
+		"description": "order matters to a list",
+		"list_one": [1, 2, 3],
+		"list_two": [3, 2, 1],
+		"expectation": "unequal"
+	}]
+}


IanWhitney:
Thanks!

I did a quick survey of the other languages that implement this problem and have some questions.

- Some languages check non-numeric sets ([Rust](http://exercism.io/exercises/rust/sublist), for example). I think that should be part of the standard test suite, but others might disagree.
- Is this API one we want to continue? Most languages will do something like this

```elixir
assert Sublist.compare([4,3,2],[5,4,3,2,1]) == :sublist
```

Where some method returns compares two lists and returns a string/constant/symbol/etc.

But [Lua does something different](http://exercism.io/exercises/lua/sublist) where the comparison returns a boolean.

I think the boolean return makes more sense here, honestly. But, not having done the problem myself, I could be missing something.

callahat:
I don't have strong opinions of the checked of nonnumeric values as well.

I do like the checking for a boolean instead of using a constant or other magic value.

IanWhitney:
My API question was about whether or not we expect students to return magic strings or booleans. This can be driven by the JSON, since some language tracks can automatically generate their tests from the JSON.

If we want to encourage boolean-returning implementations, then the json should be updated to look something like this:

```json
{
  "#": [
  ],
  "sublist": {
    "description": "",
    "list_one": [1,2,3],
    "list_two": [1,2,3,4,5],
    "expected": true
  },
  "equal": {
    "description": "",
    "list_one": [1,2,3],
    "list_two": [1,2,3],
    "expected": true
  },
  "superlist": {
    "description": "",
    "list_one": [1,2,3,4,5],
    "list_two": [1,2,3],
    "expected": true
 }
}
```
By breaking out the expected methods (superlist, sublist, equal) and declaring that we expect them to return booleans, we suggest to track maintainers that their tests should look for booleans, not magic strings.

kytrinyx:
To extend @IanWhitney's suggestion, we would have three sections in the JSON, one for each function, and each section would have multiple tests. I think.

petertseng:
Yeah there are several approaches for an API here. One approach taken by languages with variant types is that you might have a type that can take on the four values `Sublist, Superlist, Equal, Unrelated` and a single function that has to return one of the four values. It seemed to make sense in [Haskell](https://github.com/exercism/xhaskell/blob/master/exercises/sublist/example.hs) and [Rust](https://github.com/exercism/xrust/blob/master/exercises/sublist/example.rs). That might make less sense for a language that doesn't have variant types. I wonder how the tests should be such that both styles are possible.

IanWhitney:
Wait...`sublist`, `superlist`....Is this exercise a duplicate of Custom Set?

callahat:
For Custom Set, the order of the members does not matter. For this exercise, it does. From one of the examples in Custom Set:
https://github.com/exercism/x-common/blob/master/custom-set.json#L37

kytrinyx:
Is there anything we can do to help people understand the difference between lists and sets (ordered vs unordered, can contain duplicates vs all unique)?

callahat:
The readme could be updated to point out this difference? Or perhaps a definition of a 'list' and 'set' in those readme files?

callahat:
@IanWhitney what about something like following? According to the readme, several of the 

```
{
  "#": [],
  "cases": [{
    "description": "empty lists",
    "list_one": [],
    "list_two": [],
    "equal": "true",
    "sublist": "true",
    "superlist": "true"
  }, {
    "description": "empty list within non empty list",
    "list_one": [],
    "list_two": [1, 2, 3],
    "equal": "false",
    "sublist": "true",
    "superlist": "false"
  },
...
  ]
}
```

IanWhitney:
@callahat 

My guess is that will not work with the code that some tracks use to automatically generate tests.

callahat:
Could you elaborate?

IanWhitney:
I have not worked on a track that auto-generates tests, so I'm not positive. But I'm guessing that the test-generation code has an expectation that the json looks like

```json
"method1": {
  "cases":  [ 
    {
    "condition": "",
    "expectation":"",
    }
  ]
},
"method2": {
  "cases":  [ 
    {
    "condition": "",
    "expectation":"",
    }
  ]
}
```

Whereas your suggestion was

```json
  "cases":  [
    {
     "condition": "",
     "method1": "",
     "method2": ""
    }
  ]
}
``` 

That might work for tracks, I don't know. But it doesn't match the standard I've seen ([example](https://github.com/exercism/x-common/blob/master/alphametics.json), [another](https://github.com/exercism/x-common/blob/master/acronym.json))

Cohen-Carlisle:
I think the README is the source of all this confusion.

We've mentioned we don't like magic strings, but the README explicitly asks for a single function that indicates the relationship between A and B. (equal, sublist, superlist, none of the previous).

On the other hand, the simplest implementation of sublist function would be to simply test if A is a sublist of B.
If A is a sublist of B, then B is a superlist of A, no?
And if A and B are sublists of each other, they are equal.
Lastly, if neither A nor B are sublists of each other, then you arrive at the "none of the previous" scenario.

README says:
```
Write a function that given two lists determines if the first list is
contained within the second list, if the second list is contained within
the first list, if both lists are contained within each other or if none
of these are true.

Specifically, a list A is a sublist of list B if by dropping 0 or more elements
from the front of B and 0 or more elements from the back of B you get a list
that's completely equal to A.

Examples:

 * A = [1, 2, 3], B = [1, 2, 3, 4, 5], A is a sublist of B
 * A = [3, 4, 5], B = [1, 2, 3, 4, 5], A is a sublist of B
 * A = [3, 4], B = [1, 2, 3, 4, 5], A is a sublist of B
 * A = [1, 2, 3], B = [1, 2, 3], A is equal to B
 * A = [1, 2, 3, 4, 5], B = [2, 3, 4], A is a superlist of B
 * A = [1, 2, 4], B = [1, 2, 3, 4, 5], A is not a superlist of, sublist of or equal to B

```
The sublists are straightforward enough. But starting here things get fuzzy:
 ```
A = [1, 2, 3], B = [1, 2, 3], A is equal to B
```
isn't straightforward. A is a sublist of B in addition to being equal, but I guess equality is taking precedence...
```
A = [1, 2, 3, 4, 5], B = [2, 3, 4], A is a superlist of B
```
but B is also a sublist of A, I guess the order matters (A is (a sublist of|a superlist of|equal to|none) B)
```
A = [1, 2, 4], B = [1, 2, 3, 4, 5], A is not a superlist of, sublist of or equal to B
```
True enough, but what are we supposed to be returning for this?

I think if we agree on what the README should be saying, it will make this PR easier.

**Comment by petertseng** [comment]
It's indeed the case that many solutions to this problem have used, as an internal function, a sublist function. They run A.sublist?(B) and B.sublist?(A) and based on what those two functions return, derive the relationship between A and B.

The function being tested is still a single function that determines the relationship (one of the four choices).

At least that's how I've seen it in Rust and Haskell, the two tracks in which I've worked this problem.

**Comment by callahat** [comment]
It does seem that a precedent had been set by 7 of the 8 language tracks which have this exercise. The majority do return essentially an enumerable value (not just boolean) that indicates their relationship. The most recently added track, Lua, returns true or false from one method call `is_sublist` as pointed out by @IanWhitney.

So it looks like decisions are needed on:
 * [] What should be returned (boolean or enumerable)
 * [] Rewording of the README to make the exercise more clear

Anything else?

**Comment by Cohen-Carlisle** [comment]
Well, since you are doing the PR, @callahat, what do _you_ think should be done?

**Comment by callahat** [comment]
I'd vote to stick with the precedent at this point. How do we come to a decision?

**Comment by Cohen-Carlisle** [comment]
I think that's perfectly acceptable. A different PR/issue should address the use of magic strings, if its deemed to be a problem, in my opinion. So unless there are any strong objections (cc: @IanWhitney, @petertseng)... let's move forward with returning magic strings here.

**Comment by callahat** [comment]
I think the magic string in the JSON could also be used to implement the actual expectation. IE, in Ruby it could reference a constant or enumerable-like thing. But that's more of a discussion for the individual tracks.

**Comment by petertseng** [comment]
> So unless there are any strong objections (cc: @IanWhitney, @petertseng)... let's move forward with returning magic strings here.

I'm not objecting, because:

> the magic string in the JSON could also be used to implement the actual expectation. IE, in Ruby it could reference a constant or enumerable-like thing.

Right, the simple fact that this is a string in JSON doesn't force any track's version of this problem to use strings. Tracks are free to take advantage of their language's features in whatever way makes sense.

To wit, both Haskell and Rust use their language's respective variant type ([Haskell's data](https://github.com/exercism/xhaskell/blob/master/exercises/sublist/example.hs#L4) and [Rust's enum](https://github.com/exercism/xrust/blob/master/exercises/sublist/example.rs#L2)) and not strings.

**Comment by IanWhitney** [comment]
Ok.

**Comment by kytrinyx** [comment]
Based on the discussion above it looks like there's only one more thing to clarify before this is ready to merge:

> Some languages check non-numeric sets (Rust, for example). I think that should be part of the standard test suite, but others might disagree. - @IanWhitney

**Comment by Cohen-Carlisle** [comment]
> Some languages check non-numeric sets (Rust, for example). I think that should be part of the standard test suite, but others might disagree. - @IanWhitney

Adding this requirement would be trivial in languages with dynamic typing... not sure if it would lead to more interesting code or just more boilerplate in Rust, Go, etc. For this reason, I lean towards not requiring this in the metadata... but I'm open to other opinions.

I'd also like ask if `unequal` is the language we want to use when list A is not equal to, a sublist of, or a superlist of list B. Perhaps instead `none` or some other alternative? While any language consuming this metadata is free to modify it, I'd like our metadata to be able to say:
> I know words, I have the best words.

**Comment by kytrinyx** [comment]
> I know words, I have the best words.

:) Yes, our metadata should have this as its slogan for sure.

Distinct? Disparate?

**Comment by petertseng** [comment]
Unrelated?

**Comment by kotp** [comment]
I think I love unrelated.

**Comment by Cohen-Carlisle** [comment]
I may have posted on a stack exchange site looking for a good word. :laughing: 
I'll report back on my (read: their) findings.

**Comment by callahat** [comment]
@kotp unrelated might be ambiguous; it might imply that the two things being compared are not actually both lists. 'Disparate' or 'dissimilar' might be more clear.

**Comment by kytrinyx** [comment]
`non-overlapping`?

**Comment by kotp** [comment]
The title reads 'list a and list b are unequal' so 'list a and list b are unrelated' can not imply that these are not lists.

But thinking about the word alone, with perhaps less distinctive wording as actually declaring them to be lists, I think unrelated holds less exclusivity than disparate.  Perhaps "totally unrelated" and "disparate" are pretty close friends.  But I think 'unrelated and disparate' are not.

You and I may be unrelated, yet we are both list... er... people.  "Unrelated" should bring up feelings of potential similarity, more connected than 'disparate'.  And more similar in characteristics than being fundamentally devoid of shared characteristics.

Unrelated has the distinct feeling to me of connected, perhaps similar in most ways, the potential to at least, on the surface, mistake them for 'cousins'.  But definitely the same species.

Disparate has a stronger feeling to it than unrelated, I think. "containing or made up of **fundamentally** different and often incongruous elements" (Merriam-Webster) though it can be as gentle as simply "different as compared together."

**Comment by Cohen-Carlisle** [comment]
I couldn't find a single term that describes when lists are not sublists/superlists/equal.

We could borrow from set terminology (as we already are) and describe the lists as either `disjoint` or `intersecting` (or perhaps some variation like `empty intersection` and `nonempty intersection`).

This would make the implementation more involved. And it would also draw more attention to the precedence of the return value.

Currently it is implicitly `equal > sublist/superlist`. If we adopt `disjoint` and `intersecting` our precedence would presumably be `equal > sublist/superlist > disjoint/intersecting` and would probably need to be explicit.

I'm in favor of either going all the way with the set-like returns or scaling it back to just detecting sublists.

**Comment by Cohen-Carlisle** [comment]
My idea hasn't gained any supporters. With that being the case, let's put it on the back burner and focus on getting this PR across the finish line. Do we have an agreement on what `unequal` should be changed to? I'm fine with pretty much anything :laughing:

**Comment by kytrinyx** [comment]
@Cohen-Carlisle I have been at a conference, and haven't looked at any PRs or discussions in the past 4 days. That said, I am not sure whether the disjoint, empty intersection, non-empty intersection thing is worth the complexity. It might be (I haven't done the exercise, so I don't have a lot of context for this one).

Either way, I think you're right: let's get this PR across the finish line.

The current suggestions to replace `unequal` are:

- `none`
- `distinct`
- `disparate`
- `unrelated`
- `dissimilar`
- `non-overlapping`

**Comment by IanWhitney** [comment]
I think `inequal` is the term I see most often in programming. That's also what's used on [wikipedia](https://en.wikipedia.org/wiki/Inequality_(mathematics)), if that means anything.

There's also [almost disjoint](https://en.wikipedia.org/wiki/Almost_disjoint_sets), which I mention not as a real suggestion but as a joke.

**Comment by kotp** [comment]
I :laugh: because I avoided that, due to thinking inequality perhaps had some nuance in regards to the problem domain.

**Comment by kytrinyx** [comment]
Almost-disjoint is hilarious (but agreed: no!)

**Comment by Insti** [comment]
non_intersecting

**Comment by Insti** [comment]
independent

**Comment by andypike** [comment]
How about `complement` from set theory: https://en.wikipedia.org/wiki/Complement_(set_theory). 

> In set theory, a complement of a set A refers to things not in (that is, things outside of) A.

I'm not sure it's clear unless you know about set theory, but I'll put it out there as a suggestion.

❤️

**Comment by Insti** [comment]
From the [sublist problem description](https://github.com/exercism/x-common/blob/master/sublist.md):

> Write a function that given two lists determines if the first list is contained within the second list, if the second list is contained within the first list, if both lists are contained within each other or if none of these are true.

Perhaps the alliterative[1] route: 

`Sublist, Superlist, Same, Separate.`

[1] I'm not trying to start a discussion about whether alliteration is a good practice.

**Comment by rsslldnphy** [comment]
My vote would be for `unequal`. I think `disjoint` is appropriate only if the two lists share *no elements* in common, and I think for a list to be a `complement` of another list it would have to contain *everything* that wasn't in the other list.

```
disjoint([1, 2], [3, 4]) // true
disjoint([1, 2], [2, 3]) // false

complement([1, 2], [3, 4]) // false
complement([1, 2], [3,4 ..] // true, assuming the domain is the positive integers, and [3, 4 ..] creates an infinitely long list
```

And as @Cohen-Carlisle pointed out, these options aren't all mutually exclusive: a list and its `complement` are also `disjoint`, but two `disjoint` lists aren't necessarily `complement`s. Two `disjoint` lists can have a `subset`, `superset` relationship, but they don't have to. Equal lists always do - unless you're talking about *strict* superset/subsets.

I think having separate, boolean returning predicate functions for each of the options would make this explicit to the person doing the exercise: for each test case, all the functions could be called, and the return values of each checked for the correct combination of trues and falses. It would remove any ambiguity around the order of precedence.

Oh, and thinking about it, it would also remove the need to worry about what to call the `unequal` case! Because you would just have an `equal` function, and expect that it returned `false` rather than dealing with the double-negative case of expecting `unequal` to return `true`. Although I guess most languages will already provide this equality checking function, so it doesn't make for that interesting an exercise.. :-/

OceansideBill:
Perhaps "parallel" - implies a type of relationship where neither has the same elements.

wobh:
TLDR: I vote for "disjoint"

In math, we have `set`s which are collections where the order doesn't matter and uniqueness does, and `tuple`s where order matters and uniqueness doesn't. The operations on each are distinct in important ways.

| set                                    | tuple                                  |
|----------------------------------------|----------------------------------------|
| `union : (set, set) → set`             | `concatenate : (tuple, tuple) → tuple` |
| `difference : (set, set) → set`        | `replace : (tuple, tuple) → tuple`     |
| `complement : (set, set) → set`        | `truncate: (tuple, posint) → tuple`    |
| `intersection : (set, set) → set`      | `search : (tuple, tuple) → posint`     |
| `member? : (t, set) → boolean`     | `find : (t, tuple) → posint`           |
| `equal? : (set, set) → boolean`    | `match : (tuple, tuple) → posint`      |
| `disjoint? : (set, set) → boolean` | `mismatch : (tuple, tuple) → posint`   |
| `empty? : (set) → boolean`          | `length` : (tuple) → posint`           |

The semantics of this function in this exercise are much more complex. We're asking for with four different outputs, it's a valid case even when the position of the arguments is swapped--i.e. this isn't a typical operation on tuples. We're also conflating comparative states of sets (`subset`, `superset`, `equal`, `disjoint`) with those of tuples (found subsequence).

I'm not fault-finding the exercise, I'm just making the case that we shouldn't over-rationalize using set descriptions to describe the comparison of tuples.

`sublist (tuple, tuple) → member { "disjoint", "equal", "subset", "superset" }`

IanWhitney:
unequal. do it. merge this sucker.

kytrinyx:
Yay. Thank you everyone for helping think through all of this.

PR #228, opened 2016-04-17
add pangram tests

### LINKED ISSUE #222: pangram: missing edge case
@helenst came across a solution that had a bug that the test suite didn't catch. It's addressed in https://github.com/exercism/xpython/pull/331

The solution worked by accident because 'the quick...' happens to be alphabetically greater than 'abcdef...'.

Here's the test case that was added:

```
def test_invalid_pangram(self):
    self.assertFalse(
        is_pangram('the quick brown fish jumps over the lazy dog'))
```

We need to:
- [x] update the canonical data - https://github.com/exercism/x-common/blob/master/pangram.json
- [x] draft a detailed issue
  - [x] describes the problem that was found
  - [x] describes the fix
  - [x] links to the canonical data (and mentions that it has been updated with the new test cases)
  - [x] suggests that the exercise should be fixed to include these test cases
  - [x] links to this specific issue (so we get a visual todo list)
- [x] submit the issue to the relevant tracks using [blazon](https://github.com/exercism/blazon)

Cohen-Carlisle:
this commit adds 3 total tests

- adds a non-pangram to resolve #222
- adds a pangram with underscores
- adds a non-pangram where some letters have been replaced by numbers

the latter two are designed to detect solutions that rely on counting
the number of word characters (which include numbers and underscoe)

### DIFF
--- a/pangram.json
+++ b/pangram.json
@@ -16,6 +16,26 @@
         "expected": false
       },
       {
+        "description": "another missing character 'x'",
+        "input": "the quick brown fish jumps over the lazy dog",
+        "expected": false
+      },
+      {
+        "description": "pangram with underscores",
+        "input": "the_quick_brown_fox_jumps_over_the_lazy_dog",
+        "expected": true
+      },
+      {
+        "description": "pangram with numbers",
+        "input": "the 1 quick brown fox jumps over the 2 lazy dogs",
+        "expected": true
+      },
+      {
+        "description": "missing letters replaced by numbers",
+        "input": "7h3 qu1ck brown fox jumps ov3r 7h3 lazy dog",
+        "expected": false
+      },
+      {
         "description": "pangram with mixed case and punctuation",
         "input": "\"Five quacking Zephyrs jolt my wax bed.\"",
         "expected": true


kotp:
Missing the tests that states that numbers are allowed in a pangram.  Such as "The quick brown fox jumped over the lazy dog for another 5 points."  The number not invalidating the pangram.  Neither does the capital T, nor the period.

The concern given before was 26 unique characters passing.  That test seems to also be missing.

Cohen-Carlisle:
> Missing the tests that states that numbers are allowed in a pangram. Such as "The quick brown fox jumped over the lazy dog for another 5 points." The number not invalidating the pangram. Neither does the capital T, nor the period.

We have one test case for `pangram with mixed case and punctuation`, as well as one for `pangram with non ascii characters`. Is there a case in which those pass but the one with numbers fails that we want to cover?

> The concern given before was 26 unique characters passing. That test seems to also be missing.

I think you are referencing https://github.com/exercism/xruby/pull/294, but its about not passing with 26 unique _word_ characters, which the latter two tests address. Checking for 26 unique characters would fail at `the quick brown fox jumps over the lazy dog` as it has a space as well.

kotp:
I guess @victorbrenders PR is not being used, so I guess it doesn't matter what his intention was.  My response may have been in confusion with the now split concerns between two tests that may be meant to cover the same problem.

The test should be a pangram that has numbers, and it should ensure that it can pass.  This is  esting that if you are missing the 't' 'i' and 'o' it will fail.  The numbers, again, are a distraction to what the test should be testing.  We already have a test for missing letters.  We don't have a test for pangram phrases that have numbers in them.

I think we should.

It also may be possible that a test should exist where you have `0123456789klmnopqrstuvwxyz' which is 26 unique (Regular Expression) word (class) characters.  I think that @victorbrender was stating that there was a solution he had seen that would have had that pass based on a regular expression word class solution, and I think it was what his test was meant to test for.  Though unsure, he has yet to confirm his intent.

Cohen-Carlisle:
I talked with @kotp and made some changes based on his feedback.

pangram.json
```diff
@@ -16,6 +16,26 @@
...
+      },
+      {
+        "description": "pangram with underscores",
+        "input": "the_quick_brown_fox_jumps_over_the_lazy_dog",
+        "expected": true
+      },
+      {
+        "description": "pangram with numbers",
+        "input": "the 1 quick brown fox jumps over the 2 lazy dogs",
+        "expected": false
```
petertseng: should this be true, since it has all letters a-z? The presence of extra numbers should not cause the expected result to be false, right?

Cohen-Carlisle:
Ugh, bad copy paste job.

Cohen-Carlisle:
@petertseng updated my derp error.

Cohen-Carlisle:
Squashed commits.

kotp:
@exercism/track-maintainers Don't forget to generate pangram tests for your track.

PR #233, opened 2016-04-21
Clarify README for wordy

s2tephen:
Added explicit reference to order of operations to make instructions slightly clearer, see exercism/xjavascript#187

### DIFF
--- a/wordy.md
+++ b/wordy.md
@@ -8,8 +8,8 @@ The program should handle large numbers and negative numbers.
 
 Remember, that these are verbal word problems, not treated as you
 normally would treat a written problem.  This means that you calculate
-as you move forward each step. In other words, you should ignore order
-of operations. 3 + 2 * 3 = 15, not 9.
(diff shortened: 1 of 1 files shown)


kotp:
Don't you ignore the PEMDAS order of operations, but pay attention to the order of operations left to right, as you would speak and hear them, as clearly shown in the example?

Because your wording introduces "order of operations" it must be clear that you define that word in this context.  Prior to that, there should have been no miscommunication, only the error of reading into it more than what is there.

kytrinyx:
I also find this confusing, but agree that if the original is also confusing, we need to fix it. Too jet lagged right now to constructively add to the conversation, I just wanted to add my support for the intent of the PR :)

kotp:
I am for bringing this PR in... enough time has gone by for objections... and I think the change is appropriate.  Consider this a bump.

IanWhitney:
Two 👍 were good enough for me to hit that merge button.

PR #236, opened 2016-04-26
sum-of-multiples: Note change in flight in README

petertseng:
As noted in https://github.com/exercism/discussions/issues/2, while #198
to remove the [3, 5] defaults is in flight, we should mark it as such in
the README.

This avoids confusion in the form of "Oh, this track doesn't implement
the defaults in the README, so I'll add them", reversing the work we've
done to make the exercise consistent across all tracks.

### DIFF
--- a/sum-of-multiples.md
+++ b/sum-of-multiples.md
@@ -4,4 +4,16 @@ multiples of either 3 or 5, we get 3, 5, 6 and 9, 10, 12, 15, and 18.
 The sum of these multiples is 78.
 
 Write a program that can find the sum of the multiples of a given set of
-numbers. If no set of numbers is given, default to 3 and 5.
+numbers.
+
+This exercise is currently under revision.
+You are still free to do this exercise and submit solutions.
+Depending on whether this track has been updated recently,
+the tests may differ between this track and other tracks.
+
+In most tracks, the set of numbers must be explicitly provided.
+In some other tracks, if no set of numbers is given, it defaults to 3 and 5.
+We would like all tracks to require the set of numbers be explicit.
+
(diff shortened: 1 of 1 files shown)


sum-of-multiples.md
```diff
@@ -4,4 +4,12 @@ multiples of either 3 or 5, we get 3, 5, 6 and 9, 10, 12, 15, and 18.
 The sum of these multiples is 78.
 
 Write a program that can find the sum of the multiples of a given set of
-numbers. If no set of numbers is given, default to 3 and 5.
+numbers.
+
+This exercise is in the process of being reworked.
```
petertseng: Will this scare people off into thinkin "maybe I shouldn't do this exercise"?

If so, I should explicitly say "You can still do this exercise and submit solutions, but the test cases may differ from the README"

sum-of-multiples.md
```diff
@@ -4,4 +4,12 @@ multiples of either 3 or 5, we get 3, 5, 6 and 9, 10, 12, 15, and 18.
...
+In most tracks, the set of numbers must be explicitly provided.
+In some other tracks, if no set of numbers is given, it defaults to 3 and 5.
+We would like all tracks to require the set of numbers be explicit.
+
+If you would like to help out, please take a look at
```
petertseng: Seems a bit vague. Should I more specific to "If your track falls into the latter category, please take a look at..."?

On the other hand, being very open may invite more contributors...?

kotp: We welcome contributors, of course, and this is a good opportunity to let them know this.  I think providing a link to get started would be great here.

PR #237, opened 2016-04-26
add bracket-push test definition

moofkit:
start working on tests for bracket-push exercise

### DIFF
--- a/bracket-push.json
+++ b/bracket-push.json
@@ -1,0 +1,64 @@
+{
+  "cases": [
+    {
+      "description": "paired square brackets",
+      "input": "[]",
+      "expected": true
+    },
+    {
+      "description": "empty string",
+      "input": "",
+      "expected": true
+    },
+    {
+      "description": "unpaired brackets",
+      "input": "[[",
+      "expected": false
+    },
+    {
+      "description": "wrong ordered brackets",
+      "input": "}{",
+      "expected": false
+    },
+    {
+      "description": "paired with whitespace",
+      "input": "{ }",
+      "expected": true
+    },
+    {
+      "description": "simple nested brackets",
+      "input": "{[]}",
+      "expected": true
+    },
+    {
+      "description": "several paired brackets",
+      "input": "{}[]",
+      "expected": true
+    },
+    {
+      "description": "paired and nested brackets",
+      "input": "([{}({}[])])",
+      "expected": true
+    },
+    {
+      "description": "unpaired and nested brackets",
+      "input": "([{])",
+      "expected": false
+    },
+    {
+      "description": "paired and wrong nested brackets",
+      "input": "[({]})",
+      "expected": false
+    },
+    {
+      "description": "math expression",
+      "input": "(((185 + 223.85) * 15) - 543)/2",
+      "expected": true
+    },
+    {
+      "description": "complex latex expression",
+      "input": "\\left(\\begin{array}{cc} \\frac{1}{3} & x\\\\ \\mathrm{e}^{x} &... x^2 \\end{array}\\right)",
+      "expected": true
+    }
+  ]
+}


kotp:
So far bug-free!  Congrats! :)

moofkit:
Now it is ready for review. As I understand for existing implementations problems is necessary to write some kind of parser for using this test suite. Is it correct?

bracket-push.json
```diff
@@ -0,0 +1,49 @@
+{
+  "cases": [
+    {
+      "description": "empty string",
+      "input": "''",
```
Cohen-Carlisle: I believe this should read `"input": "",`

bracket-push.json
```diff
@@ -0,0 +1,49 @@
...
+      "expected": true
+    },
+    {
+      "description": "paired and nested with non-bracket chars",
+      "input": "([{f00: bar}({}[baz])])",
+      "expected": true
+    },
+    {
+      "description": "unpaired and nested with non-bracket chars",
+      "input": "([{this: sh0ld({}[be] false)])",
```
kotp: Any reason the 0 should be in here or in "foo" above?  Or is it just distraction?  Nothing in the test title is telling me what the zeros are for, what problem they are on exhibit for.

Cohen-Carlisle:
> Now it is ready for review.

Please see the comments on the diff.

> As I understand for existing implementations problems is necessary to write some kind of parser for using this test suite. Is it correct?

All tracks will be able to consume this data with tools that should exist in their repos. For example, xruby uses `JSON#parse` on the file to get a `Hash` object back, and then a generator will need to be written to generate the test file. However, that is the concern of each language and not this PR directly. After this is merged, you are more than welcome to work on generating the exercise, of course! :smile:

kotp:
See [exercism-common/CONTRIBUTING.md](https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md) especially the section "Updating a Generated Test Suite, Implementing a Completely New Exercise, and Writing a New Test Suite Generator.

bracket-push.json
moofkit: I've just want to add some numbers to test.
May be I need to write separate test with only numbers or rewrite test to be more obvious.

kotp: I think adding numbers to test would be worthwhile to do.  But those numbers may look more like expressions you may see like "((5 -3) * 12)" or perhaps "(((3*(12.223+ 15)) - 7)*21)/7" 

Or maybe a statement such as:

```latex
latexM =
\left(\begin{array}{cc} \frac{1}{3} & x\\ \mathrm{e}^{x} &...
 x^2 \end{array}\right)
```

Nothing that is tricky due to the shape of a character compared to another character, visually, just a specific set of patterns you are testing for that may be an edge case for a problem being solved.

The latex statement shows that slashes do not need to be matched, but brackets do, noting that there are an odd number of slashes in that statement.  The first example showing that forward slashes also do not need to be matched.

Those have numbers, look like something you would write, and could be broken down to individual tests specifically addressing the incidental presence of numbers.

Cohen-Carlisle:
This looks good to me. I vote to merge, but will also give some time for anyone else to chime in with additional cases or feedback.

bracket-push.json
```diff
@@ -0,0 +1,49 @@
...
+      "expected": true
+    },
+    {
+      "description": "math expression",
+      "input": "(((185 + 223.85) * 15) - 543)/2",
+      "expected": true
+    },
+    {
+      "description": "complex latex expression",
+      "input": "\right(\begin{array}{cc} \frac{1}{3} & x^2 \\end{array})",
```
petertseng: extra \ in `\\end`?

petertseng:
Interesting to see that of the languages that have this exercise, up till now only Elixir and Lua had any characters that were not brackets, parentheses, braces.

I don't oppose their addition (the problem becomes slightly more interesting but it shouldn't cause so much extra code complexity), but I'm surprised nobody else brought it up.

bracket-push.json
moofkit: Originally there are not extra `\` character. But without this escaping json is not valid. Same story for `\left` so I have changed it to `\right`.

petertseng: Oh! Okay. Hmm, if this should be a latex string then should it be the case that all the `\` get escaped, then? Otherwise the `\r` in `\right` is going to become a carriage return, right?

moofkit: I thought that way too, but as you see there are no warnings. Same result for `bin/jsonlint`. So I am little confused.

kotp: No warnings because there is nothing that you are doing 'wrong'.  It is just taking what you give it, and working with it, without any way to know you wanted something different.

moofkit: I've got it. The correct **latex** expression is `\\right(\begin{array}{cc} \frac{1}{3} & x^2 \\end{array})`. Is it?

kotp: Missing the escaped slash somewhere.  Missing the dots, this is a very different statement than the original.  Missing the whole mathrm statement.


The original...
```latex
latexM =
\left(\begin{array}{cc} \frac{1}{3} & x\\ \mathrm{e}^{x} &...
 x^2 \end{array}\right)
```

bracket-push.json
```diff
@@ -0,0 +1,49 @@
...
+      "expected": true
+    },
+    {
+      "description": "math expression",
+      "input": "(((185 + 223.85) * 15) - 543)/2",
+      "expected": true
+    },
+    {
+      "description": "complex latex expression",
+      "input": "\\left(\begin{array}{cc} \frac{1}{3} & x\\ \\mathrm{e}^{x} &... x^2 \\end{array}\right)",
```
Cohen-Carlisle: Every `\` that you want to appear in the string as a character must be escaped (i.e., `\\`).
For example, `\b` is backspace, `\f` is form feed, and `\r` is carriage return here.
Getting closer, though :smile:

moofkit:
Thank's all for patience...

Cohen-Carlisle:
I think we have this all sorted out.
@moofkit, your latex string now looks like this when printed:
`\left(\begin{array}{cc} \frac{1}{3} & x\ \mathrm{e}^{x} &... x^2 \end{array}\right)`
Is that show it should look?

bracket-push.json
```diff
@@ -0,0 +1,49 @@
...
+      "expected": true
+    },
+    {
+      "description": "math expression",
+      "input": "(((185 + 223.85) * 15) - 543)/2",
+      "expected": true
+    },
+    {
+      "description": "complex latex expression",
+      "input": "\\left(\\begin{array}{cc} \\frac{1}{3} & x\\ \\mathrm{e}^{x} &... x^2 \\end{array}\\right)",
```
Cohen-Carlisle: `x\\` looks like it should be `x\\\\` inside the JSON string. When its printed `x\\\\` will look like `x\\`.

moofkit: Sorry for my misunderstanding about this.

petertseng:
Oof, that was a bit of an ordeal. May have been nice to instead do some application with bracks and without backslashes, but as far as I can tell that string, when unescaped, would be valid LaTeX now (I tried it), so we are done with that now!

Do you maybe think there should be some simple true cases? There is a complex true case in `([{}({}[])])` but would simple ones early on like `{}`, `{}[]`, `{[]}` help guide to a solution? Another reason I ask is the Go track has those cases in as well

petertseng:
Reminder (really just for us) to close https://github.com/exercism/todo/issues/77 when this is merged

moofkit:
add some simple true cases

petertseng:
OK I am pretty sure I am :+1: on this. Given @Cohen-Carlisle also say yes earlier I'll give just a little bit of time for others but I think we are pretty set.

Cohen-Carlisle:
My only remaining question/concern is

**Do we like the ordering?**

I don't think that
```   
"description": "empty string",
"input": "",
"expected": true
```
should be first... its really more of a special case that may be unhandled. It would be better to have this later after the coder has implemented the basic functionality, in my opinion.

Other than that, I give it a :+1:

kytrinyx:
I didn't think about the empty case; I think you're right @Cohen-Carlisle.

We could look at it in two ways:

1. All of the existing brackets in the string (which is zero of them) are perfectly matched.
2. No brackets--> potentially an error state.

I'm tempted to simply define "" as being the simplest possible case of perfectly matched brackets.

petertseng:
tracks with this exercise implemented seem to agree on empty string representing matched brackets

moofkit:
I agree that empty string is some kind of special case and have changed ordering as you recommended.

petertseng:
:+1:

Cohen-Carlisle:
Thanks for sticking with it @moofkit! :confetti_ball:

moofkit:
:tada: Thank's to all! This was a really hard one...

PR #239, opened 2016-04-28
new variable length quantity problem

mattetti:
Unfortunately JSON doesn't support the byte types so I couldn't add test cases :(

This PR goes hand in hand with https://github.com/exercism/xgo/pull/322

### DIFF
--- a/variable-length-quantity.md
+++ b/variable-length-quantity.md
@@ -1,0 +1,29 @@
+The goal of this exercise is to implement [VLQ](https://en.wikipedia.org/wiki/Variable-length_quantity) encoding/decoding.
+
+In short, the goal of this encoding is to save encode integer values in a way that would save bytes.
+Only the first 7 bits of each byte is significant (right-justified; sort of like an ASCII byte). 
+So, if you have a 32-bit value, you have to unpack it into a series of 7-bit bytes. 
+Of course, you will have a variable number of bytes depending upon your integer. 
+To indicate which is the last byte of the series, you leave bit #7 clear.
+In all of the preceding bytes, you set bit #7. 
+
+So, if an integer is between `0-127`, it can be represented as one byte. 
+The largest integer allowed is `0FFFFFFF`, which translates to 4 bytes variable length. 
+Here are examples of delta-times as 32-bit values, and the variable length quantities that they translate to:
+
+
+```
+ NUMBER        VARIABLE QUANTITY
+00000000              00
+00000040              40
+0000007F              7F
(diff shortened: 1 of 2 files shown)


kytrinyx:
This is a fun-looking exercise. Brings back fond memories of working on binary data formats with you :)

The problem with byte types in JSON is definitely an interesting edge-case for us. I wonder if we should support an additional format for common test data.

@petertseng I'd love a second pair of eyes on this, if you have time.

petertseng:
Re: test data. Perhaps we can get close enough by having arrays of JSON numbers rather than arrays of bytes? I think it would be valuable to have those tests, and fundamentally the bytes are indeed numbers anyway. Far as I know though, JSON won't support the `0x12` literals so the numbers will have to be written in base 10, but I would still find that valuable.

I don't currently have time to review other parts of this. I believe the most important part to think about would be "does the README provide enough information to get a student started on the exercise" so for that I'll defer to others.

petertseng:
(going to edit in) or I can take a look again this weekend. But the door is open for anyone else to take a look in the meantime.

mattetti:
These are fair comments, I personally deal with a lot of byte level code and if the Go generated tests had numbers instead of bytes, I would probably be slightly confused and would have to convert/rewrite the tests. A lot of crypto exercises also deal with byte level data so that's another factor to consider.

An alternative idea would be to support `bson`, but requiring that all languages have a bson implementation seems a bit far fetched. 

That said, I do think the text itself is pretty self explanatory and taking a look at the Go tests should help anyone wanting to implement the problem.

petertseng:
>  if the Go generated tests had numbers instead of bytes,

We can probably arrange such that the Go generated tests will output the numbers using hexadecimal literals. I think a note in the JSON file recommending that languages generating tests use hexadecimal literals (if supported by the target language) would be sufficient here.

kytrinyx:
Yeah, a "#" entry in the JSON that says to transform to hexadecimal literals if supported in the language would be enough, I think.

On a second read-through, I think that you're right that the README might not have enough information for someone to get started if they're not familiar with the problem domain at all. It would make sense to give a bit more information here.

mattetti:
I guess I am too close to the problem to see what's missing in the
explanation :( if someone can add what's missing, it would be great

kytrinyx:
I don't have the headspace to sit down and solve it right now. I am going to merge this as is, and then we can work on adding more context later if/when someone get stuck on it.

petertseng:
The test cases have been added in #486.

PR #242, opened 2016-04-30
added metadata for flatten-array question

cloudleo:
first time adding metadata, would like people to review it and let me know if any problem exist

### DIFF
--- a/flatten-array.json
+++ b/flatten-array.json
@@ -1,0 +1,29 @@
+{
+   "cases": [
+      {
+         "description": "flattens array with just integers present",
+         "input": [1,[2,3,4,5,6,7],8],
+         "expected": [1,2,3,4,5,6,7,8]
+      },
+      {
+         "description": "5 level nesting",
+         "input": [0, 2, [[2, 3], 8, 100, 4,[[[50]]]], -2],
+         "expected":[0, 2, 2, 3, 8, 100, 4, 50, -2]
+      },
+      {
+         "description": "6 level nesting",
+         "input": [1,[2,[[3]],[4,[[5]]],6,7],8],
+         "expected":[1,2,3,4,5,6,7,8]
+      },
+      {
+         "description": "6 level nest list with null values",
+         "input": [0, 2, [[2, 3], 8, [[100]], null, [[null]]], -2],
+         "expected":[0,2,2,3,8,100,-2]
+      },
+      {
+         "description": "all values in nested list are null",
+         "input": [null,[[[null]]],null,null,[[null,null],null],null],
+         "expected":[]
+      }
+   ]
+}
--- a/flatten-array.md
+++ b/flatten-array.md
@@ -1,0 +1,8 @@
+The challenge is to write a function that accepts an arbitrarily-deep nested list-like structure and returns a flattened structure without any null/nill values. 
+ 
+For Example
+
+input: [1,[2,3,null,4],[null],5]
+
+output: [1,2,3,4,5]
+
--- a/flatten-array.yml
+++ b/flatten-array.yml
@@ -1,0 +1,4 @@
+---
+blurb: "Write a program that will take a nested list and returns a single list with all values except nill/null"
+source: "Interview Question"
+source_url: "https://reference.wolfram.com/language/ref/Flatten.html"


flatten-array.md
```diff
@@ -0,0 +1,19 @@
+The challenge is to write a function that accepts an arbitrarily-deep nested Array-like structure and returns a flattened structure with any null values removed without using the reduce method.
```
kytrinyx: Not all languages have a reduce method. We'll need a way to phrase this so that it can be used for any languages.

Likewise, I think maybe we should say _list_ rather than _Array-like_. In Go, for example, we would probably use slices, not arrays for this.

flatten-array.md
```diff
@@ -0,0 +1,19 @@
...
+
+For Example
+
+flattener.flatten([0, 2, [[2, 3], 8, 100, null, [[null]]], -2]);
```
kytrinyx: Here we would need to use some form of pseudo-code rather than something that can be taken for actual syntax.

flatten-array.md
```diff
@@ -0,0 +1,19 @@
...
+
+returns [0, 2, 2, 3, 8, 100, -2]
+
+
+flattener.clear()
+
+returns []
```
kytrinyx: I don't understand the purpose of `clear`. Can you elaborate?

flatten-array.md
```diff
@@ -0,0 +1,19 @@
...
+
+Tips
+
+* Define clear first before the function flatten
+* Think about closures and know what you have access to within a function
```
kytrinyx: Not all languages have closures. If these tips are language or implementation-specific, they could be added to a `HINTS.md` file inside the exercise directory alongside the test suite and the reference solution. This would get included into the README for just that language track.

flatten-array.yml
```diff
@@ -0,0 +1,4 @@
+---
+blurb: "Write a program that will take a nested array and returns a single with all values but null "
```
kytrinyx: Let's use _list_ instead of _array_. List feels like the more generic computer science-y term for this data structure.

Not all languages have nulls. Should we say _null or nil_ here?

flatten-array.md
```diff
@@ -0,0 +1,19 @@
...
+
+Have Fun, very popular interview question
```
kytrinyx: Let's put this in source instead of "wolfram definition", but keep the source url pointing to the wolfram definition. The inspiration for this sounds like it was the interview question, rather than finding the definition on Wolfram.

flatten-array.json
```diff
@@ -0,0 +1,19 @@
+{
+   "cases": [
+      {
```
kytrinyx: I wonder if there is a gentler build-up that we could use.

For example, what's the simplest possible thing? Perhaps `[]` -> `[]`. Then maybe `[[]]` -> `[]`.

flatten-array.json
```diff
@@ -0,0 +1,19 @@
...
+         "description": "flattens array with null and integers present",
+         "input": [0, 2, [[2, 3], 8, 100, null, [[null]]], -2],
+         "expected": [0, 2, 2, 3, 8, 100, -2]
+      },
+      {
+         "description": "flattens array with just integers present",
```
kytrinyx: This seems like a simpler example than the previous one (no nulls vs has nulls). I think it should go before it.

flatten-array.json
```diff
@@ -0,0 +1,19 @@
...
+         "input": [1,[2,3,4,5,6,7],8],
+         "expected": [1,2,3,4,5,6,7,8]
```
kytrinyx: The smallest example of this would be `[1, [2], 3]` or, if you're worried about having a nested array of one element, maybe `[1, [2, 3], 4]`.

flatten-array.json
```diff
@@ -0,0 +1,19 @@
+{
+   "cases": [
+      {
+         "description": "flattens array with null and integers present",
+         "input": [0, 2, [[2, 3], 8, 100, null, [[null]]], -2],
+         "expected": [0, 2, 2, 3, 8, 100, -2]
```
kytrinyx: Here we have:

- an array
- a nested array
- a nested array inside the nested array
- nulls

I think we should build up the tests so that they are progressively more complex.

cloudleo:
I might have committed the same solution from earlier , please disregard that and close it.  I also updated the yml, md, and json.  Also simplified the example.js and the corresponding test with it

flatten-array.json
```diff
@@ -0,0 +1,29 @@
...
+         "input": [[]],
+         "expected": []
+      },
+      {
+         "description": "flattens array with just integers present",
+         "input": [1,[2,3,4],5],
+         "expected": [1,2,3,4,5]
+      },
+      {
+         "description": "flattens array with just null",
```
kytrinyx: This says there's null, but there are no nulls.

flatten-array.json
```diff
@@ -0,0 +1,29 @@
...
+         "input": [1,[2,3,4],5,[6,[7,8]]],
+         "expected":[1, 2, 3, 4, 5, 6, 7, 8]
+      },
+      {
+         "description": "flattens array with just null",
```
kytrinyx: This has the same description as the previous one. In Ruby if you have several tests with the same name, then only the last one will be run.

Also, this doesn't have nulls in it.

flatten-array.json
```diff
@@ -0,0 +1,29 @@
...
+         "input": [0, 2, [[2, 3], 8, 100, 4,[[[50]]]], -2],
+         "expected":[0, 2, 2, 3, 8, 100, 4, 50, -2]
+      },
+      {
+         "description": "flattens array with just null",
```
kytrinyx: Same here. Same name, and no nulls.

There used to be nulls in the data, did that get lost?

cloudleo: Sorry had problems last night trying to change the pull request from the console, decided to keep it more simple and get rid of null to be more general for all other language

kytrinyx: I think that removing nulls while flattening is an interesting add-on. We just have to make sure to say null or nil in the problem description to make it language-unspecific.

flatten-array.json
```diff
@@ -0,0 +1,29 @@
...
+         "input": [[]],
+         "expected": []
+      },
+      {
+         "description": "flattens a simple nested list with integers ",
+         "input": [1,[2,3,4],5],
+         "expected": [1,2,3,4,5]
+      },
+      {
+         "description": "flattens a deep nested list",
```
kytrinyx: How about specifying "two levels" of nesting?

flatten-array.json
```diff
@@ -0,0 +1,29 @@
...
+         "input": [1,[2,3,4],5,[6,[7,8]]],
+         "expected":[1, 2, 3, 4, 5, 6, 7, 8]
+      },
+      {
+         "description": "flattens a deeper nested list",
```
kytrinyx: Five levels of nesting.

flatten-array.json
```diff
@@ -0,0 +1,29 @@
...
+         "input": [0, 2, [[2, 3], 8, 100, 4,[[[50]]]], -2],
+         "expected":[0, 2, 2, 3, 8, 100, 4, 50, -2]
+      },
+      {
+         "description": "flattens a deeper nested list part 2",
```
kytrinyx: Four levels of nesting. It seems like four should come before five.

flatten-array.json
cloudleo: Got it! I just made changes to it and added null back in , thanks again @kytrinyx , if you see anything more else, please  don't hesitate to let me know, this is very fun!

kytrinyx:
This looks good to me. Would you mind rebasing it onto master?

cloudleo:
Hey @kytrinyx  I am not sure how lol

kytrinyx:
Not a problem!

First figure out which remotes you have.

```
git remote -v
```

You'll probably have two lines, that both say `origin` and point to your fork (one for 'fetch', and one for 'push'). You'll need to add another origin, for this repository:

```
git remote add upstream https://github.com/exercism/x-common.git
```

Then pull down a local shadow copy of the upstream master branch:

```
git fetch upstream master
```

Then rebase what you have onto this local shadow copy:

```
git rebase upstream/master
```

You shouldn't have any conflicts.

That should be it. Holler if you run into trouble.

kytrinyx:
BTW: I mentioned this conversation in today's "Behind the Scenes" newsletter. http://tinyletter.com/exercism/letters/exercism-it-s-about-the-conversations

cloudleo:
@kytrinyx  So cool! I highly appreciate that, I rebase the code then pushed it, but its not showing up, not sure if I did it right or if you received should i do another pull request towards the master instead?

petertseng:
you pushed it to the "leo" branch, but this PR is from your repo's master branch, so you would to push to your repo's master branch before they are visible on this PR. That would be preferred approach rather than a new PR, since we can keep all the conversation in one place rather than spread out over two.

cloudleo:
Thanks @petertseng ! I was told it was good practice to do it in your own branch and I totally forgot to switch it back to master

kotp:
@cloudleo Any update on pushing master with this change (and squash?)?

cloudleo:
Its done, please check if everything is good @kotp and thanks for reviewing!!

kotp:
Thanks for all the hard work @cloudleo

PR #243, opened 2016-04-30
simplified the question so it can be more general to all languages

cloudleo:


Cohen-Carlisle:
It looks like this PR addresses @kytrinyx's comments in https://github.com/exercism/x-common/pull/242/. If that's the case, it would be better to make these changes in that PR instead of opening a new one.

kytrinyx:
@cloudleo If you push the new commits to the same branch as the one you created the pull request from, it will update that pull request. I'm closing this one so we can continue in #242.

PR #247, opened 2016-05-04
Add Isogram exercise

ecrmnn:
This PR adds an exercise called isogram. Found on [Wikipedia](https://en.wikipedia.org/wiki/Isogram)

[JavaScript exercise PR](https://github.com/exercism/xjavascript/pull/198)

### DIFF
--- a/isogram.json
+++ b/isogram.json
@@ -1,0 +1,30 @@
+{
+  "cases": [{
+    "input": "duplicates",
+    "expected": true
+  }, {
+    "input": "eleven",
+    "expected": false
+  }, {
+    "input": "subdermatoglyphic",
+    "expected": true
+  }, {
+    "input": "Alphabet",
+    "expected": false
+  }, {
+    "input": "thumbscrew-japingly",
+    "expected": true
+  }, {
+    "input": "Hjelmqvist-Gryb-Zock-Pfund-Wax",
+    "expected": true
+  }, {
+    "input": "Heizölrückstoßabdämpfung",
+    "expected": true
+  }, {
+    "input": "the quick brown fox",
+    "expected": false
+  }, {
+    "input": "Emily Jung Schwartzkopf",
+    "expected": true
+  }]
+}
--- a/isogram.md
+++ b/isogram.md
@@ -1,0 +1,10 @@
+Determine if a word or sentence is an isogram.
+An isogram (also known as a "nonpattern word") is a word or phrase without a repeating letter.
+
+Examples of isograms:
+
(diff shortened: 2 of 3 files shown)


kotp:
I like it.

Should we narrow the tests down to three valid and good examples of words that are isograms, and then two that aren't for the reasons described in the wikipedia?

And then a couple of tests to cover words with hyphens or punctuation?

isogram.json
```diff
@@ -0,0 +1,49 @@
+{
+  "cases": [
+    {
+      "description": "duplicates",
+      "input": "duplicates",
+      "expected": true
+    },
+    {
+      "description": "eleven",
+      "input": "eleven",
```
kytrinyx: We don't need to have a description if the input is identical... in that case the people writing generators can use the `input` key to create their description.

isogram.md
```diff
@@ -0,0 +1,6 @@
+"Determine if a word or sentence is an isogram.
+An isogram (also known as a "nonpattern word") is a word or phrase without a repeating letter.
+Examples of isograms
+- lumberjacks
+- background
+- downstream
```
kytrinyx: Let's have an example of _not an isogram_. Maybe something like:

> _isograms_, however, is not an isogram, because the `s` repeats._

isogram.yml
```diff
@@ -0,0 +1,4 @@
+---
+blurb: "Determine if a word or phrase is an isogram"
```
kytrinyx: Can we end on a period to make it a full sentence?

isogram.md
```diff
@@ -0,0 +1,6 @@
+"Determine if a word or sentence is an isogram.
+An isogram (also known as a "nonpattern word") is a word or phrase without a repeating letter.
+Examples of isograms
```
kytrinyx: I think this needs a `:`

Also, the site currently has a problem with displaying lists if there is not blank line. The original markdown spec is ambiguous about this behavior, but GitHub handles it, and so people have come to expect it to work.

In the READMEs themselves it won't be a problem, but we also display the READMEs on the site, which would trigger the issue.

ecrmnn: I had github flavored markdown in mind. Updating this

ecrmnn:
@kotp Added tests with hyphen(s) and removed unnecessary tests.

@kytrinyx 
- Removed description in ``.json``.
- Added period in ``.yml``.
- Added ``:`` and example of a non-isogram word in ``.md``

kotp:
This looks good to me.

PR #249, opened 2016-05-11
Add change problem data

petemcfarlane:
As suggested here https://github.com/exercism/xphp/pull/79#issuecomment-218271354 by @kytrinyx


IanWhitney:
👍 Thanks!

PR #254, opened 2016-05-17
Update Tournament readme with more details

IanWhitney:
Based on the conversation here: https://github.com/exercism/xrust/issues/122

This addresses a few problems that @ijanos had

- What do the column headers mean?
- How are points awarded?
- What is the input format?

There's still a need for a default test json file, but that can come in a different PR.

### DIFF
--- a/tournament.md
+++ b/tournament.md
@@ -20,9 +20,7 @@ What do those abbreviations mean?
 
 A win earns a team 3 points. A draw earns 1. A loss earns 0.
 
-The outcome should be ordered by points, descending. Ties are broken by
-first looking at the number of wins then by ordering the team names
-alphabetically.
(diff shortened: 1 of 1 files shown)


tournament.md
```diff
@@ -1,15 +1,66 @@
...
+- MP: Matches Played
+- W: Matches Won
+- D: Matches Drawn (Tied)
+- L: Matches Lost
+- P: Points
+
+A win earns a team 3 points. A draw earns 1. A loss earns 0.
+
+The outcome should be ordered by points, descending. Ties are broken by
+first looking at the number of wins then by ordering the team names
```
petertseng: if there's one thing I would say it would be that "looking at" doesn't make clear that it's descending in the number of wins.

petertseng:
small thing, but generally :+1:

PR #161, opened 2015-12-12
Add commonised JSON to scrabble score problem

SwamWithTurtles:

PR #255, opened 2016-05-19
Implement canonical dataset for scrabble-score problem

kytrinyx:
This supersedes PR #161 (preserving the work done by @SwamWithTurtles, thereby giving credit where credit is due).

Question: Should we have a case that has punctuation (which should not be scored)?

I don't like that the description for the non-English letter is a sentence whereas all the others are sentence-fragments, but that's the only case where more explanation is needed, I think.

### DIFF
--- a/scrabble-score.json
+++ b/scrabble-score.json
@@ -1,0 +1,49 @@
+{
+  "cases": [
+    {
+      "description": "lowercase letter",
+      "input": "a",
+      "expected": 1
+    },
+    {
+      "description": "uppercase letter",
+      "input": "A",
+      "expected": 1
+    },
+    {
+      "description": "valuable letter",
+      "input": "f",
+      "expected": 4
+    },
+    {
+      "description": "short word",
+      "input": "at",
+      "expected": 2
+    },
+    {
+      "description": "short, valuable word",
+      "input": "zoo",
+      "expected": 12
+    },
+    {
+      "description": "medium word",
+      "input": "street",
+      "expected": 6
+    },
+    {
+      "description": "medium, valuable word",
+      "input": "quirky",
+      "expected": 22
+    },
+    {
+      "description": "long, mixed-case word",
+      "input": "OxyphenButazone",
+      "expected": 41
+    },
+    {
+      "description": "empty input",
+      "input": "",
+      "expected": 0
+    }
+  ]
+}
--- a/scrabble-score.json
+++ b/scrabble-score.json
@@ -41,6 +41,16 @@
       "expected": 41
     },
     {
+      "description": "english-like word",
+      "input": "pinata",
+      "expected": 8
+    },
+    {
+      "description": "non-english letter is not scored",
+      "input": "piñata",
+      "expected": 7
+    },
+    {
       "description": "empty input",
       "input": "",
       "expected": 0


Cohen-Carlisle:
> Question: Should we have a case that has punctuation (which should not be scored)?

Do you mean something like `that's` as a contraction of `that is`?
If so, I think we may as well add it since it's simpler than the `pinata` example.

kotp:
Scrabble doesn't allow for sentences.  I think Scrabble Cubes may.

kytrinyx:
Realism isn't the lens through which I see this problem, necessarily (though an argument can certainly be made for it). The context I have been thinking about is data entry, or hypothetical scrabble.

What would a scrabble scoring algorithm produce if you fed it different sorts of input, given the constraint that the scoring system is not configurable, and uses the English (or US English, if there's a difference) data set for scores?

There are any number of reasons why this might be fed things that are not legitimate US English tiles as input--and what should the behavior be.

I don't think it needs to be robust, but it's often interesting to allude to certain corner cases that would not occur in an analog board game.

kotp:
This is true, realism was too strong a word, since we also don't do word validations, no check against a dictionary of any language, indeed, only letters and values.

Since the README limits the letters to exactly (and literally) uppercase letters, and values for those, perhaps all others could be valued, explicitly, with 0?  And a 'length limit' such as we hit with the "too large a word"?  That might hit closer to data-entry.  Though I think 'data entry' is a different problem.

The README also, thankfully, doesn't strictly restrict to only those 26 letters, and so leaves it open for discussions, on the site.

kytrinyx:
The problem, as it stands, is case insensitive.

scrabble-score.json
```diff
@@ -0,0 +1,59 @@
...
+    },
+    {
+      "description": "english-like word",
+      "input": "pinata",
+      "expected": 8
+    },
+    {
+      "description": "non-english letter is not scored",
+      "input": "piñata",
+      "expected": 7
```
kotp: Expected, 7, so it is scored, the title is not truthful.  The ñ'' is not scored.

IanWhitney: I'm not sure I follow. If the ñ is not scored then `piñata` should score 7.

IanWhitney:
Personally, I'd move the empty-string case to the first test. It's the easiest to implement, so the easiest to get to green. 

When I was implementing this for Rust, I felt that there were a few too many test cases. To really cover it, I think you need:

- empty string test
- case insensitivity test
- a couple of words tests
- Non-letter (either space or non-English scrabble characters) test

And, for a minimal scrabble simulacrum, you're fine.

I think that's a fine place to stop for this problem. If you wanted to add advanced scrabble problems, then those could cover things like:

- test of word length (2-15 characters)
- character count (can't have more than 1 Z?)
- blank tiles
- score modifers

But I don't see any of that being necessary for this problem as it is written.

scrabble-score.json
kotp: Ah, yes... never mind.. :)  All is good in the word... (sic)

kytrinyx:
@kotp what happened here? @IanWhitney's comments are really good, but I don't see a response from you, just a merge.

I'd like to get the canonical data sets to be optimal for the problem at hand, and these are really good observations.

kotp:
The json covers everything in the list checked off:

  - empty string test :white_check_mark: 
  - case insensitivity test :white_check_mark: 
  - a couple of words tests 
  - Non-letter (either space or non-English scrabble characters) test :white_check_mark: 

I did merge this (having touched it a few times today) a bit early, thinking I had moved my comments from the files locally to here.

I did mean to mention the edge case of 'empty input' may depend on the solution, the decisions made while consuming the tests as laid out.  So the test for empty input being first, last or late should work.

I do hope to have a continuation of this, with maybe a board being utilized, maybe, with scoring modification, based on placement, doing some kind of validation with lengths, or dictionary check, valid tile counts.

kytrinyx:
Ian said:

> When I was implementing this for Rust, I felt that there were a few too many test cases. 

When I went through this, I had the same thought.

PR #256, opened 2016-05-21
Add space-age.json, extracted from Haskell track.

jonboiser:


IanWhitney:
I don't know if it's a standard, but in the test json files I've worked on I call out the expected method and inputs.

```json
{ "#": [],
  "age_on": {
    "cases": [
      {
        "description": "Age on Earth",
        "planet": "earth",
        "seconds": 1000000000,
        "expected": 31.69
      }
  ]
}
```

jonboiser:
Yeah, I've seen that pattern too, but it's not consistent. I was writing a code generator for the R problems and was having some trouble because of this.

Most problems, if it assumes that there is only one function/method, the shape is `{ cases : [...] }`.

If there are multiple, then it might follow the pattern you showed.

However, the luhn problem seems to require multiple functions/methods, but it's not divided out like that.

kytrinyx:
Eventually we'll want consistency, I think, but I also don't want to settle on something without having seen a bunch of corner cases. On the other hand, once we discover corner cases and settle on something consistent, there will be downstream changes as we fix up all the inconsistencies.

So... I guess the question is, what are the trade-offs here? Is it worth calling out the method explicitly even when there's only one method? If most of the data sets have this, then I would lean towards doing it here, too (and perhaps also updating the other one, though we'd probably need to update a couple of generators too).

jonboiser:
I think it would be best if these JSONs were standardized to make code generation easier. The spec would have to take into account a couple things, I think

1. Whether language is more object-oriented or functional. In ruby, java, coffeescript, solutions normally involve defining a class, would the JSON need to suggest a name for the class too?
2. Naming conventions. In R, the conventions is snake_case, many others use camelCase, clojure uses dash-case (?). For code generation, it would be nice to provide names that are easy to convert to the preferred style.

I think I'll go a head and change this PR to name the method as `ageOf`. Will it be good to merge then? I can do a quick scan to see which other problems (e.g. 'bob') don't provide names for the method.

kytrinyx:
> Whether language is more object-oriented or functional. 

Each individual exercise gets a custom-written generator, so I don't think that we need to worry about class names and method names, so much. The inputs and expectations are one thing, and different groups of functionality are another.

> Naming conventions

The easiest thing would be to use regular whitespace as delimitors in the JSON, that way people can turn that into snake case, mixed case, dash case, etc.

Based on this, how about: "age of"?

jonboiser:
I made the change. Also I forgot to put the planet names in, so fixed that.

How do these generators work? For xr, I wrote one from scratch using yeoman. Once it's ready, would it work for the system that's in place for the other tracks?

kytrinyx:
Nice!

Go and Ruby write one generator per exercise at the moment (though they both have some shared/code for all the generators). This would work with those generators, as far as I can tell.

This looks good to me. Can I get a second opinion before we merge?

space-age.json
```diff
@@ -0,0 +1,54 @@
...
+      {
+        "description": "Age on Earth",
+        "planet": "Earth",
+        "seconds": 1000000000,
+        "expected": 31.69
+      },
+      {
+        "description": "Age on Mercury",
+        "planet": "Mercury",
+        "seconds": 2213483568,
```
petertseng: for some reason this number is slightly different than what it is on https://github.com/exercism/xhaskell/blob/master/exercises/space-age/space-age_test.hs - 2213483568 as opposed to 2134835688 - this causes the answer to be 291.23 instead of 280.88 - one of the two should be changed

petertseng:
I wonder a bit whether having both these lines adds value:

```json
        "description": "Age on Earth",
        "planet": "Earth",
```

and the same for all the tests. It might be the case that we only need one of the two for each, what do you think?

space-age.json
jonboiser: Thanks for pointing that out. That's a typo.

jonboiser:
It's redundant, but the format follows a convention I see in the other specs:
```
description
arg1
arg2
...
argn
expectation
```
This makes it easy to generate the spec, i.e. in mocha
```
it({{description}}, () => {
  expect(myFunc({{arg1}}, {{arg2}})).to.equal({{expected}})
})
```

kytrinyx:
I don't think we need the description. I have on my list to ditch the descriptions in all the problems where it's redundant.

jonboiser:
So in the code generation use case above, how would one automatically give a human-readable description for the test?

kytrinyx:
I would use the expected input and output to craft a human-readable sentence, like we do in the ruby leap year generator: https://github.com/exercism/xruby/blob/master/lib/leap_cases.rb#L15

jonboiser:
Made the change.

I still think that something like a metadata property would make codegen easier. In the 'Bob' problem, for example, the descriptions are pretty high-level and probably couldn't created as a function of the inputs.

Also, I was assuming a more generic codegen strategy, i.e. a generator that would work the same for all problems instead of having to write new gen for each problem.

kytrinyx:
> Also, I was assuming a more generic codegen strategy, i.e. a generator that would work the same for all problems instead of having to write new gen for each problem.

Yeah, eventually that might be possible, but I don't want to try to generalize too early, since it would be really, really onerous to fix after the fact.

space-age.json
```diff
@@ -0,0 +1,46 @@
...
+  "age on": {
+    "cases": [
+      {
+        "planet": "Earth",
+        "seconds": 1000000000,
+        "expected": 31.69
+      },
+      {
+        "planet": "Mercury",
+        "seconds": 22134835688,
```
kytrinyx: Are you sure about this number? I have 2,134,835,688 not 22,134,835,688

jonboiser:
Yeah, that's a typo too. Did you see any others?

kytrinyx:
Nope, that was the only thing I found. Thanks!

PR #257, opened 2016-05-21
Reorder custom-set tests to improve flow

ryanplusplus:
The goal of this change is to ensure that no tests that require set equality in an assertion precede the tests for set equality. This allows each set of tests to introduce as little new functionality as possible (assuming that tests are completed from top to bottom).

For example, the `add` tests assert set equality to determine whether `add` is implemented correctly, but these tests precede the `equal` tests which explicitly test the set equality method. Without this change, set equality and adding new elements to a set are both introduced in the `add` tests. After this change, the `equal` tests precede the `add` tests so that set equality is implemented prior to its use in the `add` tests.

Note that the `equal` tests were maintained below the `subset` tests to gently encourage set equality to be implemented using subsets (see https://github.com/exercism/x-common/pull/232).

### DIFF
--- a/custom-set.json
+++ b/custom-set.json
@@ -47,29 +47,6 @@
          }
       ]
    },
-   "add": {
-      "description": "Unique elements can be added to a set",
-      "cases": [
-         {
-            "description": "add to empty set",
-            "set": [],
-            "element": 3,
-            "expected": [3]
-         },
-         {
-            "description": "add to non-empty set",
-            "set": [1, 2, 4],
-            "element": 3,
-            "expected": [1, 2, 3, 4]
-         },
-         {
-            "description": "adding an existing element does not change the set",
-            "set": [1, 2, 3],
-            "element": 3,
-            "expected": [1, 2, 3]
-         }
-      ]
-   },
    "subset": {
       "description": "A set is a subset if all of its elements are contained in the other set",
       "cases": [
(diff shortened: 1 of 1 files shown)


kytrinyx:
> The allows each set of tests to introduce as little new functionality as possible

I think this is important.

IanWhitney:
@ryanplusplus, do you want to make a blazon issue that lets tracks know about this change?

ryanplusplus:
@IanWhitney: Sure, but you'll have to teach me how :)

IanWhitney:
The main thing to do is write up the issue text. Here's the one I just did for Pangram.

> New tests for the Pangram problem

> We have found that the Pangram tests miss edge cases allowing students to pass all of the current tests with an incorrect implementation.

> To cover these cases we have added new tests to the Pangram test set. Those new tests were added [in this commit](https://github.com/exercism/x-common/commit/1cc803d847d146faf460982617fad2c2194d3dd2#diff-aadf160c690615b0ad89e2f12b8e779b)

> Since this track implements Pangram, please [take a look at the new pangram.json file](https://github.com/exercism/x-common/blob/master/pangram.json) and see if your track should update its tests.

> If you do need to update your tests, please refer to this issue in your PR. That helps us see which tracks still need to update their tests.

> If your track is already up to date, go ahead and close this issue.

> More details on this change are available in [x-common issue 222](https://github.com/exercism/x-common/issues/222).

The blazon tool will publish that issue to all tracks that implement Custom Set. Installing blazon isn't that big a deal if your machine is already set up to run Go programs. Mine was not, so I had to do some extra work.

- I installed go (using brew, since I'm on OSX)
- I created a directory to put go stuff in
- I set my `GOPATH` variable to be that path
- I then [installed blazon](https://github.com/exercism/blazon#installation)
- And [configured blazon](https://github.com/exercism/blazon#configuration)

I think blazon runs best if you have your issue text in a file. So, for pangram, I would run

`blazon -file blazon/pangram.txt -exercise 'pangram'`

And that would show me the results of a dry-run.

Once I was happy with my file and wanted to create the issue for real, I would pass the `--yes` flag

`blazon -file blazon/pangram.txt -exercise 'pangram' --yes`

If you don't want to do any of the Go stuff, that's fine. I can submit the issue once you've written it.

Thanks!

kytrinyx:
Be sure to read the "tips" section in the blazon readme: https://github.com/exercism/blazon#tips

ryanplusplus:
Thanks @IanWhitney, @kytrinyx. I finally got some time to set up and use blazon and it looks like everything went smoothly. Very neat.

PR #259, opened 2016-05-28
Add canonical tests for Wordy

IanWhitney:
Currently 13 tracks implement this problem. The full list of each track's tests can be found [over at this gist](https://gist.github.com/IanWhitney/3fe2e8ca98d2c6ef37dccb313172f69b)

Of the 13, 11 tracks share the exact same test suite. Those are the tests that I have canonized in this json file.

Python and Scala each have their own tests. But the behavior under test is mostly the same.

Python checks for [some additional failure cases](https://github.com/exercism/xpython/blob/master/exercises/wordy/wordy_test.py#L52)

Scala [allows students to implement powers](https://github.com/exercism/xscala/blob/master/exercises/wordy/src/test/scala/WordProblemTest.scala#L63) as extra credit.

I can see putting some of the Python failure tests in here, but I wanted to get more feedback before making that change.

### DIFF
--- a/wordy.json
+++ b/wordy.json
@@ -1,0 +1,89 @@
+{
+  "#": [
+    "The tests that expect 'false' should be implemented to raise",
+    "an error, or indicate a failure. Implement this in a way that",
+    "makes sense for your language."
+  ],
+    "cases": [
+      {
+            "description": "addition",
+            "input": "What is 1 plus 1?",
+            "expected": 2
+      },
+      {
+            "description": "more addition",
+            "input": "What is 53 plus 2?",
+            "expected": 55
+      },
+      {
+            "description": "addition with negative numbers",
+            "input": "What is -1 plus -10?",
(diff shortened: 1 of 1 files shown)


kytrinyx:
This looks good to me. If someone else also thinks so, go ahead and merge it.

wordy.json
```diff
@@ -0,0 +1,89 @@
...
+            "expected": -11
+      },
+      {
+            "description": "large addition",
+            "input": "What is 123 plus 45678?",
+            "expected": 45801
+      },
+      {
+            "description": "subtraction",
+            "input": "What is 4 minus -12",
```
petertseng: question mark or not?

petertseng:
just one comment otherwise :+1:

wordy.json
kytrinyx: I think we should have the question mark.

PR #262, opened 2016-05-31
Adding Markdown refactoring exercise

devonestes:
It would be great to have some more refactoring exercises, and one that is easy
to write both poorly and well is a Markdown parser. This is just the outline
for that, but it does include a pretty basic set of tests that cover a small
subset of Markdown syntax.

This is in response to the discussion [here](https://github.com/exercism/discussions/issues/24) about adding more refactoring
exercises.

### DIFF
--- a/markdown.json
+++ b/markdown.json
@@ -1,0 +1,62 @@
+{
+  "#": [
+    "Markdown is a shorthand for creating HTML from text strings."
+  ],
+  "methods": {
+    "description": [
+      "Check the public API is correct."
+    ],
+    "cases": [{
+      "description": "must be able to parse a Markdown string",
+      "method": "parse",
+      "arity": 1
+    }]
+  },
+  "cases": [
+    {
+       "description": "parses normal text as a paragraph",
+       "input": "This will be a paragraph",
+       "expected": "<p>This will be a paragraph</p>"
+    },
+    {
+      "description": "parsing italics",
+      "input": "_This will be italic_",
+      "expected": "<p><i>This will be italic</i></p>"
+    },
+    {
+      "description": "parsing bold text",
+      "input": "__This will be bold__",
+      "expected": "<p><em>This will be bold</em></p>"
+    },
+    {
+      "description": "mixed normal, italics and bold text",
+      "input": "This will _be_ __mixed__",
+      "expected": "<p>This will <i>be</i> <em>mixed</em></p>"
+    },
+    {
+      "description": "with h1 header level",
+      "input": "# This will be an h1",
+      "expected": "<h1>This will be an h1</h1>"
+    },
+    {
+      "description": "with h2 header level",
+      "input": "## This will be an h2",
+      "expected": "<h2>This will be an h2</h2>"
+    },
+    {
+      "description": "with h6 header level",
+      "input": "###### This will be an h6",
+      "expected": "<h6>This will be an h6</h6>"
+    },
+    {
+      "description": "unordered lists",
+      "input": "* Item 1\n* Item 2",
+      "expected": "<ul><li><p>Item 1</p></li><li><p>Item 2</p></li></ul>"
+    },
+    {
+      "description": "With a little bit of everything",
+      "input": "# Header!\n* __Bold Item__\n* _Italic Item_",
+      "expected": "<h1>Header!</h1><ul><li><em>Bold Item</em></li><li><i>Italic Item</i></li></ul>"
+    }
+  ]
+}
--- a/markdown.md
+++ b/markdown.md
@@ -1,0 +1,11 @@
+The markdown exercise is a refactoring exercise. There is code that parses a
+given string with [Markdown
+syntax](https://guides.github.com/features/mastering-markdown/) and returns the
+associated HTML for that string. Even though this code is confusingly written
+and hard to follow, somehow it works and all the tests are passing! Your
+challenge is to re-write this code to make it easier to read and maintain
+while still making sure that all the tests keep passing.
+
+It would be helpful if you made notes of what you did in your refactoring in
+comments so reviewers can see that, but it isn't strictly necessary. The most
+important thing is to make the code better!
--- a/markdown.yml
+++ b/markdown.yml
@@ -1,0 +1,2 @@
+---
+blurb: "Refactor a Markdown parser"


markdown.json
```diff
@@ -0,0 +1,57 @@
...
+    },
+    {
+      "description": "with h6 header level",
+      "input": "###### This will be an h6",
+      "expected": "<h6>This will be an h6</h6>"
+    },
+    {
+      "description": "unordered lists",
+      "input": "* Item 1\n* Item2",
+      "expected": "<ul><li><p>Item 1</p></li><p><li>Item 2</p></li></ul>"
```
bernardoamc: The last `<p>` and `<li>` tags are in the wrong order.

markdown.md
```diff
@@ -0,0 +1,11 @@
...
+given string with [Markdown
+syntax](https://guides.github.com/features/mastering-markdown/) and returns the
+associated HTML for that string. Even though this code is confusingly written
+and hard to follow, somehow it works and all the tests are passing! Your
+challenge is to re-write this code to make it easier to read and maintain
+while still making sure that all the tests keep passing.
+
+It would be helpful if you made notes of what you did in your refactoring in
+comments so reviewers can see that, but it isn't strictly necessary. The most
+important thing is the make the code better!
```
bernardoamc: Awesome text! ❤️

markdown.json
devonestes: Good catch - thanks!

markdown.md
kytrinyx: > somehow it works and all the tests are passing!

OMG. So good. This is my life, in one sentence fragment.

kytrinyx:
I really like this. I'm going to leave it open for a bit to let other @exercism/track-maintainers comment if they have any thoughts, ideas, objections etc.

markdown.json
```diff
@@ -0,0 +1,57 @@
...
+      "expected": "<h2>This will be an h2</h2>"
+    },
+    {
+      "description": "with h6 header level",
+      "input": "###### This will be an h6",
+      "expected": "<h6>This will be an h6</h6>"
+    },
+    {
+      "description": "unordered lists",
+      "input": "* Item 1\n* Item2",
```
petemcfarlane: Missing a space at the end of line 48 between `Item` and `2`, for anyone like me who was copying/pasting samples!

petemcfarlane:
I like it! I've just made the dirtiest piece of code to implement it too which I'll submit as a PR to the xphp track.
I would suggest adding a first test to see it can parse a paragraph, 
```json
{
    "description": "parses normal text as a paragraph",
    "input": "This will be a paragraph",
    "expected": "<p>This will be a paragraph</p>"
}
```
as that seemed to be one of the first things I had to implement.

ErikSchierboom:
More refactoring exercises are always welcome, as I think it's a very important skill to have.

petemcfarlane:
I've added a sample in this PR https://github.com/exercism/xphp/pull/87. At the moment the example is the same as the given file (to refactor). Do you think the example should be a cleaned up version, or does it not matter so long as it passes the tests?

markdown.json
devonestes: Good catch - thanks!

devonestes:
@petemcfarlane I've added the other test you recommend - that's a good one!

markdown.md
ryanplusplus: `important thing is the make the code better!` => `important thing is to make the code better!`

Sorry, grammar police!

devonestes: No need to be sorry - thanks for finding the typo!

devonestes:
@petemcfarlane If I'm not mistaken, the point of the example is to ensure that the tests actually _can_ pass with a certain implementation, so since this is a refactoring example I don't think that the example needs to be any different than the original. However, it couldn't hurt to have a cleaned up version if you felt like it!

markdown.md
```diff
@@ -0,0 +1,11 @@
...
+given string with [Markdown
+syntax](https://guides.github.com/features/mastering-markdown/) and returns the
+associated HTML for that string. Even though this code is confusingly written
+and hard to follow, somehow it works and all the tests are passing! Your
+challenge is to re-write this code to make it easier to read and maintain
+while still making sure that all the tests keep passing.
+
+It would be helpful if you made notes of what you did in your refactoring in
+comments so reviewers can see that, but it isn't strictly necessary. The most
+important thing is to the make the code better!
```
petertseng: I'm thinking "to make the code better", rather than "to **the** make the code better"

kytrinyx: Your ability to catch the tiny inconsistencies is a bit spooky. Thank you!

devonestes: Good find @petertseng  - thanks! It's fixed now.

kytrinyx:
Thank you!

markdown.json
```diff
@@ -0,0 +1,62 @@
...
+    },
+    {
+      "description": "with h6 header level",
+      "input": "###### This will be an h6",
+      "expected": "<h6>This will be an h6</h6>"
+    },
+    {
+      "description": "unordered lists",
+      "input": "* Item 1\n* Item 2",
+      "expected": "<ul><li><p>Item 1</p></li><li><p>Item 2</p></li></ul>"
```
petertseng: To me it seems a bit inconsistent: There are `<p>` on the list items, but there are no `<p>` on list items in the next case ("With a little bit of everything"). This could plausibly be explained by the rule "If there are any tags on list items, don't add any, otherwise do add `<p>`" but it seems weird because I don't usually see `<p>` for list items.

What's more, I looked at http://parsedown.org/demo and https://daringfireball.net/projects/markdown/ and both do not emit `<p>` for the list items. I think they should be removed.

NobbZ: As far as I understand the common mark spec, `<p>` should be there, if the list item contains more than a single paragraph:

http://spec.commonmark.org/0.28/#example-218 vs http://spec.commonmark.org/0.28/#example-219
----

218:
```md
- one

 two
```

```html
<ul>
<li>one</li>
</ul>
<p>two</p>
```

----

219
```md
- one

  two
```

```html
<ul>
<li>
<p>one</p>
<p>two</p>
</li>
</ul>
```

Issue #266, opened 2016-06-10
pangram: missing case-sensitivity edge case


sdavids13:
Check out: https://github.com/exercism/xkotlin/pull/10

The implementation that passed all existing tests didn't need to lowercase the input and was able to double count the same upper-case and lower-case characters. A new test was implemented to swap out `dog` in `the quick brown fox jumped over the lazy dog` with `FOX` which catches the error after asserting it's _not_ a pangram.

_edit:_
Test case to add: `the quick brown fox jumped over the lazy FOX`

kytrinyx:
Nevermind, I went back to the original issue. Yes, that is the problem.

It looks like there might actually be two possible test cases that we've not covered:

1. the quick brown fox jumped over the lazy fox
1. the quick brown fox jumped over the lazy dog fox

In the second it's not a pangram because we have duplicate letters.

sdavids13:
> A pangram (Greek: παν γράμμα, pan gramma, "every letter") is a sentence using every letter of the alphabet at least once.

I believe your second example is a pangram and is covered by other tests with duplicate letters.

sdavids13:
Oh, and to clarify, your example 1 needs to have one of the `FOX`s uppercased.

Cohen-Carlisle:
As I understand it, we aren't catching faulty implementations that use case sensitive letter counts that depend on there being 26 **or more** letter characters.
If it is asserted that exactly 26 letter characters exist (again, when case sensitive), then the following test will fail:
```json
{
  "description": "pangram with non ascii characters",
  "input": "Victor jagt zwölf Boxkämpfer quer über den großen Sylter Deich.",
  "expected": true
}
```
That being the case, `the quick brown fox jumped over the lazy FOX` would indeed catch this.
So would `The quick brown frog jumped over the lazy dog`, which I kinda like :smile:

kytrinyx:
Ah, I thought that pangrams were sentences that used each letter exactly once, not at least once.

The definition of the alphabet would presumably differ depending on which language the example in. A norwegian sentence would be required to have all of the English letters, plus `æ`, `ø`, and `å`. I don't know enough German to know whether they view `ß` as a separate letter, or just a symbol that represents `ss`. In French, there is only one letter `e`, but it can occur with a number of accents, (`é`, `è`, `ê`, `ë`). From the above definition of pangram, a French text that had only e, whichever accent, would qualify as a pangram.

In light of this, I would suggest that we clarify the requirements:

- English.
[...]

PR #267, opened 2016-06-11
Pangram: Updated test to verify case insensitivity count

sdavids13:
Change for #266


Cohen-Carlisle:
It works, but I don't think it fits in right now. The title states what happened relative to a previous pangram, but not the most relevant thing (missing character `g`). In this vein, I think it should be with the other tests for missing characters, with a title similar to them.

kytrinyx:
I'm closing this until we get https://github.com/exercism/x-common/issues/266 sorted. Thanks!

PR #274, opened 2016-06-16
removes an elixir specific clause that is not necessary

jedschneider:
As shown in the [example solution](https://github.com/exercism/xelixir/blob/master/exercises/accumulate/example.exs) it can be done in Elixir without using a named module, like Enum.reduce. In fact its probably confusing to suggest so.

### DIFF
--- a/accumulate.md
+++ b/accumulate.md
@@ -18,8 +18,5 @@ Keep your hands off that collect/map/fmap/whatchamacallit functionality
 provided by your standard library!
 Solve this one yourself using other basic tools instead.
 
-Elixir specific: it's perfectly fine to use `Enum.reduce` or
-`Enumerable.reduce`.
-
 Lisp specific: it's perfectly fine to use `MAPCAR` or the equivalent,
 as this is idiomatic Lisp, not a library function.


Cohen-Carlisle:
I don't think this should be removed without substantial input from @exercism/elixir as not allowing `Enum.reduce` may make the problem a lot harder.
It may also be the case that since `reduce` was implemented in the `list-ops` exercise, the library function is allowed as we don't want to make them reinvent the wheel/copy and paste.
There has also been some talk of language specific add-ons to the README... if that's ready then this could go ahead and any message could be moved to there.

NobbZ:
I dont think, that the exercise will be harder by forbidding `Enum.reduce`.
As I have realized during Haskell classes, most people seem to find it very
hard to write this exercises function in terms of reducing/folding the list
instead of simply following the apply-cons-recurse-pattern.

Also I do think, if we leave Elixirs `Enum.reduce` in there and mention it,
why don't we take Haskells `fold`s, Rubys `Enum.reduce`/`inject` and all
the other languages as well into this list? As it is now, it seems a little
bit like using a fold is fine in elixir but forbidden on all the other
tracks, while I do personally think that it is an important step for a
learner to get from apply-cons-recurse to a higher order fold.

Cohen Carlisle <notifications@github.com> schrieb am Fr., 17. Juni 2016
03:18:

jedschneider:
I agree with @NobbZ, which is why I offer the edit. It seems silly to make an exemption for Elixir specifically when the problem is not really a problem that lends itself to reduction over list construction (cons) and Elixir supports that in its core language functionality. No copy and paste from list-ops is necessary. As a side note, the order of the exercises in the elixir module seem out of order, with the accumulate being an exercise that would be excellent way to introduce recursion earlier in the challenge set. IMHO the bracket push, and sublist problems were much harder and would benefit from more recursion and cons practice highlighted by this exercise.

parkerl:
👍 

I don't think this adds to the difficulty of the exercise. I think it is actually helpful to reinforce that `Enum.reduce` at its simplest is head/tail recursion like `reduce([h | t], acc, fun)` and `reduce([], acc, _fun)`.

PR #275, opened 2016-06-18
Remove track specific wording from triangle readme

tgecho:
We had a confused user of the Elm track because elm-test does not have a "skip" concept (https://github.com/exercism/xelm/issues/66). It seems to me that this sort of instruction would be more suitable as a comment in the test file for now.

The other option might be just qualify this note a bit.

### DIFF
--- a/triangle.md
+++ b/triangle.md
@@ -1,7 +1,5 @@
 The program should raise an error if the triangle cannot exist.
 
-Tests are provided, delete one `skip` at a time.
-
 ## Hint
 
 The sum of the lengths of any two sides of a triangle always exceeds the


Cohen-Carlisle:
:+1:

petertseng:
And in fact any languages that do have skips are probably usingn that approach in all their exercises, which means it might best go somewhere global rather than per-exercise.

As just one example see how Ruby put their `skip` documentation into https://github.com/exercism/xruby/blob/master/docs/TESTS.md

tgecho:
Thanks! 👍

PR #277, opened 2016-06-18
Alphametics: add two cases to test invalid puzzles

repinel:
1. case to check puzzles without a solution.
2. case to check puzzles with leading zero values.

### DIFF
--- a/alphametics.json
+++ b/alphametics.json
@@ -21,6 +21,16 @@
                 "description": "solve puzzle with exponents",
                 "puzzle": "PI * R ^ 2 == AREA",
                 "expected": { "P": 9, "I": 6, "R": 7, "A": 4, "E": 0 }
+            },
+            {
+                "description": "solution must have unique value for each letter",
+                "puzzle": "A == B",
+                "expected": null
+            },
+            {
+                "description": "leading zero solution is invalid",
+                "puzzle": "ACA + DD == BD",
+                "expected": null
             }
         ]
     }


kotp:
This looks good to me.  It covers the information led to from the "Readme" concerning leading zeros, which is good.

alphametics.json
```diff
@@ -21,6 +21,16 @@
                 "description": "solve puzzle with exponents",
                 "puzzle": "PI * R ^ 2 == AREA",
                 "expected": { "P": 9, "I": 6, "R": 7, "A": 4, "E": 0 }
+            },
+            {
+                "description": "invalid solution returns null",
```
petertseng: it was good that the other case explained why it was invalid ("leading zero") - it would be good to say that here too, that A and B can't be the same digit

kotp: Good catch, the reason that this title was 'returns null' was to set up for the 'leading zero' language, the explanation as to why it returns null, is due to being invalid. 

Should he title be: "invalid solution of non-unique letter/value pair returns null"?  It is much more limiting, implying that is the reason for the null, rather than general "invalidness"?

petertseng: Ah, so what I hear from this is that there are things we wish to convey:

1. "This puzzle has no valid solutions - any solution to it must have the same value assigned to two letters"
2. "Any puzzle with no valid solutions should have an expectation of null."

I would note that point 2 is not specific to this puzzle or any specific puzzle at all; it could apply to any puzzle. So perhaps that is an indication it should go elsewhere.

Another thing to consider is that explicitly putting null in the description may throw off languages for which null doesn't exist (if their generators directly use the description without modification).

I am thinking a title of "solutions can't have the same digit assigned to multiple letters".

kotp: You are thinking right in that it could go anywhere, the communication is not exactly pinned down yet for the titles.

About the null, yes, this is true, but JSON is used in all languages using this, and so... is there a probl... yes, the person reading from the exercism site, not seeing or realizing the tie-in with JSON... never mind. :)

I like your title suggestion, can it be just as descriptive, but more concise?

'solutions must have unique value for each letter' which naturally restricts to 10 letters, in a base 10 numbering system?

petertseng: > 'solutions must have unique value for each letter'

seems good to me!

kotp: @repinel Looks like changes can be made, if you agree?

repinel: Agree. Thank you all for the feedback

repinel:
@petertseng @kotp JSON updated with suggestion.

Cohen-Carlisle:
Looks good, but I think we should update the markdown file to explicitly state these rules rather than having the exercist discover them at the end of the test file.

kotp:
Doesn't this do that? 

> Each letter must represent a different digit, and the leading digit of a multi-digit number must not be zero.

It is already there.

Cohen-Carlisle:
I feel like I'm missing something but I don't see that quoted string anywhere.

kotp:
It is in the ... yep, I had it ready to go the other day, and it was still sitting on my local system in a PR that wasn't pushed up yet.  It is sitting in #278

petertseng:
:+1: from me

PR #282, opened 2016-06-26
Common Test Suite for Robot Simulator

IanWhitney:
Looking at the tracks that currently implement this problem the core tests seemed
to be

- Turn and check bearing/direction
- Place and check coordinate/position
- Move and check bearing & coordinates

The [F# track](http://exercism.io/exercises/fsharp/robot-simulator) (along with some others) had what seemed to be the most thorough yet small test suite, so I've largely copied their approach here.

There were some variant tests that I haven't included, but that I mentioned in the descriptin.

- Handling invalid directions
- Handling invalid instructions
- Defaults for position and direction

These are good things to have, but they aren't currently mentioned in the readme, so I left those tests out.

The [Go implementation](http://exercism.io/exercises/go/robot-simulator/readme) is a particular outlier because it was written for a [totally different version of the Readme](https://github.com/exercism/x-common/commit/8aaf6530d79655ca027ebd29cc1144cd900c59ca), so I didn't include any of its additional tests.

### DIFF
--- a/robot-simulator.json
+++ b/robot-simulator.json
@@ -162,7 +162,9 @@
           "position": "(0,0)",
           "direction": "north"
         },
-        "expected": "(0,1)"
+        "expected": {
+          "position": "(0,1)"
+        }
       },
       {
         "description": "decreases the y coordinate by one when facing south",
@@ -170,7 +172,9 @@
           "position": "(0,0)",
           "direction": "south"
         },
-        "expected": "(0,-1)"
+        "expected": {
+          "position": "(0,-1)"
+        }
       },
       {
         "description": "increases the x coordinate by one when facing east",
@@ -178,7 +182,9 @@
           "position": "(0,0)",
           "direction": "east"
         },
-        "expected": "(1,0)"
+        "expected": {
+          "position": "(1,0)"
+        }
       },
       {
         "description": "decreases the x coordinate by one when facing west",
@@ -186,7 +192,9 @@
           "position": "(0,0)",
           "direction": "west"
         },
-        "expected": "(-1,0)"
+        "expected": {
+          "position": "(-1,0)"
+        }
       }
     ]
   },
--- a/robot-simulator.json
+++ b/robot-simulator.json
@@ -34,7 +34,7 @@
     ]
   },
   "turn_right": {
-    "description": "rotates the robot's position 90 degrees clockwise",
+    "description": "rotates the robot's direction 90 degrees clockwise",
     "cases" : [
       {
         "description": "does not change the position",
@@ -53,7 +53,7 @@
           "direction": "north"
         },
         "expected": {
-          "position": "east"
+          "direction": "east"
         }
       },
       {
@@ -63,7 +63,7 @@
           "direction": "east"
         },
         "expected": {
-          "position": "south"
+          "direction": "south"
         }
       },
       {
@@ -73,7 +73,7 @@
           "direction": "south"
         },
         "expected": {
-          "position": "west"
+          "direction": "west"
         }
       },
       {
@@ -83,13 +83,13 @@
           "direction": "west"
         },
         "expected": {
-          "position": "north"
+          "direction": "north"
         }
       }
     ]
   },
   "turn_left": {
-    "description": "rotates the robot's position 90 degrees counter-clockwise",
+    "description": "rotates the robot's direction 90 degrees counter-clockwise",
     "cases" : [
       {
         "description": "does not change the position",
@@ -108,7 +108,7 @@
           "direction": "north"
         },
         "expected": {
-          "position": "west"
+          "direction": "west"
         }
       },
       {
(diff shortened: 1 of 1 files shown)


robot-simulator.json
```diff
@@ -0,0 +1,234 @@
...
+        }
+      },
+      {
+        "description": "changes the direction from north to east",
+        "robot": {
+          "position": "(0,0)",
+          "direction": "north"
+        },
+        "expected": {
+          "position": "east"
```
petertseng: direction

robot-simulator.json
```diff
@@ -0,0 +1,234 @@
...
+        }
+      },
+      {
+        "description": "changes the direction from east to south",
+        "robot": {
+          "position": "(1,-1)",
+          "direction": "east"
+        },
+        "expected": {
+          "position": "south"
```
petertseng: direction

robot-simulator.json
```diff
@@ -0,0 +1,234 @@
...
+        }
+      },
+      {
+        "description": "changes the direction from south to west",
+        "robot": {
+          "position": "(-1,-1)",
+          "direction": "south"
+        },
+        "expected": {
+          "position": "west"
```
petertseng: direction

robot-simulator.json
```diff
@@ -0,0 +1,234 @@
...
+        }
+      },
+      {
+        "description": "changes the direction from west to north",
+        "robot": {
+          "position": "(-1,-1)",
+          "direction": "west"
+        },
+        "expected": {
+          "position": "north"
```
petertseng: direction

robot-simulator.json
```diff
@@ -0,0 +1,234 @@
...
+        }
+      },
+      {
+        "description": "changes the direction from north to west",
+        "robot": {
+          "position": "(0,0)",
+          "direction": "north"
+        },
+        "expected": {
+          "position": "west"
```
petertseng: direction

robot-simulator.json
```diff
@@ -0,0 +1,234 @@
...
+        }
+      },
+      {
+        "description": "changes the direction from west to south",
+        "robot": {
+          "position": "(1,-1)",
+          "direction": "west"
+        },
+        "expected": {
+          "position": "south"
```
petertseng: direction

robot-simulator.json
```diff
@@ -0,0 +1,234 @@
...
+        }
+      },
+      {
+        "description": "changes the direction from south to east",
+        "robot": {
+          "position": "(-1,-1)",
+          "direction": "south"
+        },
+        "expected": {
+          "position": "east"
```
petertseng: direction

robot-simulator.json
```diff
@@ -0,0 +1,234 @@
...
+        }
+      },
+      {
+        "description": "changes the direction from east to north",
+        "robot": {
+          "position": "(-1,-1)",
+          "direction": "east"
+        },
+        "expected": {
+          "position": "north"
```
petertseng: direction

petertseng:
Tests look reasonable. What do you/everyone think about having the coordinates be e.g. `[-1, -1]` instead of `"(-1, -1)"` to avoid making tracks parse out the parentheses and numbers?

IanWhitney:
It doesn't matter to me. I picked parens because that's what I used in [queen-attack.json](https://github.com/exercism/x-common/blob/master/queen-attack.json#L17).

But if square brackets are better for languages that auto-generate tests, then that works for me.

petertseng:
Now you're making me wonder why I didn't ask the same in queen-attack.json. I do think `[1, 2]` would be better, wonder what others think (what tracks are generating their tests? I only know of Ruby and Go right now)

IanWhitney:
You can't catch everything, Peter.

@kotp or @kytrinyx, opinions on this? Knowledge of other languages that are auto-generating tests?

kotp:
One communicates to me that the passed in argument needs to be an Array, the other suggests it could be an array or two arguments.  Hopefully, the tests end up not forcing either, to keep the ways it can be solved as free as possible.  The "(1, 2)", to me suggests arguments, where "[1, 2]" suggests using an array as the argument.

This is weighted against the title of the test and any 'message' given for when the test does not pass, as well as any comment for that section.

They are generated, so really, we can manipulate them as we need to.  Clarity in the communication, even little things like parenthesis and brackets, counts though.

FYI @exercism/ruby is the 'ping' trigger that was setup for the Ruby track.

kytrinyx:
I don't have a preference. I would think that any track that implements this will think through what they want the API to be whether we choose square brackets or parentheses.

IanWhitney:
@kotp The Rust implementation will be treating them as parameters. When I did Queen Attack I passed the coordinates in as a tuple [and people didn't like it](https://github.com/exercism/xrust/pull/119). So I'm sticking passing them in separately.

Sounds like the parens/bracket thing isn't a big concern. I'll let this sit for a few days before merging. Or, if any other maintainers want to merge, go for it. I'm not the boss of you!

PR #283, opened 2016-06-27
Explicitly describe robot name format

JPBetley:
The definition of this problem led to some confusion in the scala exercise's test case. Some languages have exercises that expect differing formats for the robots names. This is an attempt to make it explicit and clear to the end-users and the exercise creators to avoid this ambiguity.

An example of this issue can be found in this PR: https://github.com/exercism/xscala/pull/123

If this PR is accepted, I will be more than happy to check the languages that have the "Robot Name" problem to make sure the tests for the "Robot Name" exercises are consistent.

### DIFF
--- a/robot-name.md
+++ b/robot-name.md
@@ -1,7 +1,7 @@
 When robots come off the factory floor, they have no name.
 
-The first time you boot them up, a random name is generated, such as
-RX837 or BC811.
+The first time you boot them up, a random name is generated in the format
+of two uppercase letters followed by three digits, such as RX837 or BC811.
 
 Every once in a while we need to reset a robot to its factory settings,
 which means that their name gets wiped. The next time you ask, it will


kytrinyx:
There's an open discussion about this: https://github.com/exercism/x-common/issues/260

@verdammelt @Insti @kotp any thoughts about this?

Insti:
Explicitly specifying the format in the description is a good improvement.
I am in favour of the uppercase only format specified by this PR.

kotp:
(background info: xruby track; not using generated tests yet for this exercise., current tests do test for uppercase, and discussions have happened with a full set of names including upper and lower case names with the solvers.)

The current example and test use the uppercase limitation, I know discussions in the solutions have talked about lower case, as well as what happens when we run out of names.  Also using solutions that don't get tested with the rand seed. (SecureRandom library).

Whatever the generated tests bring on, I think the discussions can and will still happen about these things.

Currently it is "strongly implied" due to the example names of "RX837 or BC811" in the readme, and explicitly with the tests that enforce it.

IanWhitney:
Everyone seems good with this. Merging.

PR #284, opened 2016-06-30
Identify the alphabet used.

kotp:
Identifies the alphabet used, which is ASCII, case insensitive, A
through Z.


kytrinyx:
Thanks, I think this is a really good clarification. We can revisit it if we get more confusion.

PR #285, opened 2016-07-02
transpose: add .md and .yml

andrey-gvrd:
In reference to rust implementation <https://github.com/exercism/xrust/pull/151>

### DIFF
--- a/transpose.md
+++ b/transpose.md
@@ -10,9 +10,9 @@ DEF
 is given by:
 
 ```
-A D
-B E
-C F
+AD
+BE
+CF
 ```
 
 Rows become columns and columns become rows. See <https://en.wikipedia.org/wiki/Transpose>.
--- a/transpose.md
+++ b/transpose.md
@@ -3,8 +3,8 @@ Write a program that takes input text and outputs it -- transposed.
 Roughly explained, the transpose of a matrix:
 
 ```
-A B C
-D E F
+ABC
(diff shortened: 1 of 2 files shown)


petertseng:
The examples given seem to imply that the elements of the input should be transposed word-wise. For example, that:

```
Aaa Bbb Ccc
Ddd Eee Fff
``` 

would tranpose to;

```
Aaa Ddd
Bbb Eee
Ccc Fff
```

If I understand the test expectations in https://github.com/exercism/xrust/pull/151, it appears the transposition is instead character-wise. It would be good to make this clear in the readme so that there is no confusion. especially because character-wise may indicate that

```
A B C
D E F
```

is supposed to actually transpose to:

```
AD

BE

CF
```

which is not the same as what the example indicates.

---

Finally, it would indeed be good to get a file with common test cases.

andrey-gvrd:
Fixed the example to remove the ambiguity.

transpose.md
```diff
@@ -0,0 +1,18 @@
...
+
+```
+ABC
+DEF
+```
+
+is given by:
+
+```
+A D
```
petertseng: why space?

IanWhitney:
I think this is ready. Will merge on Monday, July 18 unless someone says otherwise.

kytrinyx:
Thank you!

Issue #286, opened 2016-07-02
tournament: Should invalid lines result in an error, or be ignored?


petertseng:
The tracks that implement tournament:

http://x.exercism.io/problems/tournament

* https://github.com/exercism/xcsharp/blob/master/exercises/tournament/TournamentTest.cs
* https://github.com/exercism/xfsharp/blob/master/exercises/tournament/TournamentTest.fs
* https://github.com/exercism/xgo/blob/master/exercises/tournament/tournament_test.go
* https://github.com/exercism/xrust/tree/master/exercises/tournament/tests
* https://github.com/exercism/xswift/blob/master/exercises/tournament/TournamentTest.swift

The README at https://github.com/exercism/x-common/blob/master/tournament.md says "Your program should only accept inputs that follow this format." courtesy of #254.

Can we get some clarity on what "should only accept" means?

If there is a file containing some valid lines plus some invalid lines, what should happen?

* Ignore the invalid lines. Output the expected output for the valid lines (csharp, fsharp, rust, swift)
* Output an error for the entire thing (go)

Currently, "should only accept" to me sounds like the approach that the go track is taking is correct!

I think we should pick one of the following:

1. Leave README as-is. All tracks other than xgo should update tests to reject inputs with invalid line.
2. Change README to read "should only process input lines that follow this format, and ignore all other lines". Then xgo will either update its test expectations... Edit: or not, because xgo explicitly calls out the behavior difference at https://github.com/exercism/xgo/blob/master/exercises/tournament/tournament_test.go#L12.
3. Change README to explicitly point out that individual tracks might choose their behavior in the presence of invalid lines. This puts more work on students, they have to figure out what behavior is used by the track they are working on currently.
4. Status quo, leave everything ambiguous and see if anyone complains. No action required on anyone's part. Until someone complains.

I do not particularly care which we pick since of the tracks I'm interested in both camps are represented. But we should pick something consistent, no? (Which means I want something other than 4)

petertseng:
Oh. Just realised that xgo is already aware that its behavior is an exception from all other tracks, and has called itself out at https://github.com/exercism/xgo/blob/master/exercises/tournament/tournament_test.go#L12 . I'm editing my second choice to reflect that. In that case xgo can stay as the outlier and we'd only need to change the README here.

PR #287, opened 2016-07-03
tournament: Add JSON file

petertseng:
For the most part, the existing implementations all used the same tests:

* https://github.com/exercism/xcsharp/blob/master/exercises/tournament/TournamentTest.cs
* https://github.com/exercism/xfsharp/blob/master/exercises/tournament/TournamentTest.fs
* https://github.com/exercism/xgo/blob/master/exercises/tournament/tournament_test.go
* https://github.com/exercism/xrust/tree/master/exercises/tournament/tests
* https://github.com/exercism/xswift/blob/master/exercises/tournament/TournamentTest.swift

xgo is the clear outlier: Whereas all other tracks ignore invalid lines,
xgo declares that inputs with invalid lines cause an error (but empty
lines and comments are okay). Please refer to discussion in #286 where
we decide how to make this consistent across tracks.

For now, these tests follow the behavior of the other tracks.

xgo also had a test for ties which no other track had, so that's been
added. This tests that ties are broken alphabetically.

Discussion point: The README specifies that ties are broken first in
favor of the team with more wins. It's not possible to get a tie with a
different number of wins in a four-team group (one team must get one
win and no draws, while another team must get three draws and no wins).
We can test this by doing one of two things:
    
- A group with five teams
- A group where a pair plays each other more than once.
    
Should we test such a thing? Note that this is not usually seen in
football tournaments so such a test may be surprising to students.

Behaviors not tested:
* What happens on an empty input? Should the header be printed, with no
  rows in the result table?
* Should a team not be allowed to play itself? Should we test this? If
  so, it should be declared in the README that it's invalid for a team
  to play itself, of course.

Closes https://github.com/exercism/todo/issues/155

### DIFF
--- a/tournament.json
+++ b/tournament.json
@@ -1,0 +1,88 @@
+{
+  "#": [
+    "The inputs and outputs are represented as arrays of strings to improve readability in this JSON file.",
+    "Your track may choose whether to present the input as a single string (concatenating all the lines) or as the list.",
+    "In most cases, it seems to make sense to expect the output as a single string."
+  ],
+  "valid_inputs": {
+    "cases": [
+      {
+        "description": "typical input",
+        "input": [
+          "Allegoric Alaskans;Blithering Badgers;win",
+          "Devastating Donkeys;Courageous Californians;draw",
+          "Devastating Donkeys;Allegoric Alaskans;win",
+          "Courageous Californians;Blithering Badgers;loss",
+          "Blithering Badgers;Devastating Donkeys;loss",
+          "Allegoric Alaskans;Courageous Californians;win"
+        ],
+        "expected": [
+          "Team                           | MP |  W |  D |  L |  P",
+          "Devastating Donkeys            |  3 |  2 |  1 |  0 |  7",
+          "Allegoric Alaskans             |  3 |  2 |  0 |  1 |  6",
+          "Blithering Badgers             |  3 |  1 |  0 |  2 |  3",
+          "Courageous Californians        |  3 |  0 |  1 |  2 |  1"
+        ]
+      },
+      {
+        "description": "incomplete competition (not all pairs have played)",
+        "input": [
+          "Allegoric Alaskans;Blithering Badgers;loss",
+          "Devastating Donkeys;Allegoric Alaskans;loss",
+          "Courageous Californians;Blithering Badgers;draw",
+          "Allegoric Alaskans;Courageous Californians;win"
+        ],
+        "expected": [
+          "Team                           | MP |  W |  D |  L |  P",
+          "Allegoric Alaskans             |  3 |  2 |  0 |  1 |  6",
+          "Blithering Badgers             |  2 |  1 |  1 |  0 |  4",
+          "Courageous Californians        |  2 |  0 |  1 |  1 |  1",
+          "Devastating Donkeys            |  1 |  0 |  0 |  1 |  0"
+        ]
+      },
+      {
+        "description": "ties broken alphabetically",
+        "input": [
+          "Courageous Californians;Devastating Donkeys;win",
+          "Allegoric Alaskans;Blithering Badgers;win",
+          "Devastating Donkeys;Allegoric Alaskans;loss",
+          "Courageous Californians;Blithering Badgers;win",
+          "Blithering Badgers;Devastating Donkeys;draw",
+          "Allegoric Alaskans;Courageous Californians;draw"
+        ],
+        "expected": [
+          "Team                           | MP |  W |  D |  L |  P",
+          "Allegoric Alaskans             |  3 |  2 |  1 |  0 |  7",
+          "Courageous Californians        |  3 |  2 |  1 |  0 |  7",
+          "Blithering Badgers             |  3 |  0 |  1 |  2 |  1",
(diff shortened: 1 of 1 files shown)


petertseng:
Things I'd like to happen before this gets merged:

- [ ] should we test empty input? If so, what should the expected output be?
- [ ] should we test what happens when a team plays itself? If so, we should edit the README
- [ ] should we explicitly test the tiebreaking logic that breaks ties in favor of team with more wins? In that case, we'd have to add a fifth team to a group, or allows pair to play each other more than once.
- [ ] I may have had copy/paste errors, and also I had to make up a new case to as not to test the tie logic earlier than necessary. It's not appropriate to merge this until I verify that the results are correct.

IanWhitney:
1.  I don't think we need to test empty input. I've definitely seen some exercises suffer from a case of Edge-itis, where we try to test every weird edge case. Some edges can be interesting, but it also helps to remember that we're not writing production-ready code here.
2. Same.
3. Or just drop that tiebreak sentence from the Readme. That also solves the problem.

tournament.json
```diff
@@ -0,0 +1,88 @@
...
+        "expected": [
+          "Team                           | MP |  W |  D |  L |  P",
+          "Devastating Donkeys            |  3 |  2 |  1 |  0 |  7",
+          "Allegoric Alaskans             |  3 |  2 |  0 |  1 |  6",
+          "Blithering Badgers             |  3 |  1 |  0 |  2 |  3",
+          "Courageous Californians        |  3 |  0 |  1 |  2 |  1"
+        ]
+      },
+      {
+        "description": "incomplete competition (not all pairs have played)",
```
IanWhitney: Wasn't there some restriction on this problem that only complete tournaments could be tallied? I remember seeing that, but now I can't find it.

IanWhitney:
I have the one question about an incomplete tournament, but the test input & expectations seem fine to me.

tournament.json
petertseng: I don't see such a thing, and most tracks are testing this case as well

petertseng:
OK, so my plan is: I'll merge this once I check that the test expectations are correct. I still need to do that. If someone wants to beat me to it feel free.

In a separate PR, the README probably needs a tweak on the tiebreak rule since we won't see ties needing to be broken by # wins in things we test.

petertseng:
Merging with this diff on top (correct Alaskians to Alaskans)

```
$ git diff
diff --git a/tournament.json b/tournament.json
index a42170c..b674454 100644
--- a/tournament.json
+++ b/tournament.json
@@ -44,11 +44,11 @@
         "description": "ties broken alphabetically",
         "input": [
           "Courageous Californians;Devastating Donkeys;win",
-          "Allegoric Alaskians;Blithering Badgers;win",
-          "Devastating Donkeys;Allegoric Alaskians;loss",
+          "Allegoric Alaskans;Blithering Badgers;win",
+          "Devastating Donkeys;Allegoric Alaskans;loss",
           "Courageous Californians;Blithering Badgers;win",
           "Blithering Badgers;Devastating Donkeys;draw",
-          "Allegoric Alaskians;Courageous Californians;draw"
+          "Allegoric Alaskans;Courageous Californians;draw"
         ],
         "expected": [
           "Team                           | MP |  W |  D |  L |  P",
@@ -81,7 +81,7 @@
       },
       {
         "description": "invalid match result",
-        "input": "Devastating Donkeys;Allegoric Alaskians;dra"
+        "input": "Devastating Donkeys;Allegoric Alaskans;dra"
       }
     ]
   }
```

petertseng:
Verified:

```ruby
require 'json'

data = JSON.parse(File.read('tournament.json'))

class Team
  attr_reader :name

  def initialize(name)
    @name = name.freeze
    @wins = 0
    @losses = 0
    @draws = 0
  end

  def plays
    @wins + @losses + @draws
  end

  def points
    @wins * 3 + @draws
  end

  def win
    @wins += 1
  end

  def lose
    @losses += 1
  end

  def draw
    @draws += 1
  end

  def to_s
    '%-30s | %2d | %2d | %2d | %2d | %2d' % [@name, plays, @wins, @draws, @losses, points]
  end
end

HEADER = ('%-30s | MP |  W |  D |  L |  P' % 'Team').freeze

def tally(lines)
  teams = Hash.new { |h, k| h[k] = Team.new(k) }
  lines.each { |line|
    parts = line.chomp.split(?;)
    next unless parts.size == 3
    t1, t2, r = parts
    t1 = teams[t1]
    t2 = teams[t2]
    case r
    when 'win'
      t1.win
      t2.lose
    when 'loss'
      t2.win
      t1.lose
    when 'draw'
      t1.draw
      t2.draw
    end
  }
  HEADER + "\n" + teams.values.sort_by { |t| [-t.points, t.name] }.join("\n")
end

data['valid_inputs']['cases'].each_with_index { |c, i|
  got = tally(c['input'])
  want = c['expected'].join("\n")
  raise "#{c} got #{got}" if got != want
  puts "#{i} OK"
}

valid = data['valid_inputs']['cases'].first['input']
want = tally(valid)
data['invalid_lines']['cases'].each_with_index { |c, i|
  got = tally(valid + [c['input']])
  raise "#{c} got #{got}" if got != want
  puts "#{i} OK"
}
```

PR #289, opened 2016-07-05
rectangles.json: Explain purpose of large input test

petertseng:
As far as I can tell, it was added to rule out inefficient solutions
since it doesn't test any edge cases not already covered in previous
tests. (And even if it did, I would want to just add previous tests that
test those edge cases).

It's useful to explain the purpose of each test so tracks can decide
whether to test for performance or not. Discussion at
https://github.com/exercism/xrust/pull/165 expands on the merits.

Currently the fact that this is a test primarily for performance is not
encoded in a machine-readable way, so tracks that auto-generate tests
will not be able to mark that it's a performance test automatically. We
can revisit that if there is a large demand for it.

### DIFF
--- a/rectangles.json
+++ b/rectangles.json
@@ -84,7 +84,7 @@
       "expected": 2
     },
     {
-      "description": "large input with many rectangles",
+      "description": "large input with many rectangles, ruling out inefficient solutions",
       "input": [
         "+---+--+----+",
         "|   +--+----+",


petertseng:
If @ryanplusplus  has a moment, I could use a hand (to understand whether I have got the purpose of this test correct)

ryanplusplus:
Hey @petertseng -- my intent with that test was more to verify that the implementation does not pass the other tests due to an invalid simplification and that the solution scales to large input without generating an invalid solution. The fact that this can serve as a performance test is a nice bonus, but wasn't my goal.

petertseng:
OK, seems like I'll have to make a change to the description in that case. Any general way to summarise the simplifications that might be seen? Is it only "assuming that the number of rectangles will be < 10" (or some number), or are there more nuanced ones like "assumes rectangles have (some property only present in the small inputs)"?

ryanplusplus:
I don't think I have a great name for it. I think of it as a "stress test", but that's not very descriptive either.

IanWhitney:
Since @petertseng didn't notice a big performance problem

https://github.com/exercism/xrust/pull/165#issuecomment-230378370

I say just stick with "find many rectangles"

petertseng:
since that's not materially different from what's already here, I'll leave it be

PR #291, opened 2016-07-09
Add ocr-numbers test suite, update description

IanWhitney:
The tests come from the existing implementations. Most languages that
currently implement this problem use the exact same tests. The tests of
failures only exist in Python and Swift, but I thought they were useful
so I included them here.

While I've kept the existing tests, I have reordered them so that they
follow the problem as described in the readme. The problem progresses
from handling just single binary numbers, then garbling, then long binary
strings, then decimal numbers, then multi-line numbers.

I have also reworked the readme, clarifying the steps of the problem.

### DIFF
--- a/ocr-numbers.json
+++ b/ocr-numbers.json
@@ -1,0 +1,187 @@
+{
+  "#": [
+    "An expectation of  -1 indicates that a failure should be raised"
+  ],
+  "convert": {
+    "description": "Converts lines of OCR Numbers to a string of integers",
+    "cases": [
+      {
+        "description": "Recognizes 0",
+        "input": [
+          " _ ",
+          "| |",
+          "|_|",
+          "   "
+        ],
+        "expected": "0"
+      },
+      {
+        "description": "Recognizes 1",
+        "input": [
+          "   ",
+          "  |",
+          "  |",
+          "   "
+        ],
+        "expected": "1"
+      },
+      {
+        "description": "Unreadable but correctly sized inputs return ?",
+        "input": [
+          "   ",
+          "  _",
+          "  |",
+          "   "
+        ],
+        "expected": "?"
+      },
+      {
+        "description": "Input with a number of lines that is not a multiple of four raises an error",
+        "input": [
+          " _ ",
+          "| |",
+          "   "
+        ],
+        "expected": -1
+      },
+      {
+        "description": "Input with a number of columns that is not a multiple of three raises an error",
+        "input": [
+          "    ",
+          "   |",
+          "   |",
+          "    "
+        ],
+        "expected": -1
+      },
+      {
+        "description": "Recognizes 110101100",
+        "input": [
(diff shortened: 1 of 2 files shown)


IanWhitney:
A quick list of current implementations and their tests. Not very exciting, since there is so much similarity.

https://gist.github.com/IanWhitney/0d847ff8d4ac713a81c3d81649d28a6b

ocr-numbers.md
```diff
@@ -1,36 +1,74 @@
-## Step 1
+# Step One
 
-A simple binary font has been constructed using only pipes and
-underscores.
+To begin with, convert a simple binary font to a string containing 0 and 1.
```
petertseng: do you mean for step one to be a single digit, or multiple digits? If it is to be a single digit, "0 and 1" seems to imply the single digit should simultaneously be both, which would not be

ocr-numbers.json
```diff
@@ -0,0 +1,187 @@
...
+      "input": [
+        "    _  _     _  _  _  _  _  _ ",
+        "  | _| _||_||_ |_   ||_||_|| |",
+        "  ||_  _|  | _||_|  ||_| _||_|",
+        "                              "
+      ],
+      "expected": "1234567890"
+      },
+      {
+      "description": "Numbers separated by empty lines are recognized. Empty lines are replaced with commas",
```
petertseng: well, it's not really that there's any additional separation, since the number of lines is still a multiple of 4, and the fourth line has always been empty. the trailing empty line was not replaced with a comma either. I may have just said "Recognizes multiple lines of numbers. The lines are joined by commas."

ocr-numbers.md
```diff
@@ -1,36 +1,74 @@
...
+If the input is the correct size, but not recognizable, your program should return '?'
+
+If the input is the incorrect size, your program should return an error.
+
+# Step Two
 
-Write a program that, given a 3 x 4 grid of pipes, underscores, and
-spaces, can determine whether the the grid represents a zero, a one, or
-garble.
+Update your program to recognize binary strings, replacing garbled numbers with ?
```
petertseng: hmm, maybe it's just me, but I guess it might benefit from being explicit that there can be more than one character now.

ocr-numbers.md
```diff
@@ -1,36 +1,74 @@
...
     | _| _||_||_ |_   ||_||_|| | # decimal numbers.
     ||_  _|  | _||_|  ||_| _||_| #
                                  # fourth line is always blank
+```
+
+Is converted to "1234567890"
+
+# Step Four
+
+Update your program to handle multiple numbers, one per line. When converting several lines, replace the empty lines with commas.
```
petertseng: same thought about empty lines, maybe just say "join the line with commas" or something

IanWhitney:
If nothing else comes up by the weekend, I'll merge.

PR #296, opened 2016-07-15
binary: Add test data for "binary" exercise

tommyschaefer:
Adds test data for "binary" exercise for exercism/xruby#394.

Thank you so much for your time!

### DIFF
--- a/binary.json
+++ b/binary.json
@@ -1,5 +1,5 @@
 {
-   "cases": [
+   "decimal": [
       {
          "description": "binary 0 is decimal 0",
          "binary": "0",
--- a/binary.json
+++ b/binary.json
@@ -1,0 +1,54 @@
+{
+   "cases": [
+      {
+         "description": "binary 0 is decimal 0",
+         "binary": "0",
+         "expected": 0
+      },
+      {
+         "description": "binary 1 is decimal 1",
+         "binary": "1",
+         "expected": 1
+      },
+      {
+         "description": "binary 10 is decimal 2",
+         "binary": "10",
+         "expected": 2
+      },
+      {
+         "description": "binary 11 is decimal 3",
+         "binary": "11",
+         "expected": 3
+      },
+      {
+         "description": "binary 100 is decimal 4",
+         "binary": "100",
+         "expected": 4
+      },
+      {
+         "description": "binary 1001 is decimal 9",
+         "binary": "1001",
+         "expected": 9
+      },
+      {
+         "description": "binary 11010 is decimal 26",
+         "binary": "11010",
+         "expected": 26
+      },
+      {
+         "description": "binary 10001101000 is decimal 1128",
+         "binary": "10001101000",
+         "expected": 1128
+      },
+      {
+         "description": "binary ignores leading zeros",
+         "binary": "000011111",
+         "expected": 31
+      },
+      {
+         "description": "invalid binary numbers raise an error",
+         "binary": ["012", "10nope", "nope10", "10nope10", "001 nope", "2"],
+         "expected": -1
+      }
+   ]
+}
--- a/binary.json
+++ b/binary.json
@@ -46,8 +46,13 @@
          "expected": 31
       },
       {
-         "description": "invalid binary numbers raise an error",
-         "binary": ["012", "10nope", "nope10", "10nope10", "001 nope", "2"],
+         "description": "numbers other than one and zero raise an error",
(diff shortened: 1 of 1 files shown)


binary.json
```diff
@@ -0,0 +1,54 @@
...
+         "expected": 1128
+      },
+      {
+         "description": "binary ignores leading zeros",
+         "binary": "000011111",
+         "expected": 31
+      },
+      {
+         "description": "invalid binary numbers raise an error",
+         "binary": ["012", "10nope", "nope10", "10nope10", "001 nope", "2"],
```
Insti: I think that the error cases should be split into individual tests, each of them deserves a description of why it is an invalid input which will help when coding the solution.

binary.json
```diff
@@ -0,0 +1,54 @@
+{
+   "cases": [
```
Insti: This could probably be called "decimal" to reflect that there will probably be a `decimal` function/method being used in the solution.

binary.json
Insti: This might be done by extracting to an "errors" array to match the "cases/decimal" one above.

binary.json
```diff
@@ -0,0 +1,54 @@
...
+      },
+      {
+         "description": "binary ignores leading zeros",
+         "binary": "000011111",
+         "expected": 31
+      },
+      {
+         "description": "invalid binary numbers raise an error",
+         "binary": ["012", "10nope", "nope10", "10nope10", "001 nope", "2"],
+         "expected": -1
```
Insti: I'm not sure -1 is the right value to expect in the case of an error, it probably should be `null`.

NobbZ:
We are currently in the process of deprecating binary along some exercises that are very similar to it. Do we really need an update of the common tests here?

kytrinyx:
I think it's fine to add it, in case it takes time for tracks to deprecate it.

binary.json
kytrinyx: I think -1 is fine as long as we note somewhere that it's a place-holder for an error. Each track can deal with it as they please.

binary.json
kytrinyx: Yeah, I think it would probably be worth having separate tests for different types of error.

tommyschaefer:
Hi Everyone! I've made a few adjustments to the test data based on the feedback above.

I'm not entirely sure that I correctly implemented the change @Insti recommended:

> I think that the error cases should be split into individual tests, each of them deserves a description of why it is an invalid input which will help when coding the solution.

Did you mean to break each error case into it's own test case, or to just split the different "kinds" of error into different cases?

So sorry for the confusion and thank you so much for the feedback! 😄

Insti:
@tommyschaefer thanks for your continued progress on this.

Each thing you're checking for should have it's own (nameable) test case.
The things being tested should NOT be in arrays.

If you cannot think of a distinct name, this is a sign that you might not be testing anything useful.

binary.json
```diff
@@ -46,8 +46,13 @@
          "expected": 31
       },
       {
-         "description": "invalid binary numbers raise an error",
-         "binary": ["012", "10nope", "nope10", "10nope10", "001 nope", "2"],
+         "description": "numbers other than one and zero raise an error",
+         "binary": ["012", "2"],
```
Insti: This should become 2 tests, 
```
         "description": "2 is not a valid binary digit and so should raise an error",
         "binary":  "2",
         "expected" :  null,
```
and
```
         "description": "A number containing an invalid binary digit is invalid and should raise an error",
         "binary":  "01201",
         "expected":  null,
```

Discussion required: I'm not 100% sure null is the correct expected value here if we want to indicate in a general way that an error should be raised.

binary.json
```diff
@@ -46,8 +46,13 @@
...
+         "expected": -1
+      },
+      {
+         "description": "containing letters raises an error",
+         "binary": ["10nope", "nope10", "10nope10", "001 nope"],
```
Insti: Binary should not be an array.
See my other comment above about splitting arrays into individual tests.

Insti:
I got a bit trigger happy on merge for the [x-ruby generator](https://github.com/exercism/xruby/pull/394) before this was finalised. In order to not have the x-ruby build break, I've merged this in it's current state and will create a new issue for the test case splitting.

kytrinyx:
> I've merged this in it's current state and will create a new issue for the test case splitting.

That's a good solution! Thank you.

PR #298, opened 2016-07-19
Correct definition for the Triangle inequality

matthewfischer:
"or equal to" was missing.

In mathematics, the triangle inequality states that for any triangle,
the sum of the lengths of any two sides must be greater than or equal to
the length of the remaining side.

Source: https://en.wikipedia.org/wiki/Triangle_inequality

### DIFF
--- a/triangle.md
+++ b/triangle.md
@@ -2,6 +2,6 @@ The program should raise an error if the triangle cannot exist.
 
 ## Hint
 
-The sum of the lengths of any two sides of a triangle always exceeds the
-length of the third side, a principle known as the _triangle
+The sum of the lengths of any two sides of a triangle always exceeds or 
+is equal to the length of the third side, a principle known as the _triangle
 inequality_.


NobbZ:
We need to be careful with this. There has been some discussion around this
some time ago. One of the problems is that some tracks do it as described
in the read me and others as it is defined in mathematics. Can't search or
link right now, because I'm on mobile right now.

Am 19.07.2016 20:45 schrieb "Matt Fischer" <notifications@github.com>:

matthewfischer:
Maybe we could just explain that in the README and post a link to the mathematical definition?

petertseng:
The previous discussion was at https://github.com/exercism/x-common/issues/202 and mainly centered around whether tracks should have tests for cases when `a + b = c`. Regardless of what the answer to *that* question is, the README should at least give the correct definition.

So I say 👍  to this.

PR #301, opened 2016-07-21
Updated bracket-push tests

ebiven:
Tests for a class of problems where an unopened bracket is closed for
solutions which only track opened brackets.

### DIFF
--- a/bracket-push.json
+++ b/bracket-push.json
@@ -41,6 +41,11 @@
       "expected": true
     },
     {
+      "description": "unopened closing brackets",
+      "input": "{[)][]}",
+      "expected": false
+    },
+    {
       "description": "unpaired and nested brackets",
       "input": "([{])",
       "expected": false


Insti:
Looks good to me.

petertseng:
I was trying to figure out what kind of solution this rejects that aren't already covered by case `"}{"`, but I guess it's that `"}{"` is paired but in the wrong order, while the new case has a bracket that isn't paired at all? Do I have the idea right?

ebiven:
Correct. This as a closing bracket that was never opened, but is otherwise valid. It identifies solutions like this that forget the third check.

* If opening bracket, push onto queue
* If closing bracket and matches last item added to queue, remove last item from queue
* If closing bracket and doesn't match last item in queue then unbalanced. <- This step
* Queue length should now equal 0.

petertseng:
OK, and the idea is that `"}{"` does not identify those solutions since `"}{"` also has the trait that there is an unclosed open.

So the test cases were missing a test with only unopened closes, and this is one. OK, 👍  then.

PR #305, opened 2016-07-28
binary: split error case tests

### LINKED ISSUE #303: binary: split error case tests
https://github.com/exercism/x-common/pull/296 added a .json data file containing the test cases for this exercise. 

However some of the error case tests had been grouped in an unhelpful way.

>         "description": "numbers other than one and zero raise an error",
>         "binary": ["012", "2"],

This should become 2 cases,

```
         "description": "2 is not a valid binary digit and so should raise an error",
         "binary":  "2",
         "expected" :  null,
```
and

```
         "description": "A number containing an invalid binary digit is invalid and should raise an error",
         "binary":  "01201",
         "expected":  null,
```

Also for
> 
         "description": "invalid binary numbers raise an error",
         "binary": ["012", "10nope", "nope10", "10nope10", "001 nope", "2"],

tommyschaefer:
Resolves #303

Does this seem like it covers all of the cases in the original? It felt to me like a number of the grouped assertions were testing the same thing, so I didn't include them all. I'm totally open to adding them, I just couldn't think of distinct names for all of them 😄!

Also, there has been some discussion about what the `expected` value should be. Both `null` and `-1` seem appropriate to me. I only chose to go with `-1` because I saw it used in the `hamming` data, so I assumed that was the standard. If this isn't the case, I'm happy to change it!

Thank you so much for your time and feedback! 😄

### DIFF
--- a/binary.json
+++ b/binary.json
@@ -46,14 +46,34 @@
          "expected": 31
       },
       {
-         "description": "numbers other than one and zero raise an error",
-         "binary": ["012", "2"],
-         "expected": -1
+        "description": "2 is not a valid binary digit",
+        "binary": "2",
+        "expected": null
       },
       {
-         "description": "containing letters raises an error",
-         "binary": ["10nope", "nope10", "10nope10", "001 nope"],
-         "expected": -1
+        "description": "a number containing a non-binary digit is invalid",
+        "binary": "01201",
+        "expected": null
+      },
+      {
+        "description": "a number with trailing non-binary characters is invalid",
+        "binary": "10nope",
+        "expected": null
+      },
+      {
+        "description": "a number with leading non-binary characters is invalid",
+        "binary": "nope10",
+        "expected": null
+      },
+      {
+        "description": "a number with internal non-binary characters is invalid",
+        "binary": "10nope10",
+        "expected": null
+      },
+      {
+        "description": "a number and a word whitespace spearated is invalid",
+        "binary": "001 nope",
+        "expected": null
       }
    ]
 }


Insti:
Thanks @tommyschaefer .

In this case I believe -1 is inappropriate because it is a valid decimal number, and it is possible to represent negative numbers in binary. (Although this exercise is only concerned with positive binary numbers.)

So `null` is a better return value to indicate invalid input.

I agree with what you've done to remove redundant test cases, nice work.

binary.json
```diff
@@ -46,14 +46,24 @@
          "expected": 31
       },
       {
-         "description": "numbers other than one and zero raise an error",
-         "binary": ["012", "2"],
-         "expected": -1
+        "description": "2 is not a valid binary digit and raises an error",
```
Insti: I'd remove 'and raises an error' from the descriptions.
The error raising is an extension added by the Ruby track.

The readme just says "The program should handle invalid inputs."

binary.json
```diff
@@ -46,14 +46,24 @@
...
+        "binary": "2",
+        "expected": -1
       },
       {
-         "description": "containing letters raises an error",
-         "binary": ["10nope", "nope10", "10nope10", "001 nope"],
```
Insti: The distinct cases we have here are:
* trailing non-binary characters
* leading non-binary characters
* internal non-binary characters
* a number and a word whitespace separated
They're probably all good to include.

Insti:
Looking good. :+1:

tommyschaefer:
@Insti Thank you! Would it be helpful if I squashed these commits?

tommyschaefer:
Commits have been squashed and I updated the PR description so that if / when this gets merged, #303 should also be closed 😀

Insti:
:crossed_fingers:

Insti:
They didn't close automatically, but since you reminded me I remembered to do it manually.
Thanks for all your work on this @tommyschaefer.

tommyschaefer:
Hmm... I think this might have just gotten closed rather than merged. Is that intentional?

And my pleasure, @Insti! 😄

Insti:
Gak! Thanks for spotting that!
(And that might have been why it didn't automatically close those others..)
More :star2: :star2: for you.

tommyschaefer:
Haha! No problem 😄 Thank you so much for all of your help and feedback :+1:

PR #306, opened 2016-07-28
connect: Add cases for 1x1 boards

petertseng:
I think these are important since they're an interesting case for when a
start space in the search is also a goal space. I advocate for this in
original #297.


Insti:
seconded :+1:

Issue #202, opened 2016-03-13
triangle: clarify confusion and adjust test suites


kytrinyx:
A few weeks ago I hastily submitted an issue about triangle to every single track that has implemented it. In addition to being hasty, it was also confusingly ambiguous in its phrasing. And wrong. See discussion in https://github.com/exercism/xerlang/issues/85

The issue is surrounding the definition of the basic triangle inequality: z ≤ x + y

Notice the **less than or equal to** in that equation. Our test suites have been implemented as though that was only **less than**. A triangle with sides 2, 2, and 4 is a triangle, it's just a degenerate triangle.

Within the context of Exercism, this is taking edge cases to the absurd. The exercise is supposed to be simple, and I don't think that it makes sense to distinguish between _not a triangle_ and _degenerate triangle_.

I think that the best solution here would be to have a test case for a triangle that obviously violates the triangle inequality, and simply avoid the edge case where it is degenerate.

This would mean that we need to check each track's implementation and make sure that any cases where x+y=z are removed, and that there is at least one test case for x+y < z, e.g. ~~5, 4, 8 or 5, 5, 9.~~ 5, 4, 10 or 5, 5, 11

@kotp @NobbZ @masters3d @hankturowski @behrtam @yurrriq Have I summarized the discussion in https://github.com/exercism/xerlang/issues/85 correctly? Does this sound like a good solution to this mess?

/cc @exercism/track-maintainers @exercism/project-management

NobbZ:
Yeah, this summary seems to be correct and I also think this solution is the best thing we con do, without either requiring to implement something wrong or making the exercise unnecessary complex.

petertseng:
> that there is at least one test case for x+y < z, e.g. 5, 4, 8 or 5, 5, 9

The examples you've given do not fit `x+y < z`; instead, they fit `x+y > z`. I believe you do mean `x+y < z` (so you want to change the examples, not the inequality)

hankturowski:
@kytrinyx You're 100% correct. We were hung up on a pedantic issue and losing sight of the purpose of the exercise. At least I was.

I like this solution as It neatly sidesteps the degenerate issue. Even better, if the user decides to add the tests for equality to their own test suite, they can decided how to handle it themselves and avoid more confusion.

We may want to leave a note for future track maintainers to avoid someone helpfully adding the tests back in.

kytrinyx:
[...]

PR #311, opened 2016-07-31
triangle: Add JSON test data

### LINKED ISSUE #264: triangle: potential new test (from PR in haskell track)
See: https://github.com/exercism/xhaskell/pull/142

Is the test relevant to all the tracks that implement triangle?

(note, we don't yet have a triangle.json - the way to resolve this issue is probably to do a survey of all of the existing implementations)

The list is in this massive JSON: http://x.exercism.io/v3/problems/data-todos (I really need to implement /v3/problems/:slug)

petertseng:
Pursuant to dicussion in #202, we won't have cases where a + b = c.
Examples are (2, 4, 2, Isosceles) and (1, 3, 4, Scalene).

As of this writing, 21 tracks implemented triangle:
* https://github.com/exercism/xcsharp/blob/master/exercises/triangle/TriangleTest.cs
* https://github.com/exercism/xcpp/blob/master/triangle/triangle_test.cpp
* https://github.com/exercism/xclojure/blob/master/exercises/triangle/test/triangle_test.clj
* https://github.com/exercism/xcoffeescript/blob/master/exercises/triangle/triangle_test.spec.coffee
* https://github.com/exercism/xlisp/blob/master/exercises/triangle/triangle-test.lisp
* https://github.com/exercism/xecmascript/blob/master/exercises/triangle/triangle.spec.js
* https://github.com/exercism/xelixir/blob/master/exercises/triangle/triangle_test.exs
* https://github.com/exercism/xelm/blob/master/exercises/triangle/TriangleTests.elm
* https://github.com/exercism/xerlang/blob/master/exercises/triangle/triangle_tests.erl
* https://github.com/exercism/xfsharp/blob/master/exercises/triangle/TriangleTest.fs
* https://github.com/exercism/xgo/blob/master/exercises/triangle/triangle_test.go
* https://github.com/exercism/xhaskell/blob/master/exercises/triangle/test/Tests.hs
* https://github.com/exercism/xjava/blob/master/exercises/triangle/src/test/java/TriangleTest.java
* https://github.com/exercism/xjavascript/blob/master/exercises/triangle/triangle.spec.js
* https://github.com/exercism/xlua/blob/master/exercises/triangle/triangle_spec.lua
* https://github.com/exercism/xnim/blob/master/triangle/triangle_test.nim
* https://github.com/exercism/xperl5/blob/master/triangle/cases.json
* https://github.com/exercism/xpython/blob/master/exercises/triangle/triangle_test.py
* https://github.com/exercism/xruby/blob/master/exercises/triangle/triangle_test.rb
* https://github.com/exercism/xscala/blob/master/exercises/triangle/src/test/scala/triangle_test.scala
* https://github.com/exercism/xswift/blob/master/exercises/triangle/TriangleTest.swift

This JSON file contains the combination of all their tests.
Most tracks have the same set of tests.

Haskell has a case specific to itself added in
exercism/xhaskell#142
This case is added to the common set.

The Go and Nim tracks have tests with +/- Infinity and NaN.
These cases are not added.

Closes #264
Closes exercism/todo#157

### DIFF
--- a/triangle.json
+++ b/triangle.json
@@ -1,0 +1,90 @@
+{
+    "#": [
+        "Pursuant to discussion in #202,",
+        "we have decided NOT to test triangles where all side lengths are positive but a + b = c.",
+        "e.g: (2, 4, 2, Isosceles), (1, 3, 4, Scalene).",
+        "It's true that the triangle inequality admits such triangles.",
+        "These triangles have zero area, however.",
+        "They're degenerate triangles with all three vertices collinear.",
+        "(In contrast, we will test (0, 0, 0, Illegal), as it is a point).",
+
+        "The expectation are given as strings,",
+        "but your language may use the appropriate representations.",
+        "For example, enums, variants, or tagged unions all are viable candidates.",
+
+        "Your track may choose to have the 'illegal' result be another member of the enum/variant/union/etc.,",
+        "or instead to signal an error/exception/etc. on an illegal triangle.",
+
+        "If appropriate for your track, you'll need to ensure that no pair of expected values are equal.",
+        "Otherwise, an implementation that always returns a constant value may falsely pass the tests.",
+        "See https://github.com/exercism/xgo/pull/208"
+    ],
+    "cases": [
+        {
+            "description": "equilateral triangle has all sides equal",
+            "sides": [2, 2, 2],
+            "expected": "equilateral"
+        },
+        {
+            "description": "larger equilateral triangle",
+            "sides": [10, 10, 10],
+            "expected": "equilateral"
+        },
+        {
+            "description": "isosceles triangle with last two sides equal",
+            "sides": [3, 4, 4],
+            "expected": "isosceles"
+        },
+        {
+            "description": "isosceles triangle with first two sides equal",
+            "sides": [4, 4, 3],
+            "expected": "isosceles"
+        },
+        {
+            "description": "isosceles triangle with first and last sides equal",
+            "sides": [4, 3, 4],
+            "expected": "isosceles"
+        },
+        {
+            "description": "isosceles triangle with unequal side larger than equal sides",
+            "sides": [4, 7, 4],
+            "expected": "isosceles"
+        },
+        {
+            "description": "scalene triangle has no equal sides",
+            "sides": [3, 4, 5],
+            "expected": "scalene"
+        },
+        {
+            "description": "larger scalene triangle",
+            "sides": [10, 11, 12],
+            "expected": "scalene"
+        },
+        {
+            "description": "scalene triangle with sides in descending order",
+            "sides": [5, 4, 2],
+            "expected": "scalene"
+        },
+        {
+            "#": "Your track may choose to skip this test and deal only with integers if appropriate",
+            "description": "small scalene triangle with floating point values",
+            "sides": [0.4, 0.6, 0.3],
+            "expected": "scalene"
+        },
+        {
+            "description": "a triangle violating the triangle inequality is illegal",
+            "sides": [7, 3, 2],
+            "expected": "illegal"
+        },
+        {
+            "description": "two sides equal, but still violates triangle inequality",
+            "sides": [1, 1, 3],
+            "expected": "illegal"
+        },
+        {
+            "description": "triangles with all sides zero are illegal",
+            "sides": [0, 0, 0],
+            "expected": "illegal"
+        }
+    ]
+}


triangle.json
```diff
@@ -0,0 +1,96 @@
...
+
+        "The expectation are given as strings,",
+        "but your language may use the appropriate representations.",
+        "For example, enums, variants, or tagged unions all are viable candidates.",
+
+        "Your track may choose to have the 'illegal' result be another member of the enum/variant/union/etc.,",
+        "or instead to signal an error/exception/etc. on an illegal triangle.",
+
+        "If appropriate for your track, you'll need to ensure that no pair of expected values are equal:",
+        "See https://github.com/exercism/xgo/pull/208"
```
petertseng: could have sworn there was another, but now I can't find it.

triangle.json
```diff
@@ -0,0 +1,98 @@
...
+        "The expectation are given as strings,",
+        "but your language may use the appropriate representations.",
+        "For example, enums, variants, or tagged unions all are viable candidates.",
+
+        "Your track may choose to have the 'illegal' result be another member of the enum/variant/union/etc.,",
+        "or instead to signal an error/exception/etc. on an illegal triangle.",
+
+        "If appropriate for your track, you'll need to ensure that no pair of expected values are equal.",
+        "Otherwise, an implementation that always returns a constant value may falsely pass the tests.",
+        "See https://github.com/exercism/xgo/pull/208"
```
petertseng: also applicable to sublist because https://github.com/exercism/xpython/issues/342 - maybe I'll add a comment to that json file

triangle.json
```diff
@@ -0,0 +1,98 @@
+{
+    "#": [
+        "Pursuant to discussion in #202,",
+        "we have decided NOT to test triangles where a + b = c.",
+        "e.g: (2, 4, 2, Isosceles), (1, 3, 4, Scalene).",
```
Insti: `[10, 10, 0]` and `[0, 0, 0]` are both triangles where a + b = c

Insti: Reductio ad absurdum: If we're not including degenerate triangles, these tests should not be present either.

kotp: Since 10, 10, 0 is a line and 0, 0, 0 is a point, and the discussion happened where we are not including the fringe of degenerate triangles, they should not be here.

kotp: They are testing for failure, not for inclusion.

triangle.json
```diff
@@ -0,0 +1,98 @@
...
+        {
+            "description": "triangles with all sides zero are illegal",
+            "sides": [0, 0, 0],
+            "expected": "illegal"
+        },
+        {
+            "description": "negative side lengths are illegal",
+            "sides": [3, 4, -5],
+            "expected": "illegal"
+        }
```
kotp: Should negative size lengths be illegal?  They should not survive as negative, but perhaps be normalized to positive numbers?

triangle.json
petertseng: One possible course of action may be to clarify "we have decided NOT to test triangles where a, b, c are all positive and a + b = c" because as I see it, the point of (0, 0, 0, Illegal) and (10, 10, 0, Illegal) are to test that 0 is not a valid side length, not just to be a case like (2, 4, 2, Isosceles) or (1, 3, 4, Scalene).

However, it may be the case that cases involving zero should also get thrown out for the same reasons as in #202: they raise questions about degeneracy. Can we call it a triangle even if some side lengths are zero? And unrelated to this exercise but related to the previous question: Can we consider every triangle to be a quadrilateral with exactly one side length zero? I don't have a definitive answer.

- If we can find a definitive reason to reject 0 as a valid side length, (0, 0, 0, Illegal) and (10, 10, 0, Illegal) should stay and the above sentence should be amended.
- If we cannot, the cases should go.

triangle.json
petertseng: Do you strongly advocate for this? If you do, I would say that accompanying this case should be also cases, both legal and illegal, with two and three negative side lengths, and covering all triangle types as well. (2, 4, -7, Illegal), (2, -2, -2, Equilateral), (-7, 7, 4, Isosceles), etc.

I argue against, because I say the function under test should be considered to receive side lengths after normalisation and so it cannot receive a negative value.

But my opposition is mostly pragmatic instead of grounded in solid mathematical foundation: I don't want to be the one coming up with all the cases. Of course, *that* problem can be solved in the obvious fashion.

triangle.json
kotp: I think we had already found a definitive reason in prior discussions, regarding degenerative triangles.  Doesn't 0 as a any length create a degenerative triangle, and therefore is rejected on that basis?

I think we have decided to test, and decided to fail them, so the sentence there does not make sense.

triangle.json
kotp: I think one test that given any given valid (negative) number in any position would suffice.  Normalize all values to the 'absolute' value. We don't, after all, test against a coordinate plane, (or, rather, a location in space) so rotation in this case shouldn't matter.

What you are saying is that the user of the Triangle class should decide to normalize or not, in which case, the burden of testing is only for positive lengths, which is also fine.

I say, then, match it to the readme, and let the discussion ensue on the exercism site.

triangle.json
petertseng: (0, 0, 0, Illegal) and (10, 10, 0, Illegal) are hereby removed.

kotp:
I just solved this in the Ruby track, and noticed that at first we test for 0's, then we test for "Triangle Equality" which provides the standard, simple, definition of a Triangle, mathematically.  Once the tests for Triangle Equality happen, the tests for zero can go away.

triangle.json
rbasso: I'm not a mathematician, so my question may be dumb.

What is the interpretation of negative sides in a triangle? Could you point me some links?

rbasso:
I agree with @petertseng. The function should receive the side's lengths, so they have to be positive.

I don't understand what would be the interpretation of a negative size, because a vector length cannot be negative:

https://en.wikipedia.org/wiki/Triangle_inequality
> In Euclidean geometry and some other geometries, the triangle inequality is a theorem about distances, and it is written using vectors and vector lengths (norms):
>     ∥ x + y ∥ ≤ ∥ x ∥ + ∥ y ∥

https://en.wikipedia.org/wiki/Norm_(mathematics)

> In linear algebra, functional analysis, and related areas of mathematics, a norm is a function that assigns a strictly positive length or size to each vector in a vector space—save for the zero vector, which is assigned a length of zero. A seminorm, on the other hand, is allowed to assign zero length to some non-zero vectors (in addition to the zero vector).

kotp:
Those symbols (The double pipes) indicate absolute values, correct?  This indicates that a "negative" distance is simply a  distance with perspective (or direction), the "length" is absolute.  The choice to normalize or not can be made in the method/function or required to be normalized outside of said function.  Which suggests that the values for those vectors could have been negative, but are normalized in that function.  Hence my view that taking a negative value may not be inappropriate.  But we are free to choose.

The formula you link to using absolute function on the given lengths, suggests that we could go either way, enforcing that we are given "absolute" values.

Given the links you provide and that you understand the formulas, I would challenge your feign of non-mathyness... to use a technical term.

rbasso:
> The formula you link to using absolute function on the given **lengths**...

I think that this is where we disagree. I may be wrong but I think that - in that formula - `x`, `y` are vectors in a space.

> ... This indicates that a "negative" distance is simply a distance with perspective (or direction)...

Isn't signed real number just a 1-dimensional vector? If we treat the values in the exercise as 1-dimensional vectors, only degenerated triangles where `a + b = c` would be possible, right?

> Given the links you provide and that you understand the formulas, I would challenge your feign of non-mathyness... to use a technical term.

I like math, but my knowledge is too shallow to allow me any confidence! :smile:

Insti:
The inspiration for this exercise is the [triangle project](https://github.com/edgecase/ruby_koans/blob/master/src/about_triangle_project.rb) from rubykoans.

The [second part](https://github.com/edgecase/ruby_koans/blob/master/src/about_triangle_project_2.rb) of that problem includes uses the following test cases:
>
    assert_raise(TriangleError) do triangle(0, 0, 0) end
    assert_raise(TriangleError) do triangle(3, 4, -5) end
    assert_raise(TriangleError) do triangle(1, 1, 3) end
    assert_raise(TriangleError) do triangle(2, 4, 2) end

I think, we *should* include degenerate triangles (above: 1, 3 & 4) in the test cases.   
I think, we should **not** include triangles with negative edge lengths (above: 2).

triangle.json
petertseng: > I think one test that given any given valid (negative) number in any position would suffice. 

*If* we decide that we should test negative values, I would argue for at least adding the three cases I describe. We wouldn't want someone to consider (2, -2, -2) an isosceles triangle rather than equilateral, after all.

> I say, then, match it to the readme

And now I have to awkwardly announce that it is not noted in the README. We can fix that, of course!!!

petertseng:
> I think, we *should* include degenerate triangles (above: 1, 3 & 4) in the test cases.

I do not care either way about (0, 0, 0) or the related case (10, 10, 0). I will gladly add (0, 0, 0) back in if all interested are agreed.

(1, 1, 3) is not a degenerate triangle; it is a violation of the triangle inequality. I say we have to include this one no matter what (and luckily it is already included).

I also do not care about (2, 4, 2), but many other people in #202 do. I only see myself as enacting the will of the people. I am happy to add it if that is what people want.
Corollary question: If we add (2, 4, 2), should we expect it to be illegal or isosceles? Note for historical context that we [asked this question of all tracks in January](https://github.com/search?q=org%3Aexercism+triangle%3A+%22incorrect+test%22&ref=searchresults&type=Issues&utf8=%E2%9C%93) and also recently merged https://github.com/exercism/x-common/pull/298.

petertseng:
> I think, we should **not** include triangles with negative edge lengths (above: 2).

My current inclination is to keep a case such as (3, 4, -5). If we do not, we may see submitted solutions that *would* say (3, 4, -100) is a valid scalene triangle, if they were posed such a question. 

I can be peruaded that we should not include such caes. For example, in https://github.com/exercism/x-common/pull/287#issuecomment-231602657 there was a concern about tests having edge-itis. If we think that's what's happening here, very glad to drop the test instead.

kotp:
If 2, 4, 2 is degenerative it should be illegal.  (and it is degenerative).

petertseng:
Time to decide!

Here is my understanding of all test cases that require a decision, and what I have done with them in the current state of this PR:

* (0, 0, 0): https://github.com/exercism/x-common/pull/311#issuecomment-236503092 indicates support, and most tracks already have. **Include**.

* (3, 4, -5): **Exclude** because https://github.com/exercism/x-common/pull/311#issuecomment-236503092 shows two votes of support for **Exclude**.  I will speak briefly about alternatives:
  * For *Include, Illegal*: I do think some tracks want to test this, for prevention of code that would declare (3, 4, -100) a legal Scalene triangle, but I am happy to remove the test case to avoid edge-itis!
  * For *Include, Legal*: If we make it a legal triangle in this JSON file as https://github.com/exercism/x-common/pull/311#discussion_r72902254 reasonably suggests, tracks will be in direct opposition to what the JSON file says until all tracks update. I will not be the person to submit the PR that puts the file into such a state. I leave that to an interested party, because I'm not interested one bit in making that change.

* (2, 4, 2), (1, 3, 4), etc.: **Exclude**. All right, I know that https://github.com/exercism/x-common/pull/311#issuecomment-236508127 supports *Include, Illegal* and probably https://github.com/exercism/x-common/pull/311#issuecomment-236503092 supports *Include, Illegal* as well, but I can't in good conscience include it in light of #202. Once again I think I have to leave it to an interested party.

However, one thing I will do is look through the existing tracks and see how they treat such cases.

* Case not present (6): C++ (via https://github.com/exercism/xcpp/pull/86), ECMAScript (via https://github.com/exercism/xecmascript/pull/198), Javascript (via https://github.com/exercism/xjavascript/pull/164), Lua (via https://github.com/exercism/xlua/pull/110), Scala (via https://github.com/exercism/xscala/pull/102), Swift (via https://github.com/exercism/xswift/pull/107)
* Case expected to say triangle is illegal (14): C#, Clojure, Coffeescript, Lisp, Elixir, Elm, Erlang, F#, Haskell, Java, Nim, Perl, Python, Ruby
* Case expected to say triangle is legal (1): Go (via https://github.com/exercism/xgo/pull/232)

Guess Go's the outlier, and that's my doing too. Nevertheless, that's moot if the case is to be removed.

kytrinyx:
> (3, 4, -5): Exclude

I would vote to Exclude, and support the suggestion that if it gets included later we should update the README.

> (2, 4, 2), (1, 3, 4), etc.: Exclude

I would also vote to Exclude here, with the reasoning being that the difference between "not a triangle" and "triangle but degenerate" is more about the math than about the interesting part of the programming challenge. I do tend to get pulled into "yes but it's WRONG" discussions, because I am a hopeless pedant (at least when I've had enough sleep to pay attention to the discussion), but I think that it is not helpful to the exercise itself to focus on this issue.

petertseng:
Your choice is made!

Thanks, everyone.

Issue #230, opened 2016-04-19
How do we express the intention of non-obvious test cases in our canonical-data.json files?


petertseng:
the TODO issues like https://github.com/exercism/todo/issues/100 point us to some examples as guidelines for what should go in a test inputs/outputs json file. It is probably useful to assemble some guide somewhere about that using what we know of our existing json files.

This is prompted by https://github.com/exercism/discussions/issues/10#issuecomment-211669920 where I muse that it could have been good if some non-obvious corner cases had been commented with their purpose. Sure, in many situations the test name is enough, but when it isn't, I'm always for more documentation.

kotp:
I had noted in [this comment](https://github.com/exercism/x-common/issues/166#issuecomment-209767629) how(what appears to be) a comment made it to the test title.  That should not happen.  If we do need comments, they should be in the json file as "notes" or something similar, rather than the title.

kytrinyx:
Agreed, it would be really great to have a comprehensive guide (that we could add to). Probably somewhere in [the contributing guide](https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md).

We haven't quite worked out the format for comments--if you have suggestions for how to improve them, that would be great.

Maybe an optional key "notes" for any test case? That way we can put notes as comments in the test suite if they exist.

kotp:
Hey!  "notes" is a good idea. :laugh:  I was thinking that messages for the tests may be helpful.  The generated tests simply fail at the moment with no clarification of what may be the problem, other than the test title.  So we have two kinds of "comments" to consider.

I am not familiar enough with other testing frameworks to know how error messaging is handled, but "test-message" as a key for the test messages when deemed appropriate, and "notes" for commentary notes, such as "explanation for a magic number" may be nice.

kytrinyx:
We did have messages for some tests, but I found that it was incredibly redundant. I get the benefit of working with both Go and Ruby which have very different approaches to testing. I think we can use the test data to generate messages where necessary (e.g. leap year test suite could have a message that says "case: 1998" or something like that).

I really like the idea of having notes that are meant for the end user, though, for places where it makes sense.

kotp:
[...]

PR #314, opened 2016-08-01
raindrops: remove redundant description from JSON file

petertseng:
In all cases, the description is just the string representation of the
input number. This seems redundant and unnecesary, and I would like to
invoke "Don't repeat yourself" and remove these messages.

### DIFF
--- a/raindrops.json
+++ b/raindrops.json
@@ -1,77 +1,62 @@
 {
    "cases": [
       {
-         "description": "1",
          "number" : 1, 
          "expected": "1"
       },
       {
-         "description": "3",
          "number" : 3, 
          "expected": "Pling"
       },
       {
-         "description": "5",
          "number" : 5, 
          "expected": "Plang"
       },
       {
-         "description": "7",
          "number" : 7, 
          "expected": "Plong"
       },
       {
-         "description": "6",
          "number" : 6, 
          "expected": "Pling"
       },
       {
-         "description": "9",
          "number" : 9, 
          "expected": "Pling"
       },
       {
-         "description": "10",
          "number" : 10, 
          "expected": "Plang"
       },
       {
-         "description": "14",
          "number" : 14, 
          "expected": "Plong"
       },
       {
-         "description": "15",
          "number" : 15, 
          "expected": "PlingPlang"
       },
       {
-         "description": "21",
          "number" : 21, 
          "expected": "PlingPlong"
       },
       {
-         "description": "25",
          "number" : 25, 
          "expected": "Plang"
       },
       {
-         "description": "35",
          "number" : 35, 
          "expected": "PlangPlong"
       },
       {
-         "description": "49",
          "number" : 49, 
          "expected": "Plong"
       },
       {
-         "description": "52",
          "number" : 52, 
          "expected": "52"
       },
       {
-         "description": "105",
          "number" : 105, 
          "expected": "PlingPlangPlong"
       }


petertseng:
Note that I would accept as an alternative "so let's actually add useful information into the description", given a suggestion of what that useful information would be.

kotp:
I think removing the descriptions will break the generators... so something should be there, rather than removed.

petertseng:
One solution might be that the generators use the number as the description (converting to a string if required) - after all, it's already the case that the number is the description.

If that is not possible, I submit for consideration my alternative idea - we need to come up with better descriptions that aren't repetitive of the input.

Insti:
> removing the descriptions will break the generators

If this is the case the generators are broken and should be fixed.

rbasso:
There is at least one json file without descriptions:

https://github.com/exercism/x-common/blob/master/sum-of-multiples.json

petertseng:
It is known to me that [the Go generator](https://github.com/exercism/xgo/blob/master/exercises/raindrops/example_gen.go) completely ignore the description.

It is also acknowledged by me that [the Ruby generator](https://github.com/exercism/xruby/blob/master/lib/raindrop_cases.rb#L3) uses the description and would have to instead use the number.

kotp:
Yes, put simply, Ruby is currently using the description, so it breaks Ruby's generator.  Not a big deal, just wanted to mention the impact.  (Simple enough fix...)

petertseng:
I searched the org for "gen" and xgo was all I got. I additionally knew that Ruby had generator, so I know that "gen" could have false negative. Nevertheless, I am not currently aware of other track generating tests, so I think those most affected by the change are ready to go with this.

kotp:
Hamming used description, which I think was the "go to" to learn how to do the generators, which, by @insti's comment that the generator is broken, then it is by virtue of the exemplar.

@insti, this is what you meant, correct? That description should not have been relied upon, and so the generator is thusly broken?

Is it wrong to use the description in the generator?

kotp:
Thanks for the work on the common data, @petertseng .

Insti:
@kotp 

I am not suggesting that a currently functioning generator is broken because it may stop working in the future if the json file it is using changes.

The dependency graph should be such that the language tracks depend on x-common
x-common should be able to do whatever it wants to the json files and the language tracks need to be able to cope with that.

It is fine to use the description in the generator, if the description disappears from the json and the generator stops working, then it is the generator that needs to be fixed, not the json.

petertseng:
At the same time, I do wonder in the future if we will start to want to settle on a more common schema (if you will) for the JSON files. Consistency may help tracks that have generators. What that would look like isn't clear to me yet. Depends on what goals we want to achieve, and maybe more tracks/exercises using generators would help discover some requirements.

I only bring this up now because this PR may have called into question a convention of "All cases should have a `description` key". I keep meaning to write some things down in #230, but I guess I  don't have a clear idea of the goals I want to achieve by writing things down.

Insti:
Some kind of schema would be good, at least a guarantee that they will all adhere to some kind of minimal guaranteed structure: 
```
{
   "cases": [
      {
         "input" : [1, 1, "+"],
         "expected": 2
      },
   ],
}
```

kytrinyx:
> At the same time, I do wonder in the future if we will start to want to settle on a more common schema (if you will) for the JSON files.

Yeah, this has come up a few times. I don't know what common schema makes sense at this point—and I wanted to avoid generalizing too early (speculatively). Now we probably have enough data and experience to make some guesses.

PR #319, opened 2016-08-03
sublist: Warn about comparing values

petertseng:
In some languages, it is necesary to ensure that no pair of the four
possible expected values are equal. Otherwise, an essentially-empty
implementation may pass tests without actually doing any useful work.

Example: https://github.com/exercism/xpython/issues/342

This same warning was added to the triangle JSON file in #311.


ErikSchierboom:
Looks good to me.

Issue #321, opened 2016-08-06
Is "write a program" accurate?


petertseng:
We have a lot of descriptions that say "write a program". As https://github.com/exercism/exercism.io/issues/2746 explains, this may set up a wrong expectation from students that they are writing e.g. something with a main function (or equivalent entry point in language of choice).

But it seems that most tracks I've seen simply ask the student to write a function. Most languages then verify the behavior with unit tests against that function.

In the interest of setting up expectations correctly, shall we replace "write a program" with "write a function"?

Things I would like to check before doing that:
- Are there any tracks that do, in fact, expect the student to write a full-fledged program?
- Are there any tracks in which the equivalent concept is not called a function? (Perhaps it's called  a procedure instead? Or something completely different?)
  - Even if there aren't, what if we add such a track in the future?
  - If there currently are or could be in the future, should we do anything about it? In other words, do we decide that we'll simply use a "function" as a generic terminology, regardless of how this concept maps onto the target language?

Anything else I'm missing?

Appendix: 

```
$ grep -li "write a program" *.md
change.md
crypto-square.md
etl.md
food-chain.md
grains.md
luhn.md
protein-translation.md
scale-generator.md
sum-of-multiples.md
transpose.md
two-bucket.md

$ grep -li "write a program" *.yml
allergies.yml
all-your-base.yml
anagram.yml
beer-song.yml
binary-search-tree.yml
binary-search.yml
binary.yml
change.yml
flatten-array.yml
food-chain.yml
gigasecond.yml
grains.yml
hamming.yml
hello-world.yml
hexadecimal.yml
house.yml
kindergarten-garden.yml
largest-series-product.yml
leap.yml
luhn.yml
matrix.yml
minesweeper.yml
nth-prime.yml
ocr-numbers.yml
octal.yml
paasio.yml
palindrome-products.yml
parallel-letter-frequency.yml
pascals-triangle.yml
phone-number.yml
point-mutations.yml
protein-translation.yml
proverb.yml
queen-attack.yml
raindrops.yml
rna-transcription.yml
robot-name.yml
saddle-points.yml
say.yml
scale-generator.yml
scrabble-score.yml
secret-handshake.yml
series.yml
sieve.yml
space-age.yml
sum-of-multiples.yml
transpose.yml
triangle.yml
trinary.yml
twelve-days.yml
two-bucket.yml
word-count.yml
wordy.yml
zebra-puzzle.yml
```

kotp:
[...]

Issue #290, opened 2016-07-06
[hello-world] Consistency of the empty string


NobbZ:
In some languages we are able to define functions with the same name and different aritys. Those languages will automatically treat `hello()` different to `hello("")`.

Other languages do not have this luxury or do allow to use default values for arguments.

So in the C track I realized, the empty string is used to signal “nothing given”, while `NULL` might be more appropriate.

Other languages have similar (perhaps more noisy) ways to differ between an empty string and “nothing given” by `Maybe` or `Option` style types.

Therefore I wanted to start a discussion about treatement of the empty string.

1. Treat it as if we had called without a name, so we create `"Hello, World!"` as output
2. Treat it as proper name and return `"Hello, !"`
3. Leave it open to language maintainers

I tried to look up at least the tracks I already solved `hello-world` in:

* Different functions by arity: Erlang, Elixir
* empty string is nothing: C
* empty string is proper name: Rust (`Option`-type)
* testsuite is unclear about the meaning of the empty string: Scala (default mechanisms used based on NULL-references), Crystal (as Scala), Scheme (as Scala), PHP (default mechanisms based on PHP stuff I deny to delve into)

Especially in the last group, it mostly depends on the student how he implements the handling of the default value. I've seen solutions in such exercises that check default `name` to `NULL` (or similar) or `""` and then do a conditional reassignment in the functions body (which in my opinion defeats the purpose of default values for argument).

Especially this wild growing stuff is in my opinion dangerous. In many tracks there are not many people who actually review, and even after `hello-world` is not hidden any more, many reviewers seem to just skip it.

kytrinyx:
> which in my opinion defeats the purpose of default values for argument

Agreed, if we can set a default, it should be "world", not nil or an empty string.

Also agreed that if the language can differentiate between empty string and nil, it should probably treat the empty string as a name.

In Go you could, potentially pass a pointer to a string so that you can get a nil, but the idiom is to pass a string, and if it's empty, conditionally reassign it, so in that case I would not bring up nils at all in the test suite.

kotp:
>  and even after hello-world is not hidden any more, many reviewers seem to just skip it.

Since the hello-world comment from RIkki is "Usually the next step is to have a conversation about your code and iterate on your solution. ... we won't be encouraging conversations and iterations... " it seems to make sense that most reviewers would skip it, since the exercise is just a "If you can get here, you can get to the next step" kind of orientation thing.

NobbZ:
[...]

PR #322, opened 2016-08-06
hello-world: Be more function-oriented, not program-oriented

petertseng:
As https://github.com/exercism/exercism.io/issues/2746 rightfully
explains, the wording of "program" implies that the student is expected
to write a full program with a main function (or equivalent entry point
in language of choice).

Further, the wording of "greets" slightly implies that the function is
expected to put the result on STDOUT.

To remedy this, "program" is replaced with "function" and we specify
that the function should return the greeting.

Note that as #321 states, this is just one of many problems that has
"write a program" in its description. However, the "greets" -> "returns"
wording is an issue not covered by #321, so we deal with it here.

Closes https://github.com/exercism/exercism.io/issues/2746

### DIFF
--- a/hello-world.md
+++ b/hello-world.md
@@ -6,11 +6,12 @@
 
 ## Specification
 
-The `Hello World!` program will greet me, the caller.
+Write a `Hello World!` function that can greet someone given their name.
+The function should return the appropriate greeting.
 
-If I tell the program my name is Alice, it will greet me by saying "Hello, Alice!".
+For an input of "Alice", the response should be "Hello, Alice!".
 
-If I neglect to give it my name, it will greet me by saying "Hello, World!"
+If a name is not given, the response should be "Hello, World!"
 
 ## Test-Driven Development
 
--- a/hello-world.yml
+++ b/hello-world.yml
@@ -1,4 +1,4 @@
 ---
-blurb: 'Write a program that greets the user by name, or by saying "Hello, World!" if no name is given.'
-source: "This is a program to introduce users to using Exercism"
+blurb: 'Write a function that greets the user by name, or by saying "Hello, World!" if no name is given.'
+source: "This is an exercise to introduce users to using Exercism"
 source_url: "http://en.wikipedia.org/wiki/%22Hello,_world!%22_program"


hello-world.md
```diff
@@ -6,11 +6,11 @@
 
 ## Specification
 
-The `Hello World!` program will greet me, the caller.
+Write a `Hello World!` function that takes a single input (my name) from me, the caller, and returns an appropriate string.
```
petertseng: I think I'm going to ditch the "from me"

hello-world.md
```diff
@@ -6,11 +6,11 @@
...
 ## Specification
 
-The `Hello World!` program will greet me, the caller.
+Write a `Hello World!` function that takes a single input, the name of someone to greet, and returns an appropriate greeting.
 
-If I tell the program my name is Alice, it will greet me by saying "Hello, Alice!".
+For an input of "Alice", the response should be "Hello, Alice!".
 
-If I neglect to give it my name, it will greet me by saying "Hello, World!"
+If a name is not given, the response should be "Hello, World!"
```
petertseng: as for whether this in particular is a good idea, that's #290, not here.

hello-world.md
```diff
@@ -6,11 +6,11 @@
 
 ## Specification
 
-The `Hello World!` program will greet me, the caller.
+Write a `Hello World!` function that takes a single input, the name of someone to greet, and returns an appropriate greeting.
```
Insti: "that takes a single input" 

The Ruby and I believe other versions also allow calling the method without an argument.

Insti:
I think these changes are definitely an improvement.

hello-world.md
petertseng: Ah, that's right, need to deal with multiple arities in a way that makes sense. Getting harder and harder to write something to write something generic over all languages!

Thinking through some possibilities:

* "Write a `Hello World!` function that can take a single input" - This, I suppose, leaves open the possibility that it can take something else other than a single input, although this wording doesn't make it explicit what that something else is.
* "Write a `Hello World!` function that can take a single optional input" - possibly, though some languages won't allow optional inputs, so this sentence may get confusing for those languages.
* "Write a `Hello World!` function that can greet someone given their name" - perhaps just don't mention the function inputs at all?

Insti: > "Write a Hello World! function that can greet someone given their name" - perhaps just don't mention the function inputs at all?

:+1:

petertseng: there it is

PR #324, opened 2016-08-08
Fix minor mistake in sgf-parsing exercise

rmunn:
In SGF, the node labeled `dd` is the 4-4 point, not the 3-3 point.

Also, SGF stands for [Smart Game Format](http://senseis.xmp.net/?SmartGameFormat), not Standard Game Format.

Neither of these will affect the actual exercise, but it's always worth getting the terms right in the README.


Insti:
Thanks @rmunn :star:

PR #327, opened 2016-08-08
Add grep test data and update readme

ErikSchierboom:
The current `grep` exercise is a bit odd. If mentions that the user should implement a function that returns pairs of line numbers and matching lines. However, this does not easily fit the latter suggestion that the `grep` function should also support flags, of which one flag is intended as only listing file names.

In this PR, I've reworded the exercise to make it more like you're re-implementing the Unix `grep` command, as now it expects just a string to be output (which should correspond to the string output by the `grep` command).

Furthermore, I've added a suite of test cases.

### DIFF
--- a/grep.json
+++ b/grep.json
@@ -1,0 +1,251 @@
+{ 
+    "#": [
+        "JSON doesn't allow for multi-line strings, so all outputs are presented ",
+        "here as arrays of strings. It's up to the test generator to join the ", 
+        "lines together with line breaks.",
+        "",
+        "The tests are divided into two groups: ",
+        "* Grepping a single file",
+        "* Grepping multiple files at once",
+        "",
+        "The language track implementing this exercise should",
+        "ensure that when the tests run, three files are created",
+        "with the following contents. The file names and their",
+        "contents are listed in the 'files' section below."
+    ],
+    "files": [
+        {
+            "name": "iliad.txt",
+            "contents": [
+                "Achilles sing, O Goddess! Peleus' son;",
+                "His wrath pernicious, who ten thousand woes",
+                "Caused to Achaia's host, sent many a soul",
+                "Illustrious into Ades premature,",
+                "And Heroes gave (so stood the will of Jove)",
+                "To dogs and to all ravening fowls a prey,",
+                "When fierce dispute had separated once",
+                "The noble Chief Achilles from the son",
+                "Of Atreus, Agamemnon, King of men."
+            ]
+        },        
+        {
+            "name": "midsummer-night.txt",
+            "contents": [
+                "I do entreat your grace to pardon me.",
+                "I know not by what power I am made bold,",
+                "Nor how it may concern my modesty,",
+                "In such a presence here to plead my thoughts;",
+                "But I beseech your grace that I may know",
+                "The worst that may befall me in this case,",
+                "If I refuse to wed Demetrius."
+            ]
+        },
+        {
+            "name": "paradise-lost.txt",
+            "contents": [
+                "Of Mans First Disobedience, and the Fruit",
+                "Of that Forbidden Tree, whose mortal tast",
+                "Brought Death into the World, and all our woe,",
+                "With loss of Eden, till one greater Man",
+                "Restore us, and regain the blissful Seat,",
+                "Sing Heav'nly Muse, that on the secret top",
+                "Of Oreb, or of Sinai, didst inspire",
+                "That Shepherd, who first taught the chosen Seed"
+            ]
+        }
+    ],
+    "single-file": {
+        "description": ["Test grepping a single file"],
+        "cases": [
+            {
+                "description": "One file, one match, no flags",
+                "pattern": "Agamemnon",
+                "flags": [],
+                "files": ["iliad.txt"],
+                "expected": [
+                    "Of Atreus, Agamemnon, King of men."
+                ]
+            },
+            {
+                "description": "One file, several matches, no flags",
+                "pattern": "may",
+                "flags": [],
+                "files": ["midsummer-night.txt"],
+                "expected": [
+                    "Nor how it may concern my modesty,",
+                    "But I beseech your grace that I may know",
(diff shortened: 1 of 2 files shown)


grep.md
```diff
@@ -1,25 +1,60 @@
...
+and the output will be:
+
+<pre>
+hello
+hello again
+</pre>
+
+The `grep` command should support multiple flags at once.
+
+For example, running `grep -l -x "hello" file1.txt file2.txt` should
```
petertseng: either the `-x` should be a `-v`, or the `do not contain the string "hello"` should be changed to match what `-x` does

grep.json
```diff
@@ -0,0 +1,231 @@
...
+        "",
+        "The tests are divided into two groups: ",
+        "* Grepping a single file",
+        "* Grepping multiple files at once",
+        "",
+        "The language track implementing this exercise should",
+        "ensure that when the tests run, three files are created",
+        "with the following contents:",
+        "",
+        " [File #1]",
```
petertseng: what would you think of putting this in own section, like

```json
"files": [
  {
    "name": "iliad.txt",
    "contents": [
      "Achilles sing, O Goddess! Peleus' son;",
      "(other lines omitted from this comment)"
    ]
  }
]
```

I know it's not meant to be executed as a test, but perhaps it would make generating the files easier.

grep.json
```diff
@@ -0,0 +1,231 @@
...
+                "description": "One file, one match, match entire lines flag",
+                "pattern": "With loss of Eden, till one greater Man",
+                "flags": "-x",
+                "files": ["paradise-lost.txt"],
+                "expected": "With loss of Eden, till one greater Man"
+            },
+            {
+                "description": "One file, one match, multiple flags",
+                "pattern": "OF ATREUS, Agamemnon, KIng of MEN.",
+                "flags": "-n -i -x",
```
petertseng: is it OK that these are a space-separated string rather than an array like `files` is?

(Well, I'm sure test generators can handle it, but I wonder why the difference?)

grep.json
```diff
@@ -0,0 +1,231 @@
...
+    ],
+    "single-file": {
+        "description": ["Test grepping a single file"],
+        "cases": [
+            {
+                "description": "One file, one match, no flags",
+                "pattern": "Agamemnon",
+                "flags": "",
+                "files": ["iliad.txt"],
+                "expected": "Of Atreus, Agamemnon, King of men."
```
petertseng: The fact that some `expected` are strings and some are arrays may mess up languyages which expect a fixed type for `expected` - I would have to ee whether it's possible. The issue can be avoided by always using array, even if there is only one element

(I know that the arrays have to be joined back together to be a string anyway, but the above comment is for the benefit of tracks that wish to automatically generate their tests by reading this JSON file)

grep.json
```diff
@@ -0,0 +1,231 @@
...
+                "pattern": "Forbidden",
+                "flags": "-l",
+                "files": ["paradise-lost.txt"],
+                "expected": "paradise-lost.txt"
+            },
+            {
+                "description": "One file, several matches, case-insensitive flag",
+                "pattern": "ACHILLES",
+                "flags": "-i",
+                "files": ["midsummer-night.txt"],
```
petertseng: achilles isn't in midsummer night. iliad?

grep.json
```diff
@@ -0,0 +1,231 @@
...
+                "flags": "-x",
+                "files": ["iliad.txt", "midsummer-night.txt", "paradise-lost.txt"],
+                "expected": "midsummer-night.txt:But I beseech your grace that I may know"
+            },
+            {
+                "description": "Multiple files, one match, multiple flags",
+                "pattern": "WITH LOSS OF EDEN, TILL ONE GREATER MAN",
+                "flags": "-n -i -x",
+                "files": ["iliad.txt", "midsummer-night.txt", "paradise-lost.txt"],
+                "expected": "paradise-lost.txt:9:Of Atreus, Agamemnon, King of men."
```
petertseng: this line doesn't seem to match the requested pattern at all

grep.json
ErikSchierboom: That is a great idea! I'll make the modification.

grep.json
ErikSchierboom: Good point. I'll use arrays everywhere for the `expected` result.

grep.json
ErikSchierboom: Whoops, my bad.

grep.json
ErikSchierboom: I'll make them arrays for consistency purposes.

grep.json
ErikSchierboom: Another copy-paste mistake.

ErikSchierboom:
@petertseng I've incorporated all your feedback (thanks by the way!) and updated the PR.

grep.md
```diff
@@ -1,25 +1,60 @@
...
-  For example, running `grep -l -x "hello" file1.txt file2.txt` should
-  print the names of files that do not contain the string "hello"
+The Unix [`grep`](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/grep.html) command can be used to search for lines in one or more files 
+that match a user-provided search query.
+
+The `grep` command takes three arguments:
+
+1. The regular expression used to match lines in a file. 
+2. Zero or more flags to customize the matching behavior.
+3. One of more files in which to search for matching lines.
```
petertseng: of -> or

petertseng:
Everything seems to be in order in the test data. I have a small fix to the README and then I think it will be OK to go.

One last thing I see is that the README says the pattern can be a regular expression but the tests don't have things like `.`, `+`, `*`. I'm OK with not testing all that if trying to prevent scope creep of the exercise (should we explicitly mention that in the README or something?)

ErikSchierboom:
The README has been fixed, and I've reworded the text to not mention regular expressions as that is indeed outside the scope.

petertseng:
:+1: from me

PR #329, opened 2016-08-10
transpose: remove trailing spaces, always use arrays, add a case

petertseng:
details in individual commit messages

I should have thought of these earlier, but so it goes that I didn't. But we have the power to fix it.


ErikSchierboom:
Seems good to me.

PR #332, opened 2016-08-11
nth-prime: Add JSON test data

cacqw7:
Adding JSON test data to go along with my [xruby generator](https://github.com/exercism/xruby/pull/411)

### DIFF
--- a/nth-prime.json
+++ b/nth-prime.json
@@ -1,0 +1,34 @@
+{
+  "#": [
+    "The tests that expect 'false' should be implemented to raise",
+    "an error, or indicate a failure. Implement this in a way that",
+    "makes sense for your language."
+  ],
+  "cases": [
+    {
+       "description": "first prime",
+       "input": 1,
+       "expected": 2
+    },
+    {
+       "description": "second prime",
+       "input": 2,
+       "expected": 3
+    },
+    {
+       "description": "sixth prime",
+       "input": 6,
+       "expected": 13
+    },
+    {
+       "description": "big prime",
+       "input": 10001,
+       "expected": 104743
+    },
+    {
+       "description": "there is no zeroth prime",
+       "input": 0,
+       "expected": false
+    }
+  ]
+}


kotp:
:+1: from me.

nth-prime.json
```diff
@@ -0,0 +1,34 @@
+{
+  "#": [
+    "The tests that expect 'false' should be implemented to raise",
+    "an error, or indicate a failure. Implement this in a way that",
+    "makes sense for your language."
+  ],
+  "cases": [
+    {
+       "description": "first",
```
Insti: Consistency in descriptions: either add 'prime' here or remove it for `sixth` and `big` 
My preference is add it.

nth-prime.json
```diff
@@ -0,0 +1,34 @@
...
+       "input": "6",
+       "expected": "13"
+    },
+    {
+       "description": "big prime",
+       "input": "10_001",
+       "expected": "104_743"
+    },
+    {
+       "description": "weird case",
```
Insti: A more descriptive description would be better.
maybe: "there is no zeroth prime" or "uses 1 based indexing"

Insti:
Looks good, I only have a couple of minor nitpicks with descriptions.

petertseng:
Numbers all look correct, anyone have opinions on representing them as numbers rather than strings? Would force implementing languages to do the string -> integer parse if string, is it OK?

Insti:
Numbers would be better.

nth-prime.json
rbasso: I would simply remove the "weird case" and leave each implementation to decide how to deal with it. That would probably be better for languages that can enforce some restrictions in the type system.

cacqw7:
Updated the descriptions and changed the strings to numbers 👍

Insti:
From https://github.com/exercism/x-common/pull/332#discussion_r74446451
> I would simply remove the "weird case" and leave each implementation to decide how to deal with it. That would probably be better for languages that can enforce some restrictions in the type system.

This is the "weird case": 
```
    {
       "description": "there is no zeroth prime",
       "input": 0,
       "expected": false
    }
```
It's better to have a test in the json that is ignored by a few tracks than not have a test-case in there that is implemented by many.
I think it's good to have some defined behaviour with regard to zero. 
Language tracks can decide to not implement the test or treat the expected value as "raise an exception" as appropriate. 


Does anyone else have an opinion on this?

kytrinyx:
If we make it optional, let's add a json key or something to indicate that so that generators can decide to keep or skip optional exercises.

I don't have an opinion in this specific case though. I tend to prefer to have the zeroth case defined, but I also like math, so I'm biased.

wobh:
It seems to me, at this level, we should have a field that describes valid inputs and only provide test cases of those inputs--"happy path" tests. In this case, inputs should be positive integers, language tracks should implement the failure cases. I think this would jive better with the track-dependent "difficulty" metadata that's coming up soon.

kytrinyx:
But if there are good examples of failure cases, then it would be useful to document them.

Perhaps under a key for just failure cases, and people can do with it what they please?

ErikSchierboom:
@kytrinyx I kinda like that, separating the "success" from the "failure" test cases.

catb0t:
Can we move this forward? It looks good to me. @kytrinyx

kytrinyx:
Yepp, this looks great. Thanks @cacqw7! (Also thanks for the ping @catb0t)

petertseng:
Ah, right, we'll need to move it from nth-prime.json to exercises/nth-prime/canonical-data.json

Issue #312, opened 2016-07-31
simple-cipher has ruby code in the README


kytrinyx:
... as noted by @petertseng in #171 

It should be cleaned up to be language independent.

petertseng:
#333 has got it.

PR #333, opened 2016-08-12
Remove Ruby code examples from simple-cipher README.

allyjweir:
Change made to make the examples language agnostic. I replaced the
code with written examples in the two cases where it made sense
and removed the final example as translating it didn't translate
well to plain English


allyjweir:
Created in reference to issue #333

allyjweir:
Created in reference to issue #312

Insti:
Looks good :+1: 
I agree with removing the last example.

kytrinyx:
Thank you!

Issue #171, opened 2016-01-24
secret-handshake has ruby code in the readme


kytrinyx:
It should be cleaned up to be language-independent.

petertseng:
simple-cipher has a ton as well

petertseng:
#334 has got it.

PR #334, opened 2016-08-12
Remove Ruby examples from secret-handshake README.

allyjweir:
Left both examples in as they were originally. It is worth noting
that the first example passes an integer and the second a string
to demonstrate that the exercise expects both types to be handled.

PR submitted in response to issue #171.

### DIFF
--- a/secret-handshake.md
+++ b/secret-handshake.md
@@ -17,7 +17,7 @@ binary decide to come up with a secret "handshake".
 Here's a couple of examples:
 
 Given the input 9, the function would return the array
-["wink","jump"]
+["wink", "jump"]
 
 Given the input "11001", the function would return the array
 ["jump", "wink"]
--- a/secret-handshake.md
+++ b/secret-handshake.md
@@ -14,13 +14,14 @@ binary decide to come up with a secret "handshake".
 10000 = Reverse the order of the operations in the secret handshake.
 ```
 
-```
-handshake = SecretHandshake.new 9
-handshake.commands # => ["wink","jump"]
+Here's a couple of examples:
 
-handshake = SecretHandshake.new "11001"
-handshake.commands # => ["jump","wink"]
-```
+Given the input 9, the function would return the array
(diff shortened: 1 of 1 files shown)


Insti:
Looks good :+1: on the example translation.

You seem to have uncovered a few other issues in the Readme.

* Be explicit about input types and when decimal/binary conversions are appropriate.
* Make it clearer in the first example that a decimal to binary conversion has happened.
* Make it clearer in the second example that a reverse has taken place.
* What to do about numbers greater than 2^5 ('11111') ("will not be tested" is fine)

Do you want to deal with them here? Or shall we make another issue for them?

allyjweir:
I agree with your observations. Reading the file having not completed the exercise myself, I definitely had to intuit the aim of the problem. I would rather open a new ticket as it probably requires a little more discussion out of the scope of this original issue.

The examples were good and I would keep them in the readme in the new codeless format

Insti:
Created a new issue to address the additional issues: #335

I believe this PR for replacing the code examples is ready to merge. :+1:

secret-handshake.md
```diff
@@ -14,13 +14,14 @@ binary decide to come up with a secret "handshake".
 10000 = Reverse the order of the operations in the secret handshake.
 ```
 
-```
-handshake = SecretHandshake.new 9
-handshake.commands # => ["wink","jump"]
+Here's a couple of examples:
+
+Given the input 9, the function would return the array
+["wink","jump"]
```
petertseng: it's minor, I know, but there is no space after this comma but is a space after the other. Consistency would be good.

petertseng:
I have one small thing, otherwise I'm also :+1:  on this. Feel free to override me if nobody else cares =D

allyjweir:
@petertseng I agree. Thanks for catching that one. New commit should fix it.

petertseng:
Seems all are agreed! Thanks.

PR #337, opened 2016-08-14
Add list of topics for richer exercise metadata

kytrinyx:
This list is intended for track maintainers to use in order to add arrays of topics to
each exercise in the track's config.json file.


ErikSchierboom:
Look good!

PR #338, opened 2016-08-15
Update pangram.json

istepura:
Based on https://github.com/exercism/xfsharp/pull/195. Test for edge case when panagram is not in target alphabet

### DIFF
--- a/pangram.json
+++ b/pangram.json
@@ -44,11 +44,11 @@
         "description": "pangram with non ascii characters",
         "input": "Victor jagt zwölf Boxkämpfer quer über den großen Sylter Deich.",
         "expected": true
-      }
+      },
       {
-         "description" : "Panagram in alphabet other than ASCII",
-         "input" : "Широкая электрификация южных губерний даст мощный толчок подъёму сельского хозяйства.",
-         "expected" : false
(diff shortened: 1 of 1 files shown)


kytrinyx:
@petertseng @IanWhitney @ErikSchierboom you are the most likely to have opinions about this, I think.

petertseng:
I feel more strongly about "the commits should be squashed before merge" than whether the test should be added.

But I would support the test to be added, since it looks to be the same as that added in https://github.com/exercism/xfsharp/pull/195/files and the reasoning for it seems sound. I assume there are solutions that pass "7h3 qu1ck brown fox jumps ov3r 7h3 lazy dog" but fail the new test? (Possibly because they remove numbers?)

ErikSchierboom:
@petertseng Did you know that you can now squash the commits in the PR from within the GitHub interface? See [this article](https://help.github.com/articles/about-pull-request-merge-squashing/).

tejasbubane:
Should we create issues in all language repos using `blazon`?

Insti:
@tejasbubane that seems excessive for a minor test case addition.

tejasbubane:
@Insti Agreed. 👍  This is an edge case anyways.

PR #343, opened 2016-08-18
clarify gigasecond description, fixes #302

### LINKED ISSUE #302: [gigasecond]: Can we clarify gigasecond description?
From @MariiaKravtsova:

> I found it hard to understand what the problem "Write a program that will calculate the date that someone turned or will celebrate their 1 Gs anniversary." was asking for. I am not sure how to re-word it, but I hope someone can think of a more clear way.

What do you think of rephrasing this to say something like this?

_Write a program that calculates the moment when someone has lived for 1 billion seconds._

Cohen-Carlisle:


kotp:
This looks good to me, given the confusion that the word 'billion' can induce.

gigasecond.md
```diff
@@ -1 +1 @@
-A gigasecond is one billion (10**9) seconds.
+A gigasecond is 10^9 (1,000,000,000) seconds.
```
NobbZ: `^` does mean "exclusive or" in some programming languages, so this is ambigous as well. Since we have some of them in the list of tracks, I'd try to avoid that completely by either writing `1e9` (afaik every language has some similar way to write floats) or by using `'SUPERSCRIPT NINE' (U+2079)` (10⁹). But the latter might cause trouble on some misconfigured systems. There are still systems not properly prepared for unicode on the market :(

Cohen-Carlisle: I think it would be a mistake to try to write exponentiation here in a programming language-like syntax. 10^9 is the most straightforward "human" way to write 10-to-the-power-of-9. The superscript 9 I'm a little more open to, but since `1,000,000,000` appears immediately after 10^9, is there really any risk of confusion?

NobbZ: Yeah, I think you are right.

Cohen Carlisle <notifications@github.com> schrieb am Fr., 19. Aug. 2016 um
02:43 Uhr:

Cohen-Carlisle:
Merging my own PR...
```
(•_•)
( •_•)>⌐■-■
(⌐■_■)
YEEEEAAAHHH
```

PR #347, opened 2016-08-18
Make spaces explicitly visible in diamond kata

### LINKED ISSUE #344: Diamond kata's README is unclear about trailing spaces
It looks like I may be the first person (other than its author, of course) to complete the diamond kata, so maybe I'm just misunderstanding the README and other people find it clear. But when I read this test's README, it led me to believe that I should create diamonds that look like this (spaces have been replaced with `·` for clarity):

```plain
····A
···B·B
··C···C
·D·····D
E·······E
·D·····D
··C···C
···B·B
····A
```

However, the unit tests (in the F# version of the exercise, which is the only language in which this exercise seems to have been implemented so far) are checking that leading whitespace is equal to trailing whitespace, which means that the diamond shape the tests are *actually* looking for would be:

```plain
····A····
···B·B···
··C···C··
·D·····D·
E·······E
·D·····D·
··C···C··
···B·B···
····A····
```

Finally, the diamond shape example that the README *actually contains* looks like this once I replace its space characters with `·`:

```plain
····A·······
···B·B···
··C···C·
·D·····D
E·······E
·D·····D···
··C···C
···B·B
····A
```

That's clearly in need of editing. :-)

I suggest that the README should be more explicit about the fact that trailing spaces are expected in this exercise, and should also use the `·` character to make the trailing spaces visible in the examples. (And, of course, the examples should match what the unit tests are expecting, but that almost goes without saying. :-) )

rmunn:
Fixes #344

### DIFF
--- a/diamond.md
+++ b/diamond.md
@@ -8,7 +8,8 @@ supplied letter at the widest point.
 
 * The first row contains one 'A'.
 * The last row contains one 'A'.
-* All rows, except the first and last, have exactly two identical letters. 
+* All rows, except the first and last, have exactly two identical letters.
+* All rows have as many trailing spaces as leading spaces. (This might be 0).
 * The diamond is horizontally symmetric.
 * The diamond is vertically symmetric.
 * The diamond has a square shape (width equals height).
@@ -19,6 +20,8 @@ supplied letter at the widest point.
 
 ## Examples
 
+In the following examples, spaces are indicated by `·` characters.
+
 Diamond for letter 'A':
 
 ```plain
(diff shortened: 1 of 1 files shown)


Insti:
If we need to make spaces visible in the description, could we also use dots to make them visible in the output?
Counting whitespace while debugging is always annoying.

rmunn:
There's a certain value to requiring space characters in the code itself, because it can (if you write the solution this way) teach you about your language's built-in string padding functions. I see two possibilities:

1) The unit tests should be set to convert spaces to dots before comparing to expected values, or

2) The tests should look for some character (say, `*`) as a border around the shape:

```plain
*********
*   A   *
*  B B  *
* C   C *
*D     D*
* C   C *
*  B B  *
*   A   *
*********
```

ErikSchierboom:
I feel that this PR is fine as it is, as it explicitly states that spaces are presented as dots in the examples.

Insti:
I agree that this PR is fine.

ErikSchierboom:
Thanks @rmunn! :tada:

Issue #244, opened 2016-05-02
[binary-search]: Do not implement it for linked lists


NobbZ:
In exercism/xelixir#176 (and issues/discussion outside of exercism) there is an agreement on the fact that a binary-search is less efficient than a linear search, since a linked list does need to iterate over the list up to the n-th element for indexed access on n. Also in many implementations we need to iterate the full list to get it's length. Also when the searched element is in the first half of the list, it is necessary to copy it for the recursive call.

Aside from elixir, I already took a look into the tracks that have binary-search available at the moment (http://x.exercism.io/v3/problems/binary-search/todo).

As far as I can tell, most of the tracks are relying on an array-like datastructure, having all elements in a consecutive range of memory. But I haven't touched the languages in question for more than a hello world (or not at all), so I can't tell exactly. But for clojure I'm pretty sure it is using a linked list as well.

Also by filing this issue, I want to try to warn all other tracks that haven't implemented binary-search so far, but also don't have an appropriate datastructure.

soniakeys:
Practical binary search requires constant-time indexing, which a linked list does not provide.  So a good solution would not use a linked list and the test program should not require a linked list solution.  But the fact that linked lists are available in Elixir surely does not imply the absence of any constant-time indexed data structure or the impossibility of creating one!  I've never tried Elixir but at a quick glance, tuples look like they have constant time indexing and so it might be appropriate to implement binary search on a tuple.  It also looks like Erlang arrays are available, which would also seem suitable and, I have no idea, but perhaps more idiomatic?  Looks like there are maps.  Lua for example gets by with just maps.  A map of indexes to values would seem to be a basis for a valid solution as well.  I would encourage a little more study and perhaps a rewrite of the test program before declaring that the exercise is not suitable for Elixir or any other language.

NobbZ:
Of course there are constant time DS available in Elixir, but using a Tuple (in fact they are C-arrays at the end) for sequential data is all but idiomatic elixir, the more idiomatic way to use Tuples in elixir is as anonymous or tagged record.

Using erlangs arrays isn't idiomatic either (at least after José Valim), as he never has used them (give me two or three days and I might even give a link to the quote). Also erlang arrays aren't read in O(1), they are O(log n) (IIRC) because of their structure as beeing nested tuples, where n is the size of that array. So one could still assume indexed access beeing constant throughout a binary search, since the array probably won't change in its shape.

[...]

Issue #234, opened 2016-04-24
Binary search should not require a check for sorted input


ryanplusplus:
See https://github.com/exercism/xclojure/issues/112

In short, the Clojure track requires a check for sorted input in the Binary Search exercise. This is confusing because the check for sorted input takes as long as a linear search such that the binary search operation takes longer than just doing a linear search. We should make sure this issue does not exist in other tracks.

kotp:
[xruby/binary-search](https://github.com/exercism/xruby/blob/master/exercises/binary-search/binary_search_test.rb#L12) does indeed check for sorted... raising an error if it is not.

kytrinyx:
Agreed. If others are also in agreement, then we should write up an issue that can be cross-posted to all the tracks with blazon.

Cohen-Carlisle:
I vote firmly in the binary search should not verify sorted input camp.
The fact that it's self defeating is interesting and should probably be mentioned in the readme for curious students as well.

kytrinyx:
> The fact that checking for sorted input defeats the purpose of binary sort is interesting and should probably be mentioned in the readme for curious students as well.

Yeah, definitely.

bglusman:
I came across this issue in Elixir exercise, which actually has another issue related to it's implementation details involving singly linked list, but when I saw the requirement to error if unsorted, I thought, OK, I can do that, I can just check if the middle pivot on recurse is ever on the wrong side of the value I recursed from...  and I'd be OK with a failing test that wanted it to error in that case, as long as the input it passed was one that would require it to recurse and easily detect the unsorted list without traversing the entire list!  So that's one option, but I admit it's a little awkward to do that, but maybe good practice.

ryanplusplus:
@bglusman I don't think that's a real solution to the problem. It will catch some cases where ordering is not maintained, but it will not catch all of them.

bglusman:
Well, it will catch all you can catch without necessarily violating the log N nature of binary search... what else makes sense really if you're going to test at all and not just falsely say not_found when it may be there but not sorted?  Could change it to fall back on linear search when unsorted I guess

petertseng:
A note against checking for sortedness had been added to #349. In addition, I'm going to recommend binary-search as a problem that is too specific in #761. This may result in changes to binary-search.

In the short term, if there are any tracks that ask for sortedness check, issues should be filed on the respective tracks.

PR #349, opened 2016-08-18
binary-search: add json test data

junedev:
Following the instructions proposed by @IanWhitney in #230, I created the missing json file with the test data for binary-search. Once this is merged https://github.com/exercism/todo/issues/73 can be closed.

There are two open issues regarding the exercise:
#234 - Following (and agreeing with) the majority there I did not include checking for sorted input. I also included a respective comment in the json file. The issue can be closed once the new test data is propagated through the tracks.
#244 - I included a comment with the summary of what was discussed there but the issue seems only partly addressed by this as there are additional ideas on how/where to address the issue.

### DIFF
--- a/binary-search.json
+++ b/binary-search.json
@@ -1,0 +1,78 @@
+{
+    "#": [
+        "Here -1 is used to indicate that the value is not included in the array.",
+        "It should be replaced with the respective expression that is idiomatic",
+        "for the language that implements the tests.",
+
+        "Following https://github.com/exercism/x-common/issues/234 the exercise",
+        "should NOT include checking whether the array is sorted as it defeats",
+        "the point of the binary search.",
+
+        "The exercise should utilize an array-like (i.e. constant-time indexed)",
+        "data structure and not a linked list. If something like an array is not",
+        "usually used in the language the problem should not be implemented.",
+        "See https://github.com/exercism/x-common/issues/244 for details."
+    ],
+    "cases": [
+        {
+            "description": "finds a value in an array with one element",
+            "array": [6],
+            "value": 6,
+            "expected": 0
+        },
+        {
+            "description": "finds a value in the middle of an array",
+            "array": [1, 3, 4, 6, 8, 9, 11],
+            "value": 6,
+            "expected": 3
+        },
+        {
+            "description": "finds a value at the beginning of an array",
+            "array": [1, 3, 4, 6, 8, 9, 11],
+            "value": 1,
+            "expected": 0
+        },
+        {
+            "description": "finds a value at the end of an array",
+            "array": [1, 3, 4, 6, 8, 9, 11],
+            "value": 11,
+            "expected": 6
+        },
+        {
+            "description": "finds a value in an array of odd length",
+            "array": [1, 3, 5, 8, 13, 21, 34, 55, 89, 144, 233, 377, 634],
+            "value": 144,
+            "expected": 9
+        },
+        {
+            "description": "finds a value in an array of even length",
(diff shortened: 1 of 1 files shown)


binary-search.json
```diff
@@ -0,0 +1,60 @@
...
+        },
+        {
+            "description": "finds a value at the end of an array",
+            "array": [1, 3, 4, 6, 8, 9, 11],
+            "value": 11,
+            "expected": 6
+        },
+        {
+            "description": "finds a value in the middle of an array of odd length",
+            "array": [1, 3, 5, 8, 13, 34, 34, 55, 89, 144, 233, 377, 634],
```
Insti: Should there be 2 34's here?

junedev: Yes, this test will fail if someone just starts at the beginning of the array and returns the index on the first occurrence of the value that should be searched.

Insti: Also you probably need test cases for when the 'correct' duplicated element is the first element as well so you can't fake it by reversing the list or taking the highest index or some other similar workaround.

Insti: (I suggest that **none** of the test arrays should have duplicate elements.)

junedev: Sorry, I am a bit confused here @Insti . First you explained how to add more test cases that check the binary search is done somewhat correctly, then you said in brackets (?) that we should not do this.

Maybe you can clarify and explain a bit why you would not add these kind of tests.
My thinking was that since most languages don't easily allow counting the number of times the array is accessed this would be an easy chance to avoid that some implementations that do no binary search at all pass all the test.
You are definitely right on the separate test case(s) though.

Insti: If there are duplicate elements the minimal case is: `[ 1, 1 ]` Is the correct answer 0 or 1?
(If 0, why is rounding up wrong and rounding down right?)

Testing implementation details can be tricky.

**IF** you do want to test implementation, then test it directly and assert that the algorithm is being implemented correctly by doing something like adding a 'depth' parameter, or asking for the array result after a certain number of iterations.

The simplest thing to do is make sure the test arrays are sorted and only contain unique values, and rely on the reviewers to call out incorrect algorithms.  This is the method used in many other exercises. (And would be my preferred method.)

junedev: Thanks for taking the time to explain. I nobody else votes in favour of keeping it, I will change it to use only unique values. I might not have a chance to update the PR before the weekend. But maybe other points will come up and I can change it all together next week.

kytrinyx: That's a really great observation, @insti - I second the change to use only unique values.

junedev: PR is updated to use only unique values

Insti:
There probably should be a test for an even length array.
Did you base this file on existing tests from one of the tracks that already implement the problem?

junedev:
Unfortunately the test sets in the different tracks were very divers for this problem and I couldn't find a single one that was just right to use it directly for the json file. So I build up the content with inspiration/parts from different tracks.
Sorry for missing the even length part, I was pretty sure I included it already. PR updated accordingly.

Insti:
No problem @junedev, all your work on this is much appreciated!

petertseng:
We good with these? I think the only potential additions I would have are

* searching for a value smaller than the array's smallest value
* searching for a value larger than the array's largest value

These seem like separate cases than the existing, so could be useful. If you agree, add them and I think that will be that. If you think they aren't (and can explain to me) I think I'd take what's here as is.

junedev:
@petertseng I like the additional test cases. I will update the PR accordingly.

ErikSchierboom:
@petertseng Is this ready to be merged?

petertseng:
I say yes!

ErikSchierboom:
Merged! 🎉

Issue #269, opened 2016-06-11
Reorganize the x-common files


kytrinyx:
At the moment all of the files are in the root directory, and this is getting pretty unweildy.

I'd like to suggest that we move to the following structure:

```
.
.
└── exercises
    ├── bob
    │   ├── canonical-data.json
    │   ├── description.md
    │   └── metadata.yaml
    ├── clock
    │   ├── canonical-data.json
    │   ├── description.md
    │   └── metadata.yaml
    └── ... etc ...
```

This will allow us to have additional files for any of the problems, should we wish to do so (e.g. some way of deprecating the problem).

This will, however, cause a bit of extra work:

- Any generators that use the `:slug.json` to create the test suite will need to be adjusted.
- The x-api (problems API) will need to look for the metadata files in the new location.
- If we have other tools that use the metadata directly from x-common rather than going through the x-api, then we'll need to update those as well. I can't think of any.

ErikSchierboom:
I love the idea of grouping everything related to an exercise in one directory. You could consider creating separate directories for current and deprecated exercises. Something like this:

```
.
└── exercises
    ├── current
    │   ├── bob
    │   │   ├── data.json
    │   │   ├── description.md
    │   │   └── metadata.yaml
    │   └── ... etc ...
    └── deprecated
          ├── clock
          │   ├── data.json
          │   ├── description.md
          │   └── metadata.yaml
          └── ... etc ...
```

That would make it easier to quickly see the difference between current and deprecated exercises, as well as make the parsing a bit simpler.

kytrinyx:
Separate directories would make it easier to see/view/manage, but parsing might get more complicated since we'll have to look in two places as long as there are language tracks that have deprecated exercises.

ErikSchierboom:
@kytrinyx Good point.

Insti:
Can/Should this be done on a problem-by-problem basis?
Whenever a problem is updated it is then moved to its own subdirectory.

Advantages:
* Only one problem should break at a time.
* As the tools get updated to fix those single breakages it will make the moving of future problems easier.
* Doesn't require simultaneous co-ordination among all language tracks.
* Less work at any one point.

Disadvantages:
* Requires management of multiple possible problem locations while transition is in progress.
* Will probably cause many minor breakages.

kytrinyx:
We can do this without breaking anything by making the API handle both cases, deploying that, and then moving all of the problems at once.

It will break the generators, but that only affects ruby and go at the moment.

[...]

PR #351, opened 2016-08-22
Reorganize the x-common files

ErikSchierboom:
This PR reorganizes the X-common files as discussed in #269.


ErikSchierboom:
For fun, this is the F# script I used to generate the new structure from the existing one:

```fsharp
open System
open System.IO

let rootDir = @"D:\programmeren\_persoonlijk\x-common"

let (</>) dir file = Path.Combine [| dir; file |]

let createDirectoryIfNotExists directory =
    if not <| Directory.Exists directory then
        Directory.CreateDirectory directory |> ignore

let copyFile source target = 
    if File.Exists source then
        File.Copy(source, target)

let createDirectory sourceMetadataFile =     
    let slug = Path.GetFileNameWithoutExtension sourceMetadataFile
    let sourceDataFile = rootDir </> (sprintf "%s.json" slug)
    let sourceDescriptionFile = rootDir </> (sprintf "%s.md" slug)
    let targetDirectory = rootDir </> "exercises" </> slug
    
    let targetMetadataFile = targetDirectory </> "metadata.yml"
    let targetDataFile = targetDirectory </> "data.json"
    let targetDescriptionFile = targetDirectory </> "description.md"
    
    createDirectoryIfNotExists targetDirectory

    copyFile sourceMetadataFile targetMetadataFile
    copyFile sourceDataFile targetDataFile
    copyFile sourceDescriptionFile targetDescriptionFile

let files = 
    Directory.EnumerateFiles(rootDir, "*.yml") 
    |> Seq.filter (fun x -> not <| x.EndsWith ".travis.yml")
    |> Seq.iter createDirectory
```

Insti:
:+1: for script.

Should this PR move the existing files rather than adding new copies?
(and attempt to maintain the git history.)

Insti:
I'd also like to see the individual filenames retain their slug prefix.

`exercises/accumulate/accumulate_data.json` rather than `exercises/accumulate/data.json`

This way, if you copy a file elsewhere to work on it you still have some idea what the data is and we don't end up with 100's of files called `data.json` that are required to be in the right place to know what they contain.

ErikSchierboom:
I'm fairly certain the existing files should not be moved, as otherwise things will break.

As for renaming the files, that seems very redundant to me. If you want to work on the files, you will probably clone the whole repository and then the directory name already tells you the context. But maybe other people disagree?

Insti:
[kytrinyx wrote:](https://github.com/exercism/x-common/issues/269#issuecomment-226472430)
> We can do this without breaking anything by making the API handle both cases, deploying that, and then moving all of the problems at once.

This PR would be the 'moving all the problems at once' step.
It's useful if this PR exists first, so that those testing the required api changes can use the branch to test against.

Insti:
> As for renaming the files, that seems very redundant to me

I agree it is redundant, but what value does the filename `data` add?

ErikSchierboom:
Not much, I agree. Perhaps it should be named `test-data.json`?

kytrinyx:
> Not much, I agree. Perhaps it should be named test-data.json?

Yeah, that's better. Or `canonical-data.json`.

ErikSchierboom:
I'll rename the files to `canonical-data.json`.

ErikSchierboom:
@kytrinyx The test data JSON files have been renamed to `canonical-data.json`.

kytrinyx:
Would you take a look at the CONTRIBUTING.md file and see if any references there need to be updated?

ErikSchierboom:
I've started updating the `CONTRIBUTING.md` file. Would you mind taking a peek?

kytrinyx:
The CONTRIBUTING.md updates look good to me.

ErikSchierboom:
Is this something we want to merge, or should it be kept unmerged for a while?

Insti:
Please keep unmerged.

The x-api changes are not complete.
+ This PR duplicates the exercise metadata files and obfuscates file histories.

kytrinyx:
I think the TODO list is something like this:

- [ ] update this PR so that it moves files rather than copying them
- [ ] finalize and merge the x-api changes, deploy
- [ ] merge this PR, deploy
- [ ] update x-api to remove the logic that handles the deprecated file locations, deploy whenever

ErikSchierboom:
@kytrinyx I've updated the PR to move the files instead of copying them.

kytrinyx:
@ErikSchierboom The API change has been merged and is deploying. Would you mind rebasing this onto master to make sure we have all the latest changes and newest files?

ErikSchierboom:
@kytrinx I've rebased this PR onto the latest master and also ran my script again to pickup the latest changes. I'll be away for a couple of days after today by the way, so I hopefully this is enough.

Insti:
It's still showing +10k lines. I would expect a move to be almost zero if the + and - balance out from a move.

ErikSchierboom:
I think it's because I forgot to remove the old files. I'll try to fix it tonight

ErikSchierboom:
@Insti I've updated the PR. This time, I think everything is okay. cc @kytrinyx

rbasso:
Great!
Any idea why `word-count.json` is the only file that seems to be copied instead of moved in *Files changed*?

petertseng:
My blind guess (without verification of whether the guess is correct)

\r\n vs \n ?

ErikSchierboom:
That would be my guess too

ErikSchierboom:
@rbasso @petertseng It was indeed due to line-ending issues. I've fixed it and updated the PR>

kytrinyx:
:tada: :tada: :tada:

Insti:
Looks great, thanks @ErikSchierboom ! :heart:

ErikSchierboom:
🎉

PR #353, opened 2016-08-22
update raindrops.md to not focus on prime factors

galiat:
The raindrops Readme is a little misleading by talking about prime factors. Raindrop-speak cares if a number is divisible by certain numbers. While those numbers do happen to be prime, their primeness shouldn't matter to solve this exercise. In the ruby track I've seen several folks require the `Prime` class to solve the exercise.

TBH, I would probably not use the word 'factor' at all & instead use 'divisible by'. Was 'divisible by' avoided to not give away the idea of using modulo too easily?

### DIFF
--- a/raindrops.md
+++ b/raindrops.md
@@ -6,10 +6,10 @@
 
 ## Examples
 
-- 28's prime-factorization is 2, 2, 7.
+- 28's factors are 2, 4, **7**, 14.
   - In raindrop-speak, this would be a simple "Plong".
-- 1755 prime-factorization is 3, 3, 3, 5, 13.
+- 30's factors are 2, **3**, **5**, 6, 15.
   - In raindrop-speak, this would be a "PlingPlang".
-- The prime factors of 34 are 2 and 17.
+- 34 only has two factors- 2 and 17.
   - Raindrop-speak doesn't know what to make of that,
     so it just goes with the straightforward "34".


kytrinyx:
I actually like the fact that it has a red herring, because it leads people to need to rethink their approach when they discover it. In the real world I do this so often. I think I need to solve something in one way, and then someone will say something and everything shifts and there is simplicity and clarity.

I'm definitely open to discussing (a) the wording and (b) whether or not the red herring is too distracting and should just be removed.

@exercism/track-maintainers if any of you have strong feelings about the trade-offs between specificity and discussion, here's your chance to chime in.

kotp:
This happens very often, a name of a variable, even that we introduce, helping us to drive an implementation that is too restrictive for no good reason, and it surely happens from the customer, in their descriptions of the things that they want to have happen.  The red herring is good.

A similar thing happens in the Gigasecond exercise, where the tests specifically test for a time, in reality, the solution will take any thing that will work with the plus method as long as it will also work with the value of `1e9`.  The Date or Time references are a red herring, even perhaps more directly.

kotp:
To further the discussion when this does happen, I will mention that there is a kinder that asks "Hey, I want to play with 8 being "Plung", as I think that would be fun!".  And there is no reason not to let the little learner down.  And converting 0 is a quick test that ensure that the translations happen in order, since 0 is an universal factor.

galiat:
👍 If intent is to be a red herring, I will close.

petertseng:
I see that this is a tension induced by #295 which changed "prime factor" to just "factor" earlier in the README - now someone may wonder why one section talks about factors and the other about prime factors. Previously, both section talked about prime factors, and if this PR were merged they would both talk about factors.

If you made me decide, I would support this PR. But I advise that my opinion be taken with a grain of salt since I haven't done this exercise in any language and therefore do not observe submitted solutions and whether they unnecessarily focus on prime factorisation.

kotp:
It seems to be a fairly low occurrence that the prime factor aspect is enforced, but common that they name variables or functions with the "prime" part of the name, with no bite.

When it is enforced, it is often because (in Ruby, at least) they are using the "prime" library, rather than doing the simplest thing possible to solve the problem.

Perhaps a late test of 22 being "Plung" (as it seems to avoid the current tests) would be good to bring this out.  But bringing this out, then, removes any discussion of this, potentially.  This may be the problem of perfect specifications in a learning/lesson environment.

kytrinyx:
I've had a chance to sleep on it, and I really like this move to `factors`, without enforcing it in the tests. I think that there is still a possibility that people will overcomplicate. And some people might even go down the prime route simply because 3, 5, and 7 are primes.

I'm in favor of merging this.

raindrops.md
```diff
@@ -6,10 +6,10 @@
 
 ## Examples
 
-- 28's prime-factorization is 2, 2, 7.
+- 28's factors are 2, 4, **7**, 14.
   - In raindrop-speak, this would be a simple "Plong".
-- 1755 prime-factorization is 3, 3, 3, 5, 13.
+- 30's factors are 2, **3**, **5**, 6, 15.
```
kotp: Last thing to consider before merge, is the highlighted/bold numbers?  Should we remove them or explain why they are highlighted/bold?

kytrinyx:
Those are the numbers we're asking you to detect. I think that's clear from the context.

raindrops.md
kotp: I see, it is highlighting the given rule numbers.  Thanks!

kotp:
I think we are ready to have it come in.

PR #357, opened 2016-08-31
Add a note to flatten-array

veelenga:
Since some people do not properly understand the exercise, i decided to add this note.

http://exercism.io/submissions/996e3df67913424b8f2294bc215388dd

image

### DIFF
--- a/flatten-array.md
+++ b/flatten-array.md
@@ -1,8 +1,11 @@
-The challenge is to write a function that accepts an arbitrarily-deep nested list-like structure and returns a flattened structure without any null/nill values. 
- 
-For Example
+The challenge is to write a function that accepts an arbitrarily-deep nested list-like structure and returns a flattened structure without any null/nill values.
+
+For example:
 
 input: [1,[2,3,null,4],[null],5]
 
 output: [1,2,3,4,5]
 
+## Note
+
+Implement the function yourself. Do not use something else to perform the flattening for you.


flatten-array.md
```diff
@@ -1,8 +1,11 @@
...
+
+For example:
 
 input: [1,[2,3,null,4],[null],5]
 
 output: [1,2,3,4,5]
 
+## Note
+
+Implement the function yourself. Do not use something else to perform the flattening for you.
```
kytrinyx: Is this true for all the language tracks?

I know that in some tracks it's useful to help people get to know the core libraries.

I guess the question here, @exercism/track-maintainers, is should this be a general guide, or should this be placed in a `HINTS.md` file in this exercise for particular languages.

kytrinyx:
Thanks @veelenga, this is an interesting case—I think we have gone back and forth on whether or not we should explicitly disallow certain approaches. I can't remember where those conversations happened, though.

flatten-array.md
veelenga: @kytrinyx Ok, I didn't know that for some tracks it might be more useful. Of course, if we can't decide whether to add such note or not, it is better to let people choose the approach they want ;)

ryanplusplus: I'll add a quick 2 cents:

I think this sort of probably usually takes care of itself. It will often come up in reviews and I frequently see people first use the library function and then implement it from scratch in later submissions.

verdammelt: And sometimes the opposite happens. People will implement things that the Common Lisp language has built in features for...

kytrinyx:
I think the solution in this case is to leave it as is and let the discussion/review guide the solution.

I'm going to close this, but I do see that you also fixed a text formatting thing (which would be good to have). If you want to submit that separately I'd gladly merge it.

Thanks so much for taking the time to submit the patch so we could have this discussion :heart:

PR #362, opened 2016-09-05
Standard test suite for pascals-triangle

IanWhitney:
In surveying the current implementations, I found 2 main approaches:

- Test 1, 2, 3, 4, 5 and 20 without any boundary checking
  - [Ruby](https://github.com/exercism/xruby/blob/e50b1cbeabc3a82c1ed7800da112b0984dbbf01c/exercises/pascals-triangle/pascals_triangle_test.rb) is a good example
- Test 4 & 6 rows, and test boundaries
  - [Kotlin](https://github.com/exercism/xkotlin/blob/02ecc5a70719f0b0e741992f33879f880f7c407f/exercises/pascals-triangle/src/test/kotlin/PascalsTriangleTest.kt)
    is a good example

Also, some languages exposed specific functions for `rows` and `last_row` while some relied on built-in functions for this behavior.

In this approach I'm trying to meld what I see as the best of these worlds.

`Rows` and `Last Row` functions
----

I think these functions should be implemented by the students, so I'm highlighting them as functions to test.

Fewer Tests
----

Once a student has gotten 4 rows to work the probably have the algorithm down. No real reason to test larger numbers that I can see.

Test Boundaries
----

An upper-limit test seemed unnecessary and very language-dependent. But catching Null and negative inputs seemed like useful things to test.

### DIFF
--- a/exercises/pascals-triangle/canonical-data.json
+++ b/exercises/pascals-triangle/canonical-data.json
@@ -1,0 +1,47 @@
+{
+    "#": [
+      "Expectations are represented here as an array of arrays.",
+      "How you represent this idiomatically in your language is up to you.",
+      "An expectation of -1 indicates some sort of failure should occur"
+    ],
+    "rows": {
+      "description": "Given a count, return a collection of that many rows of pascal's triangle",
+      "cases": [
+        {
+            "description": "no rows",
+            "count": 0,
+            "expected": []
+        },
+        {
+            "description": "single row",
+            "count": 1,
+            "expected": [[1]]
+        },
+        {
+            "description": "two rows",
+            "count": 2,
+            "expected": [[1], [1, 1]]
+        },
+        {
+            "description": "three rows",
+            "count": 3,
+            "expected": [[1], [1, 1], [1, 2, 1]]
+        },
+        {
+            "description": "four rows",
+            "count": 4,
+            "expected": [[1], [1, 1], [1, 2, 1], [1, 3, 3, 1]]
+        },
+        {
+            "description": "negative rows",
+            "count": -1,
+            "expected": -1
+        },
+        {
+            "description": "no rows",
+            "count": null,
+            "expected": -1
+        }
+      ]
+    }
+}


petertseng:
In commit message and PR description, link that should go to Kotlin goes to Ruby instead

petertseng:
A curiousity I found when I did this exercise in Java was that [Java's tests](https://github.com/exercism/xjava/blob/master/exercises/pascals-triangle/src/test/java/PascalsTriangleTest.java) have a `isTriangle` function. (Though it only tested arrays of arrays where the subarrays were all the right size! If adding isTriangle I would also test things like `[[1], [1]]`). Anyway, I assume it is deliberate to not add these, since the one-liner in https://github.com/exercism/x-common/blob/master/pascals-triangle.yml just says "computes Pascal's triangle up to a given number of rows". Please confirm you agree with that.

IanWhitney:
I didn't notice it in Java, but I did spot it in..Perl, I think. Yeah, I think it's outside the scope of the problem.

pascals-triangle.json
```diff
@@ -0,0 +1,47 @@
...
+            "expected": [[1], [1, 1], [1, 2, 1]]
+        },
+        {
+            "description": "negative rows",
+            "count": -1,
+            "expected": -1
+        },
+        {
+            "description": "no rows",
+            "count": null,
```
petertseng: in some languages it will not be possible to pass null in place of an integer, but I assume they will just skip that.

Speaking of "no rows", should the input of 0 be tested?

rbasso: I think that, to keep consistency, zero could be a valid input and return an empty list `[]`, if we really want to test it. It's probably the most natural base case.

Negative inputs should raise an error if the type system cannot enforce natural numbers, so they make sense for `rows`.

rbasso: I think that, to keep consistency, zero could be a valid input returning an empty lists `[]`, if we really want to test it. It seems the natural base case to me.

Number of lines must be natural, so I agree that negatives can be considered invalid input if the type system cannot enforce positive numbers. That said, I would not criticize a solution returning an empty list for negative inputs.

Should we test those cases :question:

pascals-triangle.json
```diff
@@ -0,0 +1,47 @@
...
+            "expected": -1
+        }
+      ]
+    },
+    "last_row": {
+      "description": "Returns the last row of the pascal's triangle",
```
rbasso: If dropping compatibility with old solutions is not a problem, maybe we could take the opportunity to replace the `last_row` function by a function `row` returning the *nth* row, according the mathematical convention, as mentioned [this Wikipedia page](https://en.wikipedia.org/wiki/Pascal%27s_triangle):

> The rows of Pascal's triangle are conventionally enumerated starting with row n = 0 at the top (the 0th row).

pascals-triangle.json
```diff
@@ -0,0 +1,47 @@
...
+      "cases": [
+        {
+            "description": "fourth row",
+            "count": 4,
+            "expected": [1, 3, 3, 1]
```
rbasso: If we follow the standard convention, this would be the third line.

pascals-triangle.json
IanWhitney: If we expect `rows` to return a list/array, then wouldn't this already be built in to the language? 

```ruby
r = PascalsTriangle.rows(4)
r[3]
#=> [1, 3, 3, 1]
```

Maybe there's no point to the `last_rows` function, if all it really does is alias built-in array indexing.

petertseng: If someone were to asked me to come up with a point, I would say: Someone might argue that having `last_row` would mean that the code to generate the row need only store the current row and the previous row, instead of all rows (as it would have to in `rows`).

Such a consideration may not be a concern for the target audience. I'm not opposed to removing `last_row`.

rbasso: In Haskell, implementing `last_row` would also be trivial:

```haskell
lastRow = last . rows
```

I agree that it doesn't add anything interesting to the exercise.

rbasso:
Seems great! :+1:

PR #369, opened 2016-09-09
Test that "match entire lines" flag works properly

### LINKED ISSUE #368: Tests for grep exercise are incomplete
I wrote a solution to the grep exercise, and submitted it when it passed all the unit tests. Only when I compared my code to the example solution did I realize that I hadn't actually handled the "match entire lines" flag at all. But the only tests that checked the "match entire lines" flag had passed, because they only checked that "match entire lines" did, indeed, match when the pattern contained the entire line. There was no test that checked that the "match entire lines" flag caused a pattern to NOT match when it matched part of a line but not the whole line.

I've written a new test case to include the `grep.json` file to remedy that lack. This test caused my grep solution to fail unit testing until I actually implemented the "match entire lines" flag properly. PR coming in a second.

rmunn:
Without this test, a solution to the "grep" exercise could get away with just ignoring the "-x" flag and having it not affect the code's behavior.

Fixes #368.

### DIFF
--- a/exercises/grep/canonical-data.json
+++ b/exercises/grep/canonical-data.json
@@ -67,6 +67,33 @@
                 ]
             },
             {
+                "description": "One file, one match, print file names flag",
+                "pattern": "Forbidden",
+                "flags": ["-l"],
+                "files": ["paradise-lost.txt"],
+                "expected": [
+                    "paradise-lost.txt"
+                ]
+            },
+            {
+                "description": "One file, one match, match entire lines flag",
+                "pattern": "With loss of Eden, till one greater Man",
+                "flags": ["-x"],
+                "files": ["paradise-lost.txt"],
+                "expected": [
+                    "With loss of Eden, till one greater Man"
+                ]
+            },
+            {
+                "description": "One file, one match, multiple flags",
+                "pattern": "OF ATREUS, Agamemnon, KIng of MEN.",
+                "flags": ["-n", "-i", "-x"],
+                "files": ["iliad.txt"],
+                "expected": [
+                    "9:Of Atreus, Agamemnon, King of men."
+                ]
+            },
+            {
                 "description": "One file, several matches, no flags",
                 "pattern": "may",
                 "flags": [],
@@ -96,15 +123,6 @@
                 "expected": []
             },
             {
-                "description": "One file, one match, print file names flag",
-                "pattern": "Forbidden",
-                "flags": ["-l"],
-                "files": ["paradise-lost.txt"],
-                "expected": [
-                    "paradise-lost.txt"
-                ]
-            },
-            {
                 "description": "One file, several matches, case-insensitive flag",
                 "pattern": "ACHILLES",
                 "flags": ["-i"],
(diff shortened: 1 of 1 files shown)


petertseng:
I say the test looks good. One thing I'd discuss briefly: should it get placed so it's either immediately before or immediately after the other `-x` test? ("One file, one match, match entire lines flag") so that the tests for that functionality are together?

kytrinyx:
I would support keeping the tests for that functionally together.

We've reorganized the exercise files. Would you mind rebasing this onto the most recent master, @rmunn?

rmunn:
Re @petertseng's comment, the structure of the test data isn't "all `-x` tests together". The structure is "one file, several matches, no flags", then "one file, several matches, flag A", then "one file, several matches, flag B", then "one file, several matches, flag C". Later on, you have "multiple files, several matches, no flags", then "multiple files, several matches, flag A", then flag B, then flag C, and so on. So I inserted the `-x` test in the right order for this structure.

Now, there are actually a few tests that are out of order in this structure: there are a few places where the "one file, one match" and "one file, multiple matches" tests were interleaved in a way that doesn't really make sense. (It's not keeping flags together, because "one file, several numbers, print line numbers flag" was followed (before I added a test) by "one file, one match, print file names flag", which was in turn followed by "one file, several matches, case-insensitive flag".)

So I'll rearrange the out-of-order tests in my PR to have a more rational ordering -- but I'll try to preserve the ordering the file was apparently intended to have, which was to group matching scenarios together, rather than to group flags together.

petertseng:
> The structure is "one file, several matches, no flags", then "one file, several matches, flag A", then "one file, several matches, flag B", then "one file, several matches, flag C".

> Now, there are actually a few tests that are out of order in this structure

What if it should turn out that that was intentional (given that the current iteration of the JSON file is the only one), and that the structure isn't the one described above? Of course, in that case the test descriptions are misleading since they seem to imply one structure but actually follow another.

Okay, I realize I have no evidence to support that the problem structure is what I guesed it should be since in both the single-file and multiple-file sections each flag is only tested by itself once! So I will not constrain myself to my guess and we should talk about proposed orderings.

If a reordering is to be done in the single file section, should one be done in the multiple file section as well?

exercises/grep/canonical-data.json
```diff
@@ -67,6 +67,33 @@
...
+                "flags": ["-x"],
+                "files": ["paradise-lost.txt"],
+                "expected": [
+                    "With loss of Eden, till one greater Man"
+                ]
+            },
+            {
+                "description": "One file, one match, multiple flags",
+                "pattern": "OF ATREUS, Agamemnon, KIng of MEN.",
+                "flags": ["-n", "-i", "-x"],
```
petertseng: I want to be careful about having this test for multiple flags before some of the flags have been tested individually. In general it seems that it would be good to allow students to get one flag working at a time, right? A more gradual progression to provide clearer steps to a full solution. But if this test is here, then the student would have to get `-n` and `-i` together (`-x` has been done before). It's also espoused in https://github.com/exercism/x-common/issues/230#issuecomment-237090138

rmunn: Good point. My latest commit adds a couple more tests so that the flags will be introduced one at a time before the multiple-flags test shows up. With this commit, I think the test is now in a sensible order.

rmunn:
FYI, I'm about to get on a plane, so I'll be unavailable for the next day or two. If there are any further suggested changes to this PR, either someone else will need to write them (I believe I left the "Allow maintainers to edit" box checked when I created the PR, so anyone with write access to the x-common repo should be able to push to this branch), or else I'll get to them after I arrive and get over jet lag, which probably means on Monday.

petertseng:
To me at least, everything look good!

I only ask my opinion be taken with a grain of salt since I haven't done this exercise myself, but if nobody who has has something to say then let's merge it!

kytrinyx:
This looks good to me, too. Thank you @rmunn!

PR #373, opened 2016-09-10
Validate json files in new location

### LINKED ISSUE #371: jsonlint should look for JSON files at their post-re-organisation locations
#351 moved files from the root of this repo to per-problem subdirectories under `exercises/`.

However, jsonlint still searches for files in the root of this repo only because its glob at https://github.com/exercism/x-common/blob/master/bin/jsonlint#L4 is `*.json`.

This means that we are not getting JSON validation on any JSON files currently.

kytrinyx:
We restructured the exercise files to live under exercises/:slug/*. This updates
the json linting script to look for the json files in their new location.

Fixes #371

### DIFF
--- a/bin/jsonlint
+++ b/bin/jsonlint
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 STATUS=0
-for f in $(ls *.json); do
+for f in $(ls exercises/**/*.json); do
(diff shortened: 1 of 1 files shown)


petertseng:
Looks right!

PR #374, opened 2016-09-10
Update documentation about problem data files

### LINKED ISSUE #372: README should point to new post-re-organisation JSON locations
README currently has:

```Each problem can have a file <problem>.json```

Rendered, this just shows up as "Each problem can have a file .json", but that's a separate issue.

The more important thing is that this should get changed to match the location after #351 moved to per-exercise directories.

kytrinyx:
We recently reorganized the problem files, and the README was still
referencing the old filenames and locations.

This makes makes the new file structure explicit, and also fixes references
to files that were referred to using their old names and locations.

Fixes #372


petertseng:
Looks good to me. :+1:

PR #375, opened 2016-09-12
Correct the definition of triangle inequality to not include degenera…

durrellchamorro:
…te triangles.

### DIFF
--- a/exercises/triangle/description.md
+++ b/exercises/triangle/description.md
@@ -2,16 +2,27 @@ The program should raise an error if the triangle cannot exist.
 
 ## Hint
 
-In mathematics, a degenerate case is a limiting case in which an element of a
-class of objects is qualitatively different from the rest of the class and hence
-belongs to another, usually simpler, class.
+The triangle inequality theorem states:
+z ≤ x + y
+where x,y, and z are the lengths of the sides of a triangle. In other words, the
+sum of the lengths of any two sides of a triangle always exceeds or is equal to
+the length of the third side.
 
-1. If the sum of the lengths of any two sides of a triangle is equal to the
-   length of the third side, there exists a degenerate triangle.
+A corollary to the triangle inequality theorem is there are two classes of
+triangles--degenerate and non-degenerate. If the sum of the lengths of any two
+sides of a triangle is greater than the length of the third side, that triangle
+is two dimensional, has area, and belongs to the non-degenerate class. In
+mathematics, a degenerate case is a limiting case in which an element of a class
+of objects is qualitatively different from the rest of the class and hence
+belongs to another, usually simpler, class. The degenerate case of the triangle
+inequality theorem is when the sum of the lengths of any two sides of a triangle
+is equal to the length of the third side. A triangle with such qualities is
+qualitatively different from all the triangles in the non-degenerate class since
+it is one dimensional, looks like a straight line, and has no area. Such
+triangles are called degenerate triangles and they belong to the degenerate
+class.
 
-2. The sum of the lengths of any two sides of a triangle always exceeds or is
-   equal to the length of the third side, a principle known as the _triangle
-   inequality_.
+## Dig Deeper
 
-Therefore, a degenerate triangle does not violate the principle of _triangle
-inequality_.
+This exercise does not test for degenerate triangles. Feel free to add your own
+tests to check for degenerate triangles.
--- a/exercises/triangle/description.md
+++ b/exercises/triangle/description.md
@@ -2,6 +2,6 @@ The program should raise an error if the triangle cannot exist.
 
 ## Hint
 
-The sum of the lengths of any two sides of a triangle always exceeds or 
-is equal to the length of the third side, a principle known as the _triangle
+The sum of the lengths of any two sides of a triangle always exceeds
+the length of the third side, a principle known as the _triangle
 inequality_.
--- a/exercises/triangle/description.md
+++ b/exercises/triangle/description.md
@@ -2,6 +2,16 @@ The program should raise an error if the triangle cannot exist.
 
 ## Hint
 
-The sum of the lengths of any two sides of a triangle always exceeds
-the length of the third side, a principle known as the _triangle
+In mathematics, a degenerate case is a limiting case in which an element of a
+class of objects is qualitatively different from the rest of the class and hence
+belongs to another, usually simpler, class.
+
+1. If the sum of the lengths of any two sides of a triangle is equal to the
+   length of the third side, there exists a degenerate triangle.
+
+2. The sum of the lengths of any two sides of a triangle always exceeds or is
+   equal to the length of the third side, a principle known as the _triangle
+   inequality_.
+
+Therefore, a degenerate triangle does not violate the principle of _triangle
 inequality_.


durrellchamorro:
I noticed the definition of triangle inequality is incorrect in the README. If the sum of two of the sides equals the sum of the third side, then that's called a degenerate triangle which looks like a straight line.

petertseng:
Is it possible for us to, while doing this, also call out the case when the sum of two sides is exactly equal to the third, so that people will not feel the need to go back and forth on this e.g. #298 ?

It is known to us by #202 and #311 that we will not test such cases - it would be useful to explicitly note that in the README, I think.

durrellchamorro:
After reading your discussion about this I can see why you wouldn't want to test for degenerate triangles and I also see my proposed definition is incorrect.  One cool thing about exercism is you learn about different subjects.  I read up on degeneracy in mathematics [here](https://en.wikipedia.org/wiki/Degeneracy_(mathematics)).

As a programmer who writes object oriented code I found the first sentence in the Wikipedia article linked above of value:
> In mathematics, a degenerate case is a limiting case in which an element of a class of objects is qualitatively different from the rest of the class and hence belongs to another, usually simpler, class.

Therefore, it might be worthwhile to programmers to explain what degeneracy is followed by what a degenerate triangle is in the README.  At least for me that would have made things more clear.

petertseng:
> It might be worthwhile to programmers to explain what degeneracy is followed by what a degenerate triangle is in the README. 

I think that would be good, yes! Can probably reuse this PR to do it.

exercises/triangle/description.md
```diff
@@ -2,6 +2,16 @@ The program should raise an error if the triangle cannot exist.
...
 ## Hint
 
-The sum of the lengths of any two sides of a triangle always exceeds or 
-is equal to the length of the third side, a principle known as the _triangle
+In mathematics, a degenerate case is a limiting case in which an element of a
+class of objects is qualitatively different from the rest of the class and hence
+belongs to another, usually simpler, class.
+
+1. If the sum of the lengths of any two sides of a triangle is equal to the
+   length of the third side, there exists a degenerate triangle.
```
petertseng: it might be better to make clear that, more than the degenerate triangle existing, it is the triangle referred to earlier in the sentence that is degenerate (something like "that triangle is degenerate")

exercises/triangle/description.md
```diff
@@ -2,6 +2,16 @@ The program should raise an error if the triangle cannot exist.
...
+
+2. The sum of the lengths of any two sides of a triangle always exceeds or is
```
petertseng: Currently, there is a sentence about a degenerate case, then a numbered list, and the first item is a degenerate case, but the second isn't, and to me it seemed to make the relationship unclear (why are the two in a numbered list?). What if we start with saying the triangle inequality first, then explain the degenerate case of the inequality, as well as explaining what a degenerate triangle looks like? That would help us to understand what is meant by "qualitatively different from the rest of the class and hence belongs to another, usually simpler, class." by giving an example relevant to the exercise.

durrellchamorro: What do you think of this?

The _triangle inequality theorem_ states:
 z ≤ x + y
where x,y, and z are the lengths of the sides of a triangle.  In other words, 
the sum of the lengths of any two sides of a triangle always exceeds or is
equal to the length of the third side.  

If the sum of the lengths of any two sides of a triangle is greater than the length
of the third side, that triangle is two dimensional and it has area. However, if the sum 
of the lengths of any two sides of a triangle is equal to the length of the third side, that triangle 
is qualitatively different from all other triangles because it is one dimensional, looks like a straight line, 
and has no area.

In mathematics, a degenerate case is a limiting case in which an element of a
class of objects is qualitatively different from the rest of the class and hence
belongs to another, usually simpler, class.  The degenerate case of the triangle 
inequality theorem is when the sum of the lengths of any two sides of a triangle 
is equal to the length of the third side.  A triangle with such qualities is called a 
degenerate triangle.

durrellchamorro: What do you think of this?

The triangle inequality theorem states:
z ≤ x + y
where x,y, and z are the lengths of the sides of a triangle. In other words, 
the sum of the lengths of any two sides of a triangle always exceeds or is
equal to the length of the third side.

If the sum of the lengths of any two sides of a triangle is greater than the length
of the third side, that triangle is two dimensional and it has area. However, if the sum 
of the lengths of any two sides of a triangle is equal to the length of the third side, that triangle 
is qualitatively different from all other triangles because it is one dimensional, looks like a straight line, 
and has no area.

In mathematics, a degenerate case is a limiting case in which an element of a
class of objects is qualitatively different from the rest of the class and hence
belongs to another, usually simpler, class. The degenerate case of the triangle 
inequality theorem is when the sum of the lengths of any two sides of a triangle 
is equal to the length of the third side. A triangle with such qualities is called a 
degenerate triangle.

petertseng: Is there a chance that it will be unclear to a reader that the two classes of triangles mentioned (not degenerate vs degenerate) are a separate classification from the one resulting from the lengths of the sides (equilateral, isosceles, scalene)? If there is, we should try to find a way to make the distinction clearer. But if others find it understandable, I think this is good.

I think the last thing is: After a student has read this, the student might think "OK, I have just learned about degenerate triangles, and how should this knowledge apply to this exercise?" and we should give an answer to that potential question in the readme (that they don't violate the triangle inequality - but we won't test for them)

durrellchamorro: Good points @petertseng.  How about, "A corollary to the triangle inequality theorem is there are two classes of triangles--degenerate and nondegenerate. If the sum of the lengths of any two sides of a triangle is greater than the length of the third side (as is the case with equilateral, isosceles, and scalene triangles), that triangle is two dimensional, has area, and belongs to the nondegenerate class."

Regarding your second point, knowledge about degenerate triangles applies to the exercise because, at least for the Ruby and Coffescript exercises, the second to last test is for the degenerate case. Interestingly enough I don't see a test for the degenerate case in the Javascript exercise.   

The main reason I wanted to point out degenerate triangles do not violate the principle of triangle inequality is the tests, at least for Ruby and Coffeescript, are currently wrong.  The combination of incorrect tests and not understanding the triangle inequality confused me.  I thought either the tests were wrong or the definition of triangle inequality in the README was wrong. I incorrectly assumed the tests were right, hence the original pull request message regarding a correction to the definition of triangle inequality in the README.  After a closer look I realized a 2,2,4 triangle satisfies the degenerate condition of the triangle inequality, but In Coffeescript for example, the second to last test reads:
```coffee
it 'is illegal when violating triangle inequality 2', ->
    expect(-> new Triangle(2,2,4)).toThrow("violation of triangle inequality")
```
If it is a mistake that the Ruby and Coffeescript exercises are testing for the degenerate case, then I'm happy to make pull requests to remove those tests.  Otherwise, I'm happy to fix the tests in the languages I know if that is ok with you all. 

If you choose not to test for the degenerate case, then a way to document in the README how this knowledge about degenerate triangles applies to the exercise would be to add a "Dig Deeper" or "Extra Credit" or "Challenge" section where you explain how none of the tests checked for a degenerate triangle, but you can challenge the reader to add those tests.  You can even put the whole explanation of the corollary to the triangle inequality in the "Challenge" section itself. For OO languages you can challenge the reader to prevent degenerate triangles from being initialized directly from the Triangle class.

petertseng: > A corollary to the triangle inequality theorem is there are two classes of triangles--degenerate and nondegenerate.

Thanks, I think this is good. Explicitly noting them will prevent the confusion.

> at least for the Ruby and Coffescript exercises, the second to last test is for the degenerate case
> Interestingly enough I don't see a test for the degenerate case in the Javascript exercise.

Ah. It appears that Javascript did have it at one point, but removed it for https://github.com/exercism/xjavascript/pull/164 .

> If it is a mistake that the Ruby and Coffeescript exercises are testing for the degenerate case, then I'm happy to make pull requests to remove those tests.

Ah, good point. Yes, the test should be removed altogether, as had been decided at https://github.com/exercism/x-common/pull/311#issuecomment-236520834 - tests for degenerate triangles are not in https://github.com/exercism/x-common/blob/master/exercises/triangle/canonical-data.json and tracks should be using this data, unless they have specific language-level concerns (and this likely isn't one).

> If you choose not to test for the degenerate case, then a way to document in the README how this knowledge about degenerate triangles applies to the exercise would be to add a "Dig Deeper" or "Extra Credit" or "Challenge" section where you explain how none of the tests checked for a degenerate triangle, but you can challenge the reader to add those tests. 

I like this! I don't have a strong opinion on which of the three proposed terms to use. If you *made* me pick I'd pick "dig deeper", but I can always be convinced otherwise.

> You can even put the whole explanation of the corollary to the triangle inequality theorem in the "Challenge" section itself

I do think this sounds like a good idea.

durrellchamorro: Ok. What do you think of this?
## Hint

The triangle inequality theorem states:
z ≤ x + y
where x,y, and z are the lengths of the sides of a triangle. In other words, the
sum of the lengths of any two sides of a triangle always exceeds or is equal to
the length of the third side.

A corollary to the triangle inequality theorem is there are two classes of
triangles--degenerate and non-degenerate. If the sum of the lengths of any two
sides of a triangle is greater than the length of the third side, that triangle
is two dimensional, has area, and belongs to the non-degenerate class. In 
mathematics, a degenerate case is a limiting case in which an element of a class 
of objects is qualitatively different from the rest of the class and hence 
belongs to another, usually simpler, class. The degenerate case of the triangle 
inequality theorem is when the sum of the lengths of any two sides of a triangle 
is equal to the length of the third side. A triangle with such qualities is 
qualitatively different from all the triangles in the non-degenerate class since 
it is one dimensional, looks like a straight line, and has no area. Such 
triangles are called degenerate triangles and they belong to the degenerate 
class.

## Dig Deeper

This exercise does not test for degenerate triangles. Feel free to add your own 
tests to check for degenerate triangles.

petertseng: > you are not testing that triangles with negative sides are illegal. It seems like a good idea to test for that. The Ruby, Javascript and Coffeescript tests are currently testing for it.

Ah! It would be good to send a PR for that, then. I will first say that at #311 we had decided against it because it's not specified clearly whether the classification function is responsible for normalising possibly-negative distances into absolute lengths as opposed to the caller being responsible. If you do feel strongly about this, it is of course open to discussion if we could get a PR on it.

> What do you think of this?

Personally I am thinking it looks good, so :+1:

petertseng: It's interesting because it leaves it open to the student what to do with the degenerate triangles. You gave a possible option earlier in our discussion: It shouldn't be possible to create them because they are different. Others may choose to allow them, and classify them as either scalene or isosceles (they can't be equilateral unless all sides are zero). Others may make degenerate a separate case altogether. We like to see variety, so leaving it open seems like a good idea, which is another reason I said :+1:

durrellchamorro:
Regarding the concern about edge-case-itis and how the goal is not to write production quailty code at https://github.com/exercism/x-common/pull/287#issuecomment-231602657,  I see a theme emerging for exercism in general.  You might consider putting edge cases you think might be overkill in a Dig Deeper section in the README for people who want to challenge themselves more.  You could also note in a Dig Deeper section where advanced techniques like metaprogramming shine such as in the Space Age exercise.  If you've gone through the work of discovering, discussing, and voting on such cases you might want to share that knowledge and work in the README in this fashion.  Those edge cases might help elucidate the problem and help people solve it as was the case for me with this triangle exercise.  

If you've already decided against negative lengths, that's fine with me. Is the protocol to vote on the proposed README before I update this PR?

petertseng:
> You might consider putting edge cases you think might be overkill in a Dig Deeper section in the README for people who want to challenge themselves more.  If you've gone through the work of discovering, discussing, and voting on such cases you might want to share that knowledge and work in the README in this fashion. Those edge cases might help elucidate the problem and help people solve it as was the case for me with this triangle exercise. 

These are good suggestions that I have noted at https://github.com/exercism/x-common/issues/230#issuecomment-247816909 - there are always places where we can improve and you're right about this - these things should be documented so that we can avoid duplication of past efforts.

> You could also note in a Dig Deeper section where advanced techniques like metaprogramming shine such as in the Space Age exercise.

This is also good - sometimes I see individual tracks note such things themselves if such a technique is specific to that language, but if it is broadly applicable and language-agnostic then it can be at x-common as well.

> Is the protocol to vote on the proposed README before I update this PR?

What I would personally do is to update the PR now, since it is cheap to do so.

I'm not aware of a specific protocol. I do note that throughout the process of the past week, if someone other than me had changes to suggest, they've had ample opportunity to do so! On the other hand if they're OK with all this they didn't need to say anything.

durrellchamorro:
In addition to updating the ruby tests to match the canonical-data.json, my PR added a test to check for a 1, 3, 1 triangle since that also violates the triangle inequality.  If that PR looks ok, I'll submit PR's to update the tests for Coffeescript, and Javascript the same way.

petertseng:
Thanks @durrellchamorro !

PR #386, opened 2016-09-23
Remove redundancy from the description of isogram

jbranchaud:
When the blurb from the metadata of isogram:

> Determine if a word or phrase is an isogram

is combined with the description text, the resulting exercise
description becomes redundant.

> Determine if a word or phrase is an isogram. Determine if a word or
> sentence is an isogram. An isogram ...

So, this commit removes the redundant part from `description.md`.


petertseng:
In case any wonders where they two get combined:

http://exercism.io/exercises/fsharp/isogram/readme

There, we combine the metadata one-liner plus the full description.

petertseng:
Thanks!

PR #387, opened 2016-09-24
Amended hello-world/description.md to be formatted to 80 chars.

swmcc:
Should allow reading of the rendered README.md easier if viewing in a restricted window.

I am not sure if this is even needed but I thought I would say anyway. I don't mind going through the rest of the ```description``` files and doing the same but I am aware this could be overkill.

When viewed through a restricted window in vim the README for a ```hello-world``` exercise was looking like this:

screen shot 2016-09-25 at 00 38 58

A quick format and it now looks like this:

screen shot 2016-09-25 at 00 44 57

### DIFF
--- a/exercises/hello-world/description.md
+++ b/exercises/hello-world/description.md
@@ -1,4 +1,5 @@
-["Hello, World!"](http://en.wikipedia.org/wiki/%22Hello,_world!%22_program) is the traditional first program for beginning programming in a new language.
+["Hello, World!"](http://en.wikipedia.org/wiki/%22Hello,_world!%22_program) is
+the traditional first program for beginning programming in a new language.
 
 **Note:** You can skip this exercise by running:
 
@@ -6,8 +7,8 @@
 
 ## Specification
 
-Write a `Hello World!` function that can greet someone given their name.
-The function should return the appropriate greeting.
+Write a `Hello World!` function that can greet someone given their name.  The
+function should return the appropriate greeting.
 
 For an input of "Alice", the response should be "Hello, Alice!".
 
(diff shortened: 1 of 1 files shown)


petertseng:
Yes, I think so. We do have predecent at https://github.com/exercism/x-common/pull/200#r55940147 indicating we would like wrapping.

---

In a more general vein (not directly related to this PR, but only as an extension of it), we might like to write somewhere, maybe in the contributing guide, that we like to wrap our descriptions at N characters.

We might also like to consider looking for all files that have long lines: `git grep -El '.{81,}' exercises/*/description.md`

kytrinyx:
This looks great! Yeah, I agree with both of you; I think we should generally reformat to 80 columns for the `description.md` files, not because terminals are narrow (they usually aren't anymore), but because narrower lines are easier for humans to parse.

The Go project has some "readability" guidelines that are used for all code reviews internally at Google, and also on the Go project itself, and it has a section on long lines: https://github.com/golang/go/wiki/CodeReviewComments#line-length

> Comments are typically wrapped before no more than 80 characters, not because it's a rule, but because it's more readable when viewing in an editor that might be sized to show hundreds of columns wide. Humans are better at following narrow text (e.g. columns in a newspaper) than giant walls of wide text, as a wide editor might show. Regardless, godoc should render it nicely either way.

kytrinyx:
I'm tempted to open a new issue that generally talks about reformatting the `description.md` file, and then generating one issue for each exercise that is to verify that it's formatted correctly, and reformat it if it isn't.

This way we could tag these with "good first patch" and "hacktoberfest" and we could get a bunch of people helping with it.

swmcc:
I have a branch in my local that formats the remaining ```description.md``` files. I might put that up and see what happens. Just need time to go through it make sure it is ok. I should have that done for tomorrow night (BST). Thanks for the merge @kytrinyx

kytrinyx:
> I have a branch in my local that formats the remaining description.md files.

Nice! I won't worry about generating lots of tickets, then :)

Issue #224, opened 2016-04-10
Allergies READ ME needs clarification 


rebelwarrior:
Hi There are three ways to interpret the Allergies problem with regards to large scores. Below are the options. The first one passes the test in Elm-lang but seems incorrect to me. The second seems correct but fails the tests. And the third is possible but tests should catch it better. 

If the score is >= 256 then 
1. **Subtract 256** from the score and calculate. 
2. Ignore numbers above 256; **treat score as 255** and calculate. 
3. Subtract 128 and mark the highest allergy and calculate, if the score is still above 128 subtract it again and calculate. You might end up with an allergy duplicated or just counted twice. 

So following the first case a score of 257 equals a score of 1 since 257 - 256 is 1. This doesn't seem right to me, but if it is right then it should be stated in the Read Me.

The second is the one the READ ME alludes to but the tests fail. 

Which is the way this should go? Should we clarify the READ ME with:

> Scores are a representation of binary and cycle, so a score of 256 is the same as 0 and a score of 257 as 1. 

Or do we need to updated the *tests* and the READ ME with:

> Scores higher than the maximum score should be treated as the maximum score.

wobh:
It's not cycle but ceiling. I think scores higher than 255 should be treated as invalid inputs--the system can't interpret those as allergies. Cycling means making a presumption about all larger values. If you add allergy to a cyclical system you will have invalidated all previous interpretations of values greater than the max.

kytrinyx:
If you interpret these as bitmasks, then it doesn't matter that the number is higher than 256. Either the bit is flipped or it's not.

wobh:
Also ceiling is also a presumption: 256 == 255.

I'm on the fence about "fixing" this design. An advanced exercise could be dealing with the fallout of the current implementation, a body of accumulated data, and the requirement of adding additional allergies.

kytrinyx:
Right 0-255 would be the valid range.

wobh:
The mask could be seen as the floor problem: 256 == 0. That would be problematic in the same way. Sometimes you just need `nil`. 😀

I think of a school importing transfer students data, whose allergy scores are based on a different system. If they can't translate a given score, they probably shouldn't enter apparently valid score from the system they have.

rebelwarrior:
The cyclical solution is the one that passes the tests in Elm. 

[...]

PR #389, opened 2016-10-01
Clarify handling scores greater than 255 in allergies.

jtigger:


jtigger:
Addresses #224.

Expressed in solution-agnostic terms.

IanWhitney:
Just to make sure I understand, this change would mean that a test of 257 would be allergic to all the items mentioned in the readme?

NobbZ:
As I understand it the read me asks for a simple modulo equivalency.

Ian Whitney <notifications@github.com> schrieb am Sa., 1. Okt. 2016 14:56:

Insti:
257 would mean allergic to eggs (1)

jtigger:
The modification of the description plus the final test in suite implied by the `canonical-data.json` would mean that a score of 257 would mean some uncodified allergen (256) plus eggs(1), the former to be ignored/masked/modulo'ed.

IanWhitney:
Ok. That's not how this change reads for me. 

> Note: a given score may include allergens not listed above (i.e. allergens that score 256, 512, 1024, etc.). Your program should ignore those components of the score

The "ignore those components" could be clearer. Maybe:

> Note: Allergens with higher scores (i.e., 256, 512, 1024) should be ignored. For example, a score of 257 would be allergic to just Eggs.

jtigger:
Added example to the description.  Thanks, @IanWhitney.

kytrinyx:
This looks good to me. Thanks for hashing out the details on this!

PR #390, opened 2016-10-01
pov: Add canonical data

petertseng:
Existing implementations:

* https://github.com/exercism/xclojure/blob/master/exercises/pov/test/pov_test.clj
* https://github.com/exercism/xfsharp/blob/master/exercises/pov/PovTest.fs
* https://github.com/exercism/xgo/blob/master/exercises/pov/pov_test.go
* https://github.com/exercism/xhaskell/blob/master/exercises/pov/test/Tests.hs
* https://github.com/exercism/xlua/blob/master/exercises/pov/pov_spec.lua

The tests are a combination of the existing tests.

The trees used:

* Singleton (present in all)
* Simple tree: X, parent, sibling (present in Clojure, Go, Lua)
* Large flat tree: X, parent, three siblings (present in all)
* Deeply nested tree: "singly-linked list" as a tree (present in all)
* Target has children: X, parent, two children (present in all*)
  (in Clojure, Go, Lua, this case is embedded in a more complex tree)
* Target has cousins (present in all)

All tracks tested looking for the node named "x" in these trees, then
looking for a nonexistent node.

The Lua track had a gentler progression of complexity in the path tests,
so that was used. This is necessary simply because even after the reroot
is done, there is still work to be done to search for the target node.

(The Go track chose not to test paths)

Closes https://github.com/exercism/todo/issues/128

### DIFF
--- a/exercises/pov/canonical-data.json
+++ b/exercises/pov/canonical-data.json
@@ -1,0 +1,356 @@
+{
+  "from_pov": {
+    "description": [
+      "Reroot a tree so that its root is the specified node.",
+      "In this way, the tree is presented from the point of view of the specified node.",
+      "The input trees used here are those in the `trees` section of this file.",
+      "",
+      "If appropriate for your track, you may test that the input tree is not modified.",
+      "",
+      "Note that when rerooting upon a target node that has both parents and children,",
+      "it does not matter whether the former parent comes before or after the former children.",
+      "Please account for this when checking correctness of the resulting trees.",
+      "One suggested method is to only check two things:",
+      "1) The root of the returned tree is the root that was passed in to from_pov.",
+      "2) The sorted edge list of the returned tree is the same as the sorted edge list of the expected tree."
+    ],
+    "cases": [
+      {
+        "description": "Results in the same tree if the input tree is a singleton",
+        "tree": "singleton",
+        "from": "x",
+        "expected": {
+          "label": "x"
+        }
+      },
+      {
+        "description": "Can reroot a tree with a parent and one sibling",
+        "tree": "simple",
+        "from": "x",
+        "expected": {
+          "label": "x",
+          "children": [
+            {
+              "label": "parent",
+              "children": [
+                {
+                  "label": "sibling"
+                }
+              ]
+            }
+          ]
+        }
+      },
+      {
+        "description": "Can reroot a tree with a parent and many siblings",
+        "tree": "large_flat",
+        "from": "x",
+        "expected": {
+          "label": "x",
+          "children": [
+            {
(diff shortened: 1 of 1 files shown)


exercises/pov/canonical-data.json
```diff
@@ -0,0 +1,348 @@
+{
+  "from_pov": {
+    "description": [
+      "Reroot a tree so that its root is the specified node.",
+      "In this way, the tree is presented from the point of view of the specified node.",
+      "The input trees used here are those in the `trees` section of this file.",
```
petertseng: I chose to the put the input trees in the `"trees"` section because they are used more than once and it seemed useful to allow people to easily see that they are the same six trees being reused.

In contrast, I put the expected output trees inline since they are only used once.

exercises/pov/canonical-data.json
```diff
@@ -0,0 +1,348 @@
...
+      "",
+      "If appropriate for your track, you may test that the input tree is not modified.",
+      "",
+      "Note that when rerooting upon a target node that has both parents and children,",
+      "it does not matter whether the former parent comes before or after the former children.",
```
petertseng: https://github.com/exercism/xhaskell/issues/92

petertseng:
A lot of this was done by hand, so it behooves me to automatically verify its correctness. I have not the strength for that now, so that will have to come later.

petertseng:
This has been verified for correctness, by https://github.com/petertseng/x-common/blob/verify/exercises/pov/verify.rb

exercises/pov/canonical-data.json
```diff
@@ -0,0 +1,348 @@
...
+      "Please account for this when checking correctness of the resulting trees.",
+      "One suggested method is to only check two things:",
+      "1) The root of the returned tree is the root that was passed in to from_pov.",
+      "2) The sorted edge list of the returned tree is the same as the sorted edge list of the expected tree."
+    ],
+    "cases": [
+      {
+        "tree": "singleton",
```
petertseng: potentially contentious point: No `"description":` for any of these. I think I would rather have descriptions.

kytrinyx: Right—in the simpler exercise a lot of the descriptions became silly, just echoing the inputs and outputs. Here it probably makes more sense to flesh out the purpose of the test in a description.

ErikSchierboom:
I love the usage of uncle, kid, and such. It greatly helps intuit what's happening.

ryanplusplus:
I'm fairly certain that I actually stole all of the Lua tests from the Ruby track and I was pretty happy with those... so this looks good to me :)

PR #391, opened 2016-10-02
Update bowling canonical-data

IanWhitney:
This is based on my work implementing this problem for Rust
https://github.com/exercism/xrust/pull/213

I'm addressing a few things here.

- The description at the top is now about implementing the tests, not
about the problem domain.

- I've reordered the test to (hopefully) make things flow more smoothly.
I found that the previous set of tests introduced concepts in kind of a
mix, instead of one at a time.

They are now:

- Score a 0 point game
- Score a game with no spares/strikes
- Spares
  - Simple spare
  - Spare with bonus
  - Spare in last frame
- Strikes
  - Simple strike
  - Strike with bonus
  - Strike in last frame
- Exceptions

Some other changes:

- Updated descriptions. My goal here was to try to clearly state what
bit of scoring logic was being tested in each test, without relying too
much on bowling-specific terms. 'strike', 'spare' and 'frame' are
unavoidable, but other terms were replaceable.

- Remove unnecessary tests. There were a few different tests of games
without spares & strikes when only one is really necessary. My goal is
that each test should lead the student to make one change to their code.
If a bunch of tests can pass because of the same change then some of
those tests can be removed.

- Explicit description of arity removed. I tried to cover the expected
API in the description text.

- Expectations of exceptions changed from specific error text to `-1`,
which is what I normally see for exception test definitions.

### DIFF
--- a/exercises/bowling/canonical-data.json
+++ b/exercises/bowling/canonical-data.json
@@ -86,6 +86,10 @@
       "rolls": [5, 6, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       "expected": -1
     }, {
+      "description": "Two bonus rolls after a strike in the last frame can not score more than 10 points",
+      "rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10, 5, 6],
+      "expected": -1
+    }, {
       "description": "An unstarted game can not be scored",
       "rolls": [],
       "expected": -1
--- a/exercises/bowling/canonical-data.json
+++ b/exercises/bowling/canonical-data.json
@@ -8,6 +8,10 @@
     "The final tests define situations where the score can not be returned",
     "The expection for these tests is -1, indicating an error",
     "In these cases you should expect an error as is idiomatic for your language",
+    "When to error is also left up to your implementation. There are two options",
+    "  - Error as soon as an invalid roll is made",
+    "  - Error when scoring a game with an invalid roll",
+    "You can also error in both cases."
   ],
   "score": {
     "description": [
--- a/exercises/bowling/canonical-data.json
+++ b/exercises/bowling/canonical-data.json
@@ -3,7 +3,7 @@
     "Students should implement roll and score methods.",
     "Roll should accept a single integer.",
     "Score should return the game's final score, when possible",
-    "For brevity the tests all the rolls in an array;",
+    "For brevity the tests display all the rolls in an array;",
     "each element of the rolls array should be passed to the roll method",
     "The final tests define situations where the score can not be returned",
     "The expection for these tests is -1, indicating an exception",
--- a/exercises/bowling/canonical-data.json
+++ b/exercises/bowling/canonical-data.json
@@ -6,8 +6,8 @@
     "For brevity the tests display all the rolls in an array;",
     "each element of the rolls array should be passed to the roll method",
     "The final tests define situations where the score can not be returned",
-    "The expection for these tests is -1, indicating an exception",
-    "In these cases you should expect an exception as is idomatic for your language"
+    "The expection for these tests is -1, indicating an error",
+    "In these cases you should expect an error as is idiomatic for your language",
   ],
   "score": {
     "description": [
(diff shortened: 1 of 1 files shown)


exercises/bowling/canonical-data.json
```diff
@@ -1,117 +1,110 @@
 {
   "#": [
-    "Bowling is game where players roll a heavy ball to knock down pins",
-    "arranged in a triangle. Write code to keep track of the score of a",
-    "game of bowling."
+    "Students should implement roll and score methods.",
+    "Roll should accept a single integer.",
+    "Score should return the game's final score, when possible",
+    "For brevity the tests all the rolls in an array;",
```
kytrinyx: This sentence might be missing a word, and I can't figure out what it should be. Holds? Keeps? Stores?

exercises/bowling/canonical-data.json
```diff
@@ -1,117 +1,110 @@
...
+    "each element of the rolls array should be passed to the roll method",
+    "The final tests define situations where the score can not be returned",
+    "The expection for these tests is -1, indicating an exception",
```
kytrinyx: Perhaps `s/exception/error state`. Some languages will implement this as an exception, but others might not (e.g. Go doesn't have exceptions).

exercises/bowling/canonical-data.json
```diff
@@ -1,117 +1,110 @@
...
     }, {
-      "description": "should not allow to take score before the game has ended",
-      "rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-      "expected": "Score cannot be taken until the end of the game"
+      "description": "An incomplete game can not be scored",
+      "rolls": [0, 0],
+      "expected": -1
     }, {
-      "description": "should not allow rolls after the tenth frame",
+      "description": "A game with more than ten frames can not be scored",
```
petertseng: The change in the description seems to change the implication of where the error should occur.

Before, to me this sounds like: The roll should be rejected before there is any attempt to take the score.

After, it sounds like: The roll should be accepted, and onlyh when taking the score do we realize there is a failure.

And I definitely think the roll should be rejected as soon as possible.

But this is a difficult proposition when the only expectation this test file expresses is the score at the end, instead of noting where the error should occur. That's a difficult concept to express though. The first idea that comes to mind (and the idea may be crazy) is:

```
"rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
"roll_should_fail": 0,
```

exercises/bowling/canonical-data.json
```diff
@@ -1,117 +1,110 @@
...
     }, {
-      "description": "should not allow two normal rolls better than strike",
-      "rolls": [5, 6],
-      "expected": "Pin count exceeds pins on the lane"
+      "description": "A roll can not score more than 10 points",
+      "rolls": [11, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+      "expected": -1
     }, {
-      "description": "should not allow two normal rolls better than strike in last frame",
-      "rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 5, 6],
```
IanWhitney: If these tests caught an edge case, then they should be kept. Thanks.

exercises/bowling/canonical-data.json
IanWhitney: For those of you who aren't Peter, we've been talking about this in the Rust PR as well. 

https://github.com/exercism/xrust/pull/213

For Rust, at least, we're going to have the `roll` method fail if the roll is invalid.

exercises/bowling/canonical-data.json
petertseng: I was indeed treating the bonus rolls in a separate place from the normal ten frames, so I do say it is worth keeping.

exercises/bowling/canonical-data.json
petertseng: Seems good.

IanWhitney:
I think I've covered the comments to date.

PR #392, opened 2016-10-03
Use nil instead of nill in description/blurb

jbranchaud:
This fixes a minor typo. Nill is the verb whereas nil is the noun.


petertseng:
My references agree, and this echoes my experiences with languages that have `nil` as a keyword (I have never seen one with `nill` as a keyword)

petertseng:
Amended commit message to prepend with "flatten-array"

Thanks!

PR #394, opened 2016-10-04
forth: Add canonical data

petertseng:
Existing tests:
* https://github.com/exercism/xelixir/blob/master/exercises/forth/forth_test.exs
* https://github.com/exercism/xfsharp/blob/master/exercises/forth/ForthTest.fs
* https://github.com/exercism/xhaskell/blob/master/exercises/forth/test/Tests.hs
* https://github.com/exercism/xrust/blob/master/exercises/forth/tests/forth.rs
* https://github.com/exercism/xscala/blob/master/exercises/forth/src/test/scala/ForthTest.scala

All tracks had the exact same tests, with two exceptions:

First, The "All non-word characters are separators test". In particular,
F# chose not to include the Ogham space mark, and I argue the same in
https://github.com/exercism/xhaskell/pull/388.

The Ogham space mark adds no additional value (it is just another space
character) and is confusing (it is visually too similar to the minus
sign), so it will not be included.

Second, Rust has a few extras in its tests:

* Parsing negative numbers (adds an extra wrinkle since now we must
  determine whether a minus is unary or binary)
* Extra error cases: +, -, *, / with not enough arguments.
* Ensuring that user-defined words are case-insensitive: defines `CoUnT`
  then executes `count COUNT`
* Ensuring that user-defined words execute their definitions in the
  right order.
* Extra error cases: Incomplete definitions (a `:` without a `;`)

After discussion in #394, we decide that we will only add tests for the
second and fourth of these.

Further, we have changed the expectations from the string representation
of the stack to simply an array/list of values on the stack.

Closes https://github.com/exercism/todo/issues/92

### DIFF
--- a/exercises/forth/canonical-data.json
+++ b/exercises/forth/canonical-data.json
@@ -1,0 +1,334 @@
+{
+  "#": [
+    "The cases are split into multiple sections, all with the same structure.",
+    "In all cases, the `expected` key is the resulting stack",
+    "after executing the Forth program contained in the `input` key."
+  ],
+  "parsing and numbers": {
+    "cases": [
+      {
+        "description": "empty input results in empty stack",
+        "input": [],
+        "expected": []
+      },
+      {
+        "description": "numbers just get pushed onto the stack",
+        "input": [
+          "1 2 3 4 5"
+        ],
+        "expected": [1, 2, 3, 4, 5]
+      },
+      {
+        "description": "all non-word characters are separators",
+        "input": [
+          "1\u00002\u00133\n4\r5 6\t7"
+        ],
+        "expected": [1, 2, 3, 4, 5, 6, 7]
+      }
+    ]
+  },
+  "addition": {
+    "cases": [
+      {
+        "description": "can add two numbers",
+        "input": [
+          "1 2 +"
+        ],
+        "expected": [3]
+      },
+      {
+        "description": "errors if there is nothing on the stack",
+        "input": [
+          "+"
+        ],
+        "expected": null
+      },
+      {
+        "description": "errors if there is only one value on the stack",
+        "input": [
+          "1 +"
+        ],
+        "expected": null
+      }
+    ]
+  },
+  "subtraction": {
+    "cases": [
+      {
+        "description": "can subtract two numbers",
+        "input": [
+          "3 4 -"
+        ],
+        "expected": [-1]
+      },
+      {
+        "description": "errors if there is nothing on the stack",
+        "input": [
+          "-"
+        ],
+        "expected": null
+      },
+      {
+        "description": "errors if there is only one value on the stack",
+        "input": [
+          "1 -"
+        ],
+        "expected": null
+      }
+    ]
+  },
+  "multiplication": {
+    "cases": [
+      {
+        "description": "can multiply two numbers",
+        "input": [
+          "2 4 *"
+        ],
+        "expected": [8]
+      },
+      {
+        "description": "errors if there is nothing on the stack",
+        "input": [
+          "*"
+        ],
+        "expected": null
+      },
+      {
+        "description": "errors if there is only one value on the stack",
+        "input": [
+          "1 *"
+        ],
+        "expected": null
+      }
+    ]
+  },
+  "division": {
+    "cases": [
+      {
+        "description": "can divide two numbers",
+        "input": [
+          "12 3 /"
+        ],
+        "expected": [4]
+      },
+      {
+        "description": "performs integer division",
+        "input": [
+          "8 3 /"
+        ],
+        "expected": [2]
+      },
+      {
+        "description": "errors if dividing by zero",
+        "input": [
+          "4 0 /"
+        ],
+        "expected": null
+      },
+      {
+        "description": "errors if there is nothing on the stack",
+        "input": [
+          "/"
+        ],
+        "expected": null
+      },
+      {
+        "description": "errors if there is only one value on the stack",
+        "input": [
+          "1 /"
+        ],
+        "expected": null
+      }
+    ]
+  },
+  "combined arithmetic": {
+    "cases": [
+      {
+        "description": "addition and subtraction",
+        "input": [
+          "1 2 + 4 -"
+        ],
+        "expected": [-1]
+      },
+      {
+        "description": "multiplication and division",
+        "input": [
+          "2 4 * 3 /"
+        ],
+        "expected": [2]
+      }
+    ]
+  },
+  "dup": {
+    "cases": [
+      {
+        "description": "copies the top value on the stack",
+        "input": [
+          "1 DUP"
+        ],
+        "expected": [1, 1]
+      },
+      {
+        "description": "is case-insensitive",
+        "input": [
+          "1 2 Dup"
+        ],
+        "expected": [1, 2, 2]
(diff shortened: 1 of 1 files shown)


petertseng:
Before merging, I would like discussion on the following.

Should we include test cases for:

- Parsing negative numbers? (I recommend no)
- Cases for +, -, *, / with not enough arguments? How far should we go? Test everything in the cartesian product of {+, -, *, /} with {0, 1} arguments (8 tests in total)? I feel like we should for completeness, since all our other words have tests for not enough things on the stack.
- Ensuring that user-defined words are case-insensitive?
- Ensuring that user-defined words execute in the right order? For example that `: hi 1 2 ; hi` results in `1 2` instead of `2 1`? You may think "who would make such a mistake?", and to that I answer "Maybe you should look at my Haskell implementation of this exercise". Yes, I made that mistake. (I'm leaning yes here)
- Incomplete word definitions? (I'm leaning no)

In addition, I would to verify this for correctness, since I did a lot by hand.

exercises/forth/canonical-data.json
```diff
@@ -0,0 +1,214 @@
...
+        "description": "addition and subtraction",
+        "input": [
+          "1 2 + 4 -"
+        ],
+        "expected": "-1"
+      },
+      {
+        "description": "multiplication and division",
+        "input": [
+          "2 4 * 3 /"
```
petertseng: note that this integer division. maybe that should be called out with a separate test that points this out in its description.

exercises/forth/canonical-data.json
```diff
@@ -0,0 +1,219 @@
+{
+  "#": [
+    "The cases are split into multiple sections, all with the same structure.",
+    "In all cases, the `expected` key is the string representation of the resulting stack",
+    "if all numbers on it are printed (top-most value on the right), separated by a space."
```
petertseng: Time to question assumptions. why is it done this way? It was done this way in https://github.com/exercism/exercism.io/pull/1188 (the first time this exercise appeared), and I guess everyone else did the same. But why not check an array/list of values on the stack? Instead of `"expected": "1 2 3 4 5"`, why not `"expected": [1, 2, 3, 4, 5]` ?

exercises/forth/canonical-data.json
```diff
@@ -0,0 +1,219 @@
...
+          "1\u00002\u00133\n4\r5 6\t7"
+        ],
+        "expected": "1 2 3 4 5 6 7"
+      }
+    ]
+  },
+  "basic arithmetic": {
+    "cases": [
+      {
+        "description": "addition and subtraction",
```
petertseng: why addition *and* subtraction in one test? Why not separate them? A test with only addition, a test with only subtraction, and then possibly a test combining the two.

exercises/forth/canonical-data.json
```diff
@@ -0,0 +1,219 @@
...
+        "input": [
+          "1 2 + 4 -"
+        ],
+        "expected": "-1"
+      },
+      {
+        "description": "multiplication and division",
```
petertseng: as above, why multiplication *and* division in one test?

exercises/forth/canonical-data.json
```diff
@@ -0,0 +1,219 @@
...
+        "input": [
+          "2 4 * 3 /"
+        ],
+        "expected": "2"
+      },
+      {
+        "description": "division by zero results in an error",
+        "input": [
+          "4 2 2 - /"
```
petertseng: why not just `"4 0 /"`?

Is there an advantage of one or the other? Should we have both? Just `4 2 2 - /`? Just `4 0 /`?

exercises/forth/canonical-data.json
```diff
@@ -0,0 +1,219 @@
+{
+  "#": [
+    "The cases are split into multiple sections, all with the same structure.",
+    "In all cases, the `expected` key is the string representation of the resulting stack",
```
petertseng: for completeness, should clarify: "the resulting stack after executing the Forth program contained in the `input`" or something to that effect.

exercises/forth/canonical-data.json
kytrinyx: Agreed, seems like an array or list would make more sense here.

exercises/forth/canonical-data.json
kytrinyx: I think you're right. Doing them separately first would let people implement this iteratively.

exercises/forth/canonical-data.json
kytrinyx: Agreed. We should separate them.

exercises/forth/canonical-data.json
kytrinyx: +1

exercises/forth/canonical-data.json
kytrinyx: I can't think of a reason to have `4 2 2 - /` in addition to or instead of `4 0 /`. I vote for just plain 0.

kytrinyx:
> Parsing negative numbers? (I recommend no)

I also recommend no. I think this exercise has enough meat in it without it.

> Cases for +, -, *, / with not enough arguments? How far should we go? Test everything in the cartesian product of {+, -, *, /} with {0, 1} arguments (8 tests in total)? I feel like we should for completeness, since all our other words have tests for not enough things on the stack.

I'm leaning _yes_ on this one.

> Ensuring that user-defined words are case-insensitive?

Huh. I can't speak to this as I never did the exercise (and I've never written any forth).

> Ensuring that user-defined words execute in the right order? For example that : hi 1 2 ; hi results in 1 2 instead of 2 1? You may think "who would make such a mistake?", and to that I answer "Maybe you should look at my Haskell implementation of this exercise". Yes, I made that mistake. (I'm leaning yes here)

Definitely vote yes on this one :)

> Incomplete word definitions? (I'm leaning no)

Meh. Agreed, I don't think we need this, since we're very clear on being a _very small subset_ of forth.

petertseng:
OK.

* all expectations are arrays now instead of strings.
* arithmetic operator tests split up into own sections
* arithmetic operators each get tests that they error if there are zero or one elements on the stack
* integer division explicitly called out
* divide by zero simplified, it's just `4 0 /` now.
* add test for user-defined word executing in the right order (`countup 1 2 3`)

petertseng:
> In addition, I would to verify this for correctness, since I did a lot by hand.

OK. https://github.com/petertseng/x-common/blob/verify/exercises/forth/verify.rb

petertseng:
All right, all comments addressed, and nobody else seems to have any, so I exercise my judgment and say it's good.

exercises/forth/canonical-data.json
```diff
@@ -0,0 +1,334 @@
...
+        "description": "numbers just get pushed onto the stack",
+        "input": [
+          "1 2 3 4 5"
+        ],
+        "expected": [1, 2, 3, 4, 5]
+      },
+      {
+        "description": "all non-word characters are separators",
+        "input": [
+          "1\u00002\u00133\n4\r5 6\t7"
```
petertseng: Garbage. This \u0013 is a "device control 3". The original https://github.com/exercism/exercism.io/pull/1188 meant for this to be \u0001, "start of heading" (whatever that is).

My fault. Sorry.

PR #400, opened 2016-10-08
contribution-guide: adding section of extracting shared test data

junedev:
This is a proposal for #230. Unfortunately none of the people that participated in the discussion there made the actual pull request. So I took the liberty of putting it all together in the hope that this does not offend the original contributors to the discussion, especially @IanWhitney and @petertseng.

The discussion about notes/comments/msg that spawned off in #230 is not solved in this PR as it should lead to a change in the [readme](https://github.com/exercism/x-common#test-data-format-canonical-datajson) etc. and is not directly connected to the contribution guide.

For discussion: I would like to add a comment under point 4 about how to think about including edge cases. Any suggestions about how to better formulate the consensus there that seems to be something like "avoid including too many random edge cases but focus on the actual problem"?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -377,8 +377,23 @@ A problem must have a unique slug. This slug is used as
   this **must not** be merged until the exercism/x-common PR is merged.
 
 ## Improving Consistency By Extracting Shared Test Data
+The goal is to have a file `exercises/<slug>/canonical-data.json` for each exercise in this repository. This JSON file should contain the generic (i.e. language independent) test data for that exercise. With this the quality of the tracks can be improved, e.g. fixes or new tests can be propagated to all the tracks. They can then be implemented manually or by running the test generator. (See [this issue](https://github.com/exercism/todo/issues/13) for more background info.)
 
-TODO: elaborate.
+The following steps help you extracting the canonical test data.
+
+1. Select the exercise for which you want to create the shared test data from the [list of open todos](https://github.com/exercism/todo/issues?q=is%3Aissue+is%3Aopen+label%3Ashared-test-data)
+2. Now you need to find the existing tests. You can get a list with links to the all of those one of two ways
+  * Search for the exercise in this list http://x.exercism.io/v3/problems/data-todos (your browser might need a JSON viewer plugin/add-on to show this)
+  * Run this command in your terminal `curl http://x.exercism.io/v3/problems/data-todos | jq '.[] | map(select(.["slug"] == "accumulate"))'`. Replace `accumulate` with the name of the exercise you have chosen.
+3. Open the links in different browser tabs, navigate to the test files and read through the tests. Look for test cases that all/many of the languages have in common and identify outliers that only appear in one or two languages. It might be helpful to make some notes on your findings.
+4. Now it's time to write the canonical-data.json file. Look at examples like [the test data for hamming](https://github.com/exercism/x-common/blob/master/exercises/hamming/canonical-data.json) and check the [readme](https://github.com/exercism/x-common#test-data-format-canonical-datajson) to find out how to best structure the JSON file.
+From what you learned in step 3 decide for a sensible set of test cases. The following considerations might help with that.
+  * Try to stick to test cases that already occur in many languages so it is feasible for the track maintainers to adapt to the canonical set.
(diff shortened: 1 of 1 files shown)


IanWhitney:
I think this is a good start. Thanks! 

I'm not a word-smith, so I'll let others chime in on that. 

As for Point 4, I think the point I would want to make is that tests shouldn't focus too much on solving this exercise as a real-world problem, but should instead focus on teaching the student an interesting concept.

CONTRIBUTING.md
```diff
@@ -377,8 +377,22 @@ A problem must have a unique slug. This slug is used as
   this **must not** be merged until the exercism/x-common PR is merged.
 
 ## Improving Consistency By Extracting Shared Test Data
-
-TODO: elaborate.
+The goal is to have a file `exercises/<slug>/canonical-data.json` for each exercise in this repository. This JSON file should contain the generic (i.e. language independent) test data for that exercise. With this the quality of the tracks can be improved, e.g. fixes or new tests can be propagated to all the tracks. They can then be implemented manually or by running the test generator. (See [this issue](https://github.com/exercism/todo/issues/13) for more background info.)
+
+The following steps help you extracting the canonical test data.
+
+1. Select the exercise for which you want to create the shared test data from the [list of open todos](https://github.com/exercism/todo/issues?q=is%3Aissue+is%3Aopen+label%3Ashared-test-data)
```
petertseng: For us (not blocking this PR), we should make sure that either this list has all the problems (what if some new problems were added since those issues were created and they have no canonical data?) or make something that makes http://x.exercism.io/v3/problems/data-todos easy to visualise and recommend that here instead of the issues.

CONTRIBUTING.md
```diff
@@ -377,8 +377,22 @@ A problem must have a unique slug. This slug is used as
...
+4. Now it's time to write the canonical-data.json file. Look at examples like [the test data for hamming](https://github.com/exercism/x-common/blob/master/exercises/hamming/canonical-data.json) and check the [readme](https://github.com/exercism/x-common#test-data-format-canonical-datajson) to find out how to best structure the JSON file.
+From what you learned in step 3 decide for a sensible set of test cases. The following hints might help with that.
+  * Try to stick to test cases that already occur in many languages so it is feasible for the track maintainers to adapt to the canonical set.
+  * The tests should cover the described problem thoroughly enough but should not contain too many test cases that do not introduce any new aspect of the problem and already pass anyway.
+  * Make sure the described problem is tested. The test cases should not require people to follow one very specific way to solve the problem, e.g. the tests should avoid checking for helper functions that might not be present in other implementations.
+  * Performance tests should not be included in the canonical test data.
+5. Check the order of the tests. The simplest tests should be first.
```
petertseng: Is there anything specific about this that calls for a separate number? I ask because this seems to relate very closely to "should not contain too many test cases that do not introduce any new aspect of the problem and already pass anyway." - the main thing we might think about in test ordering is how easy it is to get to the first passing test. Then each test successively after that would ideally add a little functionality. All this so that the student can work iteratively and solve the problem a little at a time.

petertseng:
question for all - how much do we care about having this line-wrapped? We have been a little inconsistent about it in this file

petertseng:
> Any suggestions about how to better formulate the consensus there that seems to be something like "avoid including too many random edge cases but focus on the actual problem"?

I wish I could just say something about being "essential to the problem" (and I have used this before to not argue for testing certain things) but I do think that's also a bit vague. Maybe the idea can spark someone else's idea though.

I think sometimes we collectively a judgment call about whether a case would actually add to the problem and make it more interesting versus just be more tedious.

CONTRIBUTING.md
kytrinyx: How about combining these into a single point about carefully considering the test cases, and then elaborating about the things to consider?

kytrinyx:
> how much do we care about having this line-wrapped

Let's leave this for now. I'm thinking about how to improve the contributing docs to be less overwhelming. We can normalize line-wrapping in that change.

> essential to the problem

I think "essential to the problem" is actually quite good. Perhaps also add something about if in doubt call it out in the PR so we can discuss it.

junedev:
Thanks for all the feedback! Based on it I compiled a new list of points to consider under point 4 and updated the PR.

Insti:
Merging these because it's better to have something rather than nothing in the documentation. Any tweaks can be easily addressed by further PRs

Thanks @junedev :heart:

PR #402, opened 2016-10-09
Simplify alphametics. Fixes #395

### LINKED ISSUE #395: Alphametics: simplify or split into two exercises
I think the current Alphamatics exercise contains two separate problems.

* The original alphametics problem where you have to find letter-number substitutions. 
* Correctly parse a strings containing arbitrary mathematical formulas, with different operators and correct precedence. 

Solving the second part correctly requires different knowledge and most likely take longer to solve for most people. I guess in a dynamic language you can do this part through `eval` trickery, but that is not elegant and not available for less-dynamic languages. 

I can see people creating a partial parser or hardcode patterns that will satisfy all the test-cases and stop there, that would send the message that the goal of tests is to pass them and don't care about the general problem.

I see the following solutions: 
- Simplify the exercise to only use addition, this makes parsing puzzles much more easier. Then possibly split the formula parser into its own exercise without the alphametics part. 
- Create a separate "formula parser" exercise and make it clear that it should be solved before and extended for the alphametics puzzle.

This problem was briefly mentioned by @Insti in another not related issue: https://github.com/exercism/x-common/issues/273#issuecomment-226616740

ijanos:
This change addresses the issues outlined in #395. Using only addition simplifies the exercise to focus on one thing.

### DIFF
--- a/exercises/alphametics/canonical-data.json
+++ b/exercises/alphametics/canonical-data.json
@@ -13,16 +13,6 @@
                 "expected": { "S": 9, "E": 5, "N": 6, "D": 7, "M": 1, "O": 0, "R": 8, "Y": 2 }
             },
             {
-                "description": "solve puzzle with multiplication",
-                "puzzle": "IF * DR == DORI",
-                "expected": { "I": 8, "F": 2, "D": 3, "R": 9, "O": 1 }
-            },
-            {
-                "description": "solve puzzle with exponents",
-                "puzzle": "PI * R ^ 2 == AREA",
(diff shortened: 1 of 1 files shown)


petertseng:
Notes: F#, Lua, Ruby implement.

Looks like we have at least some agreement, and no dissent. And if you ask me personally I think just having addition is plenty too.

So I hope I'm not being too presumptuous here, but let's merge.

petertseng:
Notes for history: Alphametics JSON added in #154 (At the time, Lua was the only implementor), updated in #277, no changes other than that.

PR #403, opened 2016-10-10
Add word count cases

pvcarrera:
Related to https://github.com/exercism/xruby/pull/448

Add four additional test cases:
- Handles cramped lists: Tests that words separated by commas and no spaces are counted.
- Handles expanded lists: Tests that words separated by commas and new lines are counted.
- With apostrophes: Checks that words with apostrophes are also counted as words.
- With quotations: Checks that words with quotations are also counted as words.

### DIFF
--- a/exercises/word-count/canonical-data.json
+++ b/exercises/word-count/canonical-data.json
@@ -13,7 +13,7 @@
         {
             "description": "multiple occurrences of a word",
             "input": "one fish two fish red fish blue fish",
-            "expected": { "one" : 1, "fish" : 4, "two" : 1, "red" : 1, "blue" : 1 }
+            "expected": { "one": 1, "fish": 4, "two": 1, "red": 1, "blue": 1 }
         },
         {
             "description": "handles cramped lists",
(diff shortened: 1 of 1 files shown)


Insti:
@pvcarrera it's useful to give a more complete description of the changes you are proposing here. Not everyone will be familiar with Ruby or the xruby track.

pvcarrera:
@Insti sure, I've expanded the PR description, do you thing is good enough?

petertseng:
Notes for me: word-count.json was first added in #103 (I am far too lazy to figure out which tracks implemented it at the time, and it was not mentioned in #103 and so I have no insight on any decision to include or not include any cases). only update since then was #128 . Another thing I'll not have time to do is go to every track that implements word-count (and there are [a lot of them](http://x.exercism.io/problems/word-count)) and see whether they have these cases.

So all I can say is what I think from looking at the cases: The new test cases do indeed test new things, and I think they are interesting. I support their addition.

kytrinyx:
This looks great, thank you!

Maybe we can/should submit an issue to the various tracks that implement it and have each track check individually (using http://github.com/exercism/blazon). Actually, I might sit down and look through which new canonical-data.json files we've gotten that we haven't submitted issues for across the tracks, and script that up with `good-first-patch` labels.

PR #412, opened 2016-10-17
Rename flatten array to flatten

treyhunner:
Opened from a discussion in exercism/xpython#374.

I think this exercise should be renamed to use more generic terminology.  `flatten` is sufficiently generic.  Arrays in Python are different than lists and are not frequently used.

Thoughts on renaming this exercise?  Would it need to be renamed in all languages that currently include it?


Insti:
Some reasons why renaming might not be a good idea:

* It requires all the language tracks that implement the problem to also update their configurations.
* There are implications for maintaining history with the solutions of 'flatten-array' which have already been completed and stored in the Exercism database.
* There are many things you could flatten, having 'array' in the name doesn't seem unreasonable.
* It is possible to add track specific information to the readme that the Python track could use to point out that they're using Lists.

parkerl:
This was also discussed in xElixir. The name does not make sense there as we use lists rather than arrays. 

https://github.com/exercism/xelixir/pull/185

At the end of the discussion @kytrinyx outlined the steps need to make this change. Sorry, @treyhunner, you may have bitten off more than you thought with this PR. 

Here is what @kytrinyx said verbatim:

______________________________________

OK, cool. So I think we need to

- [ ] duplicate the exercise files here in x-common (so we can migrate individual language tracks independently of one another
- [ ]  create a blazon issue that can be submitted to all the languages that have implemented `flatten_array` (javascript, kotlin, ruby) to describe how to migrate the problem. I think this requires
    - renaming the problem directory
    - renaming the files
    - updating any code that refers to the old name
    - updating any comments or documentation that refer to the old name
- [ ] make sure we have a rake task that can be run in production to migrate existing solutions to the new name as soon as we deploy the migrated (see below)

We used to have a rake task that kind of did this, but it was all hard coded:

```

    desc "migrate deprecated problems"
    task :deprecated_problems do
      require 'bundler'
      Bundler.require
      require_relative '../exercism'
      {
        'point-mutations' => 'hamming'
      }.each do |deprecated, replacement|
        UserExercise.where(slug: deprecated).each do |exercise|
          if UserExercise.where(slug: replacement, user_id: exercise.user_id, language: exercise.language).count > 0
            next
          end

          exercise.slug = replacement
          exercise.save
          exercise.submissions.each do |submission|
            submission.slug = replacement
            submission.save
          end
          ACL.authorize(exercise.user, exercise.problem)
        end
      end
    end
```

It would be nice if we could have this task, but it would take two command-line arguments, one for the track_id and two from the problem slugs (src, dest).

Insti:
This is going to take some work, can you create a separate x-common issue with all that info @parkerl

kytrinyx:
The most important thing to note is that we need to copy, not move the files.

We should also add a `.deprecated` file to the `flatten-array` directory (we can add text to it to explain that it's been renamed to `flatten`, since any text in the file gets ignored. Or not, the `git blame` will end up pointing to this discussion, I think).

kotp:
Would it be simpler to introduce this as a new exercise, and deprecate the old?  Seems like a lot of coordination for something that if done as a new exercise would avoid all of it.

Insti:
It's clear that the work involved in changing the name of the exercise is more than the contents of this Pull Request. 

I will close this pull request, but I have created an issue: https://github.com/exercism/x-common/issues/454 about this issue if anyone wishes to continue working on it.

Issue #318, opened 2016-08-03
Potential new test in anagram: Same bytes different characters.


petertseng:
We got a new test in https://github.com/exercism/xrust/pull/175 . The test ensures that implementations do not take the approach of "sort the bytes then make sure the bytes are the same" which can give false positives in the case of multiple-byte characters.

Seems to make sense to add here.

petertseng:
Hello and thank you for filing this issue.

Unfortunately as you can see in #414 we decided to remove all non-ascii cases, therefore this case that you have described in this issue is no longer appropriate.

Have a nice day.

Issue #413, opened 2016-10-18
Anagram test data


fredrb:
https://github.com/exercism/x-common/blob/master/exercises/anagram/canonical-data.json#L106

Is this really a valid anagram? 
```
"candidates": ["ΒΓΑ", "ΒΓΔ", "γβα"],
"expected": ["ΒΓΑ", "γβα"]
```

Is seems to me that `"γβα"` should not be considered a valid anagram of `"ΑΒΓ"`. 
Also, this test failed when ruby generator was implemented exercism/xruby#464.

Thanks!

petertseng:
Seems correct, if it should be the case (as in earlier tests) that anagrams are case insensitive. "ΑΒΓ" lowercased is indeed "αβγ", which is a rearrangement of the "γβα"

rbasso:
Technically, I think it is an anagram, @fredb; because - surprisingly - it seems that `βγα` is `ΒΓΑ` lowercased.

That said, I don't think we should test extreme cases like this one, as it is *not an essential part of the problem* to deal with obscure case conversions.

NobbZ:
I do not think, that we should test for anything that is not in ASCII.

1. Most languages do not cope very well with codepoints beyond ASCII
2. There are letters beyond ASCII, that would need another step of normalization. Eg the german "es-zett" (ß) which is is only allowed as lowercase (but there is a capital version available which is allowed to be used on titlepages and headlines), on capitalisation it is usually turned into "SS". There has been the [long-s](https://en.wikipedia.org/wiki/Long_s), and similar characters in ancient greek (gamma was one of them AFAIK).

So we wozuld need to think about that characters before allowing them blindly.

Insti:
> I do not think, that we should test for anything that is not in ASCII.

I agree with this. Handling non-ASCII casing is a whole other problem.
:+1: for removing this particular test case from the canonical-data

PR #414, opened 2016-10-18
remove non-ascii test cases from anagram

fredrb:
Removes unicode specific test cases for anagram. This refers to #413.


petertseng:
Remind me to close #318 too.

PR #415, opened 2016-10-19
crypto-square: Clarifying how whitespace should be handled

smaclell:
It is hip to be square

I was a little confused by the description for how extra whitespace should be handled in the `crypto-square` problem. Based on the tests in the [xecmascript](https://github.com/exercism/xecmascript) instead of evenly spacing the extras whitespace across multiple rows, the whitespace appeared to be added to the final row.

I have updated the wording to try to better reflect how the whitespace was handled. Thank you for such an awesome project and being so open to contributions.

### DIFF
--- a/exercises/crypto-square/description.md
+++ b/exercises/crypto-square/description.md
@@ -44,10 +44,12 @@ imtgdvsfearwermayoogoanouuiontnnlvtwttddesaohghnsseoau
 
 Output the encoded text in chunks.  Phrases that fill perfect squares
 `(r X r)` should be output in `r`-length chunks separated by spaces.
-Imperfect squares will have `n` empty spaces.  Those spaces should be distributed evenly across the last `n` rows.
+Imperfect squares will have `n` empty spaces.  The final row should
+not have any trailing whitespace instead of including `n` empty
+spaces at the end.
 
 ```plain
-imtgdvs fearwer mayoogo anouuio ntnnlvt wttddes aohghn sseoau
+imtgdvs fearwer mayoogo anouuio ntnnlvt wttddes aohghns seoau
 ```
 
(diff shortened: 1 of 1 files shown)


rbasso:
> Imperfect squares will have `n` empty spaces.  Those spaces should be distributed evenly across the last `n` rows.

From my understanding, this is referring to the square after transposition.

If I'm not getting this wrong, this change will make the description incompatible with the current tests, right?

exercises/crypto-square/description.md
```diff
@@ -60,6 +62,6 @@ mayoogo
 anouuio
 ntnnlvt
 wttddes
-aohghn
-sseoau
+aohghns
```
petertseng: now reading the text top to bottom, it will be misaligned - before this change it was "if man was meant to stay on the ground god would have given us roots", and now it is "if man was meant to etay on tho ground gad would huve given s rootss" - is that okay? One can argue it is a measure to obscure the plain text, in which case we should just pick a group size unrelated to the square size, such as "always 5"

You can read previous discussion about it in #190, but it does appear the https://github.com/exercism/x-common/blob/master/exercises/crypto-square/canonical-data.json file matches the version *before* this PR rather than *after* it.

smaclell:
Thanks everyone for your feedback! @Insti, I will take your suggestion and go submit a PR to fix the bad specs in the ecmascript track. Have a great day!

exercises/crypto-square/description.md
smaclell: I am going to trust the canonical data and will instead go fix the specs in the `ecmascript` track. You are also correct that my description change would have misaligned the data. Thank you for responding.

Insti:
Great. Thanks for taking the initiative to try and fix the problem you found @smaclell

PR #416, opened 2016-10-20
Fix bowling description text case and typo

parkerl:
I noticed some inconsistent casing in the test descriptions for this exercise.

PR #418, opened 2016-10-21
[Bowling] Add edge case with fill balls in a last-frame strike

IanWhitney:
The previous test does not quite capture the behavior.

If the first fill fill is a non-strikes, then the total fill-ball score
must be less than 10.

But, if the first fill ball is a strike, then the total fill-ball score
must be less than 20. Because that first fill-ball strike resets the
pins.

I had what I thought was a working implementation but it totally missed
the 2nd case. Adding a test to cover that case.

### DIFF
--- a/exercises/bowling/canonical-data.json
+++ b/exercises/bowling/canonical-data.json
@@ -15,7 +15,7 @@
   ],
   "score": {
     "description": [
-      "Returns the final score of a bowling game"
+      "returns the final score of a bowling game"
     ],
     "cases": [{
       "description": "should be able to score a game with all zeros",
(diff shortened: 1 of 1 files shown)


IanWhitney:
I'm also wondering about adding 

```json
 }, {
      "description": "Two bonus rolls after a strike in the last frame can have a strike on the second roll",
      "rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10, 6, 10],
      "expected": -1
```

To check for logic that looks for either of the fill balls being == 10 and then checking that the fill score is <= 20

petertseng:
Case that is currently added 10, 10, 6 seems good.

You might think 10, 6, 10 seems to overlap with our existing 10, 5, 6, so it would only fail if someone is special-casing 10. Admittedly, someone might, given the before test. Given that we can think of an implementation that would fail it that it seems like someone could write, seems good to add it, I think.

IanWhitney:
`10,6,10` definitely failed for me. I had some code that checked if any of the fill balls were a 10. And if so, it checked that the total of the fill balls was less than 20. That was incorrect.

Insti:
Oh dear I've made a mess of this :cry: sorry!

Insti:
@IanWhitney can you make another PR for this? I can't work out how to re-apply this PR through github.

I merged this PR, then saw that it caused #416 to have conflicts and thought if I merged that one first I could merge this one without conflicts, so I reverted this merge and applied #416 but now I cant seem to re-apply this one.

Insti:
I think I've recovered this now, and don't need @IanWhitney to make another PR.

Sorry everyone for the trouble. :frowning_face:

Issue #294, opened 2016-07-14
Raindrops; prime factors wording is too specific.


kotp:
The example for the raindrops happen to be prime numbesr, but the focus should be on factors.

The game generally is exampled with 3, 5, and then continued on with further primes, but there should be no reason that someone couldn't introduce a non-prime number as a rule.

At some point, there was a mention of only the factors, while 'dangling' the "prime factor" phrasing, but not requiring it, providing a red herring which brought about some useful discussions.

PR #427, opened 2016-10-28
raindrops: include 1 and n in factor list in examples

common-nighthawk:
i believe that technically 1 and n are always factors of number n.

but more importantly--not including them in the examples leads to failures in the tests.

`raindrops_test.go:14: Convert(3) = "3", expected "Pling".`

that is, one can say the factors of 3 are either--
`[]int{1, 3}`
or
`[]int{}`

the tests expect 3 to be in the factor list, so i think it makes sense to include 1 and n in the examples :)

### DIFF
--- a/exercises/raindrops/description.md
+++ b/exercises/raindrops/description.md
@@ -6,10 +6,10 @@
 
 ## Examples
 
-- 28's factors are 2, 4, **7**, 14.
+- 28's factors are 1, 2, 4, **7**, 14, 28.
   - In raindrop-speak, this would be a simple "Plong".
-- 30's factors are 2, **3**, **5**, 6, 15.
+- 30's factors are 1, 2, **3**, **5**, 6, 15, 30.
   - In raindrop-speak, this would be a "PlingPlang".
-- 34 only has two factors- 2 and 17.
+- 34 only has four factors: 1, 2, 17, and 34.
   - Raindrop-speak doesn't know what to make of that,
     so it just goes with the straightforward "34".
--- a/exercises/raindrops/description.md
+++ b/exercises/raindrops/description.md
@@ -10,6 +10,6 @@
   - In raindrop-speak, this would be a simple "Plong".
 - 30's factors are 1, 2, **3**, **5**, 6, 15, 30.
   - In raindrop-speak, this would be a "PlingPlang".
-- 34 only has four factors: 1, 2, 17, and 34.
+- 34 has four factors: 1, 2, 17, and 34.
   - Raindrop-speak doesn't know what to make of that,
     so it just goes with the straightforward "34".
--- a/exercises/raindrops/description.md
+++ b/exercises/raindrops/description.md
@@ -11,5 +11,4 @@
 - 30's factors are 1, 2, **3**, **5**, 6, 15, 30.
   - In raindrop-speak, this would be a "PlingPlang".
 - 34 has four factors: 1, 2, 17, and 34.
-  - Raindrop-speak doesn't know what to make of that,
-    so it just goes with the straightforward "34".
+  - In raindrop-speak, this would be "34".


exercises/raindrops/description.md
```diff
@@ -6,10 +6,10 @@
...
 ## Examples
 
-- 28's factors are 2, 4, **7**, 14.
+- 28's factors are 1, 2, 4, **7**, 14, 28.
   - In raindrop-speak, this would be a simple "Plong".
-- 30's factors are 2, **3**, **5**, 6, 15.
+- 30's factors are 1, 2, **3**, **5**, 6, 15, 30.
   - In raindrop-speak, this would be a "PlingPlang".
-- 34 only has two factors- 2 and 17.
+- 34 only has four factors: 1, 2, 17, and 34.
```
Insti: Or even better, use a prime number so it does have only 2.

kotp: Don't use a prime number, as that suggests a restriction that the rules can only be prime.  The 34 is here to make that distinction a little more clear.

kotp:
There is no requirement for any list to exist, only that three is determined to be a factor.  By including some "list" you are expanding on the problem, unnecessarily.

exercises/raindrops/description.md
kotp: 1 and self are special cases.  The problem is such that you have a factor with a replacement, it matters not that there are other factors.  This problem is not about primes (the definition of a number that is divisible by one and itself only) but about finding out if a number has a certain factor, and making that replacement.  

The lists of numbers here seem to be least common denominators, and so are accurate as listed, I think.

common-nighthawk:
Hm.  I respectfully disagree.  It is not a "good" solution, but I don't see anything wrong with finding a list of the factors and then checking if 3 and then 5 and then 7 are in that list.

For reference--
http://exercism.io/submissions/5b6a9ff35f7a441787555c75fa4de7e6

Additionally, while it is true that "no requirement for any list to exist"--the instructions talk a lot about "the number's factors" and not numbers being divisible.

If the examples said, "Let's check to see if 5 is a factor of 25 by doing this..." I would agree.  But the examples illustrate how to find all the factors.

Because the expected result of the number 3 is "Pling" (and this in the test spec), I'm not clear on the argument to exclude self from the list of factors.

With all that said--you are welcome to close the PR if you disagree.  The issue seems small enough :)

kotp:
It's good to talk about it.  I just always took it as "building a string" which can be done by applying to the list the string that corresponds with the factor, rather than finding all the factors... since in order to get "PlingPlangPlong" you could present the number 0, and that can be divided by all the numbers, which is a pretty large list.

exercises/raindrops/description.md
kotp: Remember to read the information above the examples, because taken outside of the context of "If the number has 3 as a factor..." then the examples are understandably incomplete.

common-nighthawk:
Fair point regarding 0.

1.  It doesn't seem to be an uncommon soltuon to find all the factors: http://exercism.io/submissions/02ec2760aeee46e6b255172cba015d35

2.  The instructions should work for all solutions, not just efficient ones.

3.  Does it make sense to edit--
`28's factors are 2, 4, 7, 14.`
to something closer to
`28's factors include 7, but neither 3 nor 5.
In raindrop-speak, this would be a simple "Plong".`

kotp:
It is good that the most efficient solutions are not necessarily the most common ones, otherwise there would be not much to discuss and discover.

Have you read #294 #353 and #295 issues/prs?

common-nighthawk:
Thanks--the history is helpful.  I still think adding 1 and n to the list is the right approach.  If anything, the fact they aren't in the list may be an incorrect holdover from the legacy instructions.  That is, they were not part of the prime factors list.  But they should be included in the list of factors.

Insti:
1 and n should definitely be in the list of factors since the 'factors' terminology is what we've settled on. 

@kotp wrote:
> Don't use a prime number, as that suggests a restriction that the rules can only be prime. 

I don't understand how you reach that conclusion.

```
37 only has two factors: 1 and 37.

    Raindrop-speak doesn't know what to make of that, so it just goes with the straightforward "37".
```

The only thing that says about the rules is that 1 and 37 are not special factors, which we already knew.

Insti:
But I really don't care which numbers get used in the example, just that if you have 4 don't say 'only 4' unless your other numbers have more than 40 factors. (which they shouldn't)

kotp:
I really meant "leave in the non-prime" in the example, more than "don't use a prime number".  It helps to alleviate the tendency to assume that the rules have to be prime, rather than are coincidentally.

Insti:
So add the 1,n factors, leave example as 34, take out the 'only' and everyone is happy?

common-nighthawk:
I agree with all the latest comments.  I updated the PR.

exercises/raindrops/description.md
```diff
@@ -6,10 +6,10 @@
...
-- 28's factors are 2, 4, **7**, 14.
+- 28's factors are 1, 2, 4, **7**, 14, 28.
   - In raindrop-speak, this would be a simple "Plong".
-- 30's factors are 2, **3**, **5**, 6, 15.
+- 30's factors are 1, 2, **3**, **5**, 6, 15, 30.
   - In raindrop-speak, this would be a "PlingPlang".
-- 34 only has two factors- 2 and 17.
+- 34 has four factors: 1, 2, 17, and 34.
   - Raindrop-speak doesn't know what to make of that,
     so it just goes with the straightforward "34".
```
kotp: It should match the language above.

`In raindrop-speak, this would be "34".`

Insti:
Thanks @common-nighthawk :heart:

Issue #428, opened 2016-10-31
Discussion: Non-Ascii test cases.


Insti:
While updating the `anagram` test cases (issue: https://github.com/exercism/x-common/issues/413)
Discussion of handling non-ascii characters came up, and we decided that we would NOT use non-ascii characters in those tests.

@NobbZ [made a good point:](https://github.com/exercism/x-common/issues/413#issuecomment-254424051)
> I do not think, that we should test for anything that is not in ASCII.

> 1.   Most languages do not cope very well with codepoints beyond ASCII
> 1. There are letters beyond ASCII, that would need another step of normalization. Eg the german "es-zett" (ß) which is is only allowed as lowercase (but there is a capital version available which is allowed to be used on titlepages and headlines), on capitalisation it is usually turned into "SS". There has been the long-s, and similar characters in ancient greek (gamma was one of them AFAIK).

[Isogram](https://github.com/exercism/x-common/blob/master/exercises/isogram/canonical-data.json) (as of 20161031) also has non-ascii test cases.

Are there other problems that have non-ascii test cases?

I've created this issue so we can discuss the general policy of whether non-ascii characters should be used in test cases and have a a thread to point to when it comes up again in the future.

**Proposal:**
All test cases should only use ASCII characters 
(Unless extended character handling is integral to the problem.)

rbasso:
Aye!

I believe that the same logic can be applied to a broader class of questions:

> Q: Should we test for ...?
> A: No! Unless that is a fundamental part of the problem.

Anytime we testing for something that is nothing fundamental, we reduce diversity in the solutions and it gets boring to review code. :grin:

ErikSchierboom:
Agreed. I think having the non-ASCII test in Isogram does not really add anything. In fact, I think most people will be confused by it. If someone could come up with another exercise in which non-ASCII character handling does make sense, I'm all for it. But for the current exercises, I think it should be removed.

NobbZ:
I do like to have some of non US-ASCII test cases as an optional part for a very small subset of exercises. They can teach you a lot of unicode-handling if and only if you are open to it and want to learn it.

Anagram is not one of them, since it requires you to normalize charakters and some of them can't be normalized. In the example of the german `ß` vs. `SS`, is `MASSE` an anagram of `Maße`? How is it the other way round?

Word-Count on the other hand does gain a lot by adding the Unicode sugar to separate words from each other. But, well, the normalisation problem persists, but its influence is by far not that strong.

[...]

PR #433, opened 2016-11-04
isogram: Revise canonical test data

behrtam:
This removes the two unicode test cases as discussed in #428. One test case was an isogram with German umlauts, the other one was a French word with a duplicated character with accent (also non-ASCII). I don't think we loose anything here without those test cases that isn't covered by the remaining ones.

I also added descriptions for the test cases and a test case with an empty string. *If you would prefer those non unicode changes in a separate PR, just ask.*

### DIFF
--- a/exercises/isogram/canonical-data.json
+++ b/exercises/isogram/canonical-data.json
@@ -1,27 +1,44 @@
 {
-  "cases": [{
-    "input": "duplicates",
-    "expected": true
-  }, {
-    "input": "eleven",
-    "expected": false
-  }, {
-    "input": "subdermatoglyphic",
-    "expected": true
-  }, {
-    "input": "Alphabet",
-    "expected": false
-  }, {
-    "input": "thumbscrew-japingly",
-    "expected": true
-  }, {
-    "input": "Hjelmqvist-Gryb-Zock-Pfund-Wax",
-    "expected": true
-  }, {
-    "input": "the quick brown fox",
-    "expected": false
-  }, {
-    "input": "Emily Jung Schwartzkopf",
-    "expected": true
-  }]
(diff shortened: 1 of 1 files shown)


Insti:
Awesome, looks good. thanks @behrtam

exercises/isogram/canonical-data.json
```diff
@@ -1,33 +1,44 @@
...
-  }]
+  "cases": [
+    {
+      "description": "empty string has no duplicates",
+      "input": "",
+      "expected": true
+    },
+    {
+      "description": "isogram with only lower case characters",
+      "input": "duplicates",
```
Insti: I'd prefer to use a different word here as 'duplicates' is misleading. (since there are no duplicates.)
`isogram` could be good, since `isogram` is an isogram

Insti:
Including a test case with spaces might also be good.

"the quick brown fox jumps over the lazy dog"
Might be interesting since it is a panagram, but not an isogram.

"no duplicates"
Is an isogram and is autological

Insti:
(sorry to hijack your PR with potentially unrelated changes.)

exercises/isogram/canonical-data.json
behrtam: It its kind of funny a bit like that isogram**s** is no isogram (in the readme). But I agree that we should change that.

behrtam:
The last test case `"Emily Jung Schwartzkopf"` has two white spaces, so you thing we should add a negative test case with white spaces?

kytrinyx:
@behrtam I think the `"Emily Jung Schwartzkopf"` case covers it.

Insti:
Thanks @behrtam :heart:

PR #434, opened 2016-11-04
run-length-encoding: Remove non-ASCII test cases

behrtam:
This removes the two unicode test cases as discussed in #428.

I also added one test case with an empty string and one with only chars that occur once. *If you would prefer those non unicode changes in a separate PR, just ask.*

### DIFF
--- a/exercises/run-length-encoding/canonical-data.json
+++ b/exercises/run-length-encoding/canonical-data.json
@@ -1,6 +1,6 @@
 {
    "#": [
-   	  "The tests deliberately check that single characters are encoded without count."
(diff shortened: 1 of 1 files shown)


kotp:
Separate commit maybe, not a separate PR.

Why the whitespace differences?

behrtam:
Okay, removed the new test cases from this PR and fixed the unintentional changed white spaces (json auto formatter in my editor).

Insti:
Thanks @behrtam :heart:

Insti:
Ref: https://github.com/exercism/x-common/issues/441

PR #435, opened 2016-11-04
hello-world: Rename language to track ID

kytrinyx:
In Exercism "language" refers to the programming language name, (e.g.
C++), whereas Track ID refers to the identifier used on Exercism (e.g.
cpp).


kytrinyx:
Right, I didn't check which languages have implemented hello world, but enough of the slugs are different from the language, and the distinction matters.

Issue #432, opened 2016-11-03
sublist: Possible missing edge case


behrtam:
Found a clever implementation ([link](http://exercism.io/submissions/576e9759b6c34816ae204812388f58c4)) in Python for the sublist exercise that shouldn't pass all tests. It basically joins all the digits (`[1, 2, 3] -> "123"`) and uses the `in` operator that checks if one string is a substring of the other.

If we would have two list `[1, 0, 1]` and `[10, 1]` that particular solution would create two strings `"101"` and `"101"` and determine incorrectly that those lists are equal. Is this way to solve this too unlikely or should we add this as a test case?

Insti:
I think that is a good test case to add. :+1: 

Part of a good test suite is detecting things that pass that should not.

PR #438, opened 2016-11-06
sublist: Add digits edge case test

behrtam:
see #432

### DIFF
--- a/exercises/sublist/canonical-data.json
+++ b/exercises/sublist/canonical-data.json
@@ -92,5 +92,10 @@
 		"list_one": [1, 2, 3],
 		"list_two": [3, 2, 1],
 		"expectation": "unequal"
+	}, {
+		"description": "same digits but different numbers",
+		"list_one": [1, 0, 1],
+		"list_two": [10, 1],
+		"expectation": "unequal"
 	}]
 }


Insti:
It can be good to copy/paste an issue description into the first post of a Pull request rather than just referencing it so reviewers can get an idea of what the issue being resolved is without having to look through external threads.

petertseng:
More importantly, it would be good if that could be summarised in the commit message so that if I lack internet access and have the repo checked out but no access to Github, I can still understand why this test was aadded.

kytrinyx:
@behrtam Would you mind amending the commit as @petertseng describes?

tejasbubane:
@behrtam Can you update the commit message so we can get it merged in? Also while you are at it, please add `Fixes #issue-number` so that the issue is closed automatically once this PR is merged: https://help.github.com/articles/closing-issues-via-commit-messages/ and we don't need to remember to close it.

behrtam:
I added the `Fixes #issue-number` reference to the commit message. Didn't know about that handy trick.
I won't duplicate the issue in the commit message. That doesn't feel right. If the issue would be a longer discussion and maybe even have references it self or images or something, where do we stop? ... copy/paste just to save one click?

tejasbubane:
> I won't duplicate the issue in the commit message. That doesn't feel right. If the issue would be a longer discussion and maybe even have references it self or images or something, where do we stop? ... copy/paste just to save one click?

I agree with @Insti and @petertseng that a commit should have some context as to why this change was made. Refer: https://www.kernel.org/pub/software/scm/git/docs/git-commit.html#_discussion 
The commit message should be short (which is what you already have) and then there should be **description** which explains the commit. 
For more tips and details for writing good commit messages see this: http://chris.beams.io/posts/git-commit/

Insti:
@behrtam it's good to be able to know what a commit is about by simply looking through the git logs on your local machine rather than requiring internet access and GitHub to be available.

petertseng:
Seems like we have agreement. Thanks to all involved!

PR #439, opened 2016-11-06
pangram: Add case-insensitive test

behrtam:
see #266

### DIFF
--- a/exercises/pangram/canonical-data.json
+++ b/exercises/pangram/canonical-data.json
@@ -39,6 +39,11 @@
         "description": "pangram with mixed case and punctuation",
         "input": "\"Five quacking Zephyrs jolt my wax bed.\"",
         "expected": true
+      },
+      {
+        "description": "upper and lower case versions of the same character should not be counted separately",
+        "input": "the quick brown fox jumped over the lazy FOX",
+        "expected": false
       }
    ]
 }


exercises/pangram/canonical-data.json
```diff
@@ -49,6 +49,11 @@
          "description": "Panagram in alphabet other than ASCII",
          "input": "Широкая электрификация южных губерний даст мощный толчок подъёму сельского хозяйства.",
          "expected": false
+      },
+      {
+        "description": "test that count is case-insensitive",
```
Insti: This needs a better description of what we're testing for.
case insensitivity is already covered by: "Five quacking Zephyrs jolt my wax bed."

Insti: How about: "upper and lower case versions of the same character should not be counted separately"

kytrinyx:
@behrtam Here also a more descriptive commit message would be really helpful.

behrtam:
> I think there's a better description than that, I just don't know what it is right now.

Any new ideas or someone else?

Insti:
No new ideas, "upper and lower case versions of the same character should not be counted separately" is the best I have for now.

stkent:
+1 to "upper and lower case versions of the same character should not be counted separately"; it's a little verbose, but perfectly clear.

behrtam:
Perfect, thanks both of you. Than I guess we are good to go.

Insti:
Thanks @behrtam :heart:

PR #440, opened 2016-11-06
pangram: Remove non-ascii test cases.

Insti:
In https://github.com/exercism/x-common/issues/428 the decision was made
to remove non-ascii test cases from exercises that are not explicitly
about extended character set handling.

This PR removes the non-ascii test cases from the `canonical_data.json`
for this exercise.


Insti:
Ref: https://github.com/exercism/x-common/issues/441

PR #444, opened 2016-11-08
bowling: Add tests for more than 10 after last frame

petertseng:
We have tests for:

* Rolling 11 in a single roll
* Rolling 5 then 6
* Rolling 5 then 6 after a last-frame strike

So it appears we are missing the test for rolling an 11 after a
last-frame strike. Let's add it.

### DIFF
--- a/exercises/bowling/canonical-data.json
+++ b/exercises/bowling/canonical-data.json
@@ -86,6 +86,10 @@
       "rolls": [5, 6, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       "expected": -1
     }, {
+      "description": "bonus roll after a strike in the last frame can not score more than 10 points",
+      "rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10, 11],
+      "expected": -1
+    }, {
       "description": "two bonus rolls after a strike in the last frame can not score more than 10 points",
       "rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10, 5, 6],
       "expected": -1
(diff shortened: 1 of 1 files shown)


exercises/bowling/canonical-data.json
```diff
@@ -98,6 +102,10 @@
       "rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10, 6, 10],
       "expected": -1
     }, {
+      "description": "second bonus roll after a strike in the last frame can not score than 10 points",
+      "rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10, 10, 11],
```
petertseng: unsure about the necessity of this one. But maybe it's useful - some people may only check the first bonus roll?

exercises/bowling/canonical-data.json
```diff
@@ -86,6 +86,10 @@
       "rolls": [5, 6, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       "expected": -1
     }, {
+      "description": "bonus roll after a strike in the last frame can not score more than 10 points",
+      "rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10, 11],
```
petertseng: My first implementation only checked the second bonus roll and would therefore allow anything in the first, oops. Gotta catch that with this 11.

Insti:
I think it's a good idea to add explicit test cases for any bugs that are seen 'in the wild'.


The only roll that is checked for 11 is the first one. (and now the last)
Should we also have tests that the second or third roll (etc..) cannot be 11?

petertseng:
It is true - if we wish to be truly thorough, we would test that each and every roll cannot be 11.

It might be a lot to put in the JSON file; I wish we had some way of compactly expressing what I have just put in words.

The current set gets us a decent amount of mileage since (it is anticipated) most students will write the code such that the first nine frames use the same logic, and the tenth frame has a bit of extra logic for the fill balls. I admit it's a bit interesting to see that indeed we are paying an awful lot of attention to the last frame, but it rightfully is a hotspot where a lot of the problems come up.

Insti:
> The current set gets us a decent amount of mileage since (it is anticipated) most students will write the code such that the first nine frames use the same logic, and the tenth frame has a bit of extra logic for the fill balls.

Yeah, I'm fine for leaving it as it is and adding individual tests whenever anyone implements a solution that doesn't get caught.

PR #446, opened 2016-11-11
[WIP] queen-attack: use Algebraic Notation for coordinates?

jtigger:
*(this PR is not intended to be merged, but start a conversation; additional changes edits to the README would be required to make this edit complete)*

Typically, chess coordinates are expressed in [AN](https://en.wikipedia.org/wiki/Algebraic_notation_(chess)).  Wouldn't it add some flavor to this exercise to use that notation?

Doing so...
- ups the fun factor 🎉  by increasing the depth of the exercise by pulling concepts from the domain of the problem space;
- makes the exercise slightly more challenging 💪  ;
- with numeric tuples, it's really easy to slip into [primitive obsession](http://wiki.c2.com/?PrimitiveObsession) (passing around arrays of integers, for example) — by using AN, it's easier to make the "argument" to address that code smell 🐽 .

### DIFF
--- a/exercises/queen-attack/canonical-data.json
+++ b/exercises/queen-attack/canonical-data.json
@@ -14,35 +14,35 @@
       {
         "description": "queen with a valid position",
         "queen": {
-          "position": "(2,2)"
+          "position": "d1"
         },
         "expected": 0
       },
       {
         "description": "queen must have positive rank",
         "queen": {
-          "position": "(-2,2)"
+          "position": "a-1"
         },
         "expected": -1
       },
       {
         "description": "queen must have rank on board",
         "queen": {
-          "position": "(8,4)"
+          "position": "f9"
         },
         "expected": -1
       },
       {
         "description": "queen must have positive file",
         "queen": {
-          "position": "(2,-2)"
+          "position": "_3"
         },
         "expected": -1
       },
       {
         "description": "queen must have file on board",
         "queen": {
-          "position": "(4,8)"
+          "position": "i3"
         },
         "expected": -1
       }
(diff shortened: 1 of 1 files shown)


rbasso:
I think that `queen-attack` used an unfortunate indexing system, as I already pointed out in https://github.com/exercism/x-common/issues/313#issuecomment-236798084, so I think that it deserves to be updated.

At least, we need to change the indexing system to be (1,1) based, right-up oriented, to match chess, but I'm still uncertain about if using chess notation in a string would be putting to much in a single exercise.

If all the exercises involve parsing, we dilute what makes each one of them interesting, so I'm usually against mixing too many things in the same problem. I believe that the problems should be short and focused, this way they don't get tiring and boring.

Maybe it is OK in this case, because it is related to chess anyway. I don't know...

Taking the opportunity to discuss another thing in the test suite, the `create` section has an interesting description:

> *Test creation of Queens with valid and invalid positions*

It seems that this was written with OO in mind. What is being *created* :question:

Insti:
I am in favour of using chess co-ordinates.

Reasoning:
- This is phrased as a chess problem. 
- Working out if the queen is on the board is irrelevant to the problem.
- Which way the board/co-ordinates are oriented is irrelevant to the problem.
- We could add a different problem 'chess-co-ordinates' that is a simple one to handle 
algebraic notation mappings that the student will have encountered before reaching this problem.

Additional proposal: 
- Remove co-ordinate validation checks from the test cases.

jtigger:
@rbasso said:
> Taking the opportunity to discuss another thing in the test suite, the create section has an interesting description:
>
> > Test creation of Queens with valid and invalid positions

> It seems that this was written with OOP in mind. What is being created ❓

... and @Insti replied:

> Working out if the queen is on the board is irrelevant to the problem.

👍

exercises/queen-attack/canonical-data.json
```diff
@@ -14,35 +14,35 @@
...
-          "position": "(8,4)"
+          "position": "f9"
         },
         "expected": -1
       },
       {
         "description": "queen must have positive file",
         "queen": {
-          "position": "(2,-2)"
+          "position": "_3"
```
stevejb71: This does not represent a negative file.

IanWhitney:
I would like us to think about the complexity we're adding to the problem by making this change. 

- Instead of checking numbers students will now parsing & validating a string.
- Instead of math boundaries for ranks students are now checking presence of a char.
- Instead of math for diagonal checking, students are (probably) converting chars to integers and doing math

Which, fine, if there's some benefit to it. But I don't see what the benefit is beyond "It makes it more chess like." It certainly gives the students more things to _do_, but I don't think it makes the problem any better.

I don't think verisimilitude should be our goal in these problems. If making the problem more realistic makes the learning experience better, then I'm all for it. But I don't see it here.

jtigger:
> I don't think verisimilitude should be our goal in these problems. If making the problem more realistic makes the learning experience better, then I'm all for it. But I don't see it here.

Yeah, well put, @IanWhitney... we don't want difficulty for difficulty's sake; that's just annoying.  Having had a chance to go away and come back to the idea, I see how the added requirement would likely just layer on more work, not change the solution fundamentally.  That is, practitioners would probably just have to do the work to interpret the AN into integer tuples and then everything would remain the same.

Thanks for the discussion!

PR #448, opened 2016-11-12
matrix: description: add newlines in sample string

stkent:


kytrinyx:
Thanks @stkent. Would you mind elaborating on the problem/fix? It looks like there were already newlines in the example string (or am I reading this wrong?).

stkent:
@kytrinyx sure thing! I think the change is best illustrated using images of the rendered markdown:

# Before

String in markdown file contains newlines, but renders without them:

screen shot 2016-11-22 at 9 12 41 pm

# After

String in markdown file contains newlines and trailing spaces, so it renders with newlines:

screen shot 2016-11-22 at 9 13 00 pm

Adding two spaces at the end of each quoted line forces a newline to be _rendered_. IMO this is a less confusing presentation for the reader (though I suppose that may not be true if you're not viewing the rendered version of the file... hmm...).

Input welcome!

Insti:
Thanks @stkent :heart:

kytrinyx:
The screenshots sealed the deal for me :-) Thank you @stkent!

Issue #441, opened 2016-11-06
Update or remove non-ASCII test cases.


Insti:
In https://github.com/exercism/x-common/issues/428 the decision was made to remove test cases containing non-ASCII characers from exercises that are not explicitly about extended character set handling.

The following tests were identified as containing test cases that need to be fixed:

- [x] exercises/atbash-cipher/canonical-data.json
- [x] exercises/bob/canonical-data.json
- [x] exercises/forth/canonical-data.json
- [x] exercises/isogram/canonical-data.json
- [x] exercises/pangram/canonical-data.json
- [x] exercises/run-length-encoding/canonical-data.json
- [x] exercises/scrabble-score/canonical-data.json

Please check before removing a test that it is not **also testing another property of the solution**. If it is, please update the test to remove the non-ascii characters while maintaining the other intended purpose.

ldwoolley:
The discussion is still open?  isn't it?  I think handling unicode where ever possible future proofs the solutions and improves the programming solutions.

robkeim:
I just sent out #527 which addresses atbash-cipher, forth, and scrabble-score.

The last exercise [isogram](https://github.com/exercism/x-common/blob/master/exercises/isogram/canonical-data.json) was already addressed in #433.

PR #449, opened 2016-11-19
Remove non-Latin alphabetic characters from Bob. Also remove the file…

stevejb71:
… comment which looks to be not relevant any more.

This is to address issue #441.


stevejb71:
Added back in the ascii whitespace character test.

PR #370, opened 2016-09-09
raindrops: include edge cases in tests

McEileen:
I added new tests for edge cases after opening issue #434 on the Ruby track.  https://github.com/exercism/xruby/issues/434

PR #450, opened 2016-11-22
Add descriptions for the raindrops tests.

stevejb71:
Unlike the other canonical-data I have seen, the tests have no description.
I'm writing a test generator for the xocaml stream - it fails due to this.

### DIFF
--- a/exercises/raindrops/canonical-data.json
+++ b/exercises/raindrops/canonical-data.json
@@ -26,7 +26,7 @@
          "expected": "Pling"
       },
       {
-        "description" : "the sound for 8 is 8",
+        "description" : "2 to the power 3 does not make a raindrop sound as 3 is the exponent not the base",
         "number" : 8,
         "expected": "8"
       },
@@ -41,7 +41,7 @@
          "expected": "Plang"
       },
       {
-         "description" : "the sound for 14 is Plong",
+         "description" : "the sound for 14 is Plong as it has a factor of 7",
          "number" : 14,
          "expected": "Plong"
       },
--- a/exercises/raindrops/canonical-data.json
+++ b/exercises/raindrops/canonical-data.json
@@ -1,74 +1,92 @@
 {
    "cases": [
       {
+         "description" : "the sound for 1 is 1",
          "number" : 1,
          "expected": "1"
       },
       {
+         "description" : "the sound for 3 is Pling",
          "number" : 3,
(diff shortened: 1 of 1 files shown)


exercises/raindrops/canonical-data.json
```diff
@@ -1,74 +1,92 @@
...
          "number" : 9,
          "expected": "Pling"
       },
       {
+         "description" : "the sound for 10 is Plang as it has a factor 5",
          "number" : 10,
          "expected": "Plang"
       },
       {
+         "description" : "the sound for 14 is Plong",
```
dvberkel: I rather like the cases for 9, and 10, where there is an explanation for the sound in the description. It is missing here, is this by design?

petertseng:
I apologise for breaking the upcoming generator - this was my fault in #314. It seems in light of #336 it is advantageous to keep things consistent. The good part: since the new descriptions are not just the string representation of the number, then it actually makes sense to have te descriptions.

exercises/raindrops/canonical-data.json
```diff
@@ -1,74 +1,92 @@
...
          "number" : 7,
          "expected": "Plong"
       },
       {
+         "description" : "the sound for 6 is Pling as it has a factor 3",
          "number" : 6,
          "expected": "Pling"
       },
       {
+        "description" : "the sound for 8 is 8",
```
petertseng: Does it make sense to include the reasoning for #370 in here? In particular, that 2^**3** is **not** "Pling" since the 3 is the power, rather than the base?

stevejb71: If it's a common mistake in implemention is this exercise, then why not. 

Something like "2 to the power 3 does not make a raindrop sound as 3 is the power not the base"?

petertseng: That proposed description seems good.

Caveat on the strength of my opinion: Whether it qualifies for common isn't known first-hand to me as I haven't done this exercise in any language. A particular way of doing it in Ruby (https://github.com/exercism/xruby/issues/434) first caused this case to be added.

But I figure... now that it's added, might as well give it the description to suit it, and I think this description does suit.

stevejb71: I updated the description for 8.

dvberkel:
Changes look good to me

PR #452, opened 2016-11-26
Provide inputs to minesweeper canonical data.

stevejb71:
Although the inputs can be backed out from the expected values by simply deleting the numbers, this has the disadvantage:

1. Needs special case code in test generators.
2. Might encourage humans writing the test to programatically back out the inputs in the tests as well. For Exercism, I think it's better to be explicit about the test inputs - being explicit will be easier to understand.


stkent:
Nice change. I recently implemented this in the Java track and plumped for explicit input rather than "backed-out" input for the tests, for exactly the reason you suggest - better to be explicit!

petertseng:
I just drop a quick note since I made https://github.com/exercism/x-common/pull/330

I did it the way I did at the time, but I don't feel so strongly about it at all (I'm about a 2/10), so I'm happy to defer to anyone who feels strongly about it.

Some tracks do currently generate the input from the output. I imagine that becomes an application of "don't repeat yourself".

However, adding the inputs to the JSON file adds support for tracks that want to be explicit without removing existing support, so adding them cannot really be a loss. The only possible risk is any danger of the input not matching the output, but this can be verified automatically, so the risk can be mitigated.

The other alternative is to say if a track wants to be explicit, it's that track's generator's job to create the input from removing numbers from the output in the JSON. Of course not all tracks have generators yet, so this option doesn't yet make sense.

petertseng:
EBWODP:

Appendix of tracks being explicit or implicit. I believe this information is not relevant to any decision making in this PR (decision can be made independent of this information), but I was interested in it anyway.

- https://github.com/exercism/xelixir/blob/master/exercises/minesweeper/minesweeper_test.exs - implicit
- https://github.com/exercism/xfsharp/blob/master/exercises/minesweeper/MinesweeperTest.fs - explicit
- https://github.com/exercism/xgo/blob/master/exercises/minesweeper/minesweeper_test.go - implicit
- https://github.com/exercism/xhaskell/blob/master/exercises/minesweeper/test/Tests.hs - implicit
- https://github.com/exercism/xlua/blob/master/exercises/minesweeper/minesweeper_spec.lua - explicit
- https://github.com/exercism/xocaml/blob/master/exercises/minesweeper/test.ml - implicit
- https://github.com/exercism/xphp/blob/master/exercises/minesweeper/minesweeper_test.php - explicit
- https://github.com/exercism/xperl5/blob/master/exercises/minesweeper/minesweeper.t - explicit
- https://github.com/exercism/xpython/blob/master/exercises/minesweeper/minesweeper_test.py - explicit
- https://github.com/exercism/xruby/blob/master/exercises/minesweeper/minesweeper_test.rb - explicit
- https://github.com/exercism/xrust/blob/master/exercises/minesweeper/tests/minesweeper.rs - implicit
- https://github.com/exercism/xscala/blob/master/exercises/minesweeper/src/test/scala/MinesweeperTest.scala - explicit
- https://github.com/exercism/xswift/blob/master/exercises/minesweeper/MinesweeperTest.swift - explicit

Insti:
Thanks @stevejb71 :heart:

PR #456, opened 2016-11-29
minesweeper: remove number from input, add mines

petertseng:
an input containing only "8" implies that the student solution is
supposed to read a board with numbers and fill in the mines, which is
contrary to all other tests (read a board with mines, fill in the
numbers)

### DIFF
--- a/exercises/minesweeper/canonical-data.json
+++ b/exercises/minesweeper/canonical-data.json
@@ -56,9 +56,9 @@
     {
       "description": "space surrounded by mines",
       "input": [
-        "   ",
-        " 8 ",
-        "   "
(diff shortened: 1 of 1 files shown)


petertseng:
Found with:

```ruby
require 'json'

json = JSON.parse(IO.read('canonical-data.json'))

json['cases'].each { |c|
  input = c['input']
  output = c['expected']
  clean = output.map { |l| l.gsub(/\d/, ' ') }

  raise "Nope on #{c}, got #{clean}" if clean != input
}
```

stevejb71:
That was a good catch. Is the verify script something that could be added to the CI build?

Issue #457, opened 2016-11-29
Fail the travis build if an exercise is missing critical files


kytrinyx:
Everything on Exercism assumes that an exercise has at the very least a `description.md` file and a `metadata.yml` file.

If one of these files is missing in an exercise, then the build should break.

See discussion here for context: https://github.com/exercism/trackler/pull/6/commits/036a2a7f579aa29e4815cf32dd285105a810de0b#r90053608

rpottsoh:
Should I be concerned about these files at this point?  Is this something I should be looking in to?


Sent from my android device.

kytrinyx:
@rpottsoh No, you only need to worry about this if you make up a completely new exercise.

rpottsoh:
@kytrinyx PR #14 (Harry Potter) is not a conversion from another Exercism track.  Tell me what I need to know, or at least point me in the right direction. J

tejasbubane:
I am working on writing a shell script similar to https://github.com/exercism/x-common/blob/master/bin/jsonlint.

kytrinyx:
@rpottsoh Ah, excellent. In that case this section of the Contributing Guide should have what you need: https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#implementing-a-completely-new-exercise

petertseng:
Please correct me if I'm wrong but I think the merging of #458 resolves this.

PR #458, opened 2016-11-29
CI check to verify required exercises files exist.

Insti:
Adds a bash script and travis configuration update to run it to check that all exercises have both a `description.md` and a `metadata.yml` file.

Note: This only checks for existence and does not validate the contents in any way.

### DIFF
--- a/bin/check_required_files_present
+++ b/bin/check_required_files_present
@@ -1,12 +1,20 @@
 #!/usr/bin/env bash
 MISSING_FILES=0
+SHOW_PROGRESS='false'
 
 if [ ! -d exercises ] ; then
 	echo "\"exercises\" directory not found, make sure you're running from the repository root" >&2
 	exit 1
 fi
 
+function show_progress {
+	if [ $SHOW_PROGRESS == 'true' ]; then
+		echo -n '.'
+	fi
+}
+
 function require_file {
+	show_progress
 	filename=$1
 	if [ ! -f $filename ]; then
 		echo "required file missing: $filename" >&2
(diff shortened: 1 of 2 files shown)


Insti:
Example Travis build failure: https://travis-ci.org/exercism/x-common/builds/179865364
Example Travis build success: https://travis-ci.org/exercism/x-common/builds/179864754

Squashing will eliminate the reverted broken exercise commit.

Fixes #457

bin/check_required_files_present
```diff
@@ -0,0 +1,22 @@
...
+		let "MISSING_FILES+=1"
+	fi
+}
+
+for exercise_directory in exercises/* ; do
+	require_file "$exercise_directory/description.md"
+	require_file "$exercise_directory/metadata.yml"
+done
+
+exit $MISSING_FILES
```
Insti: That is a good point.
I should not try to be so clever with return values.

kytrinyx:
The links to passing and failing builds is so great—note to self: do stuff like that more!

I really like how clean this is with the `require_file` function, too.

ErikSchierboom:
This has been open for quite a while. Is this ready to be merged?

Insti:
Yes, this is ready to be merged.

There was some confusion between this PR and #459 which did essentially the same thing and were committed at almost the same time.

Either is probably better than nothing but I like this one the best. (I'm definitely biased though.)

I think we either need @tejasbubane to close his request, or have an independent reviewer   (@ErikSchierboom?) decide which to merge.

ErikSchierboom:
I think #458 has the edge here, as it offers the same functionality as #459 and more.

Issue #462, opened 2016-12-05
leap: The test cases/descriptions  should get improved


guntbert:
Related PR #453 
The discussion on that PR led to the conclusion that the test cases
- are not well chosen
- should get a better description as to *why* the test case is useful/necessary

I promised to work on them, will start now.

Insti:
Thanks @guntbert :heart:

PR #463, opened 2016-12-05
leap: Rewrite the test cases and their descriptions

guntbert:
- only test cases remain that are necessary
- they have a clear description to help providing good names for tests

related: #462

### DIFF
--- a/exercises/leap/canonical-data.json
+++ b/exercises/leap/canonical-data.json
@@ -1,37 +1,22 @@
 {
    "cases": [
       {
-         "description": "leap year in twentieth century",
-         "input": 1996,
+         "description": "year not divisible by 4: common year",
+         "input": 2015,
+         "expected": false
+      },
+      {
+         "description": "year divisible by 4, not divisible by 100: leap year",
+         "input": 2016,
          "expected": true
       },
       {
-         "description": "odd standard year in twentieth century",
-         "input": 1997,
+         "description": "year divisible by 100, not divisible by 400: common year",
+         "input": 2100,
          "expected": false
       },
       {
-         "description": "even standard year in twentieth century",
-         "input": 1998,
-         "expected": false
-      },
-      {
-         "description": "standard year in nineteenth century",
-         "input": 1900,
-         "expected": false
-      },
-      {
-         "description": "standard year in eighteenth century",
-         "input": 1800,
-         "expected": false
-      },
-      {
-         "description": "leap year twenty four hundred",
-         "input": 2400,
-         "expected": true
-      },
-      {
-         "description": "leap year two thousand",
+         "description": "year divisible by 400: leap year",
          "input": 2000,
          "expected": true
       }


exercises/leap/canonical-data.json
```diff
@@ -1,39 +1,29 @@
...
-         "input": 2400,
+         "description": "year divisible by 400: leap year",
+         "input": 2000,
          "expected": true
       },
       {
-         "description": "leap year two thousand",
-         "input": 2000,
-         "expected": true
+         "description": "year before validity of Gregorian Calendar (no leap years at all): standard year",
```
stevejb71: I don't think Gregorian calendar checks should be introduced in an exercise as early as this. 

Also, 1581 is not a leap year by the check on divisibility by 4.

exercises/leap/canonical-data.json
```diff
@@ -1,39 +1,29 @@
...
-      },
-      {
-         "description": "odd standard year in twentieth century",
-         "input": 1997,
-         "expected": false
-      },
-      {
-         "description": "even standard year in twentieth century",
-         "input": 1998,
+         "description": "year not divisible by 4: standard year",
```
stevejb71: Where does the term standard year come from? I see common year on Wikipedia. I would think just "not leap year"  would be best though.

kytrinyx:
Would you mind amending the commit so that the slug of the exercise is included? It makes it a lot easier to dig through the logs later.

exercises/leap/canonical-data.json
guntbert: "standard year" was not from me, "not a leap year" is fine by me.

exercises/leap/canonical-data.json
guntbert: Do I understand correctly that you would omit the boundary check completely, although there were no leap years before 1582 - thus only relying on the three rules?

guntbert:
@kytrinyx sorry I have no idea what the "slug of the exercise" might be.

exercises/leap/canonical-data.json
```diff
@@ -1,37 +1,22 @@
...
-      },
-      {
-         "description": "odd standard year in twentieth century",
-         "input": 1997,
-         "expected": false
-      },
-      {
-         "description": "even standard year in twentieth century",
-         "input": 1998,
+         "description": "year not divisible by 4: not a leap year",
```
Insti: Agreeing with Wikipedia: "A year that is not a leap year is called a common year." is a good idea.
"not a leap year" is a little too self referential.

exercises/leap/canonical-data.json
```diff
@@ -1,37 +1,22 @@
...
-         "input": 1900,
-         "expected": false
+         "description": "year divisible by 4, not divisible by 100: leap year",
+         "input": 2016,
+         "expected": true
       },
       {
-         "description": "standard year in eighteenth century",
-         "input": 1800,
+         "description": "year divisible by 100, not divisible by 400: not a leap year",
```
Insti: The "not divisible by 400" is not relevant here.
What is important: "this year is divisible by 4 and also by 100."

Insti:
@guntbert the 'slug' is the name of the exercise, in this case `leap` I have updated the title for you.

exercises/leap/canonical-data.json
guntbert: I disagree: every year divisible by 100 is divisible by 4 too, so to me the important part is that in this case rule 3 does not apply.

Insti: But we do not need to implement rule 3 at all for this test to pass.

kotp:
Updating the title won't update the commit though.  You can amend the commit and do a force push on the branch, it will be OK.

When you make the changes, if you also squash the commits, that might be a good time to make that change.

exercises/leap/canonical-data.json
Insti: Perhaps switching the order of the 3rd and 4th tests would be better.

guntbert:
@kotp thanks for the heads up - does this repo allow squash commits from the webUI or should I squash on the command line?

kotp:
Yes, to both questions.  Either way you are comfortable with.

exercises/leap/canonical-data.json
```diff
@@ -1,39 +1,29 @@
...
-         "input": 1998,
+         "description": "year not divisible by 4: standard year",
+         "input": 2015,
          "expected": false
       },
       {
-         "description": "standard year in nineteenth century",
-         "input": 1900,
-         "expected": false
+         "description": "year divisible by 4, not divisible by 100: leap year",
```
guntbert: Well, only years that are not divisible by 100 are leap years, aren't they? So IMO "year divisible by 4" alone is not enough, we should also state that the exception does not apply.

Insti: The only rule that applies here is 'year divisible by 4' you do not need any other rules to determine if 2016 is a leap year.

The 100 rule is only relevant when we have a year that is divisible by 100, which 2016 is not.

Why do you not also reference the 400 year rule here?

What if there were more rules? ("only in evenly numbered decades, unless the decade is 1") Would you include them all in the description?

exercises/leap/canonical-data.json
stevejb71: I would think that is better for this exercise. It is the first one on some tracks, better to keep it simple.

exercises/leap/canonical-data.json
guntbert: I still disagree - we have a decision tree, where the **next** rule ("unless") leads to an exemption of the current rule.

So 
-  rule 1 fails => common year: only one rule to check, so *only one rule in the description*
- rule 1 succeeds => possibly a leap year, we need to check rule 2
  - rule 2 fails: no exemption, so the result of rule 1 holds => leap year, no need to check further roles, so *rule 1 and rule 2 in the description* 
  - rule 2 succeeds => exemption, possibly a common year, we need to check rule 3
    - rule 3 fails: no exemption, the result of rule 2 holds => common year, I mention the *two really relevant rules* (divisibility by 4 is obvious from mathematics if the number is divisible by 100)
    - rule 3 succeeds: exemption =>leap year *only the last rule is really relevant* (divisibility by 4 and by 100 is (again) obvious from mathematics)

:+1 I really appreciate how you force me to rethink the issue.

Insti: Thanks, that decision tree was very helpful for me to be able to see the way you're approaching the problem, and I can now see how the test descriptions you've used make sense in that case.

Insti: I also appreciate the effort you've put into this and hope my differing opinions have not discouraged you.

guntbert: No worries - its good to make well thought decisions - especially when aiming at relatively inexperienced people - and in the end many future tests will draw from the templates in x-common.
How do I proceed from now on? I intend to squash all my commits, thereby amending the commit description. Should I wait on @stevejb71 ?

Insti: Do the squash and commit message changes and then @stevejb71 can review the "final" version.

kytrinyx:
> sorry I have no idea what the "slug of the exercise" might be.

Sorry about that @guntbert, it's the url-friendly identifier that each exercise has. It's the same name that we would use in the `config.json` for it, and the name of the directory containing the metadata files. In this case, `leap`.

When looking through a few thousand commits, the subject line of the commit message _Rewrite the test cases and their descriptions_ doesn't say which exercise it is referring to, which makes it harder to find things later.

exercises/leap/canonical-data.json
kotp: @guntbert from this perspect, you will want to do a squash locally, and then force push your branch, which will set this up like you want it to be.

guntbert:
I guess I am finished with **this one** now - thank you all for being clear, patient and attentive for details.

A big praise for the quick responses I got from the team - those made me feel welcome and supported.

guntbert:
@kytrinyx thank you for the explanation - in hindsight it's obvious :-)

exercises/leap/canonical-data.json
stevejb71: This now looks fine to me.

kotp:
Thanks @guntbert and @stevejb71 and @Insti

PR #465, opened 2016-12-06
New Exercise book-store

rpottsoh:
Attempting to create slug for new xpascal exercise.


petertseng:
Oh, this exercise looks interesting. As the README makes clear, a few choices for the optimisation. I am interested to see how it will go.

It would be nice if we can have the test cases in a file in this repo too, as https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#improving-consistency-by-extracting-shared-test-data has it. In this case, the task should only require using the same tests as those in https://github.com/exercism/xpascal/pull/14

rpottsoh:
@petertseng thanks for the tip.  The json file seems pretty straight forward.

exercises/harry-potter/description.md
```diff
@@ -0,0 +1,57 @@
+This exercise is based on the exercise of the same name found at cyber-dojo.org
```
Insti: This information is provided in the source section of the metadata and does not need to be repeated here.

Insti:
I'd like to see this exercise called something more generic like 'book-discounts'. The books being Harry Potter books is irrelevant to the problem.

exercises/harry-potter/description.md
```diff
@@ -0,0 +1,57 @@
...
+One copy of any of the five books costs 8 EUR. 
+
+If, however, you buy two different books, you get a 5% 
+discount on those two books.
+
+If you buy 3 different books, you get a 10% discount. 
+
+If you buy 4 different books, you get a 20% discount.
+
+If you go the whole hog, and buy all 5, you get a huge 25%
```
Insti: There are 7 books in the Harry Potter series.

exercises/book-store/description.md
```diff
@@ -0,0 +1,49 @@
...
+If you buy 3 different books, you get a 10% discount. 
+
+If you buy 4 different books, you get a 20% discount.
+
+If you go the whole hog, and buy all 5, you get a huge 25% discount. 
+
+Note that if you buy, say, four books, of which 3 are different titles, you get a 10% discount on the 3 that 
+form part of a set, but the fourth book still costs 8 EUR. 
+
+Your mission is to write a piece of code to calculate the price of any conceivable shopping basket (containing only books from the same series), giving as big a discount as possible.
```
rpottsoh: I see your point, however I don't really see a better way to word this.  At this point its is just copy/pasted from cyber-dojo.  I suppose, as it is, it adds to the challenge of the problem.  I know when I originally worked this problem I ended up disregarding this piece, but I think it is needed initially to help the reader understand the requirements.

rpottsoh:
@Insti thanks, I'm trying.  It's the first "original" exercise that I have added to exercism.  Had I only known....  Kidding, wouldn't be doing this if I didn't want to learn too.

rpottsoh:
I think I have completed what needs to be done.  Let me know where I've errored and I'll take care of it.

Insti:
It thinks your json is invalid.

exercises/book-store/canonical-data.json
```diff
@@ -0,0 +1,76 @@
...
+         "expected": 60.00
+      },
+      {
+         "description": "basket 11223344551",
+         "basket": "11223344551",
+         "expected": 68.00
+      },
+      {
+         "description": "basket 112233445512",
+         "basket": "112233445512",
```
Insti: It would be nice if the basket was an array rather than a string.

exercises/book-store/canonical-data.json
```diff
@@ -0,0 +1,76 @@
...
+   "book-store": {
+      "description": "Calculate lowest price for shopping basket only containing books from a single series",
+      "cases": [
+         {
+            "description": "TestSingleBookPrice",
+            "expected": 8.0
+         },
+         {
+            "description": "TestBasketContainsMyBooks",
+		    "mybasket": "3"
```
Insti: missing comma here

Insti:
Read this comment I wrote on another PR, about test descriptions: https://github.com/exercism/x-common/pull/451#issuecomment-263070019

Insti:
Having a good `canonical-data.json` file can be harder than writing the original specification.
It might be good to split this off into a separate PR so we can get the initial exercise merged in and once a few tracks have implemented it we can work out what the canonical test data should be.

exercises/book-store/canonical-data.json
```diff
@@ -0,0 +1,76 @@
+{
+   "#": [
+     "Not sure what to type here, come back to this later."
```
petertseng: if there is nothing in particular to add (any notes for anyone who wishes to port this exercise to a track), the entire thing can be deleted.

exercises/book-store/canonical-data.json
```diff
@@ -0,0 +1,76 @@
...
+   ],
+   "cases": [
+      {
+         "description": "TestSingleBookPrice",
```
petertseng: what is the difference between this and "TestTotalBasketPriceSingleBook"?

If the only function under test in this exercise is "Given this basket, what is its price?" then it seems there is no need to have this one.

exercises/book-store/canonical-data.json
```diff
@@ -0,0 +1,76 @@
...
+         "expected": 8.0
+      },
+      {
+         "description": "TestBasketContainsMyBooks",
```
petertseng: What does this case do?

> If the only function under test in this exercise is "Given this basket, what is its price?" then it seems there is no need to have this one.

This statement also seems true for this case.

rpottsoh:
@petertseng I've gone ahead and moved the canonical-data.json to another PR.  I am not really clear on how the inner workings of the machine (exercism.io) operate so don't have a solid grasp on what the exact role the three files play (canonical-data.json,  description.md, and metadata.yml).  At this point I am trying to format the files as best I can to examples from other exercises.  In the case of the json file I am not sure what tests should be listed and which shouldn't.  I've assumed that they should all be listed.  I uess further discussion related to the json file should take place here #470

kytrinyx:
I think all the files for the exercise should go into the same PR, (and in a single commit) since they all belong to one, atomic change.

rpottsoh:
That is the way I had it but was encouraged to pull the canonical data out into a separate PR.


Sent from my android device.

Insti:
@kytrinyx I asked @rpottsoh to move the canonical data to another PR.

Reasoning: 
There are many pre-existing exercises that do not have canonoical-data files.
The `description.md` and `metadata.yml` files for this exercise were almost complete and ready to be merged. 

By getting the exercise in place it gives multiple tracks an opportunity to implement the problem and help find/explore the required test cases, and find/resolve any issues with the description.

exercises/book-store/canonical-data.json
rpottsoh: In retrospect this probably isn't needed.  I have a property that just returns the price of a single book.  The test is just confirming that it equals 8.

exercises/book-store/canonical-data.json
rpottsoh: May be too simple of a test.  It is simply echoing back the contents of its basket.  It should equal what was placed in the basket.

exercises/book-store/canonical-data.json
rpottsoh: String type comes with the added benefit of TStringHelper which makes parsing the basket a little easier.  For the sake of my example solution I could possibly convert the input to an array, I would just convert it to a string with in the class.

petertseng:
I think things are OK. A few minor things but I am not sure if they are blocking.

exercises/book-store/description.md
```diff
@@ -0,0 +1,49 @@
...
+
+For example, how much does this basket of books cost?
+
+2 copies of the first book
```
petertseng: Note that at https://github.com/rpottsoh/x-common/blob/rpottsoh-patch-1/exercises/book-store/description.md the lines get mashed together. Bullet points?

Or do we expect people to always be reading the raw md file in a text editor rather than rendering? (I always read raw, but I don't know about others)

exercises/book-store/description.md
```diff
@@ -0,0 +1,49 @@
...
+2 copies of the second book
+2 copies of the third book
+1 copy of the fourth book
+1 copy of the fifth book
+
+One way of group these 8 books is:
+ 1 group of 5 --> 25% discount (1st,2nd,3rd,4th,5th)
```
petertseng: These also get mashed together. Not sure of the best way to deal with it - use backticks?

exercises/book-store/description.md
kytrinyx: > At this point its is just copy/pasted from cyber-dojo.

What is the licensing on that text? We may need to rewrite the text (or at least heavily edit it), even though we're referring to cyber-dojo as the source.

kytrinyx:
> I asked @rpottsoh to move the canonical data to another PR.

@Insti fair enough. I saw the other PR first, and it didn't have a [WIP] to indicate that it would have to be merged after a different PR, so I was unaware of the fact that there was a different PR in progress until I got further down my notifications list.

exercises/book-store/description.md
rpottsoh: @kytrinyx I've been scouring the cyber-dojo site and I haven't found anything at this point that makes the statement one way or the other that the exercise is protected.  In general the only licensing that I see is regarding the comercial use of the site as a whole.  Some of the other exercises there site other sites as sources for complete instructions.  Some of the exercises have NO instructions and only reference the source site.  

Should we ask for permission to use the text?

kytrinyx: > I've been scouring the cyber-dojo site and I haven't found anything at this point that makes the statement one way or the other that the exercise is protected.

Anything that doesn't have a public license is copyrighted by default, and we can't use it.

The cyber-dojo website is released under a couple of variants of the BSD license. Does it cite another source for the book sale exercise?

rpottsoh:
Good question.  I will look again with this in mind as well as check other sources that might have the same/similar exercise.

Sent from my android device.

exercises/book-store/description.md
rpottsoh: @kytrinyx   I am still looking.  Here is a link to a license, looks like boiler plate.  https://github.com/cyber-dojo/start-points-exercises/blob/master/LICENSE.md

I am still looking around.  I don't know if this license really grants permission for use of exercises or just relating to the site as a whole.

rpottsoh:
I see.  So is the exercise dead?  


Sent from my android device.

exercises/book-store/description.md
```diff
@@ -0,0 +1,67 @@
...
+- +1 group of 3 --> 10% discount (1st,2nd,3rd)
+
+This would give a total of:
+
+- 5 books at a 25% discount
+- +3 books at a 10% discount
+
+Resulting in:
+
+- 5 x (8 2.00) == 5 x 6.00 == $30.00
```
petertseng: missing minus signs, the `(8 2.00)` doesn't look right

exercises/book-store/description.md
```diff
@@ -0,0 +1,67 @@
...
+- +3 x (8 0.80) == 3 x 7.20 == $21.60
```
petertseng: also missing minus sign, should be `(8 - 0.80)` right?

exercises/book-store/description.md
```diff
@@ -0,0 +1,67 @@
...
+
+- 1 group of 4 books --> 20% discount  (1st,2nd,3rd,4th)
+- +1 group of 4 books --> 20% discount  (1st,2nd,3rd,5th)
+
+This would give a total of:
+
+- 4 books at a 20% discount
+- +4 books at a 20% discount
+
+Resulting:
```
petertseng: should this be "Resulting in" instead of just "Resulting"?

petertseng:
I do not think there are other changes that I would suggest.

I am not competent to discuss copyright, so I make no comment on that.

Before I forget to mention it:

Before merging this PR, **please squash commits**.

kytrinyx:
In terms of copyright, as long as we don't exactly copy their text word for word, and link to them as a source, I think it is fine.

rpottsoh:
Text was changed a little.  Looks like C# and F# have already implemented the exercise. 👍

robkeim:
@rpottsoh thanks to @ErikSchierboom we now have two people that have submitted implementations on the site for the this exercise in both C# and F# 👍

ErikSchierboom:
And that's thanks to @robkeim who implemented the problem!

rpottsoh:
@ErikSchierboom I noticed!  Thanks to you and @robkeim for taking the time to get it out there.

PR #466, opened 2016-12-06
Change all-your-base canonical data to use an expected key.

stevejb71:
This conforms with the majority of the canonical data files I have seen.


Insti:
Seems reasonable. This will require tracks to update their generators, but the change required should be straightforward and obvious.

stevejb71:
I made this change to avoid a special case in the ocaml generator that I'm writing. Consistency is likely to help simplify any other generators too.

PR #467, opened 2016-12-07
wordy: trim-up description

jtigger:
- replace "step" with the slightly more accurate "iteration" in headers.
- remove language-specific instructions (referring to "delete `skip`")
- remove redundant examples; the canonical data should be the source of
  variation.
- we're trying to avoid referring to "the program" or "write a
  program"


kytrinyx:
Thanks @jtigger!

PR #468, opened 2016-12-07
Fix incorrect expected values for all-your-base.

stevejb71:
3 of the error cases in all-your-base do not look correct to me.

### DIFF
--- a/exercises/all-your-base/canonical-data.json
+++ b/exercises/all-your-base/canonical-data.json
@@ -77,19 +77,19 @@
         , "input_base"   : 10
         , "input_digits" : [0]
         , "output_base"  : 2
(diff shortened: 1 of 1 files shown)


Insti:
From the top of the file:
```
          "It's up to each track do decide:"

            , "3. Are leading zeroes allowed?"
            , ""
            , "4. How should invalid input be handled?"
            , ""
            , "All the undefined cases are marked as null."
```
implies `null` is the correct value.

stevejb71:
Whoops, I missed that. Will close.

Insti:
There was a big discussion about all this when the exercise was created: https://github.com/exercism/x-common/pull/280

PR #470, opened 2016-12-08
exercise: book-store, canonical data

rpottsoh:
initial canonical-data.json for book-store.


kytrinyx:
This looks like a fun exercise. We're going to need a `metadata.yml` file that has a short blurb of the exercise, and the source of the exercise.

Here's an example: https://github.com/exercism/x-common/blob/master/exercises/nucleotide-count/metadata.yml

Insti:
> This looks like a fun exercise. We're going to need a metadata.yml file that has a short blurb of the exercise, and the source of the exercise.

Already provided here: https://github.com/exercism/x-common/pull/465

exercises/book-store/canonical-data.json
```diff
@@ -0,0 +1,76 @@
+{
+   "book-store": {
+      "description": "Calculate lowest price for shopping basket only containing books from a single series",
+      "cases": [
+         {
+            "description": "TestSingleBookPrice",
```
petertseng: previous discussion https://github.com/exercism/x-common/pull/465#discussion_r91590493

> In retrospect this probably isn't needed. I have a property that just returns the price of a single book. The test is just confirming that it equals 8.

I would agree. Not every implementation necessarliy needs to have this property in order to be correct. We would [prefer not to drive the implementation](https://github.com/exercism/discussions/issues/41).

exercises/book-store/canonical-data.json
```diff
@@ -0,0 +1,76 @@
...
+            "expected": 8.0
+         },
+         {
+            "description": "TestBasketContainsMyBooks",
```
petertseng: previous discussion: https://github.com/exercism/x-common/pull/465#discussion_r91590666

I was confused about this one since the key is "mybasket" instead of "basket". Is this a significant difference?

In addition, I don't think it is particularly necessary for this problem that a basket can report what it contains - the test code is already telling the basket what it contains. It seems the tests should be interested in finding out the price of a basket (as all the other tests are).

exercises/book-store/canonical-data.json
```diff
@@ -0,0 +1,76 @@
...
+            "expected": 60.00
+         },
+         {
+            "description": "basket 11223344551",
+            "basket": "11223344551",
+            "expected": 68.00
+         },
+         {
+            "description": "basket 112233445512",
+            "basket": "112233445512",
```
petertseng: previous discussion: https://github.com/exercism/x-common/pull/465#discussion_r91584069 - would be good if this were an array

> String type comes with the added benefit of TStringHelper which makes parsing the basket a little easier.

Sorry, I'm not familiar with what a TStringHelper is nor what language it's associated with. We'd like the canonical JSON files to be independent of any one language.

exercises/book-store/canonical-data.json
rpottsoh: @petertseng  I agree.  I am going to clean this up a bit.  This initial version of the json file is essentialy just a dump of the tests.  I was not sure at the time what exactly should appear in the json file or what its purpose is.  As I have stated before this is my first time trying to add an exercise, so learning the ropes.

exercises/book-store/canonical-data.json
rpottsoh: @petertseng understand

exercises/book-store/canonical-data.json
```diff
@@ -3,74 +3,87 @@
...
             "basket": "12345",
             "expected": 30.00
          },
          {
-            "description": "basket 11223345",
+            "description": [ 
+		             "Test a basket containing two copies of book 1, ",
+		             "two copies of book 2, two copies of book 3, ",
+		             "and one copy each of books 4 and 5."
+		           ],
```
Insti: This isn't describing why this test is useful. It's just a verbose English language description of the input.

exercises/book-store/canonical-data.json
```diff
@@ -3,74 +3,87 @@
...
          {
-            "description": "TestTotalBasketPriceEmptyBasket",
+            "description": "Test total basket price if basket is empty.",
             "basket": "",
             "expected": 0.0
          },
          {
-            "description": "TestBasketWith2DifferentBooks",
+            "description": "Test basket with two 2 different books.",
             "basket": "12",
```
Insti: Basket should be an array rather than a string.

exercises/book-store/canonical-data.json
```diff
@@ -0,0 +1,84 @@
...
+         },
+         {
+	    "description": "Should arrive at best discount and correct total for a basket containing five different books.",
+            "basket": [1, 2, 3, 4, 5],
+            "expected": 30.00
+         },
+         {
+	    "description": [
+		             "Should arrive at best discount and correct total for a basket containing 8 books.",
+		             "The basket will consist of two each of the first 3 books and one each of books 4 and 5."
```
rpottsoh: @petertseng when/how is the canonical data presented to the student?  I think it would help me if I knew the context of when each of the three (description.md, canonical-data.json, and metadata.yml) files is used.

petertseng: > when/how is the canonical data presented to the student?

Never.

It is to be used by anyone who wishes to implement this exercise for a track for which it is not implemented yet.

The old way of doing it was that everyone would have to look at every other track's test files and figure out what the best set of tests is. With this new way, everyone only has to look in one place: here.

rpottsoh:
@peterseng thanks this is helpful, thanks.

rpottsoh:
How is the decision made to go ahead and merge the PR?

exercises/book-store/canonical-data.json
```diff
@@ -0,0 +1,86 @@
...
+   "book-store": {
+      "description": "Calculate lowest price for shopping basket only containing books from a single series",
+      "cases": [
+         {
+            "description": "A basket containing only a single book is charged full price.",
+            "basket": [1],
+            "expected": 8.0
+         },
+         {
+	    "description": "A basket containing only two of the same book is charged full price.",
```
petertseng: sorry, mind making this align? there's one space missing, compared to "basket" below.

exercises/book-store/canonical-data.json
```diff
@@ -0,0 +1,86 @@
...
+         },
+         {
+	    "description": "A basket of five different books is cheaper than grouping the books any other way.",
+            "basket": [1, 2, 3, 4, 5],
+            "expected": 30.00
+         },
+         {
+	    "description": [
+		             "A basket containing three pairs of three different books plus ",
+		             "one copy each of the other two books, not paired, is the least expensive ",
```
petertseng: I don't think the "other two books, not paired" in this description is accurate. The other two books *do* get grouped - the 4 gets grouped with one [1, 2, 3] set, and the 5 gets grouped with the other [1, 2, 3] group. The wording seems to imply that the groups in this basket would be [1, 2, 3], [1, 2, 3], [4], [5], when in fact it is [1, 2, 3, 4], [1, 2, 3, 5]. I would like to see it made clear that [1, 2, 3, 4], [1, 2, 3, 5] is the best grouping, rather than [1, 2, 3, 4, 5], [1, 2, 3] (or any other).

exercises/book-store/canonical-data.json
```diff
@@ -0,0 +1,86 @@
...
+		             "grouping."
+		           ],
+            "basket": [1, 1, 2, 2, 3, 3, 4, 5],
+            "expected": 51.20
+         },
+         {
+	    "description": [
+		             "A basket containing four pairs of four different books plus ",
+		             "one copy of a book, not already paired, is the least expensive grouping."
```
petertseng: this "not already paired" also seems to imply that the grouping for this basket would be [1, 2, 3, 4], [1, 2, 3, 4], [5] when in fact it is [1, 2, 3, 4, 5], [1, 2, 3, 4].

ErikSchierboom:
Great that you're working on a set of canonical test data @rpottsoh!

exercises/book-store/canonical-data.json
rpottsoh: @petertseng thanks for the feedback.  Spacing issue should be resolved.  I was totally missing that the descriptions should contain the solution that yields the expected result, duh.  I have re-written the descriptions.  I left them wordy.  If you think being too wordy is harder to follow I could, for example, rewrite to be like given basket [1,1,2,2,3,3,4,4,5] group as [1,2,3,4,5] and [1,2,3,4]....  Let me know what you think.

exercises/book-store/canonical-data.json
```diff
@@ -0,0 +1,93 @@
...
+            "expected": 25.60
+         },
+         {
+	        "description": "A basket of five different books is the least expensive as a single group.",
+            "basket": [1, 2, 3, 4, 5],
+            "expected": 30.00
+         },
+         {
+	        "description": [
+		                    "For a basket containing eight books consisting of a pair each of the first three books ",
```
petertseng: This could be shortened to "Two groups of four is cheaper than a group of five plus a group of three". This tells us why we have this test - to catch a common mistake students might make (always make the largest group possible)

Can probably make similar changes for all others.

exercises/harry-potter/description.md
```diff
@@ -0,0 +1 @@
+
```
rpottsoh: @petertseng help me out here.  This file doesn't appear [here](https://github.com/rpottsoh/x-common/tree/rpottsoh-patch-2/exercises/book-store).  I clearly see the empty file in the PR, but do not know how to remove it other than to close the PR and open a new one.

petertseng: You will not find it in https://github.com/rpottsoh/x-common/tree/rpottsoh-patch-2/exercises/book-store .

You will find it in https://github.com/rpottsoh/x-common/tree/rpottsoh-patch-2/exercises/harry-potter .

rpottsoh: Got it.  Thanks

ErikSchierboom:
@petertseng Did you know you can do the squashing from GitHub? Just use the "Squash and merge" option when merging the PR.

rpottsoh:
@petertseng I am working to cut down on the verbage of each description.  I am using the .json file for the Change exercise.  I am introducing a new field, "targetgrouping" that has array data for the least expensive grouping to achieve the "expected" outcome.

petertseng:
> Did you know you can do the squashing from GitHub?

I've got that distinct feeling of deja vu. https://github.com/exercism/x-common/pull/338#issuecomment-240019020

exercises/book-store/canonical-data.json
```diff
@@ -0,0 +1,101 @@
...
+   "book-store": {
+      "description": [
+                       "Calculate lowest price for shopping basket only ",
+                       "containing books from a single series"
+                     ],
+      "cases": [
+         {
+            "description": "A basket containing only a single book.",
+            "basket": [1],
+            "targetgrouping": [1],
```
petertseng: compare the type of `targetgrouping` in the below case: an array of arrays of integers: `[[2], [2]]`, versus this one here, an array of integers: `[1]`.

Should it be `[[1]]` for consistency?

Same question for all other cases where the type is array of integer rather than array of arrays of integers.

exercises/book-store/canonical-data.json
```diff
@@ -0,0 +1,101 @@
...
+         },
+         {
+            "description": "A basket of five different books.",
+            "basket": [1, 2, 3, 4, 5],
+            "targetgrouping": [1, 2, 3, 4, 5],
+            "expected": 30.00
+         },
+         {
+            "description": [
+                            "A basket containing eight books consisting of a ",
```
petertseng: Any way to make it clear either in the description or elsewhere that we test this case to make sure that the student finds `[[1, 2, 3, 4], [1, 2, 3, 5]]` rather than `[[1, 2, 3, 4, 5], [1, 2, 3]]`?

It's interesting to me because I think this is the only case in this JSON file where the greedy solution won't work, so it's particularly important to me that the description point that out.

exercises/book-store/canonical-data.json
rpottsoh: Yes it should be [[1]] for consistency.  Change has been made.

exercises/book-store/canonical-data.json
rpottsoh: I am making two changes to address that having more than any one book in a group is of no advantage to the customer and the other change to draw the readers attention to this special case, that is counter-intuitive to how one might group the books, as you have pointed out.

exercises/book-store/canonical-data.json
```diff
@@ -0,0 +1,106 @@
...
+         {
+            "description":  "A basket containing only two of the same book.",
+            "basket": [2, 2],
+            "targetgrouping": [[2], [2]],
+            "expected": 16.0
+         },
+         {
+            "description": "No charge to carry around an empty basket.",
+            "basket": [],
+            "targetgrouping": [[]],
```
petertseng: I actually think `[]` is OK here (and better than `[[]]`), since `[]` still is of the type "array of arrays of integers", it just happens to be an *empty* array of arrays of integers (it could be an empty array of anything,, really). Why do I think `[]` is better than `[[]]`? Well, we don't say `[[2], [2], []]` right? So we never have empty inner arrays. Thus, `[]` rather than `[[]]`.

rpottsoh: I can do that.  It kind of looked weird to me too.

PR #472, opened 2016-12-17
Reword tree building problem and add example tree

ErikSchierboom:
The current description uses a rather hard to read description. I've tried to reword it to be more clear. Furthermore, I've added a sample tree to show how it can look like.

### DIFF
--- a/exercises/tree-building/description.md
+++ b/exercises/tree-building/description.md
@@ -15,9 +15,9 @@ An example tree:
 <pre>
 root (ID: 0, parent ID: 0)
 |-- child1 (ID: 1, parent ID: 0)
-|    |-- grandchild1 (ID: 4, parent ID: 1)
-|    +-- grandchild2 (ID: 8, parent ID: 1)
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
Whoops! Fixed that.

robkeim:
Looks great, thanks @ErikSchierboom!

robkeim:
@ErikSchierboom - I guess I should double check is there any special process around merging PRs in x-common? I've only ever merged in the F# track.

ErikSchierboom:
@robkeim Not that I know of. Usually someone like @Insti also checks the change, which he already did. So I think it's cool.

robkeim:
Ok sounds good, thanks for the quick confirmation @ErikSchierboom!

ErikSchierboom:
It indeed *does* invert the meaning, but I figured the old one was incorrect. The new one also makes more sense to me. A parent ID should always be lower than a record ID.

PR #474, opened 2016-12-21
luhn: Update Luhn tests and description

IanWhitney:
I'm addressing a few different things here:

1. The tests exposed functionality that is not needed

The `addends` and `check_digit` methods aren't needed by clients of this
code, and I didn't see that students got any benefit from explicitly
implementing them. The presence of these tests pushed students towards a
particular implementation, which we [try to
avoid](https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#improving-consistency-by-extracting-shared-test-data)

> The test cases should not require people to follow one very specific
way to solve the problem, e.g. the tests should avoid checking for
helper functions that might not be present in other implementations.

To fix this I've removed tests of `addends` and `check_digit`.

2. The tests were duplicative.

There's not a way (that I could see, at least) of implementing Luhn in
tiny steps. It's either working or its not. Once you have a couple valid
tests and a couple invalid tests your are pretty much covered. (note: I
guess you could test the cases where Luhn can be wrong, but that seemed
outside the scope of this exercise.)

Again, from the guide

> All tests should be essential to the problem. Ensure people can get
the first test passing easily. Then each following test should ideally
add a little functionality. People should be able to solve the problem a
little at a time. Avoid test cases that don't introduce any new aspect
and would already pass anyway.

To fix this I've removed most duplicative tests that would pass as soon
as the student got a working Luhn implementation.

3. The examples and the description didn't match

Luhn can be hard to grasp. At least it can be for me. What helps me is
to have explicit, real-world examples. To help with this I've updated
the description.md file to have examples of credit cards and Canadian
SINs, two places where Luhn is used in the real world.

To re-enforce the descriptions, I've used the same examples in the
tests. So students can see step-by-step breakdowns of how to get their
tests passing.

Note:

I'm still on the fence about the `generate_check_digit` method. It seems
to be to be outside the scope of this exercise, but I also think
implementing it is interesting and useful. I could go either way on
keeping it or removing it.

### DIFF
--- a/exercises/luhn/description.md
+++ b/exercises/luhn/description.md
@@ -3,6 +3,8 @@ a simple checksum formula used to validate a variety of identification
 numbers, such as credit card numbers and Canadian Social Insurance
 Numbers.
 
+The task is to write a function that checks if a given string is valid.
+
 Validating a Number
 ------
 
--- a/exercises/luhn/description.md
+++ b/exercises/luhn/description.md
@@ -44,7 +44,7 @@ An example of an invalid Canadian SIN where we've changed the final digit
 Double the second digits, starting from the right
 
 ```
-086 858 276
+086 858 277
 ```
 
 Sum the digits
--- a/exercises/luhn/description.md
+++ b/exercises/luhn/description.md
@@ -1,53 +1,112 @@
-The Luhn formula is a simple checksum formula used to validate a variety
-of identification numbers, such as credit card numbers and Canadian
-Social Insurance Numbers.
+The [Luhn algorithm](https://en.wikipedia.org/wiki/Luhn_algorithm) is
+a simple checksum formula used to validate a variety of identification
+numbers, such as credit card numbers and Canadian Social Insurance
+Numbers.
 
-The formula verifies a number against its included check digit, which is
-usually appended to a partial number to generate the full number. This
-number must pass the following test:
+Validating a Number
+------
 
-- Counting from rightmost digit (which is the check digit) and moving
-  left, double the value of every second digit.
-- For any digits that thus become 10 or more, subtract 9 from the
-  result.
-  - 1111 becomes 2121.
-  - 8763 becomes 7733 (from 2×6=12 → 12-9=3 and 2×8=16 → 16-9=7).
-- Add all these digits together.
-  - 1111 becomes 2121 sums as 2+1+2+1 to give a check digit of 6.
-  - 8763 becomes 7733, and 7+7+3+3 is 20.
+As an example, here is a valid (but fictitious) Canadian Social Insurance
+Number.
 
-If the total ends in 0 (put another way, if the total modulus 10 is
-congruent to 0), then the number is valid according to the Luhn formula;
-else it is not valid. So, 1111 is not valid (as shown above, it comes
-out to 6), while 8763 is valid (as shown above, it comes out to 20).
+```
+046 454 286
+```
 
-Write a program that, given a number
+The first step of the Luhn algorithm is to double every second digit,
+starting from the right. We will be doubling
 
-- Can check if it is valid per the Luhn formula. This should treat, for
-  example, "2323 2005 7766 3554" as valid.
-- Can return the checksum, or the remainder from using the Luhn method.
-- Can add a check digit to make the number valid per the Luhn formula and
-  return the original number plus that digit. This should give "2323 2005 7766
-  3554" in response to "2323 2005 7766 355".
+```
+_4_ 4_4 _8_
+```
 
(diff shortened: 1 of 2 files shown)


exercises/luhn/description.md
```diff
@@ -1,53 +1,112 @@
...
+An example of an invalid Canadian SIN where we've changed the final digit
+
+```
+046 454 287
+```
+
+Double the second digits, starting from the right
+
+```
+086 858 276
```
petertseng: the 7 in the final position has changed back to a 6 here.

exercises/luhn/description.md
```diff
@@ -1,53 +1,112 @@
...
+| account |   |check digit|
+```
+
+```
+8273 1232 7352 056   9
+|    account     |   |check digit|
+```
+
+Given an account number your code should be able to return a check digit
+that can be appended to the account to generate a valid Luhn
```
petertseng: > a valid Luhn

You will have to excuse a question from someone unfamiliar with the domain, but is it valid to use here "Luhn" as a shorthand for "number that is valid according to Luhn algorithm"?

IanWhitney: Yeah, that's a bad sentence. I was hoping people wouldn't notice.

Insti:
I like what you're doing here, and appreciate the detailed description in the OP. :heart:

I'm slightly against including the `generate_check_digit` tests, since it does force an implementation detail. This could be an opportunity to encourage the student to write their own tests for that function.

Some suggestions:
- Move the check digit tests before the validation tests.
- Rename `generate_check_digit` to `check_digit`
- Include simple tests for the minimal case of 2+1 digit numbers.
- What about tests for degenerate cases? 1 digit numbers?
- Include test for `0`?

IanWhitney:
> I'm slightly against including the `generate_check_digit` tests, since it does force an implementation detail. This could be an opportunity to encourage the student to write their own tests for that function.

I'm not sure how this would look. Include the tests but mark them as optional? Or just say in the readme, "Maybe you want to write this function to generate a check digit."?

>  Include simple tests for the minimal case of 2+1 digit numbers.

Normally I would start off with tests of small inputs, and the old test suite had some tests that did this. But with this exercise the size of the input doesn't matter much. Small or long, the algorithm is exactly the same. So tests of short strings weren't any easier for students to implement. 

> What about tests for degenerate cases?

I'm not sure what a degenerate case would be here. There are strings that can pass validation even if certain numbers are transposed, but I didn't see much value in testing that here. I guess we could have a test of strings that contain non-digits.

> 1 digit numbers?
> Include test for `0`?

1 digit inputs should always be false. Makes sense to test that.

IanWhitney:
I think I've addressed the concerns so far.

exercises/luhn/description.md
```diff
@@ -1,53 +1,78 @@
...
-It doesn't actually tell you if it's a real credit card number, only that it's
-a plausible one. It's the same thing with the bytes that get transmitted --
-you could have the right number of bytes and still have a garbled message. So
-checksums are a simple sanity-check, not a real in-depth verification of the
-authenticity of some data. It's often a cheap first pass, and can be used to
-quickly discard obviously invalid things.
+The [Luhn algorithm](https://en.wikipedia.org/wiki/Luhn_algorithm) is
+a simple checksum formula used to validate a variety of identification
+numbers, such as credit card numbers and Canadian Social Insurance
+Numbers.
```
Insti: It would be nice to add a sentence or two describing what we're supposed to be doing.

"The task is to write a function to check if a given number is valid."

IanWhitney: Done

Insti:
It concerns me that the entire test suite can be passed by:
```pseudocode
return input == "046 454 286"
```

IanWhitney:
We could add more positive cases...and the students could pass them with a switch statement. 

To me this seems like a problem solved by the nitpickers.

IanWhitney:
Any more updates here? If not, I'll merge on Sunday, January 8.

jtigger:
(way late to the party...)

❤️  this!  Thanks for tackling this, @IanWhitney.  It was a less-than-delightful experience to go through this exercise as it was.  🙇 

also...

@Insti said:
> It concerns me that the entire test suite can be passed by...

@IanWhitney replied
> We could add more positive cases...and the students could pass them with a switch statement.
>
> To me this seems like a problem solved by the nitpickers.

Oh yeah, we've got tons of exercises where there are sufficiently few test cases that they can be "defeated" by pure conditional logic (aka "evil coder").

I'm assuming we're working with practitioners who understand you get what you put in.

PR #475, opened 2016-12-22
Raindrops update: 10 is also a factor of 30

mondok:
`10` is a factor of `30` and was left out of the factors list in the problem description.


petertseng:
Yeah, seems easy enough to just merge. Thanks!

PR #476, opened 2016-12-22
sum-of-multiples: update description

bordeltabernacle:
description updated to mention limits,
https://github.com/exercism/xgo/issues/340#issuecomment-238150900

### DIFF
--- a/exercises/sum-of-multiples/description.md
+++ b/exercises/sum-of-multiples/description.md
@@ -3,5 +3,5 @@ multiples of either 3 or 5, we get 3, 5, 6 and 9, 10, 12, 15, and 18.
 
 The sum of these multiples is 78.
 
-Write a program that can find the sum of the multiples of a given set of
-numbers.
+Write a program that, given a number, can find the sum of the multiples
+of a given set of numbers, up to but not including that number.


exercises/sum-of-multiples/description.md
```diff
@@ -3,5 +3,5 @@ multiples of either 3 or 5, we get 3, 5, 6 and 9, 10, 12, 15, and 18.
 
 The sum of these multiples is 78.
 
-Write a program that can find the sum of the multiples of a given set of
-numbers.
+Write a program that, given a number, can find the sum of the multiples
```
petertseng: This same line is in https://github.com/exercism/x-common/blob/master/exercises/sum-of-multiples/metadata.yml#L2 - and since the blurb gets prepended to the README (I discuss at https://github.com/exercism/x-common/issues/469#issuecomment-266439658), that means this line will sandwich the README.

**That is still better** than the current situation where the README has two contradictory lines. There is also talk of [NOT prepending the blurb](https://github.com/exercism/x-common/issues/469#issuecomment-266441156).

One alternative could be to remove this line completely, leaving the prepend to do the explaining. Or we could leave it. Someone may eventually request to remove the redundancy, as happened in #386

PR #477, opened 2016-12-22
Phone-Number: Add Canonical data

tejasbubane:
Phone Number exercise tests have some variations across tracks. eg. for invalid numbers [ruby](https://github.com/exercism/xruby/blob/master/exercises/phone-number/phone_number_test.rb) expects a string of all(10) zeroes while [rust](https://github.com/exercism/xrust/blob/master/exercises/phone-number/tests/phone-number.rs) expects `None`.

It also differs from the Readme with implementing getting `country-code` from number and `pretty-printing` the number. I feel these should not be part of the exercise. Would like to know what others think.

Closes https://github.com/exercism/todo/issues/124

### DIFF
--- a/exercises/phone-number/canonical-data.json
+++ b/exercises/phone-number/canonical-data.json
@@ -1,0 +1,57 @@
+{
+  "number": {
+    "description": "Cleanup user-entered phone numbers",
+    "cases": [
+      {
+        "description": "cleans the number",
+        "phrase": "(123) 456-7890",
+        "expected": "1234567890"
+      },
+      {
+        "description": "cleans numbers with dots",
+        "phrase": "123.456.7890",
+        "expected": "1234567890"
+      },
+      {
+        "description": "cleans numbers with multiple spaces",
+        "phrase": "123 456   7890   ",
+        "expected": "1234567890"
+      },
+      {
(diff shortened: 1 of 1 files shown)


rpottsoh:
@tejasbubane  Interesting, there are more tests here than what I found for [C#](https://github.com/exercism/xcsharp/blob/master/exercises/phone-number/PhoneNumberTest.cs).  I am working on the XPascal track and will see about adding the tests you listed that I am missing.

Insti:
I'd prefer json `null` for invalid phone numbers.

petertseng:
Pretty printing and extracting subparts of numbers: Yeah, doesn't seem that important to me, and the README doesn't mention it, so I would not add them to the JSON file. It would make sense if people agreed to add it and it got added to the README and JSON file together. But that can come later.

I do think JSON `null` is a slightly better sentinel. Because how will you tell the difference between a valid `000 000 0000` versus something invalid? I encourage all tracks to make this distinction too! But I suppose this is something for the individual tracks to consider, not here.

tejasbubane:
Thanks @Insti and @petertseng. Changed the invalid output to be `null` instead of 10 zeroes.

Insti:
Thnks @tejasbubane :heart:

PR #479, opened 2016-12-23
phone-number: 2 more tests

rpottsoh:
Added two tests that are included in the csharp track.

### DIFF
--- a/exercises/phone-number/canonical-data.json
+++ b/exercises/phone-number/canonical-data.json
@@ -51,6 +51,16 @@
         "description": "invalid with right number of digits but letters mixed in",
         "phrase": "1a2b3c4d5e6f7g8h9i0j",
         "expected": null
+      },
+      {
+        "description": "has an area code",
+        "phrase": "1234567890",
+        "expected": "123"
+      },
+      {
+        "description": "formats a number",
+        "phrase": "1234567890",
+        "expected": "(123) 456-7890"
       }
     ]
   }


petertseng:
We argued against including these tests in #477:

> implementing getting `country-code` from number and `pretty-printing` the number. I feel these should not be part of the exercise.

> Pretty printing and extracting subparts of numbers: Yeah, doesn't seem that important to me, and the README doesn't mention it, so I would not add them to the JSON file

So, shall we reverse this decision?

If we do, we need to add it to the README as well.

Also, I would like for these tests not to get mixed in with the parsing tests, because then the only way to tell what function is under test is the description, which is too much work for automated test generators. We wouldn't want to make them build in natural language parsing to figure out what function to test based on the description. We would rather put them in a separate section. An example is https://github.com/exercism/x-common/blob/master/exercises/difference-of-squares/canonical-data.json, where there are separate section for `sum_of_squares` and `square_of_sum`.

I haven't changed my mind yet, so I still suggest not having these tests.

rpottsoh:
I'm fine with whatever decision the group makes.  I was not aware of the previous discussion.

There were tests in the json that were not in csharp and vice versa.

If these two extra tests are nixed I can easily remove them from xpascal.

*csharp was the reference I used for xpascal.



Sent from my android device.

Insti:
I vote for excluding anything other than validation from the test suite.

Insti:
sorry, misclicked close

Issue #447, opened 2016-11-12
[perfect-numbers] Description is not clear


snahor:
The description isn't very clear, I mean it's not clear what to implement. I've been looking at the implementations and there's no consensus. Some of them implement the classification and the others just check whether or not the number is perfect.

Insti:
[Perfect numbers description](https://github.com/exercism/x-common/blob/master/exercises/perfect-numbers/description.md)

Insti:
I would suggest doing something like what we have recently done for triangles. [Triangles test data](https://github.com/exercism/x-common/blob/master/exercises/triangle/canonical-data.json)

Where there are 3 different functions that return a boolean value depending on if the number is  Perfect, Deficient, or Abundant

guntbert:
Looking at _Triangles_ I see that `metadata.md` contains a clear statement what to do 

> Write a program that can tell you if a triangle is equilateral, isosceles, or scalene.

This statement is missing here.

guntbert:
@Insti this might be a good "first patch"

petertseng:
Naturally, one might ask: Is #481 sufficient to close this issue?

It certainly has helped greatly in this regard.

I think a little help in the README would be good - so perhaps https://github.com/exercism/x-common/blob/master/exercises/perfect-numbers/description.md could also say that we need to be able to determine whether a given number is perfect, abundant, deficient. 

The sentence on the Aliquot sum also seems a bit out of place.

petertseng:
> perhaps https://github.com/exercism/x-common/blob/master/exercises/perfect-numbers/description.md could also say that we need to be able to determine whether a given number is perfect, abundant, deficient.

#518 takes care of that.

PR #481, opened 2016-12-28
metadata: replace blurb with instruction

rootulp:
### The old blurb was a repeat of the first line from the description so I replaced it with a clear statement on what to do.
- references #447
- As @snahor pointed out, some language implementations only ask for `is_perfect` (e.g [Python](https://github.com/exercism/xpython/blob/master/exercises/perfect-numbers/perfect_numbers_test.py)) whereas some languages ask for the whole clarification (e.g [Ruby](https://github.com/exercism/xruby/blob/master/exercises/perfect-numbers/perfect_numbers_test.rb))
- I agree with @Insti and think we should standardize on asking for the whole clasification.
- Next steps: create issues in the language repos that only ask for `is_perfect`


IanWhitney:
These get easier to track if the name & commit message include the exercise name. I've updated the PR; if you could update the commit message, that'd be awesome.

Thanks for the PR!

Insti:
It's standard to prefix with the exercise slug.

> perfect-numbers: replace blurb with instruction

exercises/perfect-numbers/metadata.yml
```diff
@@ -1,4 +1,4 @@
 ---
-blurb: "The Greek mathematician Nicomachus devised a classification scheme for natural numbers."
+blurb: "Determine if a number is perfect, abundant, or deficient."
```
Insti: Some more context would be nice here:
> Determine if a number is perfect, abundant, or deficient based on Nicomachus' (60 - 120 CE) classification scheme for natural numbers,

PR #483, opened 2017-01-04
word-count: replace underscore with space in description

bordeltabernacle:


Insti:
Thanks @bordeltabernacle

PR #484, opened 2017-01-05
topics: change if/else to conditionals

bordeltabernacle:
The switch statement is a distinct concept from if-else control-flow.

### DIFF
--- a/TOPICS.txt
+++ b/TOPICS.txt
@@ -2,7 +2,7 @@ Basic concepts
 --------------
 Abstract classes
 Callbacks
-Control-flow (if-else statements)
+Control-flow (conditionals)
 Control-flow (loops)
 Events
 Exception handling


Insti:
Switch statements seem to me to be an implementation detail rather than a topic\*.

I'd suggest  `Control-flow (if-else statements)` become `Control-flow (conditionals)` rather than adding topics for every statement a language could support.

\* I may be misunderstanding how topics are being used.

TOPICS.txt
```diff
@@ -3,6 +3,7 @@ Basic concepts
 Abstract classes
 Callbacks
 Control-flow (if-else statements)
+Control-flow (switch statements)
```
rpottsoh: Perhaps too nit-pickey, but how about 
`Control-flow (switch/case statements)`
 instead?

rpottsoh:
If @Insti is correctly considering how TOPICS is used then I have to agree and you can disregard my previous comment.

bordeltabernacle:
Yes, I'd have to agree as well, `conditionals` is a much more succinct way of describing it. I did wonder if it was a bit fine-grained, but updating the `xgo` config.json for an exercise that many people had used a `switch` statement for, I was worried `if/else` was inaccurate, almost misleading.
If you're happy with `conditionals` then I'll update the PR, otherwise I totally understand if you'd rather leave as is.

bordeltabernacle:
@Insti @rpottsoh I've updated the PR

bordeltabernacle:
Thanks both for the input & merge!

rpottsoh:
I don't mean to beat a dead horse here (since the PR is merged), but shouldn't we replace the two `control-flow` statements with a single statement `conditionals`?  It seems unecessary to provide an example of control-flow when the other topics are not providing examples.  Or just say `control-flow` (once) without the paranthetical example that follows.

PR #486, opened 2017-01-06
variable-length-quantity: Add canonical data

petertseng:
Implementing tracks:

* https://github.com/exercism/xcsharp/blob/master/exercises/variable-length-quantity/VariableLengthQuantityTest.cs
* https://github.com/exercism/xfsharp/blob/master/exercises/variable-length-quantity/VariableLengthQuantityTest.fs
* https://github.com/exercism/xgo/blob/master/exercises/variable-length-quantity/variable_length_quantity_test.go
* https://github.com/exercism/xlua/blob/master/exercises/variable-length-quantity/variable-length-quantity_spec.lua
* https://github.com/exercism/xphp/blob/master/exercises/variable-length-quantity/variable-length-quantity_test.php
* https://github.com/exercism/xrust/blob/master/exercises/variable-length-quantity/tests/variable-length-quantity.rs

All have pretty much the same tests.

Notable exceptions:

* Only PHP has the encode->decode test
* Only C#, F#, PHP, Rust had the decode->encode test
* Only C#, F#, PHP, Rust had the overflow test. I noted that tracks may
  choose to exclude it.

### DIFF
--- a/exercises/variable-length-quantity/canonical-data.json
+++ b/exercises/variable-length-quantity/canonical-data.json
@@ -138,12 +138,6 @@
         "expected": null
       },
       {
-        "description": "overflowing a 32-bit integer causes error",
-        "#": "Your track may exclude this test if it is reasonable to give implementations the freedom to encode/decode larger numbers.",
-        "input": [255, 255, 255, 255, 127],
-        "expected": null
-      },
-      {
         "description": "multiple values",
         "input": [192, 0, 200, 232, 86, 255, 255, 255, 127, 0, 255, 127, 129, 128, 0],
         "expected": [8192, 1193046, 268435455, 0, 16383, 16384]
--- a/exercises/variable-length-quantity/canonical-data.json
+++ b/exercises/variable-length-quantity/canonical-data.json
@@ -149,23 +149,5 @@
         "expected": [8192, 1193046, 268435455, 0, 16383, 16384]
       }
     ]
-  },
-  "decode_encode": {
-    "cases": [
-      {
-        "description": "Round-trip through decode then encode should give original bytes.",
-        "input": [192, 0, 200, 232, 86, 255, 255, 255, 127, 0, 255, 127, 129, 128, 0],
-        "expected": [192, 0, 200, 232, 86, 255, 255, 255, 127, 0, 255, 127, 129, 128, 0]
-      }
-    ]
-  },
-  "encode_decode": {
-    "cases": [
-      {
-        "description": "Round-trip through encode then decode should give original integers.",
-        "input": [8192, 1193046, 268435455, 0, 16383, 16384],
-        "expected": [8192, 1193046, 268435455, 0, 16383, 16384]
-      }
-    ]
   }
 }
--- a/exercises/variable-length-quantity/canonical-data.json
+++ b/exercises/variable-length-quantity/canonical-data.json
@@ -1,0 +1,185 @@
+{
+  "#": [
+    "JSON doesn't allow hexadecimal literals.",
+    "All numbers are given as decimal literals instead.",
+    "It is highly recommended that your track's test generator display all numbers as hexadecimal literals."
+  ],
+  "encode": {
+    "single": {
+      "description": ["Encode a single integer, producing a series of bytes."],
+      "cases": [
+        {
+          "description": "zero",
+          "input": 0,
+          "expected": [0]
+        },
+        {
+          "description": "arbitrary single byte",
+          "input": 64,
+          "expected": [64]
+        },
+        {
+          "description": "largest single byte",
+          "input": 127,
+          "expected": [127]
+        },
+        {
+          "description": "smallest double byte",
+          "input": 128,
+          "expected": [129, 0]
+        },
+        {
+          "description": "arbitrary double byte",
+          "input": 8192,
+          "expected": [192, 0]
+        },
+        {
+          "description": "largest double byte",
+          "input": 16383,
+          "expected": [255, 127]
+        },
+        {
+          "description": "smallest triple byte",
+          "input": 16384,
+          "expected": [129, 128, 0]
+        },
+        {
+          "description": "arbitrary triple byte",
+          "input": 1048576,
+          "expected": [192, 128, 0]
+        },
+        {
+          "description": "largest triple byte",
+          "input": 2097151,
+          "expected": [255, 255, 127]
+        },
+        {
+          "description": "smallest quadruple byte",
+          "input": 2097152,
+          "expected": [129, 128, 128, 0]
+        },
+        {
+          "description": "arbitrary quadruple byte",
+          "input": 134217728,
+          "expected": [192, 128, 128, 0]
+        },
+        {
+          "description": "largest quadruple byte",
+          "input": 268435455,
+          "expected": [255, 255, 255, 127]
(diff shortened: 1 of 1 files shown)


exercises/variable-length-quantity/canonical-data.json
```diff
@@ -0,0 +1,186 @@
...
+          "input": [192, 0, 200, 232, 86, 255, 255, 255, 127, 0, 255, 127, 129, 128, 0],
+          "expected": [8192, 1193046, 268435455, 0, 16383, 16384]
+        }
+      ]
+    },
+    "decode_encode": {
+      "cases": [
+        {
+          "description": "Round-trip through decode then encode should give original bytes.",
+          "input": [192, 0, 200, 232, 86, 255, 255, 255, 127, 0, 255, 127, 129, 128, 0],
```
petertseng: since `input == expected`, I would be very happy to delete this line entirely, but then #452.

exercises/variable-length-quantity/canonical-data.json
```diff
@@ -0,0 +1,186 @@
...
+          "expected": [192, 0, 200, 232, 86, 255, 255, 255, 127, 0, 255, 127, 129, 128, 0]
+        }
+
+      ]
+    },
+    "encode_decode": {
+      "cases": [
+        {
+          "description": "Round-trip through encode then decode should give original integers.",
+          "input": [8192, 1193046, 268435455, 0, 16383, 16384],
```
petertseng: > since `input == expected`, I would be very happy to delete this line entirely, but then #452.

same.

exercises/variable-length-quantity/canonical-data.json
```diff
@@ -0,0 +1,185 @@
...
+          "expected": [8192, 1193046, 268435455, 0, 16383, 16384]
+        }
+      ]
+    }
+  },
+  "decode_encode": {
+    "cases": [
+      {
+        "description": "Round-trip through decode then encode should give original bytes.",
+        "input": [192, 0, 200, 232, 86, 255, 255, 255, 127, 0, 255, 127, 129, 128, 0],
```
petertseng: since `input == expected`, I would be very happy to delete this line entirely, but then #452.

exercises/variable-length-quantity/canonical-data.json
```diff
@@ -0,0 +1,185 @@
...
+        "expected": [192, 0, 200, 232, 86, 255, 255, 255, 127, 0, 255, 127, 129, 128, 0]
+      }
+    ]
+  },
+  "encode_decode": {
+    "cases": [
+      {
+        "description": "Round-trip through encode then decode should give original integers.",
+        "input": [8192, 1193046, 268435455, 0, 16383, 16384],
```
petertseng: > since `input == expected`, I would be very happy to delete this line entirely, but then #452.

same

petertseng:
Q of hypothetical reviewer: How to be sure these values are correct, now that you converted them to decimal? Maybe you did it wrong.

A: [Build Status](https://travis-ci.org/petertseng/x-common)

exercises/variable-length-quantity/canonical-data.json
```diff
@@ -0,0 +1,185 @@
+{
+  "#": [
+    "JSON doesn't allow hexadecimal literals.",
```
petertseng: a shame. I suppose I could have them as strings (so, instead of `[255, 127]` the test data contains `["0xff", "0x7f"]`, but that seems a little wasteful.

exercises/variable-length-quantity/canonical-data.json
```diff
@@ -0,0 +1,185 @@
...
+          "expected": [143, 248, 128, 128, 0]
+        },
+        {
+          "description": "maximum 32-bit integer input",
+          "input": 4294967295,
+          "expected": [143, 255, 255, 255, 127]
+        }
+      ]
+    },
+    "multiple": {
```
Insti: Do "single" and "multiple" need to be different cases?
Surely single is just a multiple with size == 1?

exercises/variable-length-quantity/canonical-data.json
Insti: I'd remove the decode_encode (and encode_decode) tests completely.
encode should test encode.
decode should test decode.
If both of these work correctly then running them sequentially is redundant.

exercises/variable-length-quantity/canonical-data.json
petertseng: I reconsidered, based on a concern that "single" and "multiple" being diferent cases could mislead language implementors into believing that their test code should test `decode_single`, `decode_multiple`, etc., whereas the intention is actually exactly as you say:

> Surely single is just a multiple with size == 1?

There should only be one `decode` function tested. Let's have just the single section, to make it clear.

exercises/variable-length-quantity/canonical-data.json
petertseng: Worse, upon review, this exact series of bytes and corresponding integers is *already tested* in both the `encode` and `decode` sections. So I seriously doubt there can be an implementation that would fail this test while passing the others - this case can add no value.

petertseng:
When the time comes to merge, I recommend squashing. Although I have written my commit messages in such a way that not-squashing would not be disastrous, there is no need to publish the intermediate drafts of this JSON file.

exercises/variable-length-quantity/canonical-data.json
```diff
@@ -5,179 +5,147 @@
...
+        "expected": null
+      },
+      {
+        "description": "incomplete sequence causes error, even if value is zero",
+        "input": [128],
+        "expected": null
+      },
+      {
+        "description": "overflowing a 32-bit integer causes error",
+        "#": "Your track may exclude this test if it is reasonable to give implementations the freedom to encode/decode larger numbers.",
```
Insti: The description only calls for being able to handle up to 32 bit integers and (rightly) doesn't specify what to do with inputs that are greater than that, leaving the behavior undefined seems better.
So leave out this test completely.

exercises/variable-length-quantity/canonical-data.json
```diff
@@ -0,0 +1,147 @@
+{
+  "#": [
+    "JSON doesn't allow hexadecimal literals.",
+    "All numbers are given as decimal literals instead.",
+    "It is highly recommended that your track's test generator display all numbers as hexadecimal literals."
```
petertseng: I'm aware not all tracks have them, but maybe this can be considered as a gentle nudge to get tracks to consider the practice... so this description was intentionally worded provocatively in this way.

I hope this doesn't make this file too inconvenient. I would be sad to store the string representations of the hexadecimal literals.

Insti: I think this is fine.
Strings are definitely wrong.

petertseng:
Given that

* it has been over 24*7 hours since the last comment
* an earlier comment has expressed approval of the test cases, with some caveats
* the caveats are addressed
* the verify script says the data is correct

I am merging it now, since everyone else has had a week to suggest changes.

Does it seem like I'm in too much of a hurry? I suppose I'm not in that much of a hurry, other than that I like to keep my in-flight PRs list small, and that the verify branch is currently in a very frankenbranch state and I would like to get it to be relatively normal again.

PR #487, opened 2017-01-06
variable-length-quantity: Limit to 32-bit int, not 28-bit int

### LINKED ISSUE #485: variable-length-quantity: why 0x0fffffff max?
readme:

> The largest integer allowed is 0FFFFFFF, which translates to 4 bytes variable length.

This echoes http://web.archive.org/web/20051129113105/http://www.borg.com/~jglatt/tech/midifile/vari.htm and http://www.music.mcgill.ca/~ich/classes/mumt306/midiformat.pdf#page=2

Does it have to be this way? the 0x0fffffff max was meant for midi files, but https://en.wikipedia.org/wiki/Variable-length_quantity says 

> A variable-length quantity (VLQ) is a universal code that uses an **arbitrary** number of binary octets (eight-bit bytes) to represent an **arbitrarily** large integer.

The emphasis on "arbitrary" is mine. Given that in general these values can be arbitrary, does it makes sense to include the sentence on limitations in the README?

And the Rust track has a test for 0xffffffff. But that violates the README, so https://github.com/exercism/xrust/issues/224

I say it doesn't make sense to include the limitation in the README.

petertseng:
### variable-length-quantity: s/delta-time/integer/

delta-time seems to only be meaningful in the context of MIDIs. We are
dealing with just arbitrary integers, with no mention of domain.

### variable-length-quantity: Limit to 32-bit int, not 28-bit int

The limit of a 28-bit int (0x0fffffff) is specific to MIDIs, whereas
variable-length quantities can deal with arbitrary sizes.

So why limit to 32-bits? Because it is a common size for languages, and
some language tracks already have tests that assume that the limit is
0xffffffff, so we might as well match them.

Closes #485
Closes exercism/xrust#224

### DIFF
--- a/exercises/variable-length-quantity/description.md
+++ b/exercises/variable-length-quantity/description.md
@@ -8,7 +8,7 @@ To indicate which is the last byte of the series, you leave bit #7 clear.
 In all of the preceding bytes, you set bit #7. 
 
 So, if an integer is between `0-127`, it can be represented as one byte. 
-The largest integer allowed is `0FFFFFFF`, which translates to 4 bytes variable length. 
+Although VLQ can deal with numbers of arbitrary sizes, for this exercise we will restrict ourselves to only numbers that fit in a 32-bit unsigned integer.
 Here are examples of integers as 32-bit values, and the variable length quantities that they translate to:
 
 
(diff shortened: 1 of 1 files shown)


petertseng:
Actually, that's a pretty short review period; it behooves me to give others more time before me merging it. I was in too much of a hurry to close a months-long issue, but really I should not have, since it's not like anyone's breathing down my neck to have the issue closed. But I expect this change to not be so controversial, because this is just making the READMEs match the current reality that tracks have already subscribed to.

Insti:
I think it's fine, but I also agreed with you about the change.
We can always revert when the huge wave of objections comes in on Monday morning.

petertseng:
> when

D=

Issue #401, opened 2016-10-09
Representation of undefined/exceptional values in canonical-data.json


Insti:
The current instructions in the [readme](https://github.com/exercism/x-common#test-data-format-canonical-datajson) recommend using -1 to indicate an exception.

> expected: the expected result (this would be -1 when we expect an exception)

Is '-1' the best value to be using? I think that `null` would be a better value to use to indicate an exceptional or undefined result.

Discuss.

dvberkel:
I don't think `null` is a very language neutral value to indicate an undefined or exceptional result. Some languages do not have a `null` value. And for languages that do, I would personally go to great lengths to avoid returning `null`, as it promotes a bad practise.

Furthermore I think that undefined and exceptional values are representing two different things. Something can be undefined, without being an exception. I would opt for having different conventions for the two cases

Insti:
It is up to the language tracks to implement undefined and exceptional values in the most appropriate manner for their language.

Using a null value in the json does not mean that the language needs to use null.

dvberkel:
> Using a null value in the json does not mean that the language needs to use null.

That is certainly true. But I still have the same objections. Let's say that someone wants to port an exercise to a language with `null`. If that person looks at the canonical data and sees a `null`, they could assume that it is an actual requirement to return an explicit `null`. You would have to about the convention that `null` doesn't mean `null` but either throw an exception or signal provide an undefined answer, which ever is appropriate for the target language.

I would rather see that the JSON is more self contained. That is doesn't rely on conventions and mores, but that it makes sense on its own.

IanWhitney:
I think this is a "Tabs vs Spaces" thing. It depends. As long as the description in the JSON file describes what's expected, I think it's fine.

rbasso:
The C programmer that still lives inside of me thinks that `-1` is perfect.
The Haskell one is screaming that this is so **$&#*<^%**... well, let's just say that he favors `null` but misses `Maybe`.

The problem with using `-1` is that we are encoding an error as a standard value, and this makes it impossible to distinguish it from a valid negative value.

Of course, there is the risk of someone misinterpreting a `null`, as @dvberkel mentioned, but there is the same risk with `-1`.

At least `null` is a special value in JSON...

petertseng:
[...]

PR #488, opened 2017-01-08
circular-buffer: Add canonical data

petertseng:
The circular-buffer exercise was first added in 2014 May at #9, with
Ruby and Javascript as the initial implementing tracks:

* exercism/xjavascript#17
* exercism/xruby#17

Implementing tracks:

* https://github.com/exercism/xcsharp/blob/master/exercises/circular-buffer/CircularBufferTest.cs
* https://github.com/exercism/xdlang/blob/master/exercises/circular-buffer/circular_buffer.d
* https://github.com/exercism/xecmascript/blob/master/exercises/circular-buffer/circular-buffer.spec.js
* https://github.com/exercism/xerlang/blob/master/exercises/circular-buffer/circular_buffer_tests.erl
* https://github.com/exercism/xfsharp/blob/master/exercises/circular-buffer/CircularBufferTest.fs
* https://github.com/exercism/xgo/blob/master/exercises/circular-buffer/circular_buffer_test.go
* https://github.com/exercism/xjavascript/blob/master/exercises/circular-buffer/circular-buffer.spec.js
* https://github.com/exercism/xlfe/blob/master/exercises/circular-buffer/test/circular-buffer-tests.lfe
* https://github.com/exercism/xlua/blob/master/exercises/circular-buffer/circular-buffer_spec.lua
* https://github.com/exercism/xpascal/blob/master/exercises/circular-buffer/uCircularBufferTests.pas
* https://github.com/exercism/xpython/blob/master/exercises/circular-buffer/circular_buffer_test.py
* https://github.com/exercism/xruby/blob/master/exercises/circular-buffer/circular_buffer_test.rb
* https://github.com/exercism/xrust/blob/master/exercises/circular-buffer/tests/circular-buffer.rs

All tracks pretty much implement the same tests, except:

* Tracks typically have the "read item just written" and "each item can
  only be read once" tests combined. These tests split the two.
* In dynamically-typed languages, buffers may be able to store
  heterogeneous types. This concept is not expressed in these tests.
* In some statically-typed languages, buffers use generics, such that
  buffers may store any one type of the client's choosing. This concept
  is also not expressed in these tests.
* The final test (ensuring that overwrite drops the right items) was more
  complex: capacity 5, 3 writes, 2 reads, write, read, 4 writes, 2
  overwrites, 5 reads. It's been simplified to capacity 3, 3 writes, 1
  read, write, overwrite, 3 reads.

Closes exercism/todo#79

### DIFF
--- a/exercises/circular-buffer/canonical-data.json
+++ b/exercises/circular-buffer/canonical-data.json
@@ -7,12 +7,15 @@
     "",
     "Tests will contain a number of operations. The operation will be specified in the `operation` key.",
     "Based on the operation, other keys may be present.",
-    "read: Reading from the buffer should produce the item located at `expected`.",
-    "write: Writing the item located at `item` should succeed.",
-    "readShouldFail: Reading from the buffer should fail.",
-    "writeShouldFail: Writing the item located at `item` should fail.",
+    "read: Reading from the buffer should succeed if and only if `should_succeed` is true.",
+    "  If it should succeed, it should produce the item at `expected`. ",
+    "  If it should fail, `expected` will not be present. ",
+    "write: Writing the item located at `item` should succeed if and only if `should_succeed` is true.",
     "overwrite: Write the item located at `item` into the buffer, removing the oldest item if necessary.",
     "clear: Clear the buffer.",
+    "",
+    "Failure of either `read` or `write` may be indicated in a manner appropriate for your language:",
+    "Raising an exception, returning (int, error), returning Option<int>, etc.",
     "",
     "Finally, note that all values are integers.",
     "If your language contains generics, you may consider allowing buffers to contain other types.",
@@ -25,7 +28,8 @@
       "capacity": 1,
       "operations": [
         {
-          "operation": "readShouldFail"
+          "operation": "read",
+          "should_succeed": false
         }
       ]
     },
@@ -35,10 +39,12 @@
       "operations": [
         {
           "operation": "write",
-          "item": 1
+          "item": 1,
+          "should_succeed": true
         },
         {
           "operation": "read",
+          "should_succeed": true,
           "expected": 1
         }
       ]
@@ -49,14 +55,17 @@
       "operations": [
         {
           "operation": "write",
-          "item": 1
+          "item": 1,
+          "should_succeed": true
         },
         {
           "operation": "read",
+          "should_succeed": true,
           "expected": 1
         },
         {
-          "operation": "readShouldFail"
+          "operation": "read",
+          "should_succeed": false
         }
       ]
     },
@@ -66,18 +75,22 @@
       "operations": [
         {
           "operation": "write",
-          "item": 1
+          "item": 1,
+          "should_succeed": true
         },
         {
           "operation": "write",
-          "item": 2
+          "item": 2,
+          "should_succeed": true
         },
         {
           "operation": "read",
+          "should_succeed": true,
           "expected": 1
         },
         {
           "operation": "read",
+          "should_succeed": true,
           "expected": 2
         }
       ]
@@ -88,11 +101,13 @@
       "operations": [
         {
           "operation": "write",
-          "item": 1
+          "item": 1,
+          "should_succeed": true
         },
         {
-          "operation": "writeShouldFail",
-          "item": 2
+          "operation": "write",
+          "item": 2,
+          "should_succeed": false
         }
       ]
     },
@@ -102,18 +117,22 @@
       "operations": [
         {
           "operation": "write",
-          "item": 1
+          "item": 1,
+          "should_succeed": true
         },
         {
           "operation": "read",
+          "should_succeed": true,
           "expected": 1
         },
         {
           "operation": "write",
-          "item": 2
+          "item": 2,
+          "should_succeed": true
         },
         {
           "operation": "read",
+          "should_succeed": true,
           "expected": 2
         }
       ]
@@ -124,26 +143,32 @@
       "operations": [
         {
           "operation": "write",
-          "item": 1
+          "item": 1,
+          "should_succeed": true
         },
         {
           "operation": "write",
-          "item": 2
+          "item": 2,
+          "should_succeed": true
         },
         {
           "operation": "read",
+          "should_succeed": true,
           "expected": 1
         },
         {
           "operation": "write",
-          "item": 3
+          "item": 3,
+          "should_succeed": true
         },
         {
           "operation": "read",
+          "should_succeed": true,
           "expected": 2
         },
         {
           "operation": "read",
+          "should_succeed": true,
           "expected": 3
         }
       ]
@@ -154,13 +179,15 @@
       "operations": [
         {
           "operation": "write",
-          "item": 1
+          "item": 1,
+          "should_succeed": true
         },
         {
           "operation": "clear"
         },
         {
-          "operation": "readShouldFail"
+          "operation": "read",
+          "should_succeed": false
         }
       ]
     },
@@ -170,17 +197,20 @@
       "operations": [
         {
           "operation": "write",
-          "item": 1
+          "item": 1,
+          "should_succeed": true
         },
         {
           "operation": "clear"
         },
         {
           "operation": "write",
-          "item": 2
+          "item": 2,
+          "should_succeed": true
         },
         {
           "operation": "read",
+          "should_succeed": true,
           "expected": 2
         }
       ]
@@ -194,10 +224,12 @@
         },
         {
           "operation": "write",
-          "item": 1
+          "item": 1,
+          "should_succeed": true
         },
         {
           "operation": "read",
+          "should_succeed": true,
           "expected": 1
         }
       ]
@@ -208,7 +240,8 @@
       "operations": [
         {
           "operation": "write",
-          "item": 1
+          "item": 1,
+          "should_succeed": true
         },
         {
           "operation": "overwrite",
@@ -216,10 +249,12 @@
         },
         {
           "operation": "read",
+          "should_succeed": true,
           "expected": 1
         },
         {
           "operation": "read",
+          "should_succeed": true,
           "expected": 2
         }
       ]
@@ -230,11 +265,13 @@
       "operations": [
         {
           "operation": "write",
-          "item": 1
+          "item": 1,
+          "should_succeed": true
         },
         {
           "operation": "write",
-          "item": 2
+          "item": 2,
+          "should_succeed": true
         },
         {
           "operation": "overwrite",
@@ -242,10 +279,12 @@
         },
         {
           "operation": "read",
+          "should_succeed": true,
           "expected": 2
         },
         {
           "operation": "read",
+          "should_succeed": true,
           "expected": 3
         }
       ]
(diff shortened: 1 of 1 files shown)


exercises/circular-buffer/canonical-data.json
```diff
@@ -0,0 +1,227 @@
...
+    "In general, these circular buffers are expected to be stateful,",
+    "and each language will operate on them differently.",
+    "Tests tend to perform a series of operations, some of which expect a certain result.",
+    "As such, this common test suite can only say in abstract terms what should be done.",
+    "",
+    "Tests will contain a number of operations. The operation will be specified in the `operation` key.",
+    "Based on the operation, other keys may be present.",
+    "read: Reading from the buffer should produce the item located at `expected`",
+    "write: Writing the item located at `item` should succeed",
+    "readShouldRail: Reading from the buffer should fail.",
```
Insti: s/Rail/Fail/

petertseng: Excuse me! I obviously meant "complain or protest strongly and persistently about", meaning the code should *rail* at its user on an attempt to perform a read.

(Fixed)

exercises/circular-buffer/canonical-data.json
```diff
@@ -0,0 +1,227 @@
...
+        {
+          "operation": "write",
+          "item": 1
+        },
+        {
+          "operation": "read",
+          "expected": 1
+        },
+        {
+          "operation": "readShouldFail"
```
petertseng: not related to reading an item just written. A separate test, "Each item can only be read once"?

exercises/circular-buffer/canonical-data.json
```diff
@@ -0,0 +1,227 @@
...
+        {
+          "operation": "read",
+          "expected": 1
+        },
+        {
+          "operation": "read",
+          "expected": 2
+        },
+        {
+          "operation": "readShouldFail"
```
petertseng: Not related to reading items in the order they are written.

exercises/circular-buffer/canonical-data.json
Insti: This is part of a different test: "cannot read past the end of buffer" or something

exercises/circular-buffer/canonical-data.json
Insti: redundant operation.

Insti:
I think what you're doing here is good. :+1:

exercises/circular-buffer/canonical-data.json
```diff
@@ -0,0 +1,227 @@
...
+          "expected": 2
+        },
+        {
+          "operation": "read",
+          "expected": 3
+        }
+      ]
+    },
+    {
+      "description": "clear fills up capacity for another write",
```
petertseng: s/fills/frees/

petertseng:
All these commits are going to be squashed and I'll amend the commit message and PR description to be reasonable.

petertseng:
New tests added:

* "items cleared out of buffer can't be read"
* "overwrite doesn't remove an already-read item" - As first commit explains, a simplified version of a test mixing reads, writes, overwrites, but the main point as I see it was to make sure that overwrite deletes the right item if a read has previosly been performend

exercises/circular-buffer/canonical-data.json
```diff
@@ -0,0 +1,300 @@
...
+    "Based on the operation, other keys may be present.",
+    "read: Reading from the buffer should produce the item located at `expected`.",
+    "write: Writing the item located at `item` should succeed.",
+    "readShouldFail: Reading from the buffer should fail.",
+    "writeShouldFail: Writing the item located at `item` should fail.",
+    "overwrite: Write the item located at `item` into the buffer, removing the oldest item if necessary.",
+    "clear: Clear the buffer.",
+    "",
+    "Finally, note that all values are integers.",
+    "If your language contains generics, you may consider allowing buffers to contain other types.",
```
petertseng: Note the addition of this, which should look suspiciously similar to https://github.com/exercism/x-common/blob/master/exercises/react/canonical-data.json#L38-L40

Is this valuable to point out? Well, some tracks already have it, and I didn't want this JSON file to imply that they should remove those tests or make their test suites discourage generics. In the latter case, tracks may even get [questions](https://github.com/exercism/xrust/issues/111).

exercises/circular-buffer/canonical-data.json
```diff
@@ -0,0 +1,300 @@
...
+          "operation": "write",
+          "item": 2
+        },
+        {
+          "operation": "read",
+          "expected": 1
+        },
+        {
+          "operation": "write",
+          "item": 2
```
petertseng: this write 2 is accidental and will be removed.

petertseng:
https://github.com/petertseng/x-common/commit/f4be093516e58a9dd0b46401446ba94cb058b9bb

exercises/circular-buffer/canonical-data.json
petertseng: (our comments remain, but this comment has been addressed, as this line now goes under "each item may only be read once")

exercises/circular-buffer/canonical-data.json
```diff
@@ -0,0 +1,296 @@
...
+    "Tests for that are not included here.",
+    ""
+  ],
+  "cases": [
+    {
+      "description": "reading empty buffer should fail",
+      "capacity": 1,
+      "operations": [
+        {
+          "operation": "readShouldFail"
```
Insti: I wonder if this should just be:
```
"operation": "read"
"expected": false
```

Ensuring that there is an expected result for each of the 4 methods.

petertseng: Okay, I think I understand.

A few things give me pause about the `"expected": false` to signal failure on `read`:

* Not a concern: ~~now we can't have a buffer of booleans and write false into it and expect to read a false out!~~ (not relevant for this JSON file since it only deals with integers)
* The explanation of read gets a little gnarly. It becomes "read: Reading from the buffer should produce the item located at `expected`, unless `expected` is false, in which case the read should fail."
* Statically-typed languages reading this JSON file may have more trouble dealing with the fact that the value at `expected` might be either a boolean or an integer.

Okay, what if it's something like this: for a successful read of 1:

```json
{
  operation: "read",
  item: "1",
  expected: true,
}
```

and for a failing read:

```json
{
  operation: "read",
  expected: false,
}
```

this way, `expected` is a boolean for all cases.

I'm not convinced I'm happy with the name `expected` in this usage, since now it doesn't make clear that the main thing we want to check is the value read, rather than whether it was successful. What if `expected` got renamed to `success` or `shouldSucceed`?

I'm not sure if this is also part of the suggestion, but I would feel odd about adding an `expected`: key to `overwrite` and `clear`, since having it there implies that it can be false, and I don't think we'll ever have a test that expects either of these two to fail (well, maybe overwrite on a size-0 buffer, but I think I'd rather argue that size-0 buffers should just not get created)

I am having a hard time with the `"expected": false` for reads, and I'm not sure that moving the item to a separate key makes things better. Any way to make this work?

petertseng: This has a little to do with #401, what's the best way to represent an error here. I am having a sufficiently hard time with fitting `"expected": true/false` into all operations that I think I should keep them separate. This is the suggestion of https://github.com/exercism/x-common/issues/401#issuecomment-253519582 - errors are simply encoded differently. I know I didn't use that in any other JSON file I've written, but I managed to make `null` work in all of those, whereas I'm unsure whether integer/null for reads versus true/false for writes gets too confusing. Even integer/false for reads seems a bit odd.

That, and I may be a bit biased by how the [go tests](https://github.com/exercism/xgo/blob/master/exercises/circular-buffer/circular_buffer_test.go#L48) were written - they pretty much use those exact six operations already.

Insti: Command Query Responsibility Segregation.

write is a command.
overwrite is a command.
clear is a command.
read is a query.

detecting failure is a query.

It seems we are missing the `failure_detected`\* query that indicates this.

```
  "cases": [
    {
      "description": "reading empty buffer should fail",
      "capacity": 1,
      "operations": [
        {
          "operation": "read",
          "expected": null
        },
        {
          "operation": "failure_detected",
          "expected": true
        }
      ]
    },
```

Individual tracks are still free to implement the `failure_detected` check however they like, either with a method or a try/catch block or whichever other system their language encourages.

\* There's probably a better name for this.

petertseng: Even though tracks are free to implement `failure_detected` how they like, I am **guessing** that `failure_detected` as a separate query will *tend* to suggest that it be a separate function. After all, someone porting this exercise to their track might simply think "one function for each operation, right?!", because what reason do they have to think that this `failure_detected` `operation` is any different than the others (`read, write, overwrite, clear`)*?

This hypothetical API with a separate `failure_detected` operation is one I do not wish to encourage porters to implement, especially since students will subsequently see it, and be encouraged to replicate it. I tend to think it's better to make failures clear right when they happen. After all, both queries and commands have the potential of failing, so it seems cumbersome to have to check failure in a separate query. ~~A separate `failure_detected` query also seems difficult to make work in potentially concurrent contexts~~, but luckily that is not a concern for *this* particular exercise!

I have a difficult time fitting in a circular buffer with command-query separation, even without `failure_detected`. I see from https://en.wikipedia.org/wiki/Command%E2%80%93query_separation that command-query separation has the formal definition of "methods should return a value only if they are referentially transparent and hence possess no side effects."

The `read` operation **has a side-effect**, since it **removes the just-read item** from the buffer!
If we wish to follow command-query separation, `read` is not allowed to both return the oldest value and also remove it. 
Instead, we would be forced to separate the `read` operation into `peek` and `dropOldestItem`.
Well, hey, I might not be opposed to that API change, but I unfortunately don't think I will try to convince others that it is the right thing to do for their tracks.

*: Hoewver, I have just realized after writing this argument against `failure_detected` that the current JSON schema also has two "special' `operations`: `readShouldFail` and `writeShouldFail`. What's to tell a porter that *those* two operations are in any way special? ARGH! So is my proposed JSON schema flawed as well?!

Insti: > The read operation has a side-effect, since it removes the just-read item from the buffer!

This is a good point.

petertseng: (I have changes planned to remove `readShouldFail` and `writeShouldFail` since in arguing against `failure_detected` I managed to argue against the forme rtwo as well)

exercises/circular-buffer/canonical-data.json
```diff
@@ -7,10 +7,10 @@
...
     "Tests will contain a number of operations. The operation will be specified in the `operation` key.",
     "Based on the operation, other keys may be present.",
-    "read: Reading from the buffer should produce the item located at `expected`.",
-    "write: Writing the item located at `item` should succeed.",
-    "readShouldFail: Reading from the buffer should fail.",
-    "writeShouldFail: Writing the item located at `item` should fail.",
+    "read: Reading from the buffer should succeed if and only if `should_succeed` is true.",
+    "  If it should succeed, it should produce the item at `expected`. ",
+    "  If it should fail, `expected` will not be present. ",
+    "write: Writing the item located at `item` should succeed if and only if `should_succeed` is true.",
```
Insti: Do we need to say anything like 'failure may be indicated by raising an exception or similar language dependent runtime error'

petertseng: I added it just to be sure. I have a nagging feeling that if similar sentences appear in too many places...

* https://github.com/exercism/x-common/blob/master/exercises/largest-series-product/canonical-data.json#L3
* https://github.com/exercism/x-common/blob/master/exercises/hamming/canonical-data.json#L8
* https://github.com/exercism/x-common/blob/master/exercises/rna-transcription/canonical-data.json#L6
* https://github.com/exercism/x-common/blob/master/exercises/triangle/canonical-data.json#L15

then we might extract them all out to the contributing guide's section on porting an exercise. But I don't know what the right answer to that is yet, so for now I'll add the sentence.

petertseng: (or maybe we are worried that a porter will not read that particular section of the contributing guide? I don't know)

Insti: anecdata: When I'm porting, I tend to just read the canonical-data and the readme.
I don't reread the contributing guide often.
(I'm probably non-typical enough for this to be irrelevant information.)

PR #492, opened 2017-01-11
Note on luhn description for single digits

stevejb71:
Add notes on edge cases for Luhn in the description, and add a 0 length string test.

### DIFF
--- a/exercises/luhn/description.md
+++ b/exercises/luhn/description.md
@@ -7,8 +7,11 @@ The task is to write a function that checks if a given string is valid.
 
 Validating a Number
 ------
+Strings of length 1 or less are not valid. Spaces are allowed in the input,
+but they should be stripped before checking. All other non-digit characters
+are disallowed.
 
-As an example, here is a valid (but fictitious) Canadian Social Insurance
+As an example of a valid string, here is a fictitious Canadian Social Insurance
 Number.
 
 ```


exercises/luhn/description.md
```diff
@@ -7,8 +7,11 @@ The task is to write a function that checks if a given string is valid.
 
 Validating a Number
 ------
+Strings of length 1 or less are not valid. Spaces are allowed in the input,
+but they should be stripped before checking. All other non-digit characters
+are disallowed.
 
-As an example, here is a valid (but fictitious) Canadian Social Insurance
+As an example of a valid string, here is fictitious Canadian Social Insurance
```
petertseng: I think "is a" rather than just "is"

exercises/luhn/description.md
```diff
@@ -7,8 +7,11 @@ The task is to write a function that checks if a given string is valid.
 
 Validating a Number
 ------
+Strings of length 1 or less are not valid. Spaces are allowed in the input,
+but they should be stripped before checking. All other non-digit characters
```
petertseng: this sentence makes me think of other cases: how about a string of only spaces?

However, it's not true that I would refuse to see this merged unless such a case was added. What do people think of it?

exercises/luhn/canonical-data.json
```diff
@@ -1,7 +1,12 @@
 {
   "valid": [
     {
-      "description": "single digit strings can not be valid",
+      "description": "zero length strings are not valid",
+      "input": "",
+      "expected": false
```
Insti: Is there an implementation that would return the wrong result for this case?

exercises/luhn/description.md
stevejb71: I think the space stripping is already tested, so that a spaces only test would be superfluous.

exercises/luhn/canonical-data.json
stevejb71: Probably not, the single character test cases should cover this. I'm happy to remove this commit unless anyone else can think of an example...

stevejb71:
Removed the commit containing an extra test for 0 length string.

PR #496, opened 2017-01-15
Luhn: Update test of non-digit

IanWhitney:
While implementing these tests for Rust I found that the final test was
not well written (my fault!)

https://github.com/exercism/xrust/pull/247

I added the alphabetical character to an already-invalid string. So the
test doesn't show anything useful.

Instead the test should show that a known-valid string is made invalid
by the inclusion of a non-digit

### DIFF
--- a/exercises/luhn/canonical-data.json
+++ b/exercises/luhn/canonical-data.json
@@ -26,7 +26,7 @@
       "expected": false
     },
     {
-      "description": "strings that contain non-digits are not valid",
+      "description": "valid strings with a non-digit added become invalid",
       "input": "046a 454 286",
       "expected": false
     }
--- a/exercises/luhn/canonical-data.json
+++ b/exercises/luhn/canonical-data.json
@@ -27,7 +27,7 @@
     },
     {
       "description": "strings that contain non-digits are not valid",
-      "input": "827a 1232 7352 0569",
+      "input": "046a 454 286",
       "expected": false
     }
   ]


IanWhitney:
As mentioned earlier, all of this can be passed with a single equality check. https://github.com/exercism/x-common/pull/474#issuecomment-269943166

But I think the current tests work well to cover the problem.

petertseng:
The difference is that the equality check would never pass code review since it is clearly not even trying to solve the problem, whereas it is the above wrong code looks to at least be making a reasonable attempt.

petertseng:
s/the problem/the *general* problem, just the specific cases/

petertseng:
Far more dangerous to allow reasonable-looking code that is still wrong, who knows what things could be hidden there... there are entire contests dedicated to underhanded code.

What I have just said is not relevant to this PR.

petertseng:
... unless we think that a student might mistakenly submit code like the above: they think is correct, but it is in fact wrong, but it passes the tests since the tests don't expose its wrongness.

petertseng:
Edited:

The difference is that the equality check would never pass code review since it is clearly not even trying to solve the general problem, just the specific cases. A student submitting the equality check will *probably* understand that they are submitting code that is just for these specific cases. They will correctly understand that their code won't work for any other cases.

In contrast, the above wrong code looks to at least be making a reasonable attempt. A student submitting that code might very reasonably and incorrectly think that the code works for all cases. They would be wrong, but there is not test case to show them they are wrong.

Far more dangerous to allow reasonable-looking code that is still wrong, who knows what things could be hidden there... there are entire contests dedicated to underhanded code.

What I've said is only relevant to this PR if we believe a student might submit code like the above.

petertseng:
> In contrast, the above wrong code looks to at least be making a reasonable attempt. A student submitting that code might very reasonably (but incorrectly) think that the just-submitted code works for all cases. 

Oh, erm, would the student actually think that? Surely they would know something is up, given that they see that their own code doesn't handle that `a` character?!?! So what I said is only a problem if the student doesn't look at the test cases at all.

exercises/luhn/canonical-data.json
```diff
@@ -27,7 +27,7 @@
     },
     {
       "description": "strings that contain non-digits are not valid",
-      "input": "827a 1232 7352 0569",
+      "input": "046a 454 286",
```
Insti: "strings that contain non-digits are not valid even if the digits would be"

petertseng:
>  `[acc + (double ? (d * 2 % 9) : d), !double]`

wow, bug. that doubles 9 -> 0. it should be 9 -> 9, so I'm just going to do `double && d != 9`

PR #497, opened 2017-01-15
rectangles: update with 1xN, Nx1, 1x1 cases

stkent:
In the final canonical test case for this exercise, we count rectangles whose sides have length 1. Such rectangles have slightly odd-looking representations:

```
++
++
```

```
+-+
+-+
```

```
++
||
++
```

etc.

I updated the example to include a rectangle of this type to avoid an unpleasant surprise at the end of the exercise. Thoughts?

### DIFF
--- a/exercises/rectangles/canonical-data.json
+++ b/exercises/rectangles/canonical-data.json
@@ -51,6 +51,31 @@
       "expected": 5
     },
     {
+      "description": "rectangle of height 1 is counted",
+      "input": [
+        "+--+",
+        "+--+"
+      ],
+      "expected": 1
+    },
+    {
+      "description": "rectangle of width 1 is counted",
+      "input": [
+        "++",
+        "||",
+        "++"
+      ],
+      "expected": 1
+    },
+    {
+      "description": "1x1 square is counted",
+      "input": [
+        "++",
+        "++"
+      ],
+      "expected": 1
+    },
+    {
       "description": "only complete rectangles are counted",
       "input": [
         "  +-+",
--- a/exercises/rectangles/description.md
+++ b/exercises/rectangles/description.md
@@ -2,13 +2,13 @@ Create a program to count the rectangles in an ASCII diagram like the one below.
 
 ```
    +--+
-   |  |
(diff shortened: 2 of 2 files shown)


Insti:
Adding tests for 1 by x rectangles would be a good idea.

stkent:
Good call! Should I take care of that in this PR or a separate one?

petertseng:
I would not object to it going in this PR. Personally, I have in the past separated PRs if I thought all of the following were true:

* The README change will go through quickly
* The JSON change might not (it might need more review, etc.)
* It is desirable to merge the README change ASAP because not doing so causes confusion for students, etc.

stkent:
Ok, thanks for the guidance! I'll add those tests and we can iterate from there 👍

stkent:
@petertseng @Insti revised with new canonical test data :)

kotp:
The last canonical test for which track?  The last canonical test I see is [here](https://github.com/exercism/x-common/blob/master/exercises/triangle/canonical-data.json#L110) so I am a little confused.  Or is it the last isosceles triangle test (that isn't optional)?

stkent:
@kotp this is the `rectangles` exercise, not the `triangle` exercise! I think you may have replied to the wrong thread.

kotp:
See, I told you I was :confused: !

kotp:
Feel free to squash these commits... I think it is :+1:

exercises/rectangles/canonical-data.json
```diff
@@ -51,6 +51,33 @@
...
     {
+      "description": "rectangle of height 1 is counted",
+      "input": [
+        "+---+",
+        "+---+"
+      ],
+      "expected": 1
+    },
+    {
+      "description": "rectangle of width 1 is counted",
```
Insti: Is there a reason why both of these rectangles have their other dimension 3?
Perhaps one of them could be 1x2.

(This may be a frivolous requirement.)

stkent: No special reason; I'll update!

Issue #352, opened 2016-08-22
Language-specific exercises


ErikSchierboom:
At the moment, there are a couple of exercises that are only relevant to one language. The best example is probably the [`lens-person`](https://github.com/exercism/x-common/blob/master/lens-person.md) exercise, which is only relevant to Haskell. Is having a language-specific exercise something that is considered good practice, or should we shy away from it? I ask this because F# has a very cool feature named `computation expressions` and I was wondering if I could make an exercise for that feature.

Insti:
Go for it. 
The more generic you can make the problem description the better.
And remember that you can always add a HINTS.md file in the language track that will add further language specific instructions to the Readme.

ErikSchierboom:
Okay cool!

petertseng:
As an aside (no action is required from this comment), the lens exercise is no longer Haskell-specific. Scala has a implementation too, which may be amusingly confusing because the metadata.yml still says "(Haskell-specific)", so seeing that in the list of Scala exercises may give a student pause.

Actually, maybe there *is* an action to be taken here... remove "Haskell specific" from lens-person. Maybe "Functional programming language specific"

petertseng:
I do want to ask whether an F#-specific exercise ever materialised (but I think I would have seen an x-common PR if it did, so I have to assume no)

petertseng:
I'm asking because the Rust track is now discussing some exercises that only make sense for it, since they extend an exercise (Luhn) with Rust-specific features (traits).

NobbZ:
> I'm asking because the Rust track is now discussing some exercises that only make sense for it, since they extend an exercise (Luhn) with Rust-specific features (traits).

I do not think that "traits" are specific to rust, but maybe named differently in other languages. As I do understand them, they are similar to Haskells classes, Elixirs protocols. Someone else told me, that rusts traits are even lent from scalas traits.

petertseng:
> I do not think that "traits" are specific to rust, but maybe named differently in other languages.

You have caught me exaggerating! I should not make such bold claims of uniqueness.

I guess the question that should be asked is - would other languages want to have the metadata for this new Luhn-based exercise in x-common? If yes, then maybe we will just add it to x-common! If no, then *for now* it is as if it were a language-specific exercise, if only the Rust track is interested in it.

Let's explain what the Rust proposals are, so that other languages can determine whether they are interested in them.

There are two proposals in the Rust track, which both use Rust traits but, off the top of my head, admit the following possibilities in other languages:

[...]

PR #503, opened 2017-01-17
lens: mark as immutable-data-specific, not Haskell specific

petertseng:
At the time it was added (#97), indeed [Haskell][haskell] was the only
implementing track, and it was marked Haskell-specific.

However, the [Scala][scala] track has also implemented this exercise.
I predict it is confusing for Scala students to see "Haskell specific"
in the list of the Scala exercises. The proposal is to change it to
"specific to languages with immutable data".

[haskell]: https://github.com/exercism/xhaskell/pull/84
[scala]: https://github.com/exercism/xscala/pull/157

Lenses can be a fit for any language that has immutable data and data
structures that can be nested.

In such a language, updating a deeply-nested field can be a lot of work,
but made easier with lenses. This exercise demonstrates the use of
lenses and is applicable to **any** language with immutable data, not
just Haskell.

The alternative considered was to use "functional programming" specific,
but the term "functional programming" is a little too broad. It's
altogether possible that a functional programming language could have
mutable data and a non-functional programming language could have
immutable data, so let's be precise about where lenses are helpful.

As of this writing, I do not believe any other exercise is
language-specific. Questions to this effect have been asked in:

* #352
* https://github.com/exercism/exercism.io/issues/3113

If you are curious about having language-specific exercises, you may be
interested in https://github.com/exercism/trackler/issues/22.

### DIFF
--- a/exercises/lens-person/metadata.yml
+++ b/exercises/lens-person/metadata.yml
@@ -1,5 +1,5 @@
 ---
-blurb: "(Haskell specific) use lenses to update nested records."
+blurb: "Use lenses to update nested records (specific to languages with immutable data)."
 common:
     type: focus
     focus: lenses


petertseng:
> We could be more specific and say "(Immutable data language specific)", but I'm not sure that makes as much sense.

Or, be more verbose? "(Specific to languages with immutable data)"

ErikSchierboom:
@petertseng I quite like the more specific "Specific to languages with immutable data" option. There's nothing stopping a non-functional language from having immutable data structures, in which case a lenses library would also make sense.

petertseng:
Indeed, I have decided to use "Specific to languages with immutable data". As I have updated the commit message to say (and will update the PR description), functional programming is quite a broad term, whereas if the message is more specific it's easier for track maintainers to decide whether the exercise makes sense for a given track.

ErikSchierboom:
One last thing. Wouldn't it make sense to change the description from "(Specific to languages with immutable data) use lenses to update nested records." to "Use lenses to update nested records (specific to languages with immutable data)." That would first focus on what the exercise is about, and then add the restriction. Makes more sense I think.

petertseng:
I could honestly go either way on that, but I'll lean slightly in favour of your proposal since it seems more consistent with other descriptions.

ErikSchierboom:
LGTM! :tada:

PR #505, opened 2017-01-23
nucleotide-count: Add canonical-data.json

bunnymatic:
Nucleotide count description: https://github.com/exercism/x-common/blob/master/exercises/nucleotide-count/description.md

### DIFF
--- a/exercises/nucleotide-count/canonical-data.json
+++ b/exercises/nucleotide-count/canonical-data.json
@@ -1,0 +1,44 @@
+{
+  "nucleotide_counts": {
+    "description": "count all nucleotides in a strand",
+    "cases": [
+      {
+        "description": "empty strand",
+        "strand": "",
+        "expected": {
+          "A": 0,
+          "C": 0,
+          "G": 0,
+          "T": 0
+        }
+      },
+      {
+        "description": "strand with repeated nucleotide",
+        "strand": "GGGGGGG",
+        "expected": {
+          "A": 0,
+          "C": 0,
+          "G": 7,
+          "T": 0
+        }
+      },
+      {
+        "description": "strand with multiple nucleotides",
+        "strand": "AGCTTTTCATTCTGACTGCAACGGGCAATATGTCTCTGTGTGGATTAAAAAAAGAGTGTCTGATAGCAGC",
+        "expected": {
+          "A": 20,
+          "C": 12,
+          "G": 17,
+          "T": 21
+        }
+      },
+      {
+        "description": "strand with invalid nucleotides",
+        "strand": "AGXXACT",
+        "expected": {
+          "error": "Invalid nucleotide in strand"
+        }
+      }
+    ]
+  }
+}


bunnymatic:
I noticed that the current Java test suite for Nucleotide includes a test that checks for exceptions being thrown.  I did not include that test here but if I should, I see a pattern in here that i could use... https://github.com/exercism/x-common/blob/master/exercises/grains/canonical-data.json

bunnymatic:
sounds good.  don't merge it yet.  I'll add a couple of those suggested
items.

bunnymatic:
So... if you're inclined to start a pattern or continue the pattern of "an exception was thrown", i'm happy to add something like:
```
   {
        "description": "this whatever raises an exception",
        "strand": "bogus",
        "expected": -1
    }
```
as needed. Should I proceed with that?  Do we need a quorum of 👍 's?

jtigger:
While it hasn't gotten marked as "policy", there was something that looks like rough consensus with regards to exceptional values in #401:

```json
{
  "description": "garbage input",
  "input": "100$#^0002",
  "expected": {
    "error": {
      "message": "Invalid input."
    }
}
```

bunnymatic:
that seems better than -1 for sure

bunnymatic:
I've added a proposal for 2 new exception cases.  One is handled (already) in the java tests for this exercise.

jtigger:
I like the simpler syntax.

jtigger:
Thanks for doing this, btw, @bunnymatic; every time someone codifies an exercise, an angel gets her wings. 👼 

But seriously, it significantly reduces the work required to port.  Even for tracks without a test generator, it's far far easier for both the contributor and the reviewer with this kind of data in hand.  🙇

Insti:
I really don't like that this problem is trying to test 2 things. 

Either it should count individual nucleotides, or count all of them.

Someone convince me that doing both is a good idea.

petertseng:
How came it to be this way?

Looks like through https://github.com/exercism/exercism.io/commit/4c438520e7e04b2d5eb7d220c3297b233e354972

Where'd the problem come from? http://rosalind.info/problems/dna/ - there, it actually just wants the four counts at once, never asking for the individuals. So asking for the individuals is our extension!

Should it stay this way?

I have no arguments to offer for either side.

I note that if going back to taking only all counts, this problem becomes quite similar to word-count. Compared to word count, there is no longer the obligation to split on words, but an added obligation to validate the input for valid nucleotides only.

petertseng:
> validates nucleotides in requested nucleotide

> validates nucleotides in the strand

Shall there also be a test for validating nucleotides in the strand for `count`, as well?

Insti:
Counting them all at once seems like it makes for a more interesting problem with a couple of different approaches to discuss.

My vote is for only test for the case where all nucleotides are counted at once.

bunnymatic:
@petertseng  - i'm down with another exception test around validate nucleotides during `count`. I'll wait until I know whether or not we're going to do the single "countAll" tests or both.

bunnymatic:
As for the two test discussion, I've seen (in reviewing this for other folks on the site) different approaches that may be a remnant of the problem statement and the test cases.   I've see folks solve it by writing `count(nucleotide) = function() {}` where that runs the strand and pulls out only that nucleotide and counts it up.  then `counts()` simply runs through all known nucleotides and calls `count(nuc)`.  The other approach has been - `counts()` runs the whole strand and counts everything and stashes it on the instance.  Then `count` just looks into this instance map.  These two solutions are good for discussion points around memoization/caching, lookup speed etc.

If we went with the "count all" as the only tests, do you think we'd still get both of these solutions and would it reduce the possibility for discussion/learning around these exercises?  food for thought.

exercises/nucleotide-count/canonical-data.json
```diff
@@ -3,31 +3,25 @@
...
       {
-        "description": "empty dna strand has no adenine",
+        "description": "empty strand",
         "strand": "",
         "nucleotide": "A",
         "expected": 0
       },
       {
-        "description": "dna strand with only guanine has no adenine",
+        "description": "strand with one nucleotide",
```
Insti: 'one repeated nucleotide'

Insti:
If we went with the "count all" as the only tests, do you think we'd still get both of these solutions and would it reduce the possibility for discussion/learning around these exercises? 

It would not reduce the discussion possibility.
Both approaches are still applicable for 'count all'

exercises/nucleotide-count/canonical-data.json
```diff
@@ -0,0 +1,75 @@
...
+        "strand": "",
+        "expected": {
+          "A": 0,
+          "C": 0,
+          "G": 0,
+          "T": 0
+        }
+      },
+      {
+        "description": "strand with single nucleotide",
```
Insti: "repeated"

bunnymatic:
Given the big rewrite, this last commit is a squish leaving only the latest version of the data file around.

bunnymatic:
Once this is merged, we could also close this https://github.com/exercism/todo/issues/116

petertseng:
Thanks!

PR #507, opened 2017-01-24
etl: Add canonical data

bunnymatic:
After scouring the existing test suites, I noticed that a few of them [Python, for example] (https://github.com/exercism/xpython/blob/master/exercises/etl/etl_test.py) include keys that are not just single letters but words.  I've included one extra case at the bottom that tests that case.  Not sure if it's something we want to include here or not.

### DIFF
--- a/exercises/etl/canonical-data.json
+++ b/exercises/etl/canonical-data.json
@@ -1,5 +1,12 @@
 {
   "transform": {
+    "#": [
+	"Note:  The expected input data for these tests should have",
+        "integer keys (not stringified numbers as shown in the JSON below",
+        "Unless the language prohibits that, please implement these tests",
+        "such that keys are integers. e.g. in JavaScript, it might look ",
+        "like `transform( { 1: ['A'] } );`"
+    ],
     "description": "transforms the a set of scrabble data previously indexed by the tile score to a set of data indexed by the tile letter",
     "cases": [
       {
--- a/exercises/etl/canonical-data.json
+++ b/exercises/etl/canonical-data.json
@@ -1,0 +1,62 @@
+{
+  "transform": {
+    "description": "transforms the a set of scrabble data previously indexed by the tile score to a set of data indexed by the tile letter",
+    "cases": [
+      {
+        "description": "a single letter",
+        "input": {
+          "1": ["A"]
+        },
+        "expected": {
+          "a" : 1
+        }
+      },
+      {
+        "description": "single score with multiple letters",
+        "input": {
+          "1": ["A", "E", "I", "O", "U" ]
+        },
+        "expected": {
+          "a" : 1,
+          "e" : 1,
+          "i" : 1,
+          "o" : 1,
+          "u" : 1
+        }
+      },
+      {
+        "description": "multiple scores with multiple letters",
+        "input": {
+          "1": ["A", "E"],
+          "2": ["D", "G"]
+        },
+        "expected": {
+          "a" : 1,
+          "e" : 1,
+          "d" : 2,
+          "g" : 2
+        }
+      },
+      {
+        "description": "multiple scores with differing numbers of letters",
+        "input": {
+          "1": [ "A", "E", "I", "O", "U", "L", "N", "R", "S", "T" ],
+          "2": [ "D", "G" ],
+          "3": [ "B", "C", "M", "P" ],
+          "4": [ "F", "H", "V", "W", "Y" ],
+          "5": [ "K" ],
+          "8": [ "J", "X" ],
+          "10": [ "Q", "Z" ]
+        },
+        "expected": {
+          "a": 1, "b": 3,  "c": 3, "d": 2, "e": 1,
+          "f": 4, "g": 2,  "h": 4, "i": 1, "j": 8,
+          "k": 5, "l": 1,  "m": 3, "n": 1, "o": 1,
+          "p": 3, "q": 10, "r": 1, "s": 1, "t": 1,
+          "u": 1, "v": 4,  "w": 4, "x": 8, "y": 4,
+          "z": 10
+        }
+      }
+    ]
+  }
+}


kytrinyx:
My initial thought is "no", but I'm willing to be convinced otherwise.

The way I'm thinking of this is that we have a scrabble alphabet, and we're changing the structure of the tile => point mapping.

If there are words (or multiple letters), that seems like a different problem.

bunnymatic:
sounds good to me.  i think i agree.  the way the problem is stated is
related to the scrabble board.  i'll update accordingly

Insti:
@bunnymatic, some Pull Request tips:

* Title format is "problem-slug: description" in this case "etl: Add canonical data"
* It's helpful to link to the description of the problem in the PR for canonical data files as this often needs to be referred to for context.
https://github.com/exercism/x-common/blob/master/exercises/etl/description.md

Insti:
Word tests should not be included.

exercises/etl/canonical-data.json
```diff
@@ -0,0 +1,62 @@
...
+        "description": "transforms one value",
+        "input": {
+          "1": ["A"]
+        },
+        "expected": {
+          "a" : 1
+        }
+      },
+      {
+        "description": "tramsforms more values",
```
Insti: "transforms" is spelled incorrectly (multiple times).

I'd rather see more useful test descriptions.
"Transforms more values" is a possible description for all these tests, what is this test actually testing?

bunnymatic: i'll fix the typos.  interestingly, the test descriptions are thin but consistently thin across the existing suites.  I can beef them up.  I was mimicking what i'd seen in the existing...

bunnymatic: Do you have an idea for the test descriptions?  I started with something more like
`transforms a map with a single key and an array of one value into a map with one key and one value`.  That works for the first one, but the next ones get pretty wordy.  Is that where you thought that might go?

If this file acted like I think of rspec describe/context/it blocks, the outer description could be kind of the prefix to the inner descriptions with something like

```
{
  "transform": {
    "description": "transforms the input map",
    "cases": [
      {
        "description": "with one key that has an array with one value into a map with  one key and one value",
        "input": {"1": ["A"] },
        "expected": { "a" : 1 }
      },
      {
        "description": "with one key and multiple values into a map with multiple keys each with a single value",
        "input": { "1": ["A", "E", "I", "O", "U" ] },
        "expected": {
          "a" : 1,
          ...
```

I'm happy to rewrite the descriptions, but i'd love a little direction or suggestion

exercises/etl/canonical-data.json
```diff
@@ -0,0 +1,62 @@
+{
+  "transform": {
+    "description": "transforms the input map from a map of value: [ keys ] to a map of key: value for each of the keys",
+    "cases": [
+      {
+        "description": "transforms one value",
```
Insti: "single value"

exercises/etl/canonical-data.json
```diff
@@ -0,0 +1,62 @@
...
+          "1": ["A"]
+        },
+        "expected": {
+          "a" : 1
+        }
+      },
+      {
+        "description": "transforms more values",
+        "input": {
+          "1": ["A", "E", "I", "O", "U" ]
```
Insti: "multiple values with the same score"

exercises/etl/canonical-data.json
```diff
@@ -0,0 +1,62 @@
...
+        "expected": {
+          "a" : 1,
+          "e" : 1,
+          "i" : 1,
+          "o" : 1,
+          "u" : 1
+        }
+      },
+      {
+        "description": "transforms more keys",
```
Insti: "multiple scores with multiple values"

exercises/etl/canonical-data.json
```diff
@@ -0,0 +1,62 @@
...
+        },
+        "expected": {
+          "a" : 1,
+          "e" : 1,
+          "d" : 2,
+          "g" : 2
+        }
+      },
+      {
+        "description": "transforms a full data set",
```
Insti: "multiple scores with differing numbers of letters"

exercises/etl/canonical-data.json
Insti: Why is there no test for "multiple scores with a single letter" ?

exercises/etl/canonical-data.json
Insti: Could also be, "single score with a single letter"

exercises/etl/canonical-data.json
Insti: See below for some suggestions.

exercises/etl/canonical-data.json
```diff
@@ -0,0 +1,62 @@
+{
+  "transform": {
+    "description": "transforms the input map from a map of value: [ keys ] to a map of key: value for each of the keys",
```
Insti: This prompting for a potential name of the method is not necessary, see: [circular-buffer](https://github.com/exercism/x-common/pull/488/files#diff-e447feb71d0e589a8d7bef71469adceb)
It has a comment, and then the list of cases.

petertseng: Despite having added the circular-buffer file, now I get to play devil's advocate: when the time comes to do #336, maybe someone will be unhappy that JSON files with only a single function-under-test have "cases" as a key under the top-level object, whereas those with multiple functions-under-test have "cases" in second-level objects.

But I have no particular reason to favour one way or the other until we come up with a proposed unified schema for all the JSON files.

bunnymatic: If there is the idea that this file could be used to generate a test suite, wouldn't you kind of need the method name to know how to invoke the method in the test call?

exercises/etl/canonical-data.json
bunnymatic: Thanks @Insti for all these.  I think i'm getting it.  I'll make updates accordingly

exercises/etl/canonical-data.json
bunnymatic: Though it could be an interesting error/edge scenario, it doesn't fit in with the problem description which says we're dealing with scrabble tile scores.  Because of scrabble rules, there is no case where one A tile is worth 1 and another is worth 2.

exercises/etl/canonical-data.json
Insti: The way the function is invoked is very language specific and will need to be specified by the language anyway.

@petertseng makes some good points above and I am sympathetic to the idea that some kind of identifier for the name of the problem (or section of the problem) can be specified in the JSON.

> until we come up with a proposed unified schema for all the JSON files.

 :+1:

So I guess what you're doing here is fine for now.

exercises/etl/canonical-data.json
```diff
@@ -1,9 +1,9 @@
 {
   "transform": {
-    "description": "transforms the input map from a map of value: [ keys ] to a map of key: value for each of the keys",
+    "description": "transforms the a set of scrabble data previously indexed by the tile score to a set of data indexed by the tile letter",
     "cases": [
       {
-        "description": "transforms one value",
+        "description": "transforms a single letter",
```
Insti: 'transforms ' is redundant in all these descriptions.

bunnymatic:
This may look like a pile of new commits, it was a rebase on the most current master.  Just FYI

exercises/etl/canonical-data.json
```diff
@@ -38,7 +38,7 @@
         }
       },
       {
-        "description": "transforms the full set of scrabble tiles (multiple scores with differing numbers of letters)",
+        "description": "the full set of scrabble tiles (multiple scores with differing numbers of letters)",
```
Insti: That they are scrabble tiles is irrelevant and leads to overly verbose test names since many tracks construct the test name from the description.
"multiple scores with differing numbers of letters" should be sufficient to describe the test.

If any additional description about the problem is required it should be in the `description.md` file

bunnymatic:
updated that description and squished this to 1 commit

exercises/etl/canonical-data.json
```diff
@@ -0,0 +1,62 @@
+{
+  "transform": {
+    "description": "transforms the a set of scrabble data previously indexed by the tile score to a set of data indexed by the tile letter",
+    "cases": [
+      {
+        "description": "a single letter",
+        "input": {
+          "1": ["A"]
```
petertseng: I know that this is a `"1":` (rather than `1:`) because JSON would only allow string keys in their objects.

Is it the case that languages will usually use integers as their keys? If so, perhaps a note saying so? Can put it on the same level as `description` and `cases`. No standard key for that note, for some reason people ahve used `"#":` in the past, but maybe `"notes":` is better.

bunnymatic:
Nope.  I get it.  I can see how with a large open source project like this, you want to lay down the rules and patterns as much as possible and that's hard given the open source ness and the scope (all the different languages). 

I'm happy to help out.  Hopefully the next ones will be closer to correct at the first commit.

exercises/etl/canonical-data.json
bunnymatic: I also thought about this when I started this PR but then quickly got distracted with other changes.  I'll add a note.  I think it's an important thing to call out.

bunnymatic: I looked at several data files.  Not a single one uses `notes`.  Though that is probably a better key (long term) I'm not sure this commit should be the one to start that. So I went with `#`.

If there is some documentation that could be updated... and if you guys want (me) to build a script to update all `#` entries with `notes` to lay down the "new" convention.  I'd be happy to help.  but for now, i'm sticking with `#`

Insti:
Thanks @bunnymatic :heart:

bunnymatic:
Someone can probably close https://github.com/exercism/todo/issues/90 now

Issue #279, opened 2016-06-23
Deprecate binary, trinary, octal, and hexadecimal in favor of new all-your-base exercise


kytrinyx:
We've settled on replacing the four almost identical exercises with one that is more interesting and comprehensive, which converts a number in an arbitrary base to the equivalent in a second arbitrary base. See https://github.com/exercism/x-common/issues/276 for details.

We need to (I think):

- [ ] wait for the pull request that adds `all-your-base` to be accepted
- [ ] draft the text for a new issue that can be submitted to all the languages (see https://github.com/exercism/blazon#tips)
- [ ] submit the issue using blazon
- [ ] wait for all the related issues to be closed
- [ ] finish working out the process for deprecating exercises in https://github.com/exercism/discussions/issues/40
- [ ] draft the text of an issues to submit to any tracks who implement binary, trinary, octal, hexadecimal
- [ ] submit the issues using blazon
- [ ] wait for all the related issues to go red
- [ ] close this issue

kytrinyx:
FYI @rbasso @NobbZ @ErikSchierboom @wobh @Cohen-Carlisle @yurrriq

Before we can submit a blazon ticket we need to finish clarifying the process for deprecating exercises. Up until now it's been messy and prone to error, and I'd rather have that all straightened out before we deprecate four at a time across a bunch of language tracks. https://github.com/exercism/discussions/issues/40

I've updated the checklist in the original post of this issue to reflect what I think we need to do to get this all sorted out.

verdammelt:
Just curious - is this still moving forward? Obviously *I* still need to implement one for xlisp...

kytrinyx:
@verdammelt yeah, we're still moving forward on this one. I've lost track of where we are in the process, though. I think we have implemented support for global deprecation.

mienaikage:
https://github.com/exercism/xperl6/pull/54

petertseng:
(In the case of some tracks) too late!

* [C# deprecated](https://github.com/exercism/xcsharp/pull/92)
* [F# deprecated](https://github.com/exercism/xfsharp/issues/169)
* [Haskell deprecated](https://github.com/exercism/xhaskell/pull/241)
* [Rust deprecated](https://github.com/exercism/xrust/commit/9c527e465fd95eef5b4243f123acec76892f212f)

stkent:
Ah, good knowledge! I see that some track deleted old implementations, while others (appear to have) kept them - is there a good reason to choose one over the other?

[...]

PR #510, opened 2017-01-24
trinary: Add canonical-data.json

bunnymatic:
Added data for trinary exercise.

Included 3 tests around invalid trinary strings.  This is basically a union of all invalid tests that I saw across different tracks and it matches with the [exercise description](https://github.com/exercism/x-common/blob/master/exercises/trinary/description.md).

This addresses this issue on the `todos` repo: https://github.com/exercism/todo/issues/158

### DIFF
--- a/exercises/trinary/canonical-data.json
+++ b/exercises/trinary/canonical-data.json
@@ -1,0 +1,62 @@
+{
+  "to_decimal": {
+    "description": "returns the decimal representation of the input trinary value",
+    "cases": [
+      {
+        "description": "trinary 1 is decimal 1",
+        "input": 1,
+        "expected": 1
+      },
+      {
+        "description": "trinary 2 is decimal 2",
+        "input": 2,
+        "expected": 2
+      },
+      {
+        "description": "trinary 10 is decimal 3",
+        "input": 10,
+        "expected": 3
+      },
+      {
+        "description": "trinary 11 is decimal 4",
+        "input": 11,
+        "expected": 4
+      },
+      {
+        "description": "trinary 100 is decimal 9",
+        "input": 100,
+        "expected": 9
+      },
+      {
+        "description": "trinary 112 is decimal 14",
+        "input": 112,
+        "expected": 14
+      },
+      {
+        "description": "trinary 222 is decimal 26",
+        "input": 222,
+        "expected": 26
+      },
+      {
+        "description": "trinary 1122000120 is decimal 32091",
+        "input": 1122000120,
+        "expected": 32091
+      },
+      {
+        "description": "invalid trinary digits returns 0",
+        "input": "1234",
+        "expected": 0
+      },
+      {
+        "description": "invalid word as input returns 0",
+        "input": "carrot",
+        "expected": 0
+      },
+      {
+        "description": "invalid numbers with letters as input returns 0",
+        "input": "0a1b2c",
+        "expected": 0
+      }
+    ]
+  }
+}


exercises/trinary/canonical-data.json
```diff
@@ -0,0 +1,62 @@
...
+      },
+      {
+        "description": "trinary 1122000120 is decimal 32091",
+        "input": 1122000120,
+        "expected": 32091
+      },
+      {
+        "description": "invalid trinary digits returns 0",
+        "input": "1234",
+        "expected": 0
```
petertseng: Ah. Looks like [binary](https://github.com/exercism/exercism.io/commit/5683133a9459890ebb8989932bf675cead52e2c3) and [trinary](https://github.com/exercism/exercism.io/commit/bcfd3fc72022f99292b22f8386f5c4e22e81aabf) started the same, but trinary didn't get the https://github.com/exercism/x-common/commit/af4e4a0d6aff6698180a00dc78ef9cd28330c920 treatment. That's too bad. I do dislike to encourage behaviour that can't distinguish between valid and invalid inputs.

bunnymatic: should we make these `{ expected: { error: "invalid input" }}`

I suppose as I've been doing this work of generating the canonical, i wonder how much I'm supposed to make it match the tests that exist or if it's supposed to become the "rule" and i should just expect that the tests that are out there that don't match will someday get updated.  It becomes about how much of these should be "what we think the canonical test suite should look like" and how much of it is "what the majority of tests say the canonical test suite is right now"

ErikSchierboom:
Ehm, isn't `trinary` a deprecated exercise? See #279.

bunnymatic:
Ha!

I just pulled it for `ruby` yesterday (with a `fetch`).  After doing the exercise and realizing that I'd also done the Java version, I thought i might as well do the `canonical`.

So once those are "deprecated", how do we remove them from the existing tracks?

What about adding a DEPRECATED.md file or something in the `x-common` repo as an indicator?

Either way, just trying to help out.

I'll close this guy.

https://media.giphy.com/media/yyp58Yx3Qk0TK/giphy.gif

bunnymatic:
I suppose we should close exercism/todo#158 if this is going to be unused

ErikSchierboom:
Oh, you didn't need to close it yet. I'll keep it open until @petertseng or @kytrinyx has something to say about.

bunnymatic:
*UNCLOSE*!

petertseng:
I take the same position as https://github.com/exercism/x-common/pull/296#issuecomment-234810532 - I don't oppose its addition. I recommend *not* taking my word as having any more authority than any other track maintainer though! =D (I say that because other commitments prevent me from pushing a cross-track initiative to have all tracks deprecate their existing base-conversion exercises. If there is such an individual pushing the initiative, that individual should recommend that this file not be added!)

> So once those are "deprecated", how do we remove them from the existing tracks?

All tracks that have added the all-your-base exercise should add the other base-conversion exercises to the deprecated section of the track config file. This stops `exercism fetch` from fetching the exercise, unless specifically requested to do so by specifying it. A month and a half ago I made that list of tracks that had alerady done so: https://github.com/exercism/x-common/issues/279#issuecomment-264938769

Global deprecation is performed by adding a .deprecated file in the directory in x-common. Someone will need to remind me what that does. https://github.com/exercism/x-api/issues/139 says it means it doesn't show up in todo lists, please confirm whether that's true, and how it interacts with the per-track deprecation (if it does at all...!)

It will be worthwhile, however, to state the impending deprecation in the todo issues for the remaining two, hexadecimal and octal, so that it is known that other problems can take priority. I have done so.

> how much of these should be "what we think the canonical test suite should look like" and how much of it is "what the majority of tests say the canonical test suite is right now"

I think of it as the former. In the investigation, we may discover that the majority test suite is a good one, in which case the two will be the same, but we would generally like to take the opportunity to make sure that the test suite is one we are happy with when we add the tests to this repo.

I know sometimes I want to argue "hey listen, I just submitted this PR to show what the current tests look like. I'm really not interested in making any changes y'all want to suggest. I just want this PR to be over with.", but I do think it is part of the work involved in doing these.

 **This may be important enough to document.**

petertseng:
EBWODP: Clarifying that my "commitments prevent me from pushing a cross-track initiative to have all tracks deprecate their existing base-conversion exercises" are **time** commitments, not any *other* sort of commitment (in case anyone was wondering whether it was a moral commitment, a monetary commitment(???), etc.)

bunnymatic:
Thanks @petertseng . I guess I'll give it a little time.  We can see If @kytrinyx weighs in.  Though based on the comment you pointed to, it sounds like we're leaning towards merging it in, and letting the deprecation (which happens later) take care of dumping it, right?  Is that what I heard?

Interestingly, after reading comments from @Insti [here](https://github.com/exercism/x-common/pull/513#issuecomment-275187804), and above, I want to rewrite the whole thing so that the test descriptions make more sense and maybe there are fewer tests because what are we testing really with all those cases..

But if the exercise is deprecated, then I think... who cares ;)

I'm happy to close this one and leave it unfinished or whatever.

Just let me know.

Insti:
Unless you really want to, I'd leave this one as it is and not worry too much about improving it. There are many other things that would be a better use of your time.

bunnymatic:
I'm happy to leave this alone.  And if it gets merged, fine.  If not, fine.

kytrinyx:
I think based on the above discussion: (1) let's not worry about improving it, but also (2) let's not throw it away. Also, eventually, we should probably take the time to figure out exactly what work needs to happen to properly deprecate everything. Are we missing issues notifying all the various tracks about it?

PR #513, opened 2017-01-24
prime-factors: add canonical data

bunnymatic:
Added canonical-data.json for the prime factors exercise [exercise description here](https://github.com/exercism/x-common/blob/master/exercises/prime-factors/description.md).

Addresses https://github.com/exercism/todo/issues/129

### DIFF
--- a/exercises/prime-factors/canonical-data.json
+++ b/exercises/prime-factors/canonical-data.json
@@ -1,0 +1,42 @@
+{
+  "for": {
+    "description": "returns prime factors for the given input number",
+    "cases": [
+      {
+        "description" : "no factors",
+        "input" : 1,
+        "expected" : []
+      },
+      {
+        "description" : "prime number",
+        "input" : 2,
+        "expected" : [2]
+      },
+      {
+        "description" : "square of a prime",
+        "input" : 9,
+        "expected" : [3, 3]
+      },
+      {
+        "description" : "cube of a prime",
+        "input" : 8,
+        "expected" : [2, 2, 2]
+      },
+      {
+        "description" : "product of primes and non-primes",
+        "input" : 12,
+        "expected" : [2, 2, 3]
+      },
+      {
+        "description" : "product of primes",
+        "input" : 901255,
+        "expected" : [5, 17, 23, 461]
+      },
+      {
+        "description" : "factors include a large prime",
+        "input" : 93819012551,
+        "expected" : [11, 9539, 894119]
+      }
+    ]
+  }
+}


Insti:
Again I have some issues with the test descriptions.

Have a read through this comment: https://github.com/exercism/x-common/pull/451#issuecomment-263070019

exercises/prime-factors/canonical-data.json
```diff
@@ -0,0 +1,62 @@
+{
+  "for": {
+    "description": "returns prime factors for the given input number",
```
Insti: What are these 2 lines trying to achieve?

It's better to stick with the standard format. Have a look at some of the [recent files added by @petertseng](https://github.com/exercism/x-common/search?utf8=%E2%9C%93&q=author%3Apetertseng+type%3Apull+canonical&type=Issues) for good examples of what to do.

bunnymatic: The format that I was following (from other exercises) had the "method name" as the key for a block,  then a description for what that method did, then the cases for that method call tests.

The examples I was following are here:
* https://github.com/exercism/x-common/tree/master/exercises/allergies/canonical-data.json
* https://github.com/exercism/x-common/tree/master/exercises/alphametics/canonical-data.json
* https://github.com/exercism/x-common/tree/master/exercises/atbash-cipher/canonical-data.json
* https://github.com/exercism/x-common/tree/master/exercises/binary/canonical-data.json
* https://github.com/exercism/x-common/tree/master/exercises/book-store/canonical-data.json
* https://github.com/exercism/x-common/tree/master/exercises/bowling/canonical-data.json
* https://github.com/exercism/x-common/tree/master/exercises/change/canonical-data.json
* https://github.com/exercism/x-common/tree/master/exercises/clock/canonical-data.json**
* https://github.com/exercism/x-common/tree/master/exercises/luhn/canonical-data.json

Somewhere in my travels through all different tickets and discussions around canonical data, I thought I read something that suggested a preference for the method name being included (not just the `cases` key).  So that's what that is for.

bunnymatic:
Agreed - the test descriptions are bad.  And the note you pointed to was good.  

I'm still having trouble figuring out exactly how to write the tests without the data in them and that indicate what we're testing.  In fact, if i go that route, then I think we have far more tests than we need.

Thinking about it from scratch, I could imagine tests like

* returns no factors for 1 which is prime
* returns a single factor for a prime number  (test 2 or 3)
* returns multiple factors that are the same for a number that is a square of a prime (test 4 or 9)
* returns multiple factors that are the same for a number that is a cube of a prime (test 8 or 27)
* returns multiple prime factors for a value that is a product of primes and non-primes ( test 6 )
* returns multiple prime factors for a value that is a product of primes
* returns multiple prime factors for a value that is a product of primes even if the value is very large

If this were the test suite, we'd could probably lose the test for 2, 9 and 27 and 625 and still have the coverage needed to show that the method works.

Thoughts?

exercises/prime-factors/canonical-data.json
Insti: Ok, It's probably fine, further discussion here: https://github.com/exercism/x-common/pull/507#pullrequestreview-18471819

Insti:
>    returns no factors for 1 which is prime
    returns a single factor for a prime number (test 2 or 3)
    returns multiple factors that are the same for a number that is a square of a prime (test 4 or 9)
    returns multiple factors that are the same for a number that is a cube of a prime (test 8 or 27)
    returns multiple prime factors for a value that is a product of primes and non-primes ( test 6 )
    returns multiple prime factors for a value that is a product of primes
    returns multiple prime factors for a value that is a product of primes even if the value is very large

This sounds much better, you're describing what you are testing without worrying about the specifics of the actual numbers used, and have identified some redundant tests that do not need to be among the cases. :+1:

Insti:
Although the descriptions could be made less verbose, so they can be more easily used as test names.

'no factors'
'prime number'
...
'product of primes'

There's a bit of a discussion about descriptions and test names happening here: https://github.com/exercism/x-common/issues/506

bunnymatic:
awesome.  thanks for the tips.

i'll get an updated version in soon

bunnymatic:
I can squish this down to 1 commit if need be, since the first commit is almost meaningless given the re-write.

exercises/prime-factors/canonical-data.json
```diff
@@ -0,0 +1,42 @@
...
+      },
+      {
+        "description" : "product of primes",
+        "input" : 901255,
+        "expected" : [5, 17, 23, 461]
+      },
+      {
+        "description" : "factors include a large prime",
+        "input" : 93819012551,
+        "expected" : [11, 9539, 894119]
```
Insti: caveat: I've not tried this problem myself recently. 

Are these numbers (93819012551, 901255) within the bounds of reasonableness (<1sec) for today's computers without having to rely on Project Euler style optimisation trickery?

Insti:
Squashing is good.

exercises/prime-factors/canonical-data.json
bunnymatic: I just completed this in elixir and ruby.  it is interesting because a naive solution in ruby will take a very long time to complete. But there is an efficient solution that does not take too long (less than 1 sec i believe).  I found it a good test to challenge the implementation.

bunnymatic:
squashed

exercises/prime-factors/canonical-data.json
Insti: Ok. Great :smile: Thanks.

Insti:
Thanks @bunnymatic :heart:

PR #514, opened 2017-01-26
book-store: shortened test descriptions

rpottsoh:
Descriptions have been shortened so as to make the json nicer for automatic test generators, so test case functions do not end up being too long.

### DIFF
--- a/exercises/book-store/canonical-data.json
+++ b/exercises/book-store/canonical-data.json
@@ -6,273 +6,82 @@
       "discount advantage for having more than one copy of ",
       "any single book in a grouping."
     ],
-    "cases": [
-      {
-        "description": "Only a single book",
-        "basket": [
-          1
-        ],
-        "targetgrouping": [
-          [
-            1
-          ]
-        ],
-        "expected": 8
-      },
-      {
-        "description": "Two of the same book",
-        "basket": [
-          2,
-          2
-        ],
-        "targetgrouping": [
-          [
-            2
-          ],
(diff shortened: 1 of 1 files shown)


kotp:
#470 mentioned shortening the names a few times... Looks like this was known, so happy to see it being addressed.

rpottsoh:
@kotp it was discussed in #470 but at the time it wasn't as high of a priority as just coming up with meaningful descriptions.  At the time I was not fully aware of how the canonical data was being used, I wasn't aware of automatic test generation at the time.

Insti:
The old array formatting without all those line breaks was nicer.
Having the expected values consistently as floats was better.
It's helpful to order the tests in a progressive fashion. (Empty basket should be the first test)

The shorter descriptions are a step in the right direction, but they still don't communicate *why* that is a useful thing to be testing.

> "Two each of first 3 books and 1 copy each of rest"

Why are we not testing "1 each of the first 3 books and 2 copies of books four and five"?

Insti:
Try and remember to include a link to the problem description when discussing canonical-data files:
https://github.com/exercism/x-common/blob/master/exercises/book-store/description.md

ErikSchierboom:
I'm with @robkeim here. The vertical grouping is significantly less readable to me. The descriptions themselves look good to me.

kotp:
Much better

ErikSchierboom:
Agreed. This looks great!

PR #517, opened 2017-01-28
beer-song: Add canonical data

bunnymatic:
Add canonical data for Beer Song.  

Problem Description: [here](https://github.com/exercism/x-common/blob/master/exercises/beer-song/description.md)

Closes: https://github.com/exercism/todo/issues/71


Insti:
The PR is missing a link to the problem description.

exercises/beer-song/canonical-data.json
```diff
@@ -0,0 +1,53 @@
...
+      "or nothing.  For the first case, in languages that include a range type ",
+      "structure, tests should use that structure.  In other languages, tests ",
+      "should take 2 values (start and stop) indicating the range boundaries."
+    ],
+    "description": "return multiple verses",
+    "cases": [
+      {
+        "description": "last 4 verses",
+        "start": 3,
+        "stop": 0,
```
Insti: A range is specified from a "beginning" to "end"

exercises/beer-song/canonical-data.json
```diff
@@ -0,0 +1,53 @@
...
+        "expected": "3 bottles of beer on the wall, 3 bottles of beer.\nTake one down and pass it around, 2 bottles of beer on the wall.\n\n2 bottles of beer on the wall, 2 bottles of beer.\nTake one down and pass it around, 1 bottle of beer on the wall.\n\n1 bottle of beer on the wall, 1 bottle of beer.\nTake it down and pass it around, no more bottles of beer on the wall.\n\nNo more bottles of beer on the wall, no more bottles of beer.\nGo to the store and buy some more, 99 bottles of beer on the wall.\n"
+      },
+      {
+        "description": "all verses",
```
Insti: The verse range should be specified here too.

exercises/beer-song/canonical-data.json
```diff
@@ -0,0 +1,53 @@
...
+        "expected": "No more bottles of beer on the wall, no more bottles of beer.\nGo to the store and buy some more, 99 bottles of beer on the wall.\n"
+      }
+    ]
+  },
+  "lyrics": {
+    "#": [
+      "Note:  the input to verses either takes the range of verses to report ",
+      "or nothing.  For the first case, in languages that include a range type ",
+      "structure, tests should use that structure.  In other languages, tests ",
+      "should take 2 values (start and stop) indicating the range boundaries."
```
Insti: I don't like this comment, it's dictating implementation details and and adding requirements not specified in the readme. Removing it completely would be good.

exercises/beer-song/canonical-data.json
bunnymatic: I was avoiding end because it's a keyword in some languages.  but I also wasn't happy with stop.

what about
`first_verse_number` and `last_verse_number`
?

exercises/beer-song/canonical-data.json
bunnymatic: The expectation here (from the problem) is that you would call either `lyrics(5,1)` or `lyrics(5..1)` for a language that supports a range or `lyrics()`.

So for the last case, do we set both range values to `null`?  Is that the *normal* way?  Does that need to get a comment in the `#` section?

exercises/beer-song/canonical-data.json
bunnymatic: fair 'nuf

exercises/beer-song/canonical-data.json
bunnymatic: Ok.   Elixir docs for their range use first/last (https://hexdocs.pm/elixir/Range.html).  Ruby uses begin/end (https://ruby-doc.org/core-2.2.0/Range.html) for two totally arbitrary examples.  I'll go with beginning and end.

exercises/beer-song/canonical-data.json
```diff
@@ -0,0 +1,47 @@
...
+    "description": "return multiple verses",
+    "cases": [
+      {
+        "description": "last 4 verses",
+        "beginning": 3,
+        "end": 0,
+        "expected": "3 bottles of beer on the wall, 3 bottles of beer.\nTake one down and pass it around, 2 bottles of beer on the wall.\n\n2 bottles of beer on the wall, 2 bottles of beer.\nTake one down and pass it around, 1 bottle of beer on the wall.\n\n1 bottle of beer on the wall, 1 bottle of beer.\nTake it down and pass it around, no more bottles of beer on the wall.\n\nNo more bottles of beer on the wall, no more bottles of beer.\nGo to the store and buy some more, 99 bottles of beer on the wall.\n"
+      },
+      {
+        "description": "all verses",
```
Insti: Missing `beginning` and `end` for this case.

exercises/beer-song/canonical-data.json
bunnymatic: I saw your not about the missing args, but did you see this discussion?  What should we specify for a method that doesn't take any args?

this is about the difference between `lyrics(5,4)` and `lyrics()`.

Do we spec begging and end as `null` ?  that doesn't seem like the same as no inputs to a function...

bunnymatic: I'll see if i can find some examples of this in other exercises.  I feel like there must be some others where we have initializers that optionally take values.

exercises/beer-song/canonical-data.json
bunnymatic: followed up in outdated discussion https://github.com/exercism/x-common/pull/517#discussion_r98341315

exercises/beer-song/canonical-data.json
bunnymatic: Here's an example - maybe not great - but https://github.com/exercism/x-common/blob/master/exercises/hello-world/canonical-data.json
With no name, there is no argument
With name, there is an argument
```
"cases": [
      {
         "description": "no name",
         "expected": "Hello, World!"
      },
      {
         "description": "sample name",
         "name": "Alice",
         "expected": "Hello, Alice!"
      },
   ...
```

exercises/beer-song/canonical-data.json
petertseng: At the risk of being met with the objection "but y'all did it in hello-world!"

What of the languages that do not support multiple calling conventions for `Verse` - those that require that `Verse` be called with two arguments, and do not allow for a function also `Verse` but taking zero arguments to be defined at the same time?

They may have to explicitly pass `99, 0`, or else have a differently-named function, such as `Song`. Does that affect the decision of what to do here, or not?

petertseng: Sigh, "calling conventions" is a reserved word for how arguments get passed in registers / memory, at the computer architecture level. I'm editing my message and changing "calling conventions" to "function overloads / variadic functions"

bunnymatic: I'll take a look at the language examples.  Maybe something there can give a clue.  

I did a cursory survey of the exercises that I've completed and I think this and Hello-World maybe the only ones that had this structure (overloaded constructors that take args or take nothing and use default args).  If we are trying to cover *all* languages, maybe putting in the defaults as args in this canonical-data file is right.  Dunno if it also should get a note that says something like "if the language track being implemented supports function overloads/variadic functions, then ... <something i can't word well but is like use those things as defaults>".

bunnymatic: I found the C example set exercises and they don't have specs for all lyrics, but they would have to have beginning and end specified in the function call.

I've come full circle and think that adding the default args is the right thing.
I've added these changes to the data file (and rebased and squashed).

ErikSchierboom:
@petertseng I think the array approach is slightly easier to view. But I don't necessarily feel that this PR should be updated accordingly.

petertseng:
Looks like y'all don't mind, so we merge as-is.

Thanks!

PR #518, opened 2017-01-28
perfect-numbers: description clarification

rootulp:
- add links to aliquot sum and Nicomachus
- provide examples of perfect, abundant, and deficient numbers
- provide instruction statement at bottom
- references #447


ErikSchierboom:
It's a great addition, although I think it would be slightly better if the added aliquot sum explanation was moved to just after "based on their aliquot sum.", which is just before their examples. That makes more sense to me.

rootulp:
@ErikSchierboom great feedback! I reordered the text following your suggestion.

ErikSchierboom:
@rootulp That's quick! LGTM!

ErikSchierboom:
Merged. Thanks @rootulp! 🎉

PR #519, opened 2017-01-30
RFC: hello-world: remove reference to "name not given"

petertseng:
This implements the second option "change the hello-world README to be
so vague..." of #520.

As stated in #520, the point of hello-world is to give a very easy
exercise. This helps give an intro to how to fetch, how to submit, and
how to run the test suite for a given language. The README included with
hello-world helps out with that.

The clause "If a name is not given" is very difficult for languages that
do not support overloading functions and/or default function parameters.

This is against the vision of a very easy hello-world. The ideal is a
function that returns "Hello, World!" unconditionally.

By not mentioning the "name not given" in the README, every track that
currently has a hello-world exercise can change it or not, as they
please.

This is a very delicate balance to strike. If we make it too vague, that
is also undesirable. See #322. I cannot be sure that I have gotten it
right in this commit.

### DIFF
--- a/exercises/hello-world/canonical-data.json
+++ b/exercises/hello-world/canonical-data.json
@@ -1,14 +1,5 @@
 {
-   "#": [
-     "Language implementations vary on the issue of missing input variables.",
-     "In this set of test cases, the input variable is left out of the definition,",
-     "and the generator would need to handle the language implementation."
-   ],
    "cases": [
-      {
-         "description": "no name",
-         "expected": "Hello, World!"
-      },
       {
          "description": "sample name",
          "name": "Alice",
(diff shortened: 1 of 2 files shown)


petertseng:
There is too much going on in this PR. I'll extract some of it in an issue, and make it applicable to all tracks, because as it stands everyone can fairly think that I'm imagining this problem and the problem has nothing to do with them, so it is not worth the time to read this PR.

rbasso:
👍 for option 1:  Change `hello-world` to be even simpler.

The purpose of a [Hello, World!](https://en.wikipedia.org/wiki/%22Hello,_World!%22_program) exercise seems to be, IMHO, just testing the tool-chain. Anything beyond returning/outputting the string makes it less suitable to that purpose.

Insti:
I don't think this is an ideal solution to the main problem with hello world.  See #520

petertseng:
Let's close to avoid accidental merges while we examine our options in #520.

PR #521, opened 2017-01-30
List Ops: add canonical data

bunnymatic:
Problem description [here](https://github.com/exercism/x-common/blob/master/exercises/list-ops/description.md) - albeit brief.

This is a first cut submit.  It most certainly needs revision.  If we can answer the following questions, I think I can make a second revision that should be pretty good.

Things to think about:
* Is the function representation clear enough? - see this discussed here: https://github.com/exercism/todo/issues/152#issuecomment-275889363
* Do we need to represent `foldl` and `foldr` in tests or is `reduce` sufficent?  After looking through 8 test suites from the existing implementations I found the following
* Many suites cover very large lists (0..100000) but that seems unreasonable to write into a json file... `"list": [0,1,2,3,4... lots more ... 99999, 100000]` ... so do those tests not show up here?  or do we include a comment (which i've done)?

| method | freq in sampled test suites |
| ------- | -----------------------|
| append | 6 |
| concat | 8|
| filter  | 7 |
| length/count | 8 |
| map | 8 |
| reverse | 8 |
| foldl & foldr | 5 |
| reduce | 3 |

Closes: https://github.com/exercism/todo/issues/109

### DIFF
--- a/exercises/list-ops/canonical-data.json
+++ b/exercises/list-ops/canonical-data.json
@@ -1,0 +1,147 @@
+{
+  "#": [
+    "Though there are no specifications here for dealing with large lists,",
+    "implementers may add tests for handling large lists to ensure that the",
+    "solutions have thought about performance concerns."
+  ],
+  "append": {
+    "description": "append entries to a list and return the new list",
+    "cases": [
+      {
+        "description": "empty lists",
+        "list1": [],
+        "list2": [],
+        "expected": []
+      },
+      {
+        "description": "empty list to list",
+        "list1": [],
+        "list2": [1,2,3,4],
+        "expected": [1,2,3,4]
+      },
+      {
+        "description": "non-empty lists",
+        "list1": [1,2],
+        "list2": [2,3,4,5],
+        "expected": [1,2,2,3,4,5]
+      }
+    ]
+  },
+  "concat": {
+    "description": "concat lists and lists of list into a new list",
+    "cases": [
+      {
+        "description": "empty list",
+        "lists": [],
+        "expected": []
+      },
+      {
+        "description": "list of lists",
+        "lists": [[1,2],[3],[],[4,5,6]],
+        "expected": [1,2,3,4,5,6]
+      }
+    ]
+  },
+  "filter": {
+    "description": "filter list returning only values that satisfy the filter function",
+    "cases": [
+      {
+        "description": "empty list",
+        "list": [],
+        "function": "value modulo 2 == 1",
+        "expected": []
+      },
+      {
+        "description": "non-empty list",
+        "list": [1,2,3,5],
+        "function": "value modulo 2 == 1",
+        "expected": [1,3,5]
+      }
+    ]
+  },
+  "length": {
+    "description": "returns the length of a list",
+    "cases": [
+      {
+        "description": "empty list",
+        "list": [],
+        "expected": 0
+      },
+      {
+        "description": "non-empty list",
+        "list": [1,2,3,4],
+        "expected": 4
+      }
+    ]
+  },
+  "map": {
+    "description": "return a list of elements whos values equal the list value transformed by the mapping function",
+    "cases": [
+      {
+        "description": "empty list",
+        "list": [],
+        "function": "value + 1",
+        "expected": []
+      },
+      {
+        "description": "non-empty list",
+        "list": [1,3,5,7],
+        "function": "value + 1",
+        "expected": [2,4,6,8]
+      }
+    ]
+  },
+  "foldl": {
+    "description": "folds (reduces) the given list from the left with a function",
+    "cases": [
+      {
+        "description": "empty list",
+        "list": [],
+        "initial": 2,
+        "function": "value / accumulator",
+        "expected": 2
+      },
+      {
+        "description": "division of integers",
+        "list": [1,2,3,4],
+        "initial": 24,
+        "function": "value / accumulator",
+        "expected": 64
+      }
+    ]
+  },
+  "foldr": {
+    "description": "folds (reduces) the given list from the right with a function",
+    "cases": [
+      {
+        "description": "empty list",
+        "list": [],
+        "initial": 2,
+        "function": "value / accumulator",
+        "expected": 2
+      },
+      {
+        "description": "division of integers",
+        "list": [1,2,3,4],
+        "initial": 24,
+        "function": "value / accumulator",
+        "expected": 9
+      }
+    ]
+  },
+  "reverse": {
+    "description": "reverse the elements of the list",
+    "cases": [
+      {
+        "description": "empty list",
+        "list": [],
+        "expected": []
+      },
+      {
+        "description": "non-empty list",
+        "list": [1,3,5,7],
+        "expected": [7,5,3,1]
+      }
+    ]
+  }
+}


NobbZ:
In my opinion, there should be a clear distinction between left and right folds:

```exs
ListOps.foldl &(&1 / &2), 1, [1,2,3,4]
#=> 2.6666666666666665

ListOps.foldr &(&1 / &2), 1, [1,2,3,4]
#=> 0.375
```

With `reduce` only, one had only one version of those, and probably it is exactly that version that I do NOT need right now.

Insti:
Suggestion:  The '*_function' json keys should all be just called 'function'

exercises/list-ops/canonical-data.json
```diff
@@ -71,40 +71,57 @@
...
       }
     ]
   },
-  "reduce": {
-    "description": "combines list elements by applying reducer_function specified",
+  "foldl": {
+    "description": "folds (reduces) the given list from the left with a function",
+    "cases": [
+      {
+        "description": "empty list",
```
stevejb71: Both folds need an "initial" parameter to supply the first input to the fold. The initial in this case would be 0.

exercises/list-ops/canonical-data.json
```diff
@@ -0,0 +1,143 @@
...
+  "reverse": {
+    "description": "reverse the elements of the list",
+    "cases": [
+      {
+        "description": "empty list",
+        "list": [],
+        "expected": []
+      },
+      {
+        "description": "normal list",
```
stevejb71: Maybe non-empty list? An empty list is perfectly normal!

exercises/list-ops/canonical-data.json
```diff
@@ -0,0 +1,143 @@
...
+  "length": {
+    "description": "returns the length of a list",
+    "cases": [
+      {
+        "description": "empty list",
+        "list": [],
+        "expected": 0
+      },
+      {
+        "description": "simple list",
```
stevejb71: Many non-empty instead of simple?

stevejb71: *maybe

bunnymatic: yeah - i'm good with non-empty.  Normal and simple are both pretty vague.

exercises/list-ops/canonical-data.json
bunnymatic: indeed it is.  my bad.

bunnymatic:
Are there any thoughts on handling large lists?

> Many suites cover very large lists (0..100000) but that seems unreasonable to write into a json file... "list": [0,1,2,3,4... lots more ... 99999, 100000] ... so do those tests not show up here? or do we include a comment (which i've done)?

bunnymatic:
Also... just rebased on master and squished the commits. FYI

exercises/list-ops/canonical-data.json
```diff
@@ -0,0 +1,147 @@
...
+        "description": "empty list to list",
+        "list1": [],
+        "list2": [1,2,3,4],
+        "expected": [1,2,3,4]
+      },
+      {
+        "description": "non-empty lists",
+        "list1": [1,2],
+        "list2": [2,3,4,5],
+        "expected": [1,2,3,4,5]
```
petertseng: hmm, given that the two lists we are appending have two `2` between them, are sure this expected list should have only one `2`? I would agree if this were a set we were talking about, but it's a list here.

bunnymatic: good catch.  I think at the last minute i decided that having a both arrays be larger than 1 element seemed like it would enhance the test a bit but I only updated `list1` and not the `expected`

exercises/list-ops/canonical-data.json
```diff
@@ -0,0 +1,147 @@
...
+      }
+    ]
+  },
+  "foldl": {
+    "description": "folds (reduces) the given list from the left with a function",
+    "cases": [
+      {
+        "description": "empty list",
+        "list": [],
+        "initial": "2",
```
petertseng: would the string 2 be confusing here? what of the number 2?

exercises/list-ops/canonical-data.json
```diff
@@ -0,0 +1,147 @@
...
+        "function": "value / accumulator",
+        "expected": 2
+      },
+      {
+        "description": "division of integers",
+        "list": [1,3,5,7],
+        "initial": "2",
```
petertseng: same, number 2?

exercises/list-ops/canonical-data.json
```diff
@@ -0,0 +1,147 @@
...
+      }
+    ]
+  },
+  "foldr": {
+    "description": "folds (reduces) the given list from the right with a function",
+    "cases": [
+      {
+        "description": "empty list",
+        "list": [],
+        "initial": "2",
```
petertseng: same, number 2?

exercises/list-ops/canonical-data.json
```diff
@@ -0,0 +1,147 @@
...
+        "function": "value / accumulator",
+        "expected": 2
+      },
+      {
+        "description": "division of integers",
+        "list": [1,3,5,7],
+        "initial": "2",
```
petertseng: same, number 2?

exercises/list-ops/canonical-data.json
```diff
@@ -0,0 +1,147 @@
...
+    "description": "concat lists and lists of list into a new list",
+    "cases": [
+      {
+        "description": "empty list",
+        "lists": [],
+        "expected": []
+      },
+      {
+        "description": "list of lists",
+        "lists": [[1,2],[3],[],[4,[5,6]]],
```
petertseng: notice that the `[4, [5, 6]]` is a array of either ints or array of ints. This will require some unfortunate contortions in languages where arrays (or the analogous type) may only contain values of only one type. In general, one might think concatenate would take an `[[a]]` and return an `[a]`. If the inner array may be arbitrarily nested, given the expected result, this operation is `concat + flatten`, not just concat - is this desirable? Is there, indeed, a language where the operation named `concat` does also flatten (and can you tell me which)?

I would recommend not expecting this function to flatten.

petertseng: Er, I would also recommend not using arrays of heterogeneous types

exercises/list-ops/canonical-data.json
```diff
@@ -0,0 +1,147 @@
...
+        "initial": "2",
+        "function": "value / accumulator",
+        "expected": 2
+      },
+      {
+        "description": "division of integers",
+        "list": [1,3,5,7],
+        "initial": "2",
+        "function": "value / accumulator",
+        "expected": 0.47619047619047616
```
stevejb71: Could you choose your numbers so the result comes to an integer. The type of number in the list looks like an int (some languages are strict about numeric conversions) and there may be a rounding error to deal with for floats.

bunnymatic: yep

exercises/list-ops/canonical-data.json
bunnymatic: That's funny.  I also didn't expect it to be a flatten operation either but the "examples" that i looked at in a couple languages included that.  

So the actual test here should be something more like
`[ [1,2,3], [], [4, 5] ] = [[1,2,3,4,5]]`
right?

This is "concat all the lists you handed me in this list of lists but don't flatten them."

petertseng: > `expect( concat([ [1,2,3], [], [4, 5] ]) ) to equal [[1,2,3,4,5]]`

I would expect the answer to be `[1, 2, 3, 4, 5]` rather than `[[1, 2, 3, 4 ,5]]` - the function took a list of list of integers and turned it into a list of integers.

Or if we want to use an existing track's test as a model (don't know whether any other tracks have this exact one) https://github.com/exercism/xhaskell/blob/master/exercises/list-ops/test/Tests.hs#L112 - `[ [1, 2], [3], [], [4, 5, 6]]` and `[1, 2, 3, 4, 5, 6]`

bunnymatic: I looked again.  Elixir has the same test.  Ruby has nothing like `concat` in the test suite.  I'm good with your example and expectation.  

After seeing that, I think *I* might have added that inner array to sort of harden the example.  Clearly that was a bad idea ;)

Hopefully this latest rev will fly.

exercises/list-ops/canonical-data.json
bunnymatic: added initial params.

stevejb71:
Haskell, say, would have to have floating point numbers in the list. Probably not a big deal.

petertseng:
Thanks!

PR #522, opened 2017-01-30
luhn: Add more tests

gibfahn:
The luhn test suite didn't cover a lot of cases, including:
 - Strings of zeros
 - Non-alphabetic characters
 - Numbers which are only valid if you reverse the string

This updates the test suite to cover these.

Let me know if I'm doing anything wrong.

Refs: https://github.com/exercism/xrust/pull/253

### DIFF
--- a/exercises/luhn/canonical-data.json
+++ b/exercises/luhn/canonical-data.json
@@ -11,6 +11,11 @@
       "expected": false
     },
     {
+      "description": "simple valid sin",
+      "input": " 5 9 ",
+      "expected": true
+    },
+    {
       "description": "valid Canadian SIN",
       "input": "046 454 286",
       "expected": true
@@ -29,6 +34,36 @@
       "description": "valid strings with a non-digit added become invalid",
       "input": "046a 454 286",
       "expected": false
+    },
+    {
+      "description": "punctuation is not allowed",
+      "input": "055-444-285",
+      "expected": false
+    },
+    {
+      "description": "symbols are not allowed",
+      "input": "055£ 444$ 285",
+      "expected": false
+    },
+    {
+      "description": "single zero with space is invalid",
+      "input": " 0",
+      "expected": false
+    },
+    {
+      "description": "lots of zeros are valid",
+      "input": " 00000",
+      "expected": true
+    },
+    {
+      "description": "another valid sin",
+      "input": "055 444 285",
+      "expected": true
+    },
+    {
+      "description": "nine doubled is nine",
+      "input": "091",
+      "expected": true
     }
   ]
 }


Insti:
Thanks @gibfahn 

Your suggestions have caused me to identify something I think is an issue with this problem #523 I don't think the string parsing tests belong in this problem.

Insti:
>   Strings of zeros

Why are strings of zeros invalid?

>   Numbers which are only valid if you reverse the string

What is this testing?

gibfahn:
>Why are strings of zeros invalid?

Looking at the README I somehow got the impression that they were invalid. I haven't found it clearly stated anywhere, but I'm willing to believe 0 is a valid answer.

>Numbers which are only valid if you reverse the string

The formula works by reversing the sequence of numbers and going through them doubling every second digit. The current set of answers all pass even if you forget to reverse the sequence.


So I guess zero strings should pass. I'll update.

stevejb71:
@gibfahn "but I'm willing to believe 0 is a valid answer" 

The readme states that strings of length 1 or less are invalid.

gibfahn:
>>>Why are strings of zeros invalid?
>>Looking at the README I somehow got the impression that they were invalid. I haven't found it clearly stated anywhere, but I'm willing to believe 0 is a valid answer.

>The readme states that strings of length 1 or less are invalid.

0 as in the result of the formula is zero, talking about strings of zeros (i.e. `0000`) where the calculation ends up being `0 mod 10 == 0`.

Insti:
Having tests for both valid and invalid zero strings would be useful.

`000` valid
`0000` invalid - missing check digit.

NobbZ:
Why are 4 zeros missing the checkdigit? I take all but the last one, sum
them up and get a sum of 0, now I substract 0 from 10 and get 10. 10 % 10
is 0, so my checkdigit is 0, which again is equal to what I took away
before summing up. so 0000 is valid as well.

Geoff Hubbard <notifications@github.com> schrieb am Di., 31. Jan. 2017 um
10:18 Uhr:

gibfahn:
My (current) interpretation is that a single zero is invalid, but multiple zeros are always valid.

exercises/luhn/canonical-data.json
```diff
@@ -29,6 +29,46 @@
...
+    },
+    {
+      "description": "single zero with space is invalid",
+      "input": " 0",
+      "expected": false
+    },
+    {
+      "description": "lots of zeros are invalid",
+      "input": " 00000",
+      "expected": false
```
petertseng: don't forget to change this one (I think we want this one to be valid, right?)

exercises/luhn/canonical-data.json
```diff
@@ -29,6 +29,46 @@
...
+      "input": "091",
+      "expected": true
+    },
+    {
+      "description": "must double from right",
+      "input": "0091",
+      "expected": true
+    },
+    {
+      "description": "simple valid sin",
```
petertseng: probably worth moving this up such that it becomes the first true case - let the student solve the simplest case first.

petertseng: also another note that this is another case that will reveal doubling from the wrong side - if doubling the wrong digits here we will get 14 instead of 10 and falsely think this is invalid.

This could possibly make the "must double from right" case redundant.

gibfahn: Makes sense, I've moved this up and removed the "must double from right" case

exercises/luhn/canonical-data.json
gibfahn: Done

exercises/luhn/canonical-data.json
```diff
@@ -29,6 +34,36 @@
...
+      "input": " 00000",
+      "expected": true
+    },
+    {
+      "description": "another valid sin",
+      "input": "055 444 285",
+      "expected": true
+    },
+    {
+      "description": "nine doubled is nine",
```
petertseng: in case anyone wonder, this caught a bug in mine (I was using mod 9 after doubling, which turns nine into zero instead of nine)

exercises/luhn/canonical-data.json
```diff
@@ -29,6 +34,36 @@
...
+      "expected": false
+    },
+    {
+      "description": "lots of zeros are valid",
+      "input": " 00000",
+      "expected": true
+    },
+    {
+      "description": "another valid sin",
+      "input": "055 444 285",
```
petertseng: can you confirm - what does this test that above cases do not? is it just here to contrast with `055-444-285`? What is an example of an implementation that would pass the above cases but fail this one?

gibfahn: Part of it was the contrast with `0555-444-285`, and part of it was the thought that having a couple of different longer valid numbers might help someone who was struggling to work out how the algorithm needed to be implemented.

I can remove it if necessary.

exercises/luhn/canonical-data.json
```diff
@@ -29,6 +34,36 @@
...
       "expected": false
+    },
+    {
+      "description": "punctuation is not allowed",
+      "input": "055-444-285",
+      "expected": false
+    },
+    {
+      "description": "symbols are not allowed",
+      "input": "055£ 444$ 285",
```
petertseng: can I get a confirmation on whether the £ is in ASCII? I am seeing bytes 0xc2 0xa3 for it

Insti: It is not.

PR #527, opened 2017-01-30
multiple problems: Remove non-ASCII characters to fix #441

### LINKED ISSUE #441: Update or remove non-ASCII test cases.

robkeim:
#441

### DIFF
--- a/exercises/atbash-cipher/canonical-data.json
+++ b/exercises/atbash-cipher/canonical-data.json
@@ -46,11 +46,6 @@
                 "description": "encode all the letters",
                 "phrase": "The quick brown fox jumps over the lazy dog.",
                 "expected": "gsvjf rxpyi ldmul cqfnk hlevi gsvoz abwlt"
-            },
-            {
-                "description": "encode ignores non ascii",
-                "phrase": "non ascii éignored",
-                "expected": "mlmzh xrrrt mlivw"
             }
         ]
     },
--- a/exercises/forth/canonical-data.json
+++ b/exercises/forth/canonical-data.json
@@ -17,13 +17,6 @@
           "1 2 3 4 5"
         ],
         "expected": [1, 2, 3, 4, 5]
-      },
-      {
-        "description": "all non-word characters are separators",
-        "input": [
-          "1\u00002\u00133\n4\r5 6\t7"
-        ],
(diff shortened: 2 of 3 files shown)


Insti:
I'm not sure putting changes to multiple problems in the same commit is a good idea.

Are you sure that these tests are not also testing another aspect of the solution?

robkeim:
@Insti should I create three different PRs then?

I looked through the tests and given the titles they are explicitly testing non-ASCII functionality. I don't have any way of validating if the original intent of the test was to cover two different things though.

Insti:
3 commits would be good.

It's good to ask yourself, "Might I need to revert part of this?".

robkeim:
Ok sounds good @Insti

I've got a couple of follow-up questions:
- What is the best way to update this to 3 separate commits? Do I simply abandon this PR and start a new one, or is there some ninja git trick I'm missing?
- Don't we do a "squash and merge" when merging into master? In that case, the intermediate commits are going to be lost right?

petertseng:
> What is the best way to update this to 3 separate commits?

The way I like to do this is `git reset origin/master`, then use `git add -p` to add each change separately, with `git commit` for each commit. You will need to `push --force` when done.

> Don't we do a "squash and merge" when merging into master?

I would recommend not doing so if there are separate commits that are meant to stay separate. Only squash and merge if there is one commit in the PR, or if the commits in the PR are meant to be squashed.

robkeim:
Thanks @petertseng for the tip, although the forced push didn't go well so I had to delete the remote branch and start again :(

The new PR is #529

PR #528, opened 2017-01-30
Removes "write a program" from descriptions to address #321

robkeim:
#321

### DIFF
--- a/exercises/anagram/metadata.yml
+++ b/exercises/anagram/metadata.yml
@@ -1,4 +1,4 @@
 ---
-blurb: "Write a program that, given a word and a list of possible anagrams, selects the correct sublist."
+blurb: "Given a word and a list of possible anagrams, select the correct sublist."
 source: "Inspired by the Extreme Startup game"
 source_url: "https://github.com/rchatley/extreme_startup"
--- a/exercises/binary-search/metadata.yml
+++ b/exercises/binary-search/metadata.yml
@@ -1,4 +1,4 @@
 ---
-blurb: "Write a program that implements a binary search algorithm."
+blurb: "Implement a binary search algorithm."
 source: "Wikipedia"
 source_url: "http://en.wikipedia.org/wiki/Binary_search_algorithm"
--- a/exercises/scale-generator/description.md
+++ b/exercises/scale-generator/description.md
@@ -1,7 +1,6 @@
-Write a program that can generate musical scales. It should accept
-a tonic, or starting note, and a set of intervals. It should be able to
-output the notes in the given scale, starting with the tonic and
-following the specified interval pattern.
+Given a tonic, or starting note, and a set of intervals, generate
+the musical scale starting with the tonic and following the
+specified interval pattern.
 
 Scales in Western music are based on the chromatic (12-note) scale.This
 scale can be expressed as the following group of pitches:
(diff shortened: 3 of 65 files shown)


kotp:
I like it.  It is all the same thing though, so could be one commit.

robkeim:
Thanks @kotp!

I would have done one commit but from [this PR](https://github.com/exercism/x-common/pull/527) it seems we favor commits not crossing exercises. Whoever merges it can squash if they only want to keep one commit.

kotp:
That PR and the suggestion for 3 different commits was for a different reason.  The change here is one problem, and does not have the potential to effect like the other could.

petertseng:
Yeah I'd have to think about that one. I don't *think* I'd ever revert part of this PR but not the others, but I certainly won't complain about them being separate.

could you amend some commit messages? cryto -> crypto, and serach -> search.

petertseng:
kindergarten-garten -> kindergarten-garden

petertseng:
otcal -> octal

exercises/scale-generator/description.md
```diff
@@ -1,7 +1,7 @@
-Write a program that can generate musical scales. It should accept
-a tonic, or starting note, and a set of intervals. It should be able to
-output the notes in the given scale, starting with the tonic and
-following the specified interval pattern.
+Generate musical scales accepting a tonic, or starting note, and
+a set of intervals. It should be able to output the notes in the
```
petertseng: Previously, "it" referred to the program, and now I'm not sure "it" makes sense any longer. Maybe it will make sense by simply removing the words "It should"?

exercises/anagram/metadata.yml
```diff
@@ -1,4 +1,4 @@
 ---
-blurb: "Write a program that, given a word and a list of possible anagrams, selects the correct sublist."
+blurb: "Given a word and a list of possible anagrams, selects the correct sublist."
```
petertseng: probably selects -> select

exercises/binary-search/metadata.yml
```diff
@@ -1,4 +1,4 @@
 ---
-blurb: "Write a program that implements a binary search algorithm."
+blurb: "Implements a binary search algorithm."
```
petertseng: Probably "Implement" here, so that the verb is consistent with other blurbs.

robkeim:
Thanks for the review @petertseng. I reviewed the changes before submitting but not the commit messages. I'll update them this evening.

exercises/twelve-days/metadata.yml
```diff
@@ -1,4 +1,4 @@
 ---
-blurb: "Write a program that outputs the lyrics to 'The Twelve Days of Christmas'"
+blurb: "Outputs the lyrics to 'The Twelve Days of Christmas'"
```
petertseng: Probably "Output" here, to match verbs.

kotp:
@petertseng is right, the same concern.  I would not expect a portion of these to want to be reverted.  It isn't a big deal about the individual commits, except a lot of short-interval work.  And as noted, lots of opportunity for small errors in typing.

exercises/scale-generator/description.md
robkeim: I decided to reword the whole sentence to the "Given a xxx" syntax.

robkeim:
@petertseng I went ahead and updated the PR with your comments, thanks again for the review

@kotp I left the individual commits for now, but I could squash them with the merge, I don't have a preference.

kotp:
Either way is good for me.  With a squash, the commit will read really well, itemized.

petertseng:
Somehow, we lost the `change` commit, so that now the exercise's description again reads "Write a program"

also, can we get grade-school too? It says "Write a small archiving program"

robkeim:
Thanks @petertseng for all of the valuable feedback!

It looks like the Travis CI builds are backed up, so I'll merge this once the build completes.

PR #532, opened 2017-02-03
nucleotide-codons: deprecate in favour of protein-translation

### LINKED ISSUE #268: The protein-translation and nucleotide-codons exercises are the same
If you look at [the](https://github.com/exercism/x-common/blob/master/nucleotide-codons.md) [docs](https://github.com/exercism/x-common/blob/master/protein-translation.md), you'll find that both these exercises do the exact same thing: they translate codons to there correct names.

One of these exercise should probably be deprecated. Personally, I think the nucleotide-codons is the one that should be deprecated, as its description is to me far less clear than that of the protein-translation exercise.

petertseng:
Closes #268

### DIFF
--- a/exercises/nucleotide-codons/.deprecated
+++ b/exercises/nucleotide-codons/.deprecated
@@ -1,0 +1,3 @@
+**NOTE: This exercise has been deprecated**
+Your track should implement protein-translation instead.
+https://github.com/exercism/x-common/issues/268


ErikSchierboom:
If I look at other [deprecated exercises](https://github.com/exercism/x-common/search?utf8=%E2%9C%93&q=deprecated), they almost all start with "**NOTE: This exercise has been deprecated.**". I like that this is the first thing you see in that case, which makes it really clear that the exercise is deprecated. Perhaps we can do that here too?

petertseng:
Let's try that, then. Since you didn't object to linking the issue nor the suggestion of what exercise to implement instead, I kept them, of course, just added the necessary line

petertseng:
Someone needs to remind me what this .deprecated file does again, whether this PR needs to be dependent on https://github.com/exercism/xrust/pull/258.

petertseng:
Okay far as I can tell from https://github.com/exercism/x-api/pull/140 , deprecating it globally just prevents it from showing up in the "to be implemented" section of any other track, so I don't think https://github.com/exercism/xrust/pull/258 and this PR need to be temporally dependent on each other in some way.

ErikSchierboom:
I think linking to the issue and the suggestion are great features, so I'm glad you kept them in.

That means this PR can be merged, right?

petertseng:
> That means this PR can be merged, right?

I'm ready for it to happen, and I *think* it is safe even while the Rust PR hasn't been merged. If we are unsure at all, we better let @kytrinyx know before the daily site deploy, in case it does cause any unexpected effects.

ErikSchierboom:
@kytrinyx would merging this PR would be safe to do?

kytrinyx:
Yes, this is safe to merge.

PR #533, opened 2017-02-03
food-chain: remove repeated blurb text from README

jdsutherland:
Previously, the README contained duplicate text of the blub from [metadata.yml](https://github.com/exercism/x-common/blob/master/exercises/food-chain/metadata.yml#L2):
image

This fix removes the identical blub text from the README.


petertseng:
Thanks!

PR #534, opened 2017-02-03
added rotational-cipher exercise

LogoiLab:
I tried to add this to xrust. Realized I needed to add it here first.

### DIFF
--- a/exercises/rotational-cipher/cannonical-data.json
+++ b/exercises/rotational-cipher/cannonical-data.json
@@ -1,0 +1,70 @@
+{
+    "#": [
+        "The tests are a series of rotation tests: "
+    ],
+    "rotate": {
+        "description": ["Test rotation from English to ROTn"],
+        "cases": [
+            {
+                "description": "rotate a by 1",
+                "text": "a",
+                "shift_key": 1,
+                "expected": "b"
+            },
+            {
+                "description": "rotate a by 26, same output as input",
+                "text": "a",
+                "shift_key": 26,
+                "expected": "a"
+            },
+            {
+                "description": "rotate a by 0, same output as input",
+                "text": "a",
+                "shift_key": 26,
+                "expected": "a"
+            },
+            {
+                "description": "rotate m by 13",
+                "text": "m",
+                "shift_key": 13,
+                "expected": "z"
+            },
+            {
+                "description": "rotate n by 13 with wrap around alphabet",
+                "text": "n",
+                "shift_key": 13,
+                "expected": "a"
+            },
+            {
+                "description": "rotate capital letters",
+                "text": "OMG",
+                "shift_key": 5,
+                "expected": "TRL"
+            },
+            {
+                "description": "rotate spaces",
+                "text": "O M G",
+                "shift_key": 5,
+                "expected": "T R L"
+            },
+            {
+                "description": "rotate numbers",
+                "text": "Testing 1 2 3 testing",
+                "shift_key": 4,
+                "expected": "Xiwxmrk 1 2 3 xiwxmrk"
+            },
+            {
+                "description": "rotate punctuation",
+                "text": "Let's eat, Grandma!",
+                "shift_key": 21,
+                "expected": "Gzo'n zvo, Bmviyhv!"
+            },
+            {
+                "description": "rotate all letters",
+                "text": "The quick brown fox jumps over the lazy dog.",
+                "shift_key": 13,
+                "expected": "Gur dhvpx oebja sbk whzcf bire gur ynml qbt."
+            }
+        ]
+    }
+}
--- a/exercises/rotational-cipher/description.md
+++ b/exercises/rotational-cipher/description.md
@@ -1,0 +1,30 @@
+# Rotational Cipher
+
+Create an implementation of the rotational cipher, also sometimes called the Caesar cipher.
+
+The Caesar cipher is a simple shift cipher that relies on
+transposing all the letters in the alphabet using an integer key
+between `0` and `26`. Using a key of `0` or `26` will always yield
+the same output due to modular arithmetic. The letter is shifted
+for as many values as the value of the key.
+
+The general notation for rotational ciphers is `ROT + <key>`.
+The most commonly used rotational cipher is `ROT13`.
+
+A `ROT13` on the Latin alphabet would be as follows:
+
+```plain
+Plain:  abcdefghijklmnopqrstuvwxyz
+Cipher: nopqrstuvwxyzabcdefghijklm
+```
+
+It is stronger than the Atbash cipher because it has 27 possible keys, and 25 usable keys.
+
+Ciphertext is written out in the same formatting as the input including spaces and punctuation.
+
+## Examples
+- ROT5  `omg` gives `trl`
+- ROT0  `c` gives `c`
+- ROT26 `Cool` gives `Cool`
+- ROT13 `The quick brown fox jumps over the lazy dog.` gives `Gur dhvpx oebja sbk whzcf bire gur ynml qbt.`
+- ROT13 `Gur dhvpx oebja sbk whzcf bire gur ynml qbt.` gives `The quick brown fox jumps over the lazy dog.`
--- a/exercises/rotational-cipher/metadata.yml
+++ b/exercises/rotational-cipher/metadata.yml
@@ -1,0 +1,4 @@
+---
+blurb: "Create an implementation of the rotational cipher, also sometimes called the Caesar cipher."
+source: "Wikipedia"
+source_url: "https://en.wikipedia.org/wiki/Caesar_cipher"


exercises/rotational-cipher/cannonical-data.json
```diff
@@ -0,0 +1,90 @@
...
+        "description": ["Test rotation from English to ROTn"],
+        "cases": [
+            {
+                "description": "rotate a by 1",
+                "text": "a",
+                "shift_key": 1,
+                "expected": "b"
+            },
+            {
+                "description": "rotate a by 26",
```
petertseng: it would probably be worth calling out in the description that this is expected to produce the same output as input.

exercises/rotational-cipher/cannonical-data.json
```diff
@@ -0,0 +1,90 @@
...
+                "text": "a",
+                "shift_key": 26,
+                "expected": "a"
+            },
+            {
+                "description": "rotate a by 0",
```
petertseng: > it would probably be worth calling out in the description that this is expected to produce the same output as input.

same

exercises/rotational-cipher/cannonical-data.json
```diff
@@ -0,0 +1,90 @@
...
+                "expected": "a"
+            },
+            {
+                "description": "rotate m by 13",
+                "text": "m",
+                "shift_key": 13,
+                "expected": "z"
+            },
+            {
+                "description": "rotate n by 13",
```
petertseng: this is the first non-trivial (trivial being any number divisible by 26) case where we wrap around the alphabet. I think that is worth pointing out in description.

exercises/rotational-cipher/cannonical-data.json
```diff
@@ -0,0 +1,90 @@
...
+            },
+            {
+                "description": "rotate punctuation",
+                "text": "Let's eat, Grandma!",
+                "shift_key": 21,
+                "expected": "Gzo'n zvo, Bmviyhv!"
+            },
+            {
+                "description": "rotate mindblowingly",
+                "text": "mindblowingly",
```
petertseng: explain - what does this case test that a previous case does not?

What implementations might pass all previous cases, but not this one?

exercises/rotational-cipher/cannonical-data.json
```diff
@@ -0,0 +1,90 @@
...
+                "shift_key": 10,
+                "expected": "wsxnlvygsxqvi"
+            },
+            {
+                "description": "rotate deep thought",
+                "text": "Truth is fiction.",
```
petertseng: > explain - what does this case test that a previous case does not?

same

exercises/rotational-cipher/cannonical-data.json
```diff
@@ -0,0 +1,90 @@
...
+                "shift_key": 16,
+                "expected": "Jhkjx yi vysjyed."
+            },
+            {
+                "description": "rotate all letters",
```
petertseng: compared to the above two, I can easily see the purpose of this case - it rotates all the letters. So I clearly understand why this case is here, whereas the same is not true for the two above.

exercises/rotational-cipher/description.md
```diff
@@ -0,0 +1,37 @@
...
+- ROT5  `omg` gives `trl`
+- ROT0  `c` gives `c`
+- ROT26 `Cool` gives `Cool`
+- ROT13 `The quick brown fox jumps over the lazy dog.` gives `Gur dhvpx oebja sbk whzcf bire gur ynml qbt.`
+- ROT13 `Gur dhvpx oebja sbk whzcf bire gur ynml qbt.` gives `The quick brown fox jumps over the lazy dog.`
+
+## BONUS!
+non-ASCII characters are downgraded this means that an `é` would become an `e`.
+
+## Source
```
petertseng: You'll want to remove these. These are automatically inserted, based on what's in metadata.yml

exercises/rotational-cipher/description.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Examples
+- ROT5  `omg` gives `trl`
+- ROT0  `c` gives `c`
+- ROT26 `Cool` gives `Cool`
+- ROT13 `The quick brown fox jumps over the lazy dog.` gives `Gur dhvpx oebja sbk whzcf bire gur ynml qbt.`
+- ROT13 `Gur dhvpx oebja sbk whzcf bire gur ynml qbt.` gives `The quick brown fox jumps over the lazy dog.`
+
+## BONUS!
+non-ASCII characters are downgraded this means that an `é` would become an `e`.
```
petertseng: I think, by #428, we wouldn't want to add these. Unless there is some particular reason it's very important to *this* particular problem (but wasn't important to, say, Atbash, whose non-ASCII cases got removed in #529)

Downgrading leads to some interesting questions - now, it's not always the case that `rot13(rot13(t)) = t`. And it leads to conflict on whether `rot26(t) = t` (if downgrading happens, maybe it won't be, unless someone special-cased 0 and 26 to always return the input unchanged, but the nwe're being inconsistent on whether to downgrade)

exercises/rotational-cipher/cannonical-data.json
```diff
@@ -32,7 +32,7 @@
                 "expected": "z"
```
rpottsoh: I think @petertseng mentioned this but I see it still here.  Shouldn't `rotate a by 0 same input as output` be worded `rotate a by 0, same output as input` and then shouldn't the `shift_key` be equal to 0 instead of 26.  If left like it is wouldn't this test also be considered a wrap around test?

rpottsoh:
Thanks @LogoiLab for this new exercise.

exercises/rotational-cipher/cannonical-data.json
```diff
@@ -0,0 +1,72 @@
+{
+    "#": [
+        "The tests are a series of rotation tests and a bonus test: ",
+        "* Rotational tests",
```
petertseng: I believe this line and the one below it can be removed.

actually, in that case, maybe the entire `"#"` section can be removed.

exercises/rotational-cipher/cannonical-data.json
```diff
@@ -0,0 +1,72 @@
+{
+    "#": [
+        "The tests are a series of rotation tests and a bonus test: ",
```
petertseng: I believe the text about "bonus test" can be removed.

exercises/rotational-cipher/cannonical-data.json
```diff
@@ -0,0 +1,72 @@
...
+                "expected": "b"
+            },
+            {
+                "description": "rotate a by 26, same output as input",
+                "text": "a",
+                "shift_key": 26,
+                "expected": "a"
+            },
+            {
+                "description": "rotate a by 0 same input as output",
```
petertseng: seems weird to have the wording be different than above. how about making them consistent?

exercises/rotational-cipher/cannonical-data.json
```diff
@@ -0,0 +1,72 @@
...
+                "expected": "a"
+            },
+            {
+                "description": "rotate m by 13",
+                "text": "m",
+                "shift_key": 13,
+                "expected": "z"
+            },
+            {
+                "description": "rotate n by 13 first wrap around alphabet",
```
petertseng: I would get rid of the word "first" as it's not clear whether it means it's the first case that does so (which is what I meant... unless you count 26 as wrapping around all the way), or the first shift_key that would, or the first letter that would (both of which depend)

exercises/rotational-cipher/description.md
```diff
@@ -0,0 +1,34 @@
...
+Ciphertext is written out in the same formatting as the input including spaces and punctuation.
+
+## Examples
+- ROT5  `omg` gives `trl`
+- ROT0  `c` gives `c`
+- ROT26 `Cool` gives `Cool`
+- ROT13 `The quick brown fox jumps over the lazy dog.` gives `Gur dhvpx oebja sbk whzcf bire gur ynml qbt.`
+- ROT13 `Gur dhvpx oebja sbk whzcf bire gur ynml qbt.` gives `The quick brown fox jumps over the lazy dog.`
+
+## Source
```
petertseng: don't forget to remove these lines (Source)

exercises/rotational-cipher/cannonical-data.json
```diff
@@ -0,0 +1,70 @@
...
+            },
+            {
+                "description": "rotate numbers",
+                "text": "Testing 1 2 3 testing",
+                "shift_key": 4,
+                "expected": "Xiwxmrk 1 2 3 xiwxmrk"
+            },
+            {
+                "description": "rotate punctuation",
+                "text": "Let's eat, Grandma!",
```
petertseng: I've shown you this already, but I wanted to show everyone else.

https://twitter.com/neilhimself/status/822312562809110529

exercises/rotational-cipher/cannonical-data.json
```diff
@@ -0,0 +1,70 @@
...
+                "expected": "b"
+            },
+            {
+                "description": "rotate a by 26, same output as input",
+                "text": "a",
+                "shift_key": 26,
+                "expected": "a"
+            },
+            {
+                "description": "rotate a by 0 same output as input",
```
petertseng: a comma here for consistency with the above?

LogoiLab: sorry I'm mediocre with git and pull requests. Never heard of squashing.

petertseng:
Everything seems to be in order here, and I see you squashed the commits as well, very nice. Let's merge.

petertseng:
Thanks!

PR #535, opened 2017-02-06
change: Add case for target > coins, but still can't make

petertseng:
The current error cases ask us to ensure that two types of targets are
invalid:
* targets smaller than the smallest coin
* negative targets

These are both reasonable.

I would react with chagrin if this led student solutions to assume that
these are the only two cases in which one could fail to meet the target
i.e. they assume that any target > the smallest coin is valid.

This commit proposes adding a case that disproves that - the target is
larger than even the largest coin, but you will not be able to get a
combination that reaches the target.

### DIFF
--- a/exercises/change/canonical-data.json
+++ b/exercises/change/canonical-data.json
@@ -49,6 +49,12 @@
                 "expected": -1
             },
             {
+                "description": "error if no combination can add up to target",
+                "coins": [5, 10],
+                "target": 94,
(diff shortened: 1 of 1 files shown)


stkent:
Makes total sense to me; Java track is poised to add this so I'll hold that exercise until this is 👍'd and merged!

ErikSchierboom:
LGTM!

PR #536, opened 2017-02-06
bowling: be more explicit about when errors should happen

petertseng:
The JSON file currently says "When to error is also left up to your
implementation", but we should instead be explicit.

Let's say I `roll 5`, then `roll 6`, then `roll 0` 18 times.
Okay, that's 20 rolls, that's a full game, so I can take the score now,
right?
Now I try to take the score and find that I get an error. Argh!

In the absolute best case, the error result from score would be so kind
to tell me that my `roll 6` way back there was invalid, but in the worst
case it might not, and I would be left to my own devices to figure out
what happened, since there was no feedback on any of the individual
rolls.

This commit proposes that invalid rolls be made explicit right when the
roll is made, not at some unspecified time in the future (when
attempting to take the score of the game). This is done via a
`roll_should_fail` key that is mutually exclusive with the `expected`
key.

Previous discussions I know of (since I was involved):
* exercism/xrust#213 (comment)
* exercism/xhaskell#440

### DIFF
--- a/exercises/bowling/canonical-data.json
+++ b/exercises/bowling/canonical-data.json
@@ -5,13 +5,21 @@
     "Score should return the game's final score, when possible",
     "For brevity the tests display all the rolls in an array;",
     "each element of the rolls array should be passed to the roll method",
-    "The final tests define situations where the score can not be returned",
-    "The expection for these tests is -1, indicating an error",
-    "In these cases you should expect an error as is idiomatic for your language",
-    "When to error is also left up to your implementation. There are two options",
-    "  - Error as soon as an invalid roll is made",
-    "  - Error when scoring a game with an invalid roll",
-    "You can also error in both cases."
+    "",
+    "If the `roll_should_fail` key is present",
+    "Then after rolling the values in the `roll` key,",
+    "attempting to roll the value in the `roll_should_fail` key should produce an error.",
+    "",
+    "If the `expected` key is present and non-negative",
+    "all rolls should succeed, and the final score should equal that value.",
+    "",
+    "If the `expected` key is present and negative,",
+    "all rolls should succeed, but attempting to score the game should result in an error.",
+    "",
+    "Each case will have exactly one of `roll_should_fail` or `expected`.",
+    "",
+    "In all error cases you should expect an error as is idiomatic for your language",
+    "whether that be via exceptions, optional values, or otherwise."
   ],
   "score": {
     "description": [
@@ -75,36 +83,36 @@
       "expected": 300
     }, {
       "description": "rolls can not score negative points",
-      "rolls": [-1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-      "expected": -1
+      "rolls": [],
+      "roll_should_fail": -1
     }, {
       "description": "a roll can not score more than 10 points",
-      "rolls": [11, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-      "expected": -1
+      "rolls": [],
+      "roll_should_fail": 11
     }, {
       "description": "two rolls in a frame can not score more than 10 points",
-      "rolls": [5, 6, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-      "expected": -1
+      "rolls": [5],
+      "roll_should_fail": 6
     }, {
       "description": "bonus roll after a strike in the last frame can not score more than 10 points",
-      "rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10, 11],
-      "expected": -1
+      "rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10],
+      "roll_should_fail": 11
     }, {
       "description": "two bonus rolls after a strike in the last frame can not score more than 10 points",
-      "rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10, 5, 6],
-      "expected": -1
+      "rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10, 5],
+      "roll_should_fail": 6
     }, {
       "description": "two bonus rolls after a strike in the last frame can score more than 10 points if one is a strike",
       "rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10, 10, 6],
       "expected": 26
     }, {
       "description": "the second bonus rolls after a strike in the last frame can not be a strike if the first one is not a strike",
-      "rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10, 6, 10],
-      "expected": -1
+      "rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10, 6],
+      "roll_should_fail": 10
     }, {
       "description": "second bonus roll after a strike in the last frame can not score than 10 points",
-      "rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10, 10, 11],
-      "expected": -1
+      "rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10, 10],
+      "roll_should_fail": 11
     }, {
       "description": "an unstarted game can not be scored",
       "rolls": [],
@@ -114,9 +122,9 @@
       "rolls": [0, 0],
       "expected": -1
     }, {
-      "description": "a game with more than ten frames can not be scored",
-      "rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-      "expected": -1
+      "description": "cannot roll if game already has ten frames",
+      "rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+      "roll_should_fail": 0
     }, {
       "description": "bonus rolls for a strike in the last frame must be rolled before score can be calculated",
       "rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10],
--- a/exercises/bowling/canonical-data.json
+++ b/exercises/bowling/canonical-data.json
@@ -7,129 +7,163 @@
     "each element of the previous_rolls array should be passed to the roll method",
     "and each of those previous rolls is expected to succeed.",
     "",
-    "If the `roll_should_fail` key is present",
-    "Then after rolling the values in the `roll` key,",
-    "attempting to roll the value in the `roll_should_fail` key should produce an error.",
+    "Two properties are tested:",
     "",
-    "If the `expected` key is present, it will either be an integer or an error object.",
-    "If it is an integer, all rolls should succeed, and the final score should equal that value.",
-    "If it is an error object, all rolls should succeed, but attempting to score the game should result in an error.",
+    "`score`: All rolls succeed, and taking the score gives the expected result.",
+    "The expected result may be an integer score or an error.",
     "",
-    "Each case will have exactly one of `roll_should_fail` or `expected`.",
+    "`roll`: All previous_rolls succeed, and rolling the number of pins in `roll` produces the expected result.",
+    "Currently, all cases of this type result in errors.",
     "",
     "In all error cases you should expect an error as is idiomatic for your language",
     "whether that be via exceptions, optional values, or otherwise."
   ],
   "cases": [{
     "description": "should be able to score a game with all zeros",
+    "property": "score",
     "previous_rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
     "expected": 0
   }, {
     "description": "should be able to score a game with no strikes or spares",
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
I think it is desirable to fail as soon as possible, as in general that is a good strategy to follow. How to specify it, I'm not sure. That's a bit harder I think.

petertseng:
OK, nobody gave a better suggestion, so I'm rolling with `roll_should_fail`.

petertseng:
Verified for correctness at https://github.com/petertseng/x-common/compare/verify...petertseng:verify-roll-should-fail?expand=1

exercises/bowling/canonical-data.json
```diff
@@ -5,13 +5,21 @@
...
-    "You can also error in both cases."
+    "",
+    "If the `roll_should_fail` key is present",
+    "Then after rolling the values in the `roll` key,",
+    "attempting to roll the value in the `roll_should_fail` key should produce an error.",
+    "",
+    "If the `expected` key is present and non-negative",
+    "all rolls should succeed, and the final score should equal that value.",
+    "",
+    "If the `expected` key is present and negative,",
```
petertseng: This is in violation of #401 and #551. I would like to suggest the `error`s from #240 be used.

petertseng:
OK, now using the error object of #401.

petertseng:
Verified for correctness at https://github.com/petertseng/x-common/compare/verify...petertseng:verify-roll-should-fail?expand=1

rbasso:
Just a comment, that shouldn't prevent merging this:

The fact that you are proposing to use a `shouldFail` or an `expected` that could contain an `error` (two distinct encodings for exceptional conditions) suggest that restricting `error` to be a `string`  if forcing alternative encodings of error-related data.

ErikSchierboom:
> The fact that you are proposing to use a shouldFail or an expected that could contain an error (two distinct encodings for exceptional conditions) suggests that restricting error to be a string is forcing alternative encodings of error exception-related data.

I think that's because there are two methods being tested here:

1. Recording a roll.
2. Getting the score.

With the [PR](https://github.com/exercism/x-common/pull/602) almost being merged, we could be using the `property` field to indicate that a different property is being tested, right?

petertseng:
So you are suggesting that the tests look something like this (I gave one example for each of the three different cases represented: Score succeeds, score fails because game is incomplete, roll fails because it is invalid (we could explicitly add cases where the roll succeeds; the expected value would probably just be `null` I suppose).

```json
    [ { "description"   : "game with no strikes or spares"
      , "property"      : "score"
      , "previous_rolls": [3, 6, 3, 6, 3, 6, 3, 6, 3, 6, 3, 6, 3, 6, 3, 6, 3, 6, 3, 6]
      , "expected"      : 90
      }   
    , { "description"   : "can't score incomplete game"
      , "property"      : "score"
      , "previous_rolls": [0, 0]
      , "expected"      : {"error": "something"}
      }   
    , { "description"   : "can't roll more pins than there are on the lane"
      , "property"      : "score"
      , "previous_rolls": [5] 
      , "roll"          : 6 
      , "expected"      : {"error": "something"}
      }   
    ]   
```

ErikSchierboom:
Actually, shouldn't the last case be:

```json
{ "description"   : "can't roll more pins than there are on the lane"
  , "property"      : "roll"
  , "previous_rolls": [5] 
  , "roll"          : 6 
  , "expected"      : {"error": "something"}
}
```

With the property being tested set to `"roll"` instead of `"score"`?

petertseng:
> With the property being tested set to `"roll"` instead of `"score"`?

You're right. Edited to be so, so that we are on the same page

ErikSchierboom:
> You're right. Edited to be so, so that we are on the same page

I think your three test-cases example looks perfect now! Very clean and understandable. Great!

rbasso:
First of all, let me say that I agree with the idea of returning which of the rolls in the lists in invalid. It is an useful information for the user.

Sorry for getting in the middle of this PR, I was just thinking out too loud. Now, let me explain what I meant.

> I think that's because there are two methods being tested here:
>
> 1. Recording a roll.
> 2. Getting the score.

This may be where we diverge. I think the exercise is just about scoring a game, so we need to test `score`, and that is also the only test group in the `canonical-data.json`.

To fulfill the new requirements of returning which roll is the invalid, this `score` may return:

- A score, for a valid and complete game.
- An exceptional value for an incomplete game.
- Another exceptional value for an invalid list of rolls, with the information of which roll is the invalid one.

A valid score should be represented normally in `expected`:

```json
"expected": 42
```

An exceptional condition would, by convention, be represented as an `error`:

```json
"expected": { "error": "Incomplete game" }
```

or

```json
"expected": { "error": "Invalid roll 6 - Cannot score more than 10" }
```

I was just thinking that, because `error` was defined as a `string` in #551, there is not way for us to include more information about the error in a structure way.

Anyway, it seems that this PR has taken a turn, and now the proposal includes also testing an *intermediary function*. In this case, I think that my original arguments doesn't apply anymore.

ErikSchierboom:
@rbasso I'd also be content with your solution, but I do really like to "fail-fast" in my code, and failing on an invalid roll instead of only failing when calculating a score seemed reasonable to me.

petertseng:
First, I will say that I will propose a new file to match the new schema, which should make clear when the errors are happening.

---

Second, I will say that the Haskell track only has a `score` function returning an `Either`, with the `Left` being a type that can either be an invalid roll (specifying precisely which roll was invalid) or an incomplete game. The discussion can be seen at https://github.com/exercism/xhaskell/issues/440 and the result at https://github.com/exercism/xhaskell/blob/master/exercises/bowling/test/Tests.hs

Now I see what we mean when we say we want a richer error than just `"error": <string>"`: it would be great to have structured data specifying which roll was invalid in the `error` object. Exception-based languages should be able to make this happen easily too; if this is seen as desirable we should revise the `error` object.

petertseng:
EBWODP:

> I will propose a new file

But not now. Maybe in 12 hours.

rbasso:
#### About when to fail

> @rbasso I'd also be content with your solution, but I do really like to "fail-fast" in my code, and failing on an invalid roll instead of only failing when calculating a score seemed reasonable to me.

Think in what I wrote more like a question than a suggestion. By the way, I also prefer to *fail-fast* most of the times, @ErikSchierboom.

I was just pondering that maybe we are letting details of *"how to solve the problem"* affect the *"how to specify the problem"*. If the solution gives the correct results, it shouldn't matter how it does it internally.

#### About more structured expected types

> Now I see what we mean when we say we want a richer error than just "error": <string>": it would be great to have structured data specifying which roll was invalid in the error object. Exception-based languages should be able to make this happen easily too; if this is seen as desirable we should revise the error object.

I just started thinking about the subject after we added the restrictions from #551 in #602, but maybe the problem can be generalized a little. Maybe `error` is just a special case of the more general problem of returning different things depending on the input data.

Let's say we have a function called `feedShark`, which can return one of the following (a sum type), depending on the input `food`:

- Case 1 - a boolean indicating if it has eaten the food.
- Case 2 - an exceptional value, indicating that the object  is something too big to even think about feeding it.
- Case 3 - another exceptional value, indicating that the feeder has lost a body part in the proccess, also indicating which part was eaten.

We have a lot of different ways to encode that in JSON:

```json
[ { "description": "Try to feed a furniture"
  , "property"   : "feedShark"
  , "food"       : "wardrobe"
  , "expected"   : false
  }
, { "description": "Try to feed a big animal"
  , "property"   : "feedShark"
  , "food"       : "elephant"
  , "expected"   : "Impossible"
  }
, { "description": "Try to feed a big animal"
  , "property"   : "feedShark"
  , "food"       : "elephant"
  , "expected"   : { "feederEaten": "right arm" }
  }
]
```

We could also be more descriptive about the cases, like this:


```json
[ { "description": "Try to feed a furniture"
  , "property"   : "feedShark"
  , "food"       : "wardrobe"
  , "expected"   : { "eaten": true }
  }
, { "description": "Try to feed a big animal"
  , "property"   : "feedShark"
  , "food"       : "elephant"
  , "expected"   : { "impossible": {} }
  }
, { "description": "Try to feed a big animal"
  , "property"   : "feedShark"
  , "food"       : "elephant"
  , "expected"   : { "feederEaten": "right arm" }
  }
]
```

We could even flatten everything, maybe for readability:

```json
[ { "description": "Try to feed a furniture"
  , "property"   : "feedShark"
  , "food"       : "wardrobe"
  , "eaten"      : true
  }
, { "description": "Try to feed a big animal"
  , "property"   : "feedShark"
  , "food"       : "elephant"
  , "impossible" : {}
  }
, { "description": "Try to feed a big animal"
  , "property"   : "feedShark"
  , "food"       : "elephant"
  , "feederEaten": "right arm"
  }
]
```

rbasso:
#### About when to fail

> @rbasso I'd also be content with your solution, but I do really like to "fail-fast" in my code, and failing on an invalid roll instead of only failing when calculating a score seemed reasonable to me.

Think in what I wrote more like a question than a suggestion. By the way, I also prefer to *fail-fast* most of the times, @ErikSchierboom.

I was just pondering that maybe we are letting details of *"how to solve the problem"* affect the *"how to specify the problem"*. If the solution gives the correct results, it shouldn't matter how it does it internally.

#### About more structured expected types

> Now I see what we mean when we say we want a richer error than just "error": <string>": it would be great to have structured data specifying which roll was invalid in the error object. Exception-based languages should be able to make this happen easily too; if this is seen as desirable we should revise the error object.

I just started thinking about the subject after we added the restrictions from #551 in #602, @petertseng, but maybe the problem can be generalized a little. Maybe `error` is just a special case of the more general problem of returning different things depending on the input data.

Let's say we have a function called `feedShark`, which can return one of the following (a sum type), depending on the input `food`:

- Case 1 - a boolean indicating if it has eaten the food.
- Case 2 - an exceptional value, indicating that the object  is something too big to even think about feeding it.
- Case 3 - another exceptional value, indicating that the feeder has lost a body part in the proccess, also indicating which part was eaten.

We have a lot of different ways to encode that in JSON:

```json
[ { "description": "Try to feed a furniture"
  , "property"   : "feedShark"
  , "food"       : "wardrobe"
  , "expected"   : false
  }
, { "description": "Try to feed a big animal"
  , "property"   : "feedShark"
  , "food"       : "elephant"
  , "expected"   : "Impossible"
  }
, { "description": "Try to feed a big animal"
  , "property"   : "feedShark"
  , "food"       : "elephant"
  , "expected"   : { "feederEaten": "right arm" }
  }
]
```

We could also be more descriptive about the cases, like this:


```json
[ { "description": "Try to feed a furniture"
  , "property"   : "feedShark"
  , "food"       : "wardrobe"
  , "expected"   : { "eaten": true }
  }
, { "description": "Try to feed a big animal"
  , "property"   : "feedShark"
  , "food"       : "elephant"
  , "expected"   : { "impossible": {} }
  }
, { "description": "Try to feed a big animal"
  , "property"   : "feedShark"
  , "food"       : "elephant"
  , "expected"   : { "feederEaten": "right arm" }
  }
]
```

We could even flatten everything, maybe for readability:

```json
[ { "description": "Try to feed a furniture"
  , "property"   : "feedShark"
  , "food"       : "wardrobe"
  , "eaten"      : true
  }
, { "description": "Try to feed a big animal"
  , "property"   : "feedShark"
  , "food"       : "elephant"
  , "impossible" : {}
  }
, { "description": "Try to feed a big animal"
  , "property"   : "feedShark"
  , "food"       : "elephant"
  , "feederEaten": "right arm"
  }
]
```

petertseng:
Same old place.

https://github.com/petertseng/x-common/compare/verify...petertseng:verify-roll-should-fail?expand=1

petertseng:
The unindent is inevitable to match the new schema. But that's why it's in its own commit, so that it is easily seen that that commit *only* unindents.

ErikSchierboom:
IMHO, the current version is great. What do you think @rbasso? Perhaps we should move the discussion about return values to a separate issue?

rbasso:
> IMHO, the current version is great. What do you think @rbasso?

I'm usually against testing for intermediary functions, but I never solved this exercise and I trust that you have your reasons. 👍 

>  Perhaps we should move the discussion about return values to a separate issue?

It was just an idea...I guess I'm spending too much time with JSON. 😄

ErikSchierboom:
@rbasso Well, the [problem description requirements](https://github.com/exercism/x-common/blob/master/exercises/bowling/description.md#requirements) actually explicitly mention two operations to be implemented:

> - roll(pins : int) is called each time the player rolls a ball. The argument is the number of pins knocked down.
> - score() : int is called only at the very end of the game. It returns the total score for that game.

:)

rbasso:
> Well, the problem description requirements actually explicitly mention two operations to be implemented

You are indisputably right about that. 😄

petertseng:
Sorry, no squash since I think the individual changes are easier to understand than a single diff that purports to change the entire file (it really doesn't)

I was a bit selfish and only served my specific goals for this PR: The goal of expressing in some way that a roll should fail fast (as soon as it is known to be erroneous).

There are other goals that are still worthy, even though I left them aside for this PR. I kept the scope small for this PR, but these are worth their own discussion:

## More structured data in `error`

This allows tracks to easily check whether the right kind of error is given. This seems easy enough to implement for a few tracks: Those with sum types can check that the constructor, and those with exceptions can check the class of the exception.

From what I know, the Swift track is very diligent about doing this. I don't have time to show y'all all the examples, but:

* https://github.com/exercism/xswift/blob/master/exercises/minesweeper/Tests/MinesweeperTests/MinesweeperTests.swift#L69
* https://github.com/exercism/xswift/blob/master/exercises/queen-attack/Tests/QueenAttackTests/QueenAttackTests.swift#L35
* https://github.com/exercism/xswift/blob/master/exercises/largest-series-product/Tests/LargestSeriesProductTests/LargestSeriesProductTests.swift#L52
* https://github.com/exercism/xswift/blob/master/exercises/bowling/Tests/BowlingTests/BowlingTests.swift#L103

Other tracks don't do that as often. I explained in https://github.com/exercism/x-common/issues/401#issuecomment-275585678 that  e.g. Rust merely checks [`Result#is_err`](https://doc.rust-lang.org/std/result/enum.Result.html#method.is_err) and Haskell merely checks [`Data.Either.isLeft`](https://hackage.haskell.org/package/base-4.9.1.0/docs/Data-Either.html#v:isLeft) (with bowling being a unique case where the error type **is** checked!)

If more tracks are interested in doing this, it is certainly possible. It may be good to see whether there is any fixed schema that will relieve us from having to write exercise-specific code for each exercise's `error`s.

# Bowling overspecifying

The https://github.com/exercism/x-common/blob/master/exercises/bowling/metadata.yml tells us this exercise hails from http://butunclebob.com/ArticleS.UncleBob.TheBowlingGameKata . Now, note that we have already morphed it from its original form since we require checking for errors whereas the original did not.

We may consider further revisions, such as not specifying the internals and only asking "the score resulting from these rolls, or error". For maximum effect it is probably good if the error describes which roll is erroneous (as the Haskell tests do).

I don't oppose that change, but I do believe it is best left until we have a better `error`.

ErikSchierboom:
:tada:

Issue #539, opened 2017-02-08
xjulia builds are running long/hanging (?)


jtigger:
In the last couple of days, I've noticed that builds on a couple of tracks I co-maintain have been hanging...  looking further, I found that there were a number of `xjulia` builds that are running for a considerable time (~45-50 mins)...

I also notice that there are multiple builds at once.  Right now there are four concurrent builds running in Travis for Exercism and all four are `xjulia`.  This apparently hits some limit and other builds are delayed until these either complete, are stopped, or timeout.

![image](https://cloud.githubusercontent.com/assets/470176/22745022/028cd3de-edd4-11e6-9b95-4408cf589c77.png)

I couldn't view the logs for the builds.   Any ideas of what might be happening?

@exercism/julia — thoughts?

SaschaMann:
The macOS builds take a really long time to start up at the moment (up to an hour). As soon as they start, they finish within about a minute.

I thought it's still a result of the recent macOS issues that travis had. I know that it happens for other julia packages too that use Travis. I'm not really sure what causes it.

I could temporarily disable the macOS testing for now; I didn't know this causes issues for other tracks.

Are there any other tracks that experience the same issues with long macOS-startup times?

petertseng:
Trying to figure out what could be happening here - if OSX plays a role, would one expect that the other tracks with OSX builds could also cause builds to back up?

* https://github.com/exercism/xcsharp/blob/master/.travis.yml
* https://github.com/exercism/xfsharp/blob/master/.travis.yml
* https://github.com/exercism/gui/blob/master/.travis.yml

In general, I had observed in the past that Travis only runs five jobs at a time, but I thought that was a per-repo limit. Hadn't seen the per-organisation limit (assuming that's what it is).

Can we try to test out a few situations, perhaps by creating some dummy repos in the org, some with OSX builds?

petertseng:
As another sample point so that we can see the influence of other factors as well: If we are looking at tracks with many jobs per build, then we might look at https://github.com/exercism/xgo/blob/master/.travis.yml#L17-L20 which hilariously attempts to stay as close to **ten** jobs as possible without going over.

Another facet of my position where I was not clear in expression:

[...]

PR #541, opened 2017-02-08
empty commit 2 - trigger travis build, but no Go build now

petertseng:
for #539


petertseng:
ugh. not what I wanted to see (this one started up very quickly)

PR #542, opened 2017-02-09
pascals-triangle: Clarify test case descriptions

jonmcalder:
pascals-triangle: Clarify test case descriptions to distinguish between zero rows and null/no rows.

### DIFF
--- a/exercises/pascals-triangle/canonical-data.json
+++ b/exercises/pascals-triangle/canonical-data.json
@@ -8,7 +8,7 @@
       "description": "Given a count, return a collection of that many rows of pascal's triangle",
       "cases": [
         {
-            "description": "no rows",
+            "description": "zero rows",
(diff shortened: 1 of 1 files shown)


petertseng:
Oh. Yeah this makes sense, since otherwise they would both say "no rows", which makes it too ambiguous.

Could we ask for the commit message to be prefixed with "pascals-triangle: " so that when we look at just log messages it's clear what exercise this is for?

Thanks!

petertseng:
Don't forget to change the commit message's contents too, not just the PR title.

I got it covered.

petertseng:
Thanks!

jonmcalder:
Apologies -  figured you probably meant I should amend the commit message itself too but was busy (and lazy). Thanks for being accommodating - will remember in future.

PR #543, opened 2017-02-10
Make canonical data for RLE test generator friendly.

stevejb71:
Change the format of the canonical data json for the
run length encoding exercise to be the same as that for
the other multi-function exercises.
This is more useful for test generators (certainly the Ocaml
one).

Descriptions also updated to be a little clearer.

### DIFF
--- a/exercises/run-length-encoding/canonical-data.json
+++ b/exercises/run-length-encoding/canonical-data.json
@@ -1,52 +1,61 @@
 {
-   "#": [
-      "The tests deliberately check that single characters are encoded without count."
-   ],
-   "cases": [
-      {
-        "description": "encode empty string",
-        "input": "",
-        "expected": ""
-      },
-      {
-        "description": "encode single characters only",
-        "input": "XYZ",
-        "expected": "XYZ"
-      },
-      {
-        "description": "decode empty string",
-        "input": "",
-        "expected": ""
-      },
-      {
-        "description": "decode single characters only",
-        "input": "XYZ",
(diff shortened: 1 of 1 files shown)


exercises/run-length-encoding/canonical-data.json
```diff
@@ -1,52 +1,61 @@
...
+            "input": "12WB12W3B24WB",
+            "expected": "WWWWWWWWWWWWBWWWWWWWWWWWWBBBWWWWWWWWWWWWWWWWWWWWWWWWB"
+          }
+      ]
+   },
+   "consistency": {
+      "cases": [
+          {
+            "description": "encode followed by decode gives original string",
+            "input": "zzz ZZ  zZ",
```
petertseng: I always wonder why this is not just two separate cases in encode and decode. Or if testing that `∀x. encode (decode x) == x` why doesn't the test suite just generate random strings?

Maybe to be discussed in a later PR, since this one is just to move the cases around.

stevejb71:
I agree with the suggestions in #336 for putting a function name in or near each test case - that would allow a more flexible order if required.

Updating to a new schema would certainly break the OCaml test generator, but it's a pretty small piece of code so would not be significant effort to fix.

exercises/run-length-encoding/canonical-data.json
stevejb71: I'd guess some standard way to specify a random input/expected would be required. But I think it's a good idea.

PR #544, opened 2017-02-10
hello-world: Change to always expect "Hello, World!"

### LINKED ISSUE #520: I posit that hello-world is too complex for a first exercise. How do we deal with it?
I learned from:

* https://github.com/exercism/xhaskell/pull/475#issuecomment-275971130
* https://github.com/exercism/todo/issues/12
* https://github.com/exercism/exercism.io/issues/2021

Hello world should be a very easy exercise. Its objectives:

* Demonstrates to the student that their`exercism fetch` and `exercism submit` are working (NB, this *could* be a language-agnostic exercise, but it is not clear where we would put it)
* Explain to the student how to run the test suite in the language
    * it's up to each track to provide this in their track-specific documentation e.g. https://github.com/exercism/xruby/blob/master/docs/TESTS.md
    * this role doesn't *have* to be filled by hello-world, but some students prefer to have the separate exercise dedicated only to making sure this step is solid before moving on. Those who don't care can skip it.
* Introduce the student to how TDD works
    * shown by https://github.com/exercism/x-common/blob/master/exercises/hello-world/description.md
    * shown by the process of getting the tests to pass

I am here to argue that **the current hello-world** ("Alice" -> "Hello, Alice!", "Bob" -> "Hello, Bob!", no input -> "Hello, World!") **does not serve this purpose well**, since it also requires the student to know enough about the language to handle the no-input case.

This effect is *especially* pronounced in languages that do not allow function overloading or default arguments. Some languages resort to passing an empty string, defining two functions, or others discussed in #290. [Elm](https://github.com/exercism/xelm/blob/master/exercises/hello-world/HelloWorld.elm), [F#](https://github.com/exercism/xfsharp/blob/master/exercises/hello-world/HelloWorldTest.fs), [Haskell](https://github.com/exercism/xhaskell/pull/475), [Kotlin](https://github.com/exercism/xkotlin/blob/master/exercises/hello-world/src/main/kotlin/HelloWorld.kt), [OCaml](https://github.com/exercism/xocaml/blob/master/exercises/hello-world/hello_world.mli), [Purescript](https://github.com/exercism/xpurescript/blob/master/exercises/hello-world/examples/src/HelloWorld.purs), and [Rust](https://github.com/exercism/xrust/blob/master/exercises/hello-world/example.rs) are all using their Option/Maybe type, which I argue is not suitable for hello-world (In [Learn You A Haskell](http://learnyouahaskell.com/), Maybe is not mentioned until chapter 7 of 14). But even in languages with function overloading or default arguments, I suggest that these two features are too much for hello-world.

I would say that the best way to achieve the goal of hello-world is to **have the function-under-test take zero arguments and return "Hello, World!" unconditionally**.

But what can we do? The current hello-world is already out there, and several languages have it.

1. If we change the hello-world README to unequivocally state that the function should return "Hello, World!" unconditionally, then all languages that currently have hello-world will be inconsistent with the README. Is that okay?
2. If we change the hello-world README to be so vague that either interpretation is possible (the current interpretation, or unconditional "Hello, world"), languages can do as they wish, but it is a delicate balance to strike - we have seen before in https://github.com/exercism/exercism.io/issues/2746, an underspecified problem is also undesirable.
3. What if we left hello-world as-is, and made a new exercise, `the-answer`, whose only function-under-test should return 42 unconditionally. Then, languages can do whichever of the two they please. The only disadvantage is that it will be confusing to track maintainers and exercise porters to have two starter exercises.
4. Maybe you disagree with my proposal: Maybe you think students can be expected to have some minimum level of knowledge upon starting the track. Or you think my proposal is rather too simple. Let me know about it!

rbasso:
As discussed in #520, the exercise `hello-world` is too complex to fulfill its purpose and we decided to simplify it.

The new test suite will simply expect the string **"Hello, World!"**, and will serve mainly as a sanity check that everything is working.

This change may disrupt some tracks that already have it implemented, so we are scheduling the merging of this PR to allow the @exercism/track-maintainers to sync track-specific changes with `x-common`.

The merging is tentatively scheduled to: **2016-02-20 00:01 UTC** +-1 day 😄

Please direct any question/critic to #520, so that we can keep the discussion focused there.

Closes #520.

### DIFF
--- a/exercises/hello-world/canonical-data.json
+++ b/exercises/hello-world/canonical-data.json
@@ -1,23 +1,11 @@
 {
-   "#": [
-     "Language implementations vary on the issue of missing input variables.",
-     "In this set of test cases, the input variable is left out of the definition,",
-     "and the generator would need to handle the language implementation."
-   ],
-   "cases": [
-      {
-         "description": "no name",
-         "expected": "Hello, World!"
-      },
-      {
-         "description": "sample name",
-         "name": "Alice",
-         "expected": "Hello, Alice!"
-      },
-      {
-         "description": "other sample name",
-         "name": "Bob",
-         "expected": "Hello, Bob!"
-      }
-   ]
+  "exercise":"hello-world",
+  "version":"0.1.0",
+  "cases":[
+    {
+      "description":"Say Hi!",
+      "property":"hello",
+      "expected":"Hello, World!"
+    }
+  ]
 }
--- a/exercises/hello-world/description.md
+++ b/exercises/hello-world/description.md
@@ -1,46 +1,11 @@
 ["Hello, World!"](http://en.wikipedia.org/wiki/%22Hello,_world!%22_program) is
-the traditional first program for beginning programming in a new language.
+the traditional first program for beginning programming in a new language
+or environment.
 
-**Note:** You can skip this exercise by running:
+The objectives are simple:
 
-    exercism skip $TRACK_ID hello-world
+- Write a function that says "Hello, World!".
+- Run the test suite and make sure that it succeeds.
+- Submit your solution and check it at the website.
 
-## Specification
-
-Write a `Hello World!` function that can greet someone given their name.  The
-function should return the appropriate greeting.
-
-For an input of "Alice", the response should be "Hello, Alice!".
-
-If a name is not given, the response should be "Hello, World!"
-
-## Test-Driven Development
-
-As programmers mature, they eventually want to test their code.
-
-Here at Exercism we simulate [Test-Driven
-Development](http://en.wikipedia.org/wiki/Test-driven_development) (TDD), where
-you write your tests before writing any functionality. The simulation comes in
-the form of a pre-written test suite, which will signal that you have solved
-the problem.
-
-It will also provide you with a safety net to explore other solutions without
-breaking the functionality.
-
-### A typical TDD workflow on Exercism:
-
-1. Run the test file and pick one test that's failing.
-2. Write some code to fix the test you picked.
-3. Re-run the tests to confirm the test is now passing.
-4. Repeat from step 1.
-5. Submit your solution (`exercism submit /path/to/file`)
-
-## Instructions
-
-Submissions are encouraged to be general, within reason. Having said that, it's
-also important not to over-engineer a solution.
-
-It's important to remember that the goal is to make code as expressive and
-readable as we can. However, solutions to the hello-world exercise will not be
-reviewed by a person, but by rikki- the robot, who will offer an encouraging
-word.
+If everything goes well, you will be ready to fetch your first real exercise.
--- a/exercises/hello-world/metadata.yml
+++ b/exercises/hello-world/metadata.yml
@@ -1,4 +1,4 @@
 ---
-blurb: 'Greet the user by name, or by saying "Hello, World!" if no name is given.'
+blurb: 'The classical introductory exercise. Just say "Hello, World!"'
 source: "This is an exercise to introduce users to using Exercism"
 source_url: "http://en.wikipedia.org/wiki/%22Hello,_world!%22_program"


exercises/hello-world/canonical-data.json
```diff
@@ -1,23 +1,12 @@
...
-         "description": "other sample name",
-         "name": "Bob",
-         "expected": "Hello, Bob!"
+  "exercise":"hello-world",
+  "version":"0.1.0",
+  "group":[
+    {
+      "hello":{
+        "description":"Say Hi!",
+        "expected":"Hello, World!"
```
rbasso: There is no standard format for the `canonical-data.json`, so I used the one I proposed [here](https://github.com/exercism/x-common/issues/336#issuecomment-278550730). If by any chance it gets accepted without change, it will be one less file to update. 😄

exercises/hello-world/metadata.yml
```diff
@@ -1,4 +1,4 @@
 ---
-blurb: 'Greet the user by name, or by saying "Hello, World!" if no name is given.'
+blurb: 'Just say "Hello, World!"'
```
rbasso: I changed it, but I really don't know what would be good to put there.
If you have any suggestion, please write it here and I'll change it immediately.

exercises/hello-world/description.md
```diff
@@ -1,46 +1,11 @@
...
-## Specification
-
-Write a `Hello World!` function that can greet someone given their name.  The
-function should return the appropriate greeting.
-
-For an input of "Alice", the response should be "Hello, Alice!".
-
-If a name is not given, the response should be "Hello, World!"
-
-## Test-Driven Development
```
rbasso: I think that the instructions where misplaced.

Not all exercises implemented the `hello-world` exercise. Moving the instructions to `leap`'s `description.md` would just reproduce the problem in another place.

The immediate solution I can think of is track-dependent: Move it to the `HINTS.md` from the first exercise that is not `hello-world`. This is more flexible than forcing it to be `leap`.

This way, each track can customize the instructions to reflect what are the best practices for testing in the language. Not all languages are so TDD focused as `ruby` seems to be, specially the strongly-typed ones, that force users to type-check everything before running a single test.

I just create a Gist of a `HINTS.md` file with the old instructions [here](https://gist.github.com/anonymous/dc6d262de44db71441d8da16c41c4c80), to make migration easier. If you think this is a good idea, we could link it in a more visible place to help the track maintainers.

ErikSchierboom: I agree that the instrutions were misplaced. Languages can use the proposed hint for their next exercise, whatever it may be.

exercises/hello-world/description.md
```diff
@@ -1,46 +1,11 @@
...
-## Instructions
-
-Submissions are encouraged to be general, within reason. Having said that, it's
-also important not to over-engineer a solution.
-
-It's important to remember that the goal is to make code as expressive and
-readable as we can. However, solutions to the hello-world exercise will not be
-reviewed by a person, but by rikki- the robot, who will offer an encouraging
-word.
+If everything goes well, you will be ready to fetch your first real exercise.
```
Dispader: Very nice.  One note (probably to myself): while the instructions we lose in this Pull seem appropriate to remove here, they might be useful somewhere else.  If we do end up choosing a `intro-to-tdd` exercise, this might be a good source for that `description.md`.

exercises/hello-world/description.md
```diff
@@ -1,46 +1,11 @@
...
-the traditional first program for beginning programming in a new language.
+the traditional first program for beginning programming in a new language
+or environment.
 
-**Note:** You can skip this exercise by running:
+The objectives are simple:
 
-    exercism skip $TRACK_ID hello-world
+- Write a function that says "Hello, World!".
+- Run the tests and make sure that they succeed.
```
ErikSchierboom: Maybe "Run the tests" should be "Run the test(s)", as I expect most languages to only have a single test.

kotp: Run the test file?  This is always the case, unless there are tracks with multiple files for testing?

rbasso: "Run the test suite..." ?

ErikSchierboom: Run the test suite is a very nice suggestion! I like that.

rbasso:
The time for merging has come (in fact we are 4 hours and 30 minutes late), so I'm merging this as promised/agreed and hope that all the tracks also merge their changes soon, to keep consistency

> The merging is tentatively scheduled to: 2016-02-20 00:01 UTC +-1 day 😄

PR #546, opened 2017-02-11
difference-of-squares: make short description consistent

stkent:
Excerpt from the fuller description for this problem (my emphasis):

> Hence the difference between **the square of the sum** of the first ten natural numbers and **the sum of the squares** is 2640

This PR updates the short description to match the component order of the full description.


ErikSchierboom:
LGTM!

PR #547, opened 2017-02-12
luhn: tweak canonical tests

stkent:


exercises/luhn/canonical-data.json
```diff
@@ -11,7 +11,7 @@
       "expected": false
     },
     {
-      "description": "simple valid sin",
+      "description": "simple valid SIN",
```
stkent: Capitalized `sin` -> `SIN` throughout.

exercises/luhn/canonical-data.json
```diff
@@ -21,6 +21,11 @@
       "expected": true
     },
     {
+      "description": "another valid Canadian SIN",
+      "input": "055 444 285",
+      "expected": true
+    },
```
stkent: In the PR that added this test, the following reasoning was given:

> Part of it was the contrast with 0555-444-285, and part of it was the thought that having a couple of different longer valid numbers might help someone who was struggling to work out how the algorithm needed to be implemented.

I moved it earlier in the suite so that it is more likely encountered by someone who is still trying to work out the algorithm, and updated the language in the test cases that contrast with this one to make that comparison more clear.

exercises/luhn/canonical-data.json
```diff
@@ -31,17 +36,17 @@
...
     {
-      "description": "valid strings with a non-digit added become invalid",
+      "description": "valid strings with a non-digit included become invalid",
       "input": "046a 454 286",
       "expected": false
     },
     {
-      "description": "punctuation is not allowed",
+      "description": "valid strings with punctuation included become invalid",
       "input": "055-444-285",
```
stkent: Compare to valid case `055 444 285`.

exercises/luhn/canonical-data.json
```diff
@@ -31,17 +36,17 @@
...
       "expected": false
     },
     {
-      "description": "symbols are not allowed",
+      "description": "valid strings with symbols included become invalid",
```
stkent: Compare to valid case `055 444 285`.

exercises/luhn/canonical-data.json
```diff
@@ -51,17 +56,12 @@
...
     },
     {
-      "description": "lots of zeros are valid",
-      "input": " 00000",
-      "expected": true
-    },
-    {
-      "description": "another valid sin",
-      "input": "055 444 285",
+      "description": "more than a single zero is valid",
```
stkent: Updated to state the limitation here more clearly, and moved the space character to indicate that a "chain" of zeros is not required.

exercises/luhn/canonical-data.json
```diff
@@ -51,17 +56,12 @@
...
+      "input": "00 000",
       "expected": true
     },
     {
-      "description": "nine doubled is nine",
+      "description": "input digit 9 is correctly converted to output digit 9",
```
stkent: Nine doubled is not nine; tried to make this a little more explicit!

exercises/luhn/canonical-data.json
```diff
@@ -51,17 +56,12 @@
...
     {
-      "description": "lots of zeros are valid",
-      "input": " 00000",
-      "expected": true
-    },
-    {
-      "description": "another valid sin",
-      "input": "055 444 285",
+      "description": "more than a single zero is valid",
+      "input": "00 000",
```
stkent: Maybe `"0 0000"` would be even better here?

stkent:
I was also tempted to update the description for this test to make it more clear that this is the only test that will fail if the "wrong" set of alternating digits are manipulated. Thoughts on that?

```
    {
      "description": "simple valid sin",
      "input": " 5 9 ",
      "expected": true
    }
```

stkent:
Good call; I'll update that now!

stkent:
@petertseng updated with a few changes:

- removed doubled "valid Canadian SIN" test cases and simplified invalid cases to all be variants of the remaining valid SIN;
- added early test using "059" and following valid test using "59";
- updated the multiple-zeros test case to reduce possibility for misinterpretation.

stkent:
@petertseng see also #549 which I believe addresses another potential source of confusion for implementors.

michaelorr:
These changes look good to me. The descriptions and the test cases make sense and there doesn't seem to be enough overlap between them where they could be condensed down without sacrificing something. I don't have merge rights here but 👍   :shipit:

ErikSchierboom:
Great work, thanks! :tada:

PR #551, opened 2017-02-12
README: bring expectation of canonical-data.json values (description, errors, no msg) up-to-date

petertseng:
Note that while there is a new schema proposed in #336, it's believed by me that everything being said in this PR is still applicable even when that schema is applied:

* msg will still be gone
* descriptions will still follow the principles
* errors will still be defined as they are in #401

So this may be evaluated without having to wait for #336

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,8 @@ variations.
 
 Each test case has the the following keys:
 - description: which will be used to name each generated test
+  - The description should not simply explain **what** each case is (that is redundant information)
+  - The description should explain **why** each case is there. For example, what kinds of implementation mistakes might this case help us find?
 - 'variable names': one or more variable names with values which will be passed to the solution method
 - expected: the expected result (this would be -1 when we expect an exception)
 - msg: a nice message for the failing case
--- a/README.md
+++ b/README.md
@@ -55,7 +55,6 @@ Each test case has the the following keys:
   - if an error is expected (because the input is invalid, or any other reason), the value at `"expected"` should be an object containing exactly one property, `"error"`, whose value is a string.
     - The string should explain why the error would occur.
     - A particular track's implementation of the exercise **need not** necessarily check that the error includes that exact string as the cause, depending on what is idiomatic in the language (it may not be idiomatic to check strings for error messages).
-- msg: a nice message for the failing case
 
 ## Automated Tests
 
--- a/README.md
+++ b/README.md
@@ -50,7 +50,11 @@ Each test case has the the following keys:
   - The description should not simply explain **what** each case is (that is redundant information)
   - The description should explain **why** each case is there. For example, what kinds of implementation mistakes might this case help us find?
 - 'variable names': one or more variable names with values which will be passed to the solution method
-- expected: the expected result (this would be -1 when we expect an exception)
+- expected: the expected result
+  - if the input is valid but there is no result for the input, the value at `"expected"` should be `null`.
+  - if an error is expected (because the input is invalid, or any other reason), the value at `"expected"` should be an object containing exactly one property, `"error"`, whose value is a string.
+    - The string should explain why the error would occur.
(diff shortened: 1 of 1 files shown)


petertseng:
Previous discussion of descriptions: https://github.com/exercism/x-common/pull/451#issuecomment-263070019

README.md
```diff
@@ -47,9 +47,14 @@ variations.
 
 Each test case has the the following keys:
 - description: which will be used to name each generated test
+  - the description should not simply explain **what** each case is (that is redundant information)
+  - the description should explain **why** each case is there
+  - for example, what kinds of incorrect implementations might this case help us find?
 - 'variable names': one or more variable names with values which will be passed to the solution method
```
ErikSchierboom: I was wondering about the "will be passed to the solution method" part. Shouldn't that be "will be passed to the test method"?

ErikSchierboom:
This PR seems like a great improvement.

README.md
petertseng: so I guess here "solution method" means "method in the solution". In leap, we'd get `"year":1996` telling us to do `is_leap 1996` or `isLeap(1996)` or similar. I do think "solution method" makes more sense than "test method". Basically I think it's to make clear that it's what the student wrote and is submitting.

If it still bears improvement, maybe I could offer "function under test", or simply "the submitted function" (maybe too specific, if not all submissions are single functions). I would have also offered "the student's function" but that canonicalises the student word that I use, whereas other people use others (practitioner, exercist, user...)

ErikSchierboom: Ah I see. Good point then.

README.md
```diff
@@ -47,9 +47,14 @@ variations.
...
-- expected: the expected result (this would be -1 when we expect an exception)
-- msg: a nice message for the failing case
+- expected: the expected result
+  - if an error is expected, the value at `"expected"` should be an object containing exactly one property, `"error"`, whose value is a string.
```
rbasso: In this proposal, how should one differentiate between a missing value (optional return) and an error? Is it possible, or both should use the same structure?

ErikSchierboom: This is also a point of attention to me. In #336 I suggest to use `null` as the value to represent a missing value. Languages with support for optional types can then use their optional type instead of null.

README.md
```diff
@@ -47,9 +47,14 @@ variations.
 
 Each test case has the the following keys:
 - description: which will be used to name each generated test
+  - the description should not simply explain **what** each case is (that is redundant information)
+  - the description should explain **why** each case is there
+  - for example, what kinds of incorrect implementations might this case help us find?
```
petertseng: you're right, there is no good reason to put it on another

README.md
petertseng: let's talk at https://github.com/exercism/x-common/issues/336#issuecomment-280229280 shall we!

petertseng: Ah, I rubber-ducked myself into finding a reasonable answer. I've added the possibility of `null`, on the assumption that my guess is correct. I'm very open to corrections, of course.

ErikSchierboom:
I think this looks great now!

README.md
```diff
@@ -47,9 +47,14 @@ variations.
 
 Each test case has the the following keys:
 - description: which will be used to name each generated test
+  - The description should not simply explain **what** each case is (that is redundant information)
+  - The description should explain **why** each case is there. For example, what kinds of implementation mistakes might this case help us find?
 - 'variable names': one or more variable names with values which will be passed to the solution method
-- expected: the expected result (this would be -1 when we expect an exception)
-- msg: a nice message for the failing case
+- expected: the expected result
+  - if the input is valid but there is no result for the input, the value at `"expected"` should be `null`.
```
rbasso: While the last proposal in #336 doesn't enforce the existence of a `expected` property, if it is present, it can now also be `null`.

ErikSchierboom:
I think we can merge this now, right?

Issue #596, opened 2017-02-14
contributing guidelines: add PR recommendations


robphoenix:
In https://github.com/exercism/xgo/pull/512 @kytrinyx made some good PR recommendations, specifically:

1.  the subject line in the PR to be a one-sentence summary. This is something that should not have the word *and* (explicitly or implied).
2. to have the detail in the body of the PR.
3. For things that are "closes #XXX" or "See #XXX" I prefer to have that in the body of the PR, simply because the subject line doesn't support markdown, and these don't get turned into links. It makes it harder to follow to go look at the related issue or PR.

If these are generally accepted recommendations across the tracks it would be good to include them in the contributing guide. Along with any other similar general preferences for PR's/Commit messages/Issues.

behrtam:
You forgot the great link: https://chris.beams.io/posts/git-commit/

robphoenix:
yes, of course!

robphoenix:
closed by #598

PR #598, opened 2017-02-14
CONTRIBUTING.md: update pull request guidelines

robphoenix:
see #596

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -591,6 +591,9 @@ scenarios in this guide.
 
 1. Put the name of the exercise in the subject line of the commit.
    E.g. `hamming: Add test case for strands of unequal length`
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -591,13 +592,22 @@ scenarios in this guide.
 
 1. Put the name of the exercise in the subject line of the commit.
    E.g. `hamming: Add test case for strands of unequal length`
+1. The subject line should be a one-sentence summary. This is something that
```
ErikSchierboom: I think "This is something that should not include the word *and*" could be simplified to something like "It should not include the word *and*"

robphoenix:
I've updated, changing the line to 
`The subject line should be a one-sentence summary, and should not include the word *and* (explicitly or implied).`

PR #602, opened 2017-02-18
Add JSON Schema for 'canonical-data.json' and update the README.md

### LINKED ISSUE #336: canonical-data.json standardisation discussion (was: Malformed data?)
It appears that all-your-base.json is malformed. Where allergies.json has the structure of:

``` json
{
    "allergic_to": {
        "description": [ ... ],
        "cases": [ { "description": "...", ... } ... ]
            }, ...
}
```

all-your-base.json has:

``` json
{
  "#": [ ... ],
  "cases": [ ... ]
```

`cases` should be wrapped in a function name, yes?

It appears that bin/jsonlint _only_ checks that the json parses, not that it has good structure.

At the very least, I think this should be patched up and the README expanded to actually **show** the desired structure. Happy to do a PR for that, assuming I understand it already. 😀

rbasso:
In exercism/x-common#336 (`canonical-data.json` standardisation discussion), it was recently discussed a JSON Schema to capture the fundamental structure needed for `canonical-data.json` files.

After a lot of discussions and a few iterations, the [resulting schema](https://github.com/exercism/x-common/issues/336#issuecomment-280550277) proposed seems stable and mature enough to become at least a recommended standard, IMHO.

This proposal was designed with the following goals:

- Keep/improve human-readability.
- Keep it flexible enough to allow designing any reasonable test suite.
- Make it more regular and machine-readable.

There where naturally some trade-offs, as it is impossible to satisfy these objectives completely, but I think we got a good balance here.

To get a first impression of what is this about, in this new schema, a simple test suite would look like this:

```json
{ "exercise": "foobar"
, "version" : "1.0.0"
, "cases":
    [ { "description": "Foo'ing a word returns it reversed"
      , "property"   : "foo"
      , "input"      : "lion"
      , "expected"   : "noil"
      }
    , { "description": "Bar'ing a name returns its parts combined"
      , "property"   : "bar"
      , "firstName"  : "Alan"
      , "lastName"   : "Smithee"
      , "expected"   : "ASlmainthee"
      }
    ]
}
```

Of course,  there are more features, like comments, test groups and error signaling!

We expect that this change will soon allow us to:

- Automatically verify the `canonical-data.json` files.
- Simplify test generation, allowing more code to be shared among all exercises.
- Pave the way for future standardization. As our understanding of test suite design improves and new best practices emerge, it will be easy to patch this schema to capture even more structure.

We know that this proposal will not make it possible to generate a test suite in a fully automatic way. To allow that we would have to significantly change all the test suites, and [there was no agreement](https://github.com/exercism/x-common/issues/336#issuecomment-262184838) about how to do it or even if that is desirable.

Considering that this is a very sensible and highly debated topic, we shouldn't merge this PR until we are sure that enough people have a clear understanding of it and think this is an improvement over what we currently have.

We would like to know what do you think about this proposal. So, after studying it for a while, please...

- Say ❤️ if you think this is great!
- Say 👍 if you think this would make `x-common` better.
- Say 👎  if you think this would make `x-common` worse.
- If you have any question about it, please ask!

Also, if you think this isn't good the way it is, please consider helping us improve it!

Closes #336.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,21 +36,76 @@ There are three metadata files:
 ## Test Data Format (canonical-data.json)
 
 This data can be incorporated into test programs manually or extracted by a
-program.  The file contains a single JSON object with a key for documentation
-and keys for various tests that may be meaningful for a problem.
+program.  The file format is described in `canonical-schema.json`, but it
+is easier to understand with a example:
 
-The documentation uses the key "#" with a list of strings as the value.
-These strings document how the problem readme (`description.md`) is generally
-interpreted in test programs across different languages.  In addition to a
-mainstream implementation path, this information can also document significant
-variations.
+```json
+{ "exercise": "foobar"
+, "version" : "1.0.0"
+, "comments":
+    [ " Comments are always optional and can be used almost anywhere.      "
+    , "                                                                    "
+    , " They usually document how the exercise's readme ('description.md') "
+    , " is generally interpreted in test programs across different         "
+    , " languages.                                                         "
+    , "                                                                    "
+    , " In addition to a mainstream implementation path, this information  "
+    , " can also document significant variations.                          "
+    ]
+, "cases":
+    [ { "comments":
+          [ " A test case must have a 'description' and a 'property'.  "
+          , " Anything else is optional.                               "
+          , "                                                          "
+          , " The 'property' is a string in lowerCamelCase identifying "
+          , " the type of test, but most of the times it is just the   "
+          , " name of a function being tested.                         "
+          , "                                                          "
+          , " Test cases can have any number of additional keys, and   "
+          , " most of them also have an 'expected' one, defining the   "
+          , " value a test should return.                              "
+          ]
+      , "description": "Foo'ing a word returns it reversed"
+      , "property"   : "foo"
+      , "input"      : "lion"
+      , "expected"   : "noil"
+      }
+    , { "description": "Bar'ing a name returns its parts combined"
+      , "property"   : "bar"
+      , "firstName"  : "Alan"
+      , "lastName"   : "Smithee"
+      , "expected"   : "ASlmainthee"
+      }
+    , { "comments":
+          [ " Test cases can be arbitrarily grouped with a description "
+          , " to make organization easier.                             "
+          ]
+      , "description": "Abnormal inputs: numbers"
+      , "cases":
+          [ { "description": "Foo'ing a number returns nothing"
+            , "property"   : "foo"
+            , "input"      : "42"
+            , "expected"   : null
+            }
+          , { "description": "Bar'ing a name with numbers gives an error"
+            , "property"   : "bar"
+            , "firstName"  : "HAL"
+            , "lastName"   : "9000"
+            , "expected"   : { "error": "You should never bar a number" }
+            }
+          ]
+      }
+  ]
+}
 
-Each test case has the the following keys:
-- description: which will be used to name each generated test
-  - The description should not simply explain **what** each case is (that is redundant information)
-  - The description should explain **why** each case is there. For example, what kinds of implementation mistakes might this case help us find?
-- 'variable names': one or more variable names with values which will be passed to the solution method
-- expected: the expected result
+```
+
+Keep in mind that the description should not simply explain **what** each case
+is (that is redundant information) but also **why** each case is there. For
+example, what kinds of implementation mistakes might this case help us find?
+
+There are also some convention about `expected` that you must follow:
+
   - if the input is valid but there is no result for the input, the value at `"expected"` should be `null`.
   - if an error is expected (because the input is invalid, or any other reason), the value at `"expected"` should be an object containing exactly one property, `"error"`, whose value is a string.
     - The string should explain why the error would occur.
--- a/canonical-schema.json
+++ b/canonical-schema.json
@@ -1,0 +1,137 @@
+{
+   "comments":
+   [ " This is a JSON Schema for 'canonical-data.json' files.     "
+   , "                                                            "
+   , " It enforces just a general structure for all exercises,    "
+   , " without specifying how the test data should be organized   "
+   , " for each type of test. We do this to keep generality and   "
+   , " allow support for tests the do not fit well in the         "
+   , " 'function (input) == output' structure, like property      "
+   , " tests.                                                     "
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -36,21 +36,74 @@ There are three metadata files:
...
+    , { "description": "Bar'ing a name returns it duplicated"
+      , "property"   : "bar"
+      , "input"      : "Oink"
+      , "expected"   : "OinkOink"
+      }
+    , { "comments":
+          [ " Test cases can be arbitrarily grouped with a description "
+          , " to make organization easier.                             "
+          ]
+      , "description": "Abnormal input: empty strings and numbers"
```
ErikSchierboom: Would it perhaps make sense to also have an example with multiple input values?

rbasso: You're absolutely right, @ErikSchierboom! I'll fix that soon.

rbasso: Just changed the previous case to use two properties, `firstName` and `lastName`, as input. Was this what you had in mind, @ErikSchierboom?

ErikSchierboom: Originally, I just thought to pass an array to `input`, but I think this is more readable as you now have input value names. Furthermore, if one was to use an array to specify multiple input values, it would become odd to pass an array.

One other option might be to make the content of `input`  be an object. So something like this:

- test method with no parameters: `"input": {}`
- test method with one parameter: `"input": { "name": "john" }`
- test method with more than one parameter: `"input": { "name": "john", "age": 18 }`
- test method with one parameter that is an array: `"input": { "allergies": ["fish", "peanuts"] }`

What do you think about that?

rbasso: It really makes sense but, *considering that we decided not to standardize the input data for now*, I used distinct key names to showcase that this is possible. This way people will not be mislead thinking that they need to have an `input` key in every test. Only `description` and `property` are mandatory.

rbasso: I'm afraid that, if we try to standardize too much now, more people will start to disagree and the chances of getting this approve will go down.

My target is to get a general structure approved, leaving all the details for other discussions.

We already added the restrictions to `error`, which are great, but they make this PR even harder for people to understand. I'm really not sure if trying to add more things now is good idea...

ErikSchierboom: Point taken. Let's leave it as it is then.

rbasso: Just for the sake of discussion, I found a few possible objection to adding an `input` object:

- We don't really need it. All additional keys in a test case are already implicitly inputs to the test.
- It decreases human-readability, adding an unneeded `input` key and increasing one level of nesting.
- Google JSON Style Guide recommends avoiding unneeded objects.

Following the same reasoning, it would also make more sense to have either a `expected` or an `error`. Maybe we shouldn't have put the `error` inside `expected`.

rbasso: Just for the sake of discussion, I found a few possible objection to adding an `input` object:

- We don't really need it. All additional keys in a test case are already implicitly inputs to the test.
- It decreases human-readability, adding an unneeded `input` key and increasing one level of nesting.
- Google JSON Style Guide recommends avoiding unneeded objects.
- It would not fit more general property tests well, and we are trying to be as general as possible here.

Following the same reasoning, it would also make ~~more~~ sense to have either a `expected` or an `error`. ~~Maybe we shouldn't have put the `error` inside `expected`~~.

**Edit:** I think that `error` is nice the way it is now, and we are only enforcing it's structure if it is present in the `expected`, so we are not really losing anything with the current proposal, as it allows still allows other representation for errors. If in the future we decide that it is inconvenient, we can change the recommended encoding, but for now we already agreed on that.

rbasso:
I forgot to mention @exercism/track-maintainers in the first message.

README.md
```diff
@@ -36,21 +36,76 @@ There are three metadata files:
...
+      , "description": "Foo'ing a word returns it reversed"
+      , "property"   : "foo"
+      , "input"      : "lion"
+      , "expected"   : "noil"
+      }
+    , { "description": "Bar'ing a name returns its parts combined"
+      , "property"   : "bar"
+      , "firstName"  : "Alan"
+      , "lastName"   : "Smithee"
+      , "expected"   : "Aslmainthee"
```
kotp: The example suggests more than the description states.  Should be "ASlmainthee" I believe.

rbasso: Why?!

I just kept the first in upper case , because is looks more like a real name.

Anyway, I'll change it if you prefer it the other way.

kotp: Because the description states that it returns the parts combined, not modified and combined.

rbasso: Fixed! Thanks! 👍

stkent:
I wonder if it would be useful to also include a schema version in canonical data? Just an idle thought. But anyway, LGTM!

catb0t:
I really love this. I was the lone collaborator on #336 arguing for more machine-readability (to strike a balance with human-readability) and this is absolutely beautiful IMO.

rbasso:
> I wonder if it would be useful to also include a schema version in canonical data? Just an idle thought. But anyway, LGTM!

Do you know what is the recommended way to do it, @stkent? The best I could find was [this link](https://github.com/snowplow/iglu/wiki/Self-describing-JSONs).

So, unless there is a more generally accepted versioning style, I'm considering following the directions on that link and add the following to schema:

```json
   "self": { "vendor" : "io.exercism"
           , "name"   : "canonical-data"
           , "format" : "jsonschema"
           , "version": "1-0-0"
           },
```

What do you think about it?

rbasso:
@stkent, I just updated the PR to including the schema version according to [this specification](https://github.com/snowplow/iglu/wiki/Self-describing-JSONs). Of course we can change the format later, but it seems a good idea to start versioning the schema right now.

---

@catb0t wrote:
> I really love this. I was the lone collaborator on #336 arguing for more machine-readability (to strike a balance with human-readability) and this is absolutely beautiful IMO.

I'm glad that you liked it!

---

Despite some minor modifications, in the last 15 days we had enough support for this schema and it seems there are no major changes needed here, so I'm planning merging this PR in a few hours, after taking a last look at it.

rbasso:
Here is [a link](https://jsonschemalint.com/#/version/draft-04/markup/json?gist=ee4f0954310119575c1737e09fff680c) which can be used to easily validate a `canonical-data.json` file against the schema.

ErikSchierboom:
🎉 !

PR #603, opened 2017-02-18
Fix etl exercise canonical-data

aawray:
Wrong order of letters in the expected output


NobbZ:
Personally I do prefer ordering by keys rather than by values. Even while I
assume that there is no ordering in the implemantation I feel more
comfortable with sorted keys.

Also I think I should make clear that by ordered I mean even grouping by
semantic, but that doesn't apply for the proposed change.

Peter Tseng <notifications@github.com> schrieb am Sa., 18. Feb. 2017, 19:06:

exercises/etl/canonical-data.json
```diff
@@ -39,8 +39,8 @@
         },
         "expected": {
           "a" : 1,
-          "e" : 1,
           "d" : 2,
+          "e" : 1,
           "g" : 2
         }
```
rbasso: Property's ordering has no meaning in JSON, but as this was the only object not *key-sorted*, it makes sense to sort it to keep consistency.

Issue #537, opened 2017-02-06
[run-length-encoding] test case conflicts with description


daveyarwood:
The [description](https://github.com/exercism/x-common/blob/master/exercises/run-length-encoding/description.md) of the run-length-encoding exercise says:

> For simplicity, you can assume that the unencoded string will only contain the letters A through Z.

However, there is [one test case](https://github.com/exercism/x-common/blob/master/exercises/run-length-encoding/canonical-data.json#L47-L49) that contains spaces and lowercase letters.

I believe the test case should be changed to something like `OEJJJJJXAAPPPL`, the intent being that all of the test cases deal with strings containing only uppercase letters. I could make a PR to fix this test case, if you'd like?

robkeim:
Thanks @daveyarwood for pointing this out!

I added that description trying to fix #238, but as you pointed out my description was not complete. I think the correct thing to do would be to update the description to say something like:
"only letters A through Z (upper and lower case) as well as white space".

daveyarwood:
I wonder if that would create more work for the individual language repos? The example solutions would all need to be updated to handle spaces in the input.

robkeim:
Tracks that have already implemented this exercise should have been taking the existing whitespace into account (assuming that they're following the canonical data). Therefore, I think it makes more sense to not modify the canonical data and to update the description.

daveyarwood:
Ah, that's a good point. 👍

robkeim:
If you'd like to send a PR with an updated description go right ahead, otherwise let me know and I'll give it another shot.

daveyarwood:
I'm happy to take a stab at it.

Question, though: what about the lowercase z's? Are we considering valid input to be only uppercase letters and whitespace?

robkeim:
My understanding, given the current canonical data, is that we're accepting all ASCII characters except for the numbers, although the tests only test A through Z both upper and lowercase.

Here's another attempt (tell me what you think):
> For simplicity, you can assume that the unencoded string only contains whitespace and the letters A through Z both upper and lowercase.

daveyarwood:
That sounds good to me 👍 

I'll make a PR soon to update the description when I have a little bit of free time.

PR #610, opened 2017-02-23
run-length-encoding: Add whitespace tests

DesHorsley:
Adds tests to check encoding and decoding of whitespace characters leaves the characters unchanged as per the description.md instructions.


behrtam:
Just discovered #537 which changed the way white space is handled.

So we really want to change all the implementations in the tracks as for example python treats white spaces like other chars, so the result would look like this `encode("  hsqq qww  ") == '2 hs2q q2w2 '`.

DesHorsley:
I agree that there is no reason to ignore it, other than my understanding of the description of the exersize, stating the whitespace should pass through unchanged. Perhaps the description should be amended to make it clear that white space should be treated as any other character?

DesHorsley:
@behrtam how would you like me to proceed? I'm happy to contribute where I can. 

Is it just a case of updating the exercise description to remove the ambiguity and state whitespace is treated like the other characters?

DesHorsley:
Closed as per #613

PR #614, opened 2017-02-27
hello-world: clarify "says"

kevinlitchfield:
I think it could be worthwhile to clarify what "says" means in the exercise description. Often, "Hello, World!" programs print `Hello, World!` to the screen, but the Exercism exercises I've seen instead ask for a function that returns the string `Hello, World!`.

Because "says" could sound like it's figuratively referring to printing to the screen, and because a learner familiar with "Hello, World!" programs might expect to write a program that prints `Hello, World!` to the screen, I think "returns the string" might be clearer here.

Thoughts?

### DIFF
--- a/exercises/hello-world/description.md
+++ b/exercises/hello-world/description.md
@@ -4,7 +4,7 @@ or environment.
 
 The objectives are simple:
 
-- Write a function that says "Hello, World!".
+- Write a function that returns the string "Hello, World!".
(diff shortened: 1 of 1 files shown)


stkent:
+1 this. I think there are other exercises that could also benefit from the same adjustment (`diamond` springs to mind; I think that explicitly uses "prints", but most (all?) implementations return a list of strings).

Issue #591, opened 2017-02-13
word-search: Implement canonical-data.json


kytrinyx:
We want to have a standard set of test inputs and outputs for each exercise to
make it easier to port them to new languages, as well as to help keep the
exercises in sync.

The Word Search exercise can be found in the [./exercises/word-search/][exercise-dir] directory.

The step-by-step instructions for how to compile a canonical-data.json file is
described in [this section of the contributing guide][contributing].

See http://exercism.io/contribute/canonical-data/word-search for the up-to-date list of
language tracks that have an implementation of the Word Search exercise.

The [tracking issue][tracking-issue] has more context.

[contributing]: https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#improving-consistency-by-extracting-shared-test-data
[tracking-issue]: https://github.com/exercism/x-common/issues/552
[exercise-dir]: https://github.com/exercism/x-common/tree/master/exercises/word-search

stkent:
I'll tackle this soon.

petertseng:
Is it the case that the C# and F# implementations deliberately test all eight directions?

If so, should that become the standard?

Does the existing example grid test all eight directions?

stkent:
@petertseng I was just diving in to that, and yes: C# and F# test all directions, and C# (where I started reading) also checks a grid of a different size:

```
qwertyuiopz
luamsicrexe
abcdefghijk
```

stkent:
C# also tests that words not in the puzzle are not found 👍

stkent:
Ok, @petertseng, you nailed it: Go checks 4 of the 8 available directions.

E
S
W
NW

I'll use the updated version of the grid for the canonical tests.

petertseng:
Perhaps the example grid can also get changed, then.

The other interesting corner case I would recommend is... literally the corner case. Check that a word is found even if one of its letters is in the bottom-right corner. Just checking those off-by-one errors.

ErikSchierboom:
@stkent I'm not completely sure, but I think I copied the existing tests from the Python test suite, where I must have failed to regard the example data in the description. I do think that testing for all directions makes sense here, as that is also how one usually solves these puzzles. Having some corner cases is always good, so I'm all for that!

stkent:
Here's a visualization of the current (newer) solutions. All corners are accounted for!

[...]

PR #617, opened 2017-03-01
word-search: update description example to match most common impl

stkent:
See #591 for context.


ErikSchierboom:
Good work @stkent! :tada:

PR #619, opened 2017-03-01
perfect-numbers: fix misleading test description

stkent:
Original description

> Non-negative integer is rejected (not a natural number)

could be interpreted as suggesting that _any_ non-negative integer should be rejected. This is not the case. New description

> Zero is rejected (not a natural number)

reflects intent much better.


ErikSchierboom:
Thanks! :tada:

PR #620, opened 2017-03-02
change: Added a positive case without 1 unit coin available

weapp:
All the current positive test cases include 1 unit coin in available coins and it's could be confusing.

This commit proposes a new test case that is positive but not includes 1 unit coin in the input.

### DIFF
--- a/exercises/change/canonical-data.json
+++ b/exercises/change/canonical-data.json
@@ -37,6 +37,12 @@
                 "expected": [2, 2, 5, 20, 20, 50, 100, 100, 100, 100, 100, 100, 100, 100, 100]
             },
             {
+                "description": "possible change without unit coins available",
+                "coins": [2, 5, 10, 20, 50],
+                "target": 21,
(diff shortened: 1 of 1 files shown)


exercises/change/canonical-data.json
```diff
@@ -37,6 +37,12 @@
                 "expected": [2, 2, 5, 20, 20, 50, 100, 100, 100, 100, 100, 100, 100, 100, 100]
             },
             {
+                "description": "posible change without one coins available",
```
rpottsoh: @weapp just a small grammatical nit pick, `coins` should be singular instead of plural.  I like the added twist of not having a penny available.  Thanks

petertseng: "coins of value one"?

"unit coins"?

petertseng: "posible" -> "possible"

ErikSchierboom:
Thanks a lot @weapp! :tada:

PR #621, opened 2017-03-03
TOPICS.txt: add number systems

robphoenix:
number systems can cover such problem areas as working with binary
and hexadecimal numbers, for instance in the secret-handshake exercise.


robphoenix:
This is covered well enough by bitwise operations.

Issue #613, opened 2017-02-26
run-length-encoding: How to deal with white spaces


behrtam:
`<history>`
To understand how we got here I tried to figure out the history of the changes the happened to the run length encoding exercise.

In #238 the problem with numerical data for run length encoding input was raised and to keep the exercise simple it was decided to not include numbers.
This was "incorrectly" fixed in #530. Instead of excluding numbers, it reduced the input data range down to only upper case A-Z. There is no reason to exclude lower case characters and there was already test input data with white space, which is also totally fine.

This discrepancy between the new description and the old test data was discovered in #537. With #538 "only A-Z" was changed to include lower case and white spaces. But this change also introduced a new inconsistency because it assumed incorrectly that white spaces are left unchanged.
It is neither necessary to leave white spaces unchanged (there are some simple ciphers/encodings that relay on this), nor was it in accordance with the existing test data. 
`</history>`

*Status quo*: I also checked some language tracks (Python, Javascript, Ruby, Scala) and they all encode white spaces.

So, here is what I would like to propose we do ...

- [x] remove the part about "unchanged white space" from the description
- [x] close #610 

and maybe

- [x] add an explicit test case with lower case characters
- [x] add an explicit test case with white spaces

DesHorsley:
Sounds good I'm happy to run with this if you'd like a hand.

daveyarwood:
Good catch! I am on board with all of this, especially the explicit test cases for whitespace and lowercase characters.

behrtam:
Okay, than lets get this ready to roll. @DesHorsley would you like to add the two test cases to the canonical test data? I would work on the description if no one else wants to do that.

DesHorsley:
Sounds good

robkeim:
Good detective work @behrtam to track this one down!

I like your proposed changes, although in #624 I don't see the description being updated for the exercise.

kotp:
Thanks @behrtam and @robkeim and @DesHorsley

PR #624, opened 2017-03-04
Adds tests for whitespace and lowercase characters for the "Run Length Encoding" exersize

DesHorsley:
As per issue #613


stevejb71:
Could you say which exercise you are changing in the commit message. 
My reading of the description is that whitespace is not encoded. Could you update it to be clearer, and also say what type of whitespace is handled (just ascii 32 I guess?)

DesHorsley:
Hi @stevejb71 , 
I've updated the commit message, @behrtam is planning to update the description to clarify whitespace usage as per #613

Issue #625, opened 2017-03-06
canonical-data.json: Compliance with JSON Schema


rbasso:
Now that #602 was merged, it would be great to have all the `canonical-data.json` files compliant with the JSON Schema, so that we could automatically check them for errors in the future.

At the moment, we have 74 `canonical-data.json` files, but only 3 are compliant.

When porting the files, I guess it would be nice to follow a few guidelines to make reviewing easier:
- Distinct commits with the following:
    1. `exercise-name: Make canonical-data.json compliant`, containing the minimal changes to make the test suite compliant.
    2. `exercise-name: Fix canonical-data.json formatting`, with the changes needed to make it readable again.
- pull-request's title: `exercise-name: Make exercise schema-compliant`
- pull-request's body: at least a line containing `Related to #625`.
- `version` should start at **1.0.0** and follow the `exercise` property.

Unless there is already a standard in place that I don't know, I suggest formatting with 2-spaces indentation in the second commit, for consistency, which is already the default when running `jq '.' canonical-data.json`.

Of course, most of the exercises will demand additional changes to a standard formatting, to make them more compact and readable.

**Edit:** Just to be clear, this issue is **not about style** in designing the test suite. *Changes that go beyond fixing the test suite to be compliant and fix it to be consistently formatted after that should be avoided here!*

---

#### `canonical-data.json` compliance

- [x] acronym #626
- [x] allergies #628
- [x] all-your-base #630
- [x] alphametics #675 
- [x] anagram #631
- [x] atbash-cipher #632 
- [x] beer-song #661 / #694 
- [x] binary #644 
- [x] binary-search #676 
- [x] bob #636
- [x] book-store #635
- [x] bowling #677 
- [x] bracket-push #680 
- [x] change #641 
- [x] circular-buffer #681 
- [x] clock #669 
- [x] connect #670 
- [x] crypto-square #652 
- [x] custom-set #672 
- [x] diamond #612
- [x] difference-of-squares #653 
- [x] dominoes #682 
- [x] etl #654 
- [x] flatten-array #683 
- [x] food-chain #684 
- [x] forth #688 
- [x] gigasecond #685 
- [x] grains #642 
- [x] grep #712
- [x] hamming #627
- [x] hello-world #629 
  - [x] Need bump to 1.0.0
- [x] isogram #686 
- [x] largest-series-product #689 
- [x] leap #633
- [x] list-ops #690 
- [x] luhn #643
- [x] markdown #691 
- [x] meetup #663 
[...]

PR #628, opened 2017-03-06
allergies: Make exercise schema-compliant

rbasso:
Related to #625.

### DIFF
--- a/exercises/allergies/canonical-data.json
+++ b/exercises/allergies/canonical-data.json
@@ -3,125 +3,147 @@
   "version": "1.0.0",
   "cases": [
     {
-        "description": "allergicTo",
-        "comments": [
-            "Given a number and a substance, indicate whether Tom is allergic ",
-            "to that substance.",
-            "Test cases for this method involve more than one assertion.",
-            "Each case in 'expected' specifies what the method should return for",
-            "the given substance."
-        ],
-        "cases": [
+      "description": "allergicTo",
+      "comments": [
+        "Given a number and a substance, indicate whether Tom is allergic ",
+        "to that substance.",
+        "Test cases for this method involve more than one assertion.",
+        "Each case in 'expected' specifies what the method should return for",
+        "the given substance."
+      ],
+      "cases": [
+        {
+          "description": "no allergies means not allergic",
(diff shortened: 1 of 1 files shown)


exercises/allergies/canonical-data.json
```diff
@@ -1,109 +1,149 @@
...
-            "Each case in 'expected' specifies what the method should return for",
-            "the given substance."
-        ],
-        "cases": [
+  "exercise": "allergies",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "allergicTo",
+      "comments": [
```
catb0t: Am I misreading the spec, or doesn't `comments` go at the top level? 

```
      "canonicalData":
          { "description": "This is the top-level file structure"
          , "type"       : "object"
          , "required"   : ["exercise" , "version", "cases"]
          , "properties" :
                { "exercise" : { "$ref": "#/definitions/exercise"  }
                , "version"  : { "$ref": "#/definitions/version"   }
                , "comments" : { "$ref": "#/definitions/comments"  }
                , "cases"    : { "$ref": "#/definitions/testGroup" }
                }
          , "additionalProperties": false
          },
```

rbasso: We can have comments in the top level, in a test group (cases), or inside a test, so that people can keep them where they fit better. It this case, It is in the test group because it pertains just that group of test.

catb0t: That makes sense, thanks

kotp:
@rbasso is this diving into the "more than just formatting and compliance"?

rbasso:
> @rbasso is this diving into the "more than just formatting and compliance"?

This PR was not supposed to cross that line, @kotp. What is the offending part?

kotp:
At first glance it appears it is grouping expectations in a different way.  I will look closer though.

rbasso:
I was forced to move the `description` to `comments`, because one of them was a list of strings. This also forced me to add other descriptions, because the schema doesn't allow test groups without descriptions (they wouldn't make sense for the student)

kotp:
That now makes sense.

PR #633, opened 2017-03-06
leap: Make exercise schema-compliant

catb0t:
Related to #625.

### DIFF
--- a/exercises/leap/canonical-data.json
+++ b/exercises/leap/canonical-data.json
@@ -1,30 +1,30 @@
 {
   "exercise": "leap",
   "version": "1.0.0",
-   "cases": [
-      {
-         "description": "year not divisible by 4: common year",
-         "property": "leapYear",
-         "input": 2015,
-         "expected": false
-      },
-      {
-         "description": "year divisible by 4, not divisible by 100: leap year",
-         "property": "leapYear",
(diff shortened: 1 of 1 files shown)


exercises/leap/canonical-data.json
```diff
@@ -1,24 +1,34 @@
...
-         "input": 2000,
-         "expected": true
-      }
-   ]
+  "exercise": "leap",
+  "version": "1.0.0",
+  "comments": [
+    "Given a Gregorian calendar year, indicate whether it is a leap year."
+  ],
+
```
rbasso: This comment was not in the original test suite and also is not needed for compliance, so maybe it should not be part of this commit/PR.

exercises/leap/canonical-data.json
```diff
@@ -1,24 +1,34 @@
...
+  "cases": [
+    {
+       "description": "year not divisible by 4: common year",
+       "property": "notLeap",
```
rbasso: #### About `isLeap` and `notLeap`

All the tests in this test suite use exactly the same test logic, only changing the test data, so they should have the same `property` value. `isLeapYear`and `leapYear` are good candidates

catb0t:
Oh, ok -- hopefully this is more what you wanted.

PR #634, opened 2017-03-06
word-count: Make exercise schema-compliant 

catb0t:
Related to #625.


kotp:
I brought it in, only as I know this will occasionally happen.

PR #640, opened 2017-03-06
run-length-encoding: Fix incorrect description

behrtam:
The description incorrectly stated that white spaces 
would be passed through unchanged (details #613).


kotp:
According to the consistency case, the spaces should go through unchanged.

```json
   "consistency": {
      "cases": [
          {
            "description": "encode followed by decode gives original string",
            "input": "zzz ZZ  zZ",
            "expected": "zzz ZZ  zZ"
          }
      ]
   }
```

PR #645, opened 2017-03-07
trinary: Make exercise schema-compliant

rbasso:
Related to #625.

### DIFF
--- a/exercises/trinary/canonical-data.json
+++ b/exercises/trinary/canonical-data.json
@@ -1,62 +1,77 @@
 {
-  "to_decimal": {
+  "exercise": "trinary",
+  "version": "1.0.0",
+  "cases": [
+  {
     "description": "returns the decimal representation of the input trinary value",
     "cases": [
       {
         "description": "trinary 1 is decimal 1",
+        "property": "toDecimal",
         "input": 1,
         "expected": 1
       },
       {
         "description": "trinary 2 is decimal 2",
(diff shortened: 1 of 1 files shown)


exercises/trinary/canonical-data.json
```diff
@@ -1,62 +1,77 @@
...
-  "to_decimal": {
+  "exercise": "trinary",
+  "version": "1.0.0",
+  "cases": [
+  {
     "description": "returns the decimal representation of the input trinary value",
     "cases": [
       {
         "description": "trinary 1 is decimal 1",
+        "property": "toDecimal",
```
kotp: Shouldn't the property simply be "Decimal" or even "Base10"?

rbasso: In the original file, the test group was named `to_decimal`, so I just adapted it to make it a valid property according to the JSON Schema: `toDecimal`.

IMHO, `decimal`, `toDecimal` and `base10` do not capture what it means. The output is a number here, not a string, so there is no sense talking about what is its base. It is represented in JSON as a decimal, but that is just a convenience for visualization. 

if I where to choose I name, it would probably be `fromTrinary`.

Summarizing, I just tried to minimize the number of changes. But I don't think the name is good by any criteria.

kotp:
As an exercise that will soon be deprecated, this is definitely good enough.

PR #655, opened 2017-03-08
queen-attack: Make exercise schema-compliant

rbasso:
Related to #625.

### DIFF
--- a/exercises/queen-attack/canonical-data.json
+++ b/exercises/queen-attack/canonical-data.json
@@ -11,132 +11,132 @@
     "tests can be offered as extra credit"
   ],
   "cases": [
-  {
-    "description": "Test creation of Queens with valid and invalid positions",
-    "cases" : [
-      {
-        "description": "queen with a valid position",
-        "property": "create",
-        "queen": {
-          "position": "(2,2)"
+    {
+      "description": "Test creation of Queens with valid and invalid positions",
+      "cases": [
+        {
+          "description": "queen with a valid position",
+          "property": "create",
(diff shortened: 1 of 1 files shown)


exercises/queen-attack/canonical-data.json
```diff
@@ -8,119 +10,133 @@
...
-        "description": "queen must have positive rank",
-        "queen": {
-          "position": "(-2,2)"
+        {
+          "description": "queen must have positive rank",
+          "property": "create",
+          "queen": {
+            "position": "(-2,2)"
+          },
+          "expected": -1
```
ErikSchierboom: Shouldn't `-1` be replaced with an `error` object? If so, the description in the "comments" key should also be updated.

rbasso: I agree you you. It should be an `error`, or maybe `null`.

A lot of other test suites have exactly the same problem, but in #625 we just making changes that do not affect the contents of the test suite, only changes in structure (with some unavoidable small exceptions), so that they it formally complies with the JSON schema.

Maybe we should open a new force after finishing this one to make sure that all the test suite follow the best practices and error encoding conventions. For now, I would really like to avoid it, because even the mechanical work of adapting them to the new schema is already taking a lot of time. 😩

ErikSchierboom: I see. Let's do what you suggested then.

rbasso: BTW, thanks for reviewing all this.

PR #661, opened 2017-03-09
beer-song: Make canonical data more logical

kytrinyx:
We were testing:
- the first verse
- (*) some random verse in the middle
- each of the special case verses
- (*) 4 consecutive verses
- all verses

The two that are marked with an asterisk are not very logical choices in
terms of TDD.

The first asterisk is about the generation of individual verses.

The structure of the algorithm is that verses 99 through 3 all have the
same implementation. Then verses 2, 1, and 0 are all unique.

Instead of testing the first generic verse and some random generic
verse, it makes more sense to test the outer bounds: the first (99) and
the last (3).

The second asterisk is about generating consecutive verses.

When testing multiples, the natural progression isn't "4" then "all".

The best test data is _the smallest number that is multiple_, which is
two. Since that can be implemented with a simple concatenation of the
upper and lower bounds (verse of N + verse of M), it's useful to add
an extra case that pushes towards a loop.

Again, we want the smallest number that suggests a loop, which is three
consecutive verses, not four.

Then, finally, we have the full integration test, which is all the verses.


kotp:
PR #662 is set to merge into this branch...

exercises/beer-song/canonical-data.json
```diff
@@ -1,49 +1,77 @@
...
-      }
-    ]
-  }
+  "exercise": "beer-run",
```
rbasso: I guess that the exercise's name should be `beer-song`, which is the slug.

kotp: meh, not sure why I put that in there, I know it is beer-song.  I can fix it now

kotp: Thanks @rbasso good catch.  Fixed.

kytrinyx:
What does is the "property" property supposed to contain?

We have the section `"verse"` which has the property `"verse"`, and the section `"lyrics"` which has property `"verses"`.

rbasso:
I'll check that now, @kytrinyx.

rbasso:
Ops! I guess left that one slip when reviewing. Sorry! Just opened #694 to fix that.

> What does is the "property" property supposed to contain?

The `property` is just a camelCase string that uniquely identifies a test type. The name `testType` was not popular at the time, so we settled with `property` because it was considered general enough to capture any kind of tests.

I agree that the name is a bit strange at first, because most people are used to tests in the form `f (input) == expected`, for explicit values of `input` and `expected`. The natural naming for these simple tests would be `function`.

But it is easy to create test that do not fit in that format: `f (inputA) == f (inputB)`, or `f (inputA) == g (inputB)`. There is artificial to use `function` in these, so we needed a more general name.

Also, there is the extreme case of property-based testing, which we don't have yet, AFAIK, but is the most general case.

Putting in simple terms, we are always testing some *property* of the solution, and hence this name Makes sense?

kytrinyx:
> Putting in simple terms, we are always testing some property of the solution, and hence this name. Makes sense?

That makes sense. Thanks for clarifying!

PR #663, opened 2017-03-09
meetup: Make exercise schema-compliant

rbasso:
Related to #625.


rbasso:
Are you looking at the individual commits, @ErikSchierboom?
The second one should only fix formatting, but sometimes the GitHub interface doesn't match/highlight the corresponding parts and changes.

PR #671, opened 2017-03-09
anagram: Rename duplicated test case description

behrtam:
There are some tracks that generate their test suits and use the description field as name for the test function/method.
If this would get merged. Do we change the version to 1.1.0 or 1.0.1?

*Note: Don't know if the new names are so great but I couldn't find different ones.*

### DIFF
--- a/exercises/anagram/canonical-data.json
+++ b/exercises/anagram/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "anagram",
-  "version": "1.0.0",
+  "version": "1.0.1",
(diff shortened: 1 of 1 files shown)


rbasso:
> If this would get merged. Do we change the version to 1.1.0 or 1.0.1?

Great question @behrtam!

I guess it is time for us to discuss how to do semantic versioning of the `canonical-data.json` files.

rbasso:
Hold this for a while! I'm opening an issue to discuss that!

PR #672, opened 2017-03-09
custom-set: Make exercise schema-compliant

rbasso:
Related to #625.

### DIFF
--- a/exercises/custom-set/canonical-data.json
+++ b/exercises/custom-set/canonical-data.json
@@ -1,318 +1,318 @@
 {
-   "exercise": "custom-set",
-   "version": "1.0.0",
-   "comments": [
-     "These tests cover the core components of a set data structure: checking",
-     "presence, adding, comparing and basic set operations. Other features",
-     "such as deleting elements, checking size, sorting are not tested, but",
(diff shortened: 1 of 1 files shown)


exercises/custom-set/canonical-data.json
```diff
@@ -1,277 +1,318 @@
...
+          "set": [1, 2, 3],
+          "element": 3,
+          "expected": [1, 2, 3]
+        }
       ]
-   },
-   "intersection": {
+    },
+    {
       "description": "Intersect returns a set of all shared elements",
```
ErikSchierboom: The property is named `intersection`, should we also change the description to refer to `Intersection`?

rbasso: Agreed! 😄 
I just opened #674 for it. Now that we have a draft of semantic versioning, it may be a good idea to start enforcing it.

kotp:
Thanks for the review @ErikSchierboom ...

exercises/custom-set/canonical-data.json
ErikSchierboom: @rbasso Good point! :)

PR #685, opened 2017-03-10
Fix "Gigasecond: Schema Compliance"

kotp:
Reverts exercism/x-common#679


kotp:
ref: #625

tejasbubane:
I think we should do `Fix Compliance` instead of revert and then add new again.

kotp:
Yeah, but I think we can work from this one.  This will mean less of a 'diff'.

kotp:
Thanks @rbasso and @tejasbubane !

PR #688, opened 2017-03-10
forth: Make exercise schema-compliant

rbasso:
Related to #625.

### DIFF
--- a/exercises/forth/canonical-data.json
+++ b/exercises/forth/canonical-data.json
@@ -1,18 +1,24 @@
 {
-  "#": [
+  "exercise": "forth",
+  "version": "1.0.0",
+  "comments": [
     "The cases are split into multiple sections, all with the same structure.",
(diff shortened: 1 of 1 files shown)


petertseng:
>  I do question the `property`

I think there only needs to be a single value for it, `"evaluate"`

rbasso:
You are completely right, @petertseng! Thanks for catching that! 👍

rbasso:
You are completely right, @petertseng! Thanks for catching that! 👍 

Any recommendation for the property name? `run`? `forth`?

petertseng:
My initial thought was `evaluate`, but your suggestions of `run` or `forth` seem reasonable as well. Since all seem reasonable, I suppose we can just pick one, since there will only be one for the exercise (unless we make changes! in which case we can rediscuss then)

rbasso:
Rewritten with `evaluate`, @petertseng!

PR #704, opened 2017-03-11
pov: Make exercise schema-compliant

petertseng:
You probably want to review commit-by-commit (but hey don't let me tell you what to do, if you want to review the whole thing at once you are free to do it!).

I suppose there can still be a reformatting too (you'll notice a `"cases": [{`) line but...

1. Since `testGroup` is necessarily and both types that might be elements of it are objects, I have absolutely no compunction about putting the `[{` on the same line!
    * Unfortunately, I did not apply the same treatment to the inner `cases`, sorry.
2. Keeping that diff small!
3. Okay, if you *really* want me to reformat, I can, but I wanted to make sure that there are no *other* changes that people want, because then I would have to rebase and blow away my reformatting.

### DIFF
--- a/exercises/pov/canonical-data.json
+++ b/exercises/pov/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "pov",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "cases": [{
     "description": "Reroot a tree so that its root is the specified node.",
     "comments": [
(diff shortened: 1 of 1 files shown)


petertseng:
I find the commit message of "replace unnecessary uses of cousin with smaller tree" lacking. I am going to amend its commit message. The contents will not otherwise change, so feel free to continue your review unimpeded.

petertseng:
That is better. By the way, you could maybe convince me to make some of those "cousins" replacements smaller. As I explained in the commit message, it's not really interesting to have a large tree for an error case.

petertseng:
I've been suggested to bump this to 1.1.0

By the way, I prefer this NOT be squashed on merge.

petertseng:
Q: You changed the test inputs. How do we know the new outputs match the new inputs? The schema check only checks for conformance to the schema, and not the actual correctness of the test data.

A: I [did it before](https://github.com/exercism/x-common/pull/390#issuecomment-250947901), and I'll do it again.

https://github.com/petertseng/x-common/blob/verify/exercises/pov/verify.rb

petertseng:
3. Okay, if you *really* want me to reformat, I can, but I wanted to make sure that there are no *other* changes that people want, because then I would have to rebase and blow away my reformatting.

PR #451, opened 2016-11-26
Add descriptions to Roman numerals canonical data.

stevejb71:
Descriptions could be worked out programatically (or by a human), but:
1. this entails more work for an automated test generator, which is not needed elsewhere as most (if not all) other data files provide descriptions.
2. descriptions won't be consistent across streams

PR #707, opened 2017-03-11
roman-numerals: add descriptions

petertseng:
We'd like some descriptions so that roman-numerals can be brought into
compliance with the schema, as in #625.

Closes #451 via replacement

### DIFF
--- a/exercises/roman-numerals/canonical-data.json
+++ b/exercises/roman-numerals/canonical-data.json
@@ -1,74 +1,92 @@
 {
    "cases": [
       {
+         "description": "1 is a single I",
          "number" : 1, 
          "expected": "I"
       },
       {
+         "description": "2 is two I's",
          "number" : 2, 
          "expected": "II"
       },
       {
+         "description": "3 is three I's",
          "number" : 3, 
          "expected": "III"
       },
       {
+         "description": "4, being 5 - 1, is IV",
          "number" : 4, 
          "expected": "IV"
       },
       {
+         "description": "5 is a single V",
          "number" : 5, 
          "expected": "V"
       },
       {
+         "description": "6, being 5 + 1, is VI",
          "number" : 6, 
          "expected": "VI"
       },
       {
+         "description": "9, being 10 - 1, is IX",
          "number" : 9, 
          "expected": "IX"
       },
       {
+         "description": "20 is two X's",
          "number" : 27, 
          "expected": "XXVII"
       },
       {
+         "description": "48 is not 50 - 2 but rather 40 + 8",
          "number" : 48, 
          "expected": "XLVIII"
       },
       {
+         "description": "50 is a single L",
          "number" : 59, 
          "expected": "LIX"
       },
       {
+         "description": "90, being 100 - 10, is XC",
          "number" : 93, 
          "expected": "XCIII"
       },
       {
+         "description": "100 is a single C",
          "number" : 141, 
          "expected": "CXLI"
       },
       {
+         "description": "60, being 50 + 10, is LX",
          "number" : 163, 
          "expected": "CLXIII"
       },
       {
+         "description": "400, being 500 - 100, is CD",
          "number" : 402, 
          "expected": "CDII"
       },
       {
+         "description": "500 is a single D",
          "number" : 575, 
          "expected": "DLXXV"
       },
       {
+         "description": "900, being 1000 - 100, is CM",
          "number" : 911, 
          "expected": "CMXI"
       },
       {
+         "description": "1000 is a single M",
          "number" : 1024, 
          "expected": "MXXIV"
       },
       {
+         "description": "3000 is three M's",
          "number" : 3000, 
          "expected": "MMM"
       }


rbasso:
I think it is great, considering how hard it is to to come up with a reason for test cases that where previously written. 👍

exercises/roman-numerals/canonical-data.json
```diff
@@ -1,74 +1,92 @@
...
          "number" : 93, 
          "expected": "XCIII"
       },
       {
+         "description": "100 is a single C",
          "number" : 141, 
          "expected": "CXLI"
       },
       {
+         "description": "60, being 50 + 10, is LX",
```
petertseng: I decided that the only new thing this case was testing was 60, but it's a bit of an illogical place to do that. It can be done earlier (near 50 and 90), without dealing with 100s. But I tried not to change tests in this PR.

exercises/roman-numerals/canonical-data.json
```diff
@@ -1,74 +1,92 @@
...
          "number" : 2, 
          "expected": "II"
       },
       {
+         "description": "3 is three I's",
          "number" : 3, 
          "expected": "III"
       },
       {
+         "description": "4, being 5 - 1, is IV",
```
petertseng: seems slightly illogical to refer to 4 as 5 - 1 before 5 has been introduced, but maybe someone thinks it's really important to use the natural ordering of numbers.

exercises/roman-numerals/canonical-data.json
```diff
@@ -1,74 +1,92 @@
...
          "number" : 5, 
          "expected": "V"
       },
       {
+         "description": "6, being 5 + 1, is VI",
          "number" : 6, 
          "expected": "VI"
       },
       {
+         "description": "9, being 10 - 1, is IX",
```
petertseng: same: 9 as 10 - 1 before 10 introduced.

exercises/roman-numerals/canonical-data.json
```diff
@@ -1,74 +1,92 @@
...
          "number" : 48, 
          "expected": "XLVIII"
       },
       {
+         "description": "50 is a single L",
          "number" : 59, 
          "expected": "LIX"
       },
       {
+         "description": "90, being 100 - 10, is XC",
```
petertseng: same: 90 as 100 - 10 before 100 introduced.

exercises/roman-numerals/canonical-data.json
```diff
@@ -1,74 +1,92 @@
...
          "number" : 141, 
          "expected": "CXLI"
       },
       {
+         "description": "60, being 50 + 10, is LX",
          "number" : 163, 
          "expected": "CLXIII"
       },
       {
+         "description": "400, being 500 - 100, is CD",
```
petertseng: same: 400 as 500 - 100 before 500 introduced.

exercises/roman-numerals/canonical-data.json
```diff
@@ -1,74 +1,92 @@
...
          "number" : 402, 
          "expected": "CDII"
       },
       {
+         "description": "500 is a single D",
          "number" : 575, 
          "expected": "DLXXV"
       },
       {
+         "description": "900, being 1000 - 100, is CM",
```
petertseng: same: 900 as 1000 - 100 before 1000 introduced.

exercises/roman-numerals/canonical-data.json
```diff
@@ -1,74 +1,92 @@
...
          "expected": "CXLI"
       },
       {
+         "description": "60, being 50 + 10, is LX",
          "number" : 163, 
          "expected": "CLXIII"
       },
       {
+         "description": "400, being 500 - 100, is CD",
          "number" : 402,
```
petertseng: this tests a useful property noted in the README: empty places are skipped. Since the tens place is zero, we don't see an X or L in this answer. It might be useful to call this out. One might see this as distinct from 10 -> X because there are numbers on both sides of the zero.

exercises/roman-numerals/canonical-data.json
```diff
@@ -1,74 +1,92 @@
...
          "expected": "DLXXV"
       },
       {
+         "description": "900, being 1000 - 100, is CM",
          "number" : 911, 
          "expected": "CMXI"
       },
       {
+         "description": "1000 is a single M",
          "number" : 1024,
```
petertseng: same, empty 100's place can be called out interestingly.

rbasso:
> I hope that characters like , and - and + in the descriptions don't mess test generators up.

At the moment we don't restricted the contents of the descriptions. The only thing I expected, which is not enforced, is that it doesn't contain control characters, newlines or non ASCII character.

Considering that the `description` is user-displayable content, it makes sense to expect the generators to deal with arbitrary text, IMHO.

exercises/roman-numerals/canonical-data.json
ErikSchierboom: Hmmm, I a bit torn on this one. Perhaps we should first do all of the individual letter case (I, V, X, etc.) and then move on to the edge cases? I think that would make more sense for this particular problem domain, even though it can seem a bit odd at first.

exercises/roman-numerals/canonical-data.json
ErikSchierboom: +1 for calling out

PR #709, opened 2017-03-11
say: add test case for 810000

csabella:
Example code had written two 'and' statements.

### DIFF
--- a/exercises/say/canonical-data.json
+++ b/exercises/say/canonical-data.json
@@ -61,6 +61,12 @@
       "expected": "one thousand two hundred thirty-four"
     },
     {
+      "description": "eight hundred and ten thousand",
+      "property": "say",
+      "input": 810000,
+      "expected": "eight hundred and ten thousand"
+    },
+    {
       "description": "one million",
       "property": "say",
       "input": 1000000,


petertseng:
can you clarify what you mean by

> Example code had written two 'and' statements.

Does that mean that you saw some code that would incorrectly read this as "eight hundred and ten thousand and" ?

That will help us figure out why it's good to include this case. What mistakes might it help us catch?

petertseng:
I don't see any cases in this entire file where the word "and" appears, but this test case adds one. Can you explain when "and" might be used?

behrtam:
Normally `and` would be in nearly every test case, but it makes the exercise more complicated, so it's just an extension: https://github.com/exercism/x-common/blob/master/exercises/say/description.md#extensions

rbasso:
For a parsing problem, it would be interesting, but in this case it seems to be arbitrary to add a single case with **and** in the middle.

How would the user decide in which cases to use the **and**, @behrtam?

csabella:
In the original example code, the solution was written as 'eight hundred and and ten thousand'.

exercises/say/canonical-data.json
```diff
@@ -61,6 +61,12 @@
       "expected": "one thousand two hundred thirty-four"
     },
     {
+      "description": "eight hundred and ten thousand",
+      "property": "say",
+      "input": 810000,
+      "expected": "eight hundred and ten thousand"
```
behrtam: The **and** needs to be removed as this is only part of the extension. The base exercise ignores the and part.

csabella:
Sorry, all the python examples had the 'and', so that's why I had added it.  

This test was fashioned from an existing python test for 987654321123 with the solution of "nine hundred and eighty-seven billion six hundred and fifty-four million three hundred and twenty-one thousand one hundred and twenty-three".

I only added this test because the original example.py code gave the wrong answer (with two ands).  If the base cases don't have any 'and' required, then I'll delete the pull request.

Thanks.

petertseng:
It's true that the tests in this repo will not have the "and". as you requested, we will close this repo. Anyone, let us know if there is still value in including it without the "and". Thanks everyone.

petertseng:
EBWODP:

> we will close this repo. 

Er, I meant this PR. Not the entire repo. That would be ridiculous.

PR #710, opened 2017-03-11
pov 1.1.1: remove comments about `trees` section

petertseng:
We had already removed these in #704, but I forgot to remove the
comments.


petertseng:
Hmm, now doubting myself on whether a comment change will want a patch level bump

petertseng:
Changing a comment is explicitly in the list of things that want a patch-level bump. So it has been bumped.

Issue #699, opened 2017-03-11
canonical-data.json: Shared data among test cases


rbasso:
While working in #625, we discovered that some `canonical-data.json` files have data shared among test cases:

- `grep` - Has a `files` object that encodes files to be "grepped" in multiples test cases.
- `pov` - Has a `trees` object shared by multiple test cases, referenced by their names.

The current JSON Schema doesn't allow that, so I plan to sidestep this issue putting the offending structures in comments, temporarily, until we decide how deal with that.

I see a few option:

- Replicate the information where it is needed: This has the advantage of keeping all test cases independent, but it hurts to ask a programmer to do it.
- Change the schema to allow a `resources`/`shared` structure: This would avoid repetition, but also increase the complexity of test generators significantly.
- Move from JSON to YAML: Using YAML anchors would be a clean and easy solution, but YAML is not well-supported as JSON, and that may be a problem for test generators in some languages.
- Change the offending exercises, so that this feature isn't needed: The simple, trustworthy non-solution.

Any other option?

petertseng:
Doh. And I know that I was at least partially involved in both:

* https://github.com/exercism/x-common/pull/390#discussion_r81459867
* https://github.com/exercism/x-common/pull/327#r73993706

---

Let me pretend for a moment that we have decided to accept duplication in pov/canonical-data.json, and see what that looks like.

```
$ grep '"tree"' pov/canonical-data.json | sort | uniq -c                                             
      6         "tree": "cousins",
      1         "tree": "deeply_nested",
      1         "tree": "kids",
      2         "tree": "large_flat",
      2         "tree": "simple",
      2         "tree": "singleton",
```

So we won't care if deeply_nested and kids go away, and we probably can survive repeating `simple` and `large_flat`. `cousins` would hurt. What are the cases that use it?

* Can reroot a tree with cousins: All right, I guess we have to keep it here.
* Errors if target does not exist in a large tree: we probably do not have to use cousins for that. We can use a smaller tree that is less painful to write out in full.
* Can find path to cousin: OK, I guess we have to keep it here as well.
* Can find path from nodes other than x: we most definitely do not have to use cousins for that. We can use a smaller tree that is less painful to write out in full.
* Errors if destination does not exist: same; doesn't require cousins.
* Errors if source does not exit: same; doesn't require cousins.

[...]

PR #658, opened 2017-03-08
Hook schema validation in CI

tejasbubane:
Using [ajv-cli](https://github.com/jessedc/ajv-cli). Check canonical-data files for all exercises are compatible with the canonical-schema.

Remove `jsonlint` from CI since the schema validator will take care of checking malformed json.

A quick attempt to get schema validation running in CI. This does not include excluding incompatible files. Hence the build will fail until we convert all files for compatibility. 

Ref: #625

PR #712, opened 2017-03-12
grep: Make exercise schema-compliant

rbasso:
This exercise has 3 text files encoded in the `canonical-data.json`, which is incompatible with the new JSON schema.

We are discussing in #699 what is the best way to deal with sharing test data among test cases, and it seems there isn't much support for adding this feature to the schema.

Anyway, that discussion doesn't need to suspend the migration of the current test data.

I proposed here to simply move the encoded files to the comments and proceed with the migration.

Related to #625.


rbasso:
This exercise has 3 text files encoded in the `canonical-data.json`, which is incompatible with the new JSON schema.

We are discussing in #699 what is the best way to deal with sharing test data among test cases, and it seems there isn't much support for adding this feature to the schema.

Anyway, that discussion doesn't need to suspend the migration of the current test data.

I proposed here simply moving the encoded files to the comments and proceed with the migration.

Related to #625.

rbasso:
Updated to fix some minor formatting issues.

rbasso:
> I wonder if any language track will complain that it's too hard to extract the files from the comments (find lines surrounded by |, strip trailing newlines), but someone comes and argues for it I can't really argue on their behalf.

I was not even considering the possibility of algorithmically extracting the lyrics from the comments, @petertseng. 😄 

I think this PR is just an immediate solution for preserving the lyrics and allow compliance. I would really like to see a rewriting of this exercise allowing more automatic test generation, but waiting for it would postpone #625, and I think that it would be good no have #658 merged soon to enforce compliance in Travis-CI.

petertseng:
Found out I missed a word:

> but someone comes and argues

but **until** someone comes and argues

PR #713, opened 2017-03-12
two-bucket: Make exercise schema-compliant

rbasso:
Related to #625.

### DIFF
--- a/exercises/two-bucket/canonical-data.json
+++ b/exercises/two-bucket/canonical-data.json
@@ -2,6 +2,7 @@
     "two_bucket": {
         "cases": [
             {
+                "description": "(3, 5) -> 1, starting with the first bucket",
                 "bucket_one": 3,
                 "bucket_two": 5,
                 "goal": 1,
@@ -13,6 +14,7 @@
                 }
             },
             {
+                "description": "(3, 5) -> 1, starting with the second bucket",
                 "bucket_one": 3,
                 "bucket_two": 5,
                 "goal": 1,
@@ -24,6 +26,7 @@
                 }
             },
             {
+                "description": "(7, 11) -> 2, starting with the first bucket",
                 "bucket_one": 7,
                 "bucket_two": 11,
                 "goal": 2,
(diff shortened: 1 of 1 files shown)


petertseng:
Remarks:

3/3, 0/5, goal 1: Only requires the "fill first" and "first -> second" actions.
5/5, 0/3, goal 1: Requires all three actions.
7/7, 0/11, goal 2: I can't tell you what this tests that the above two don't (because the second case already covers all three actions), other than "different bucket sizes"
11/11, 0/7, goal 2: Same.

petertseng:
And this just made me realize: Unless the goal amount is exactly one bucket's capacity, by necessity the other bucket will be full when the goal is reached.

rbasso:
Sorry, @petertseng . I saw the approved and almost immediately merged. Shall I revert?

petertseng:
> Shall I revert?

No real reason to. The only thing that will come out of my comments is maybe it will help you find a better description (but it didn't help me). If it helps you, just update the descriptions.

exercises/two-bucket/canonical-data.json
```diff
@@ -1,50 +1,58 @@
...
-                "goal": 2,
-                "start_bucket": "two",
-                "expected": {
-                    "moves": 18,
-                    "goal_bucket": "two",
-                    "other_bucket": 7
-                }
-            }
-        ]
+  "exercise": "two-buckets",
```
petertseng: the singular two-bucket perhaps? rather than the plural two-buckets

unless it doesn't need to match the exercise slug

maybe that should be checked, I certainly don't have anything that checks it

rbasso: I screwed it! It should be `two-bucket`.

Thanks for catching that (my third big mistake in 24 hour). 😁

rbasso: Should I also bump the version number to `1.0.1`?

rbasso: #715

PR #718, opened 2017-03-12
New exercise: complex numbers

SaschaMann:
This exercise is about writing your own complex numbers implementation.

What does this teach?

- Operator overloading in those languages that support it.
- How testing for float operations works because you cannot assume an exact result (for some of the tests).
- Complex numbers themselves, which may be used in other exercises. (I have some ideas but nothing concrete yet.)

---
The tests are fairly basic for now and there are probably many cases that aren't covered so this is just a starting point.

### DIFF
--- a/exercises/complex-numbers/canonical-data.json
+++ b/exercises/complex-numbers/canonical-data.json
@@ -44,8 +44,8 @@
     {
       "description": "Absolute value",
       "property": "abs",
-      "input": "(1+i)*(1+i)",
-      "expected": "2"
+      "input": "3+4i",
+      "expected": "5"
     },
     {
       "description": "Complex conjugate",
@@ -68,7 +68,8 @@
     {
       "description": "Imaginary unit",
       "property": "mul",
-      "input": "i*i",
+      "z1": "i",
+      "z2": "i",
       "expected": "-1"
     },
     {
--- a/exercises/complex-numbers/canonical-data.json
+++ b/exercises/complex-numbers/canonical-data.json
@@ -1,0 +1,88 @@
+{
+  "exercise": "complex-numbers",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "comments": [
+        " The canonical data assumes mathematically correct real   ",
+        " numbers. The testsuites should consider rounding errors  ",
+        " instead of testing for exact values for any non-integer  ",
+        " tests.                                                   "
+      ],
+      "description": "Arithmetic",
+      "cases": [
+        {
+          "description": "Addition",
+          "property": "add",
+          "z1": "1+2i",
+          "z2": "3+4i",
+          "expected": "4+6i"
+        },
+        {
+          "description": "Subtraction",
+          "property": "sub",
+          "z1": "1+2i",
+          "z2": "3+4i",
+          "expected": "-2-2i"
+        },
+        {
+          "description": "Multiplication",
+          "property": "mul",
+          "z1": "1+2i",
+          "z2": "3+4i",
+          "expected": "-5+10i"
+        },
+        {
+          "description": "Division",
+          "property": "div",
+          "z1": "1+2i",
+          "z2": "3+4i",
+          "expected": "11/25+2i/25"
+        }
+      ]
+    },
+    {
+      "description": "Absolute value",
+      "property": "abs",
+      "input": "(1+i)*(1+i)",
+      "expected": "2"
+    },
+    {
+      "description": "Complex conjugate",
+      "property": "conjugate",
+      "input": "1+i",
+      "expected": "1-i"
+    },
+    {
+      "description": "Real part",
+      "property": "real",
+      "input": "1+2i",
+      "expected": "1"
+    },
+    {
+      "description": "Imaginary part",
+      "property": "imaginary",
+      "input": "1+2i",
+      "expected": "2"
+    },
+    {
+      "description": "Imaginary unit",
+      "property": "mul",
+      "input": "i*i",
+      "expected": "-1"
+    },
+    {
+      "comments": [
+        " Defining the exponential function can be optional.       ",
+        " If the language used does not have sine and cosine       ",
+        " functions in the standard library, this will be          ",
+        " significantly more difficult than the rest of the exer-  ",
+        " cise and should probably not be part of the problem.     "
+      ],
+      "description": "Euler identity",
+      "property": "exp",
+      "input": "pi*i",
+      "expected": "-1"
+    }
+  ]
+}
--- a/exercises/complex-numbers/canonical-data.json
+++ b/exercises/complex-numbers/canonical-data.json
@@ -1,21 +1,23 @@
 {
   "exercise": "complex-numbers",
   "version": "1.0.0",
+  "comments": [
+    " The canonical data assumes mathematically correct real   ",
+    " numbers. The testsuites should consider rounding errors  ",
+    " instead of testing for exact values for any non-integer  ",
+    " tests.                                                   ",
+    " Complex numbers z are represented as arrays [x, y] so    ",
+    " that z = x + i * y.                                      "
+  ],
   "cases": [
     {
       "description": "Imaginary unit",
       "property": "mul",
-      "z1": "i",
-      "z2": "i",
-      "expected": "-1+0i"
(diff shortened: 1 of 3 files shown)


exercises/complex-numbers/canonical-data.json
```diff
@@ -0,0 +1,88 @@
...
+    {
+      "description": "Imaginary part",
+      "property": "imaginary",
+      "input": "1+2i",
+      "expected": "2"
+    },
+    {
+      "description": "Imaginary unit",
+      "property": "mul",
+      "input": "i*i",
```
petertseng: seems a bit confusing - this `mul` has an `input`, but the `mul` above has a `z1` and a `z2`. This means that an automatic test generator can't simply rely on `property` to decide what to generate; it must also check whether `input` is present.

exercises/complex-numbers/canonical-data.json
```diff
@@ -0,0 +1,88 @@
...
+          "z1": "1+2i",
+          "z2": "3+4i",
+          "expected": "11/25+2i/25"
+        }
+      ]
+    },
+    {
+      "description": "Absolute value",
+      "property": "abs",
+      "input": "(1+i)*(1+i)",
```
petertseng: seems to obscure the purpose of the test if the `abs` test also has multiplication in it. at the end of the day, this is just testing that `abs(2i)` is 2, right? if wanting to test that `(1 + i) * (1 + 1)` is `2i`, that seems to be a `mul` test not an `abs` test, right?

petertseng: EBWODP: 

> `(1 + i) * (1 + 1)` is `2i`

I of course meant `(1 + i) * (1 + i)` is `2i`

SaschaMann: Good point, I chose that because it gave a round result that would be easy to check by hand. Changing it to `abs(3+4i)=5`.

exercises/complex-numbers/canonical-data.json
SaschaMann: I originally had all cases with just input and mathematical notation but changed them to make them easier to parse. Forgot to update this one to also have `z1`, `z2`, will do so.

exercises/complex-numbers/canonical-data.json
```diff
@@ -68,7 +68,8 @@
     {
       "description": "Imaginary unit",
       "property": "mul",
```
petertseng: now I'm thinking... surely `i * i = -1` would be an easier test than `(1+2i) * (3+4i) = -5 + 10i` right? any chance it should get moved before?

SaschaMann: I put more general properties first and the special cases afterwards. But since i^2=-1 is so fundamental, it would make sense to have it as the very first test, indeed.

SaschaMann:
I chose the string representation because it's very readable for humans. Parsing that string to a complex number could be quite annoying because the parser would have to infer the types of `a` and `b` or just convert everything to a default type which is not necessarily the best solution (depends on the language used, though). I don't think that would still fit this exercise.

The human-readable representation could be moved to the description. I'm not sure how the testdata would have to be written for the alternative you outlined. Could you give me an example, please?

petertseng:
> I'm not sure how the testdata would have to be written for the alternative

It would probably have to be that instead of `"z1": "1+2i",` we have `"z1": [1, 2]` or `"z1": {"real": 1, "imaginary", 2}`. TBD how to do the 11/25 and 2/25

petertseng:
> `"z1": {"real": 1, "imaginary", 2}`

Wrong. that is not a JSON object. it should be `"z1": {"real": 1, "imaginary": 2}`

SaschaMann:
11/25=0.44, 2/25=0.08; json supports Floats so that would work. Writing it as an array seems reasonable to me. I'll change it later.

SaschaMann:
Test generators may still have issues with the `exp` test since that will likely not have an exact result. I can't think of a way to fix that, however, unless the generators have some sort of `test_approximate` flag that one can set for a case?

petertseng:
I don't think we have a good solution for that, given that we also do not have such a solution in https://github.com/exercism/x-common/blob/master/exercises/space-age/canonical-data.json

exercises/complex-numbers/canonical-data.json
```diff
@@ -0,0 +1,89 @@
...
+    },
+    {
+      "description": "Absolute value",
+      "property": "abs",
+      "input": "3+4i",
+      "expected": "5"
+    },
+    {
+      "description": "Complex conjugate",
+      "property": "conjugate",
```
stevejb71: What about conjugate on a real number. And a purely imaginary one.

exercises/complex-numbers/canonical-data.json
```diff
@@ -0,0 +1,89 @@
...
+    {
+      "comments": [
+        " Defining the exponential function can be optional.       ",
+        " If the language used does not have sine and cosine       ",
+        " functions in the standard library, this will be          ",
+        " significantly more difficult than the rest of the exer-  ",
+        " cise and should probably not be part of the problem.     "
+      ],
+      "description": "Euler identity",
+      "property": "exp",
```
SaschaMann: I was thinking of an implementation based on Euler's formula. I can add an explanation in the test-description.

exercises/complex-numbers/canonical-data.json
```diff
@@ -242,8 +244,8 @@
       ],
       "description": "Euler's identity/formula",
       "property": "exp",
-      "input": "pi*i",
-      "expected": "-1"
+      "input": [0, "pi"],
```
SaschaMann: I'm not happy with this but I can't think of a better way without using an approximation of π.

exercises/complex-numbers/canonical-data.json
```diff
@@ -0,0 +1,251 @@
...
+    " The canonical data assumes mathematically correct real   ",
+    " numbers. The testsuites should consider rounding errors  ",
+    " instead of testing for exact values for any non-integer  ",
+    " tests.                                                   ",
+    " Complex numbers z are represented as arrays [x, y] so    ",
+    " that z = x + i * y.                                      "
+  ],
+  "cases": [
+    {
+      "description": "Imaginary unit",
```
stevejb71: Shouldn't this be under the ```Multiplication``` group?

exercises/complex-numbers/canonical-data.json
```diff
@@ -0,0 +1,251 @@
...
+        " Defining the exponential function can be optional.       ",
+        " If the language used does not have sine and cosine       ",
+        " functions in the standard library, this will be          ",
+        " significantly more difficult than the rest of the exer-  ",
+        " cise and should probably not be part of the problem.     ",
+        " The recommended implementation uses Euler's formula      ",
+        " exp(ix) = cos(x) + i * sin(x). This is not an ideal sol- ",
+        " ution but works for the purpose of teaching complex      ",
+        " numbers.                                                 "
+      ],
```
stevejb71: This is quite a limited test - perhaps add tests for exponentiation on real numbers, i^2, etc.

petertseng:
ignore notifications just received from me in the past 3 minutes, just testing.

Insti:
@SaschaMann are you still working on this?

SaschaMann:
@Insti Sorry, I forgot about this PR after going on vacation. I've added some tests for the complex exponential as @stevejb71 requested.

exercises/complex-numbers/canonical-data.json
SaschaMann: I can move it under Multiplication but I think it should be seperate and on top because it's the most important property of complex numbers.

ErikSchierboom:
@petertseng You said "Probably would be good to squash.", does that mean it is good enough to be merged? If so, the squashing can also be done directly from the GitHub interface.

petertseng:
> does that mean it is good enough to be merged

yes. I promise that I will never Approve (in the github sense) a pull request unless that is the case.

petertseng:
I'm going to reword that into a logically equivalent statement but one that is less likely to appear to mean the wrong thing at first glance.

I promise that if I Approve (in the GitHub sense) a pull request, I do think it is good enough to be merged.

PR #719, opened 2017-03-12
phone-number: Correct test to more-than check vs equal-to check

### LINKED ISSUE #608: phone-number: Does not properly specifiy test for numbers with length greater than 11
The specification for the phone-number exercise says

> - If the phone number is more than 11 digits assume that it is a bad number

But the phone-number/canonical-data.json does not specify checking this properly. It checks for length of 11 without leading double 1 to be invalid and for length of 12 to be invalid. It should instead (according to the brief in the readme)check for any number greater than 11 to be invalid.

wolf99:
Fix #608

### DIFF
--- a/exercises/phone-number/canonical-data.json
+++ b/exercises/phone-number/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "phone-number",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "cases": [
     {
       "description": "Cleanup user-entered phone numbers",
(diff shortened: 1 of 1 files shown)


wolf99:
Thanks for the heads-up @petertseng (and the title help @kotp). 

Just to note that I did take a quick glance at the CONTRBUTING.md but found it very long to try and find the relevant information in (i.e. the bit Peter linked)... Maybe that, or something like it, could be split out to the same directory and linked to to make it easier to find? Just a suggestion.

PR #724, opened 2017-03-14
phone-number: Clears up contradiction re: why 11 digits is invalid

rpottsoh:
There are two tests concerning 11 digit numbers.  One test says they are valid when the digits start with one.  The other test states that 11 digits is invalid, but doesn't state why.  The two tests seem to be contradicting each other.


rpottsoh:
I have to add, @kotp pointed this out to me as I was updating this exercise in the Delphi track.

PR #726, opened 2017-03-16
word-count: Simplify by removing quotation aspect

behrtam:
In #403 the exercise was extended with a completely new aspect (*the removal of single quotations but not apostrophes*) which is not mentioned in it's minimalistic description.

Especially in the last months we tried to simplify exercises in a way that they try to tackle at best one problem (or at least not too many different ones). Right now we test `splitting on different characters`, `counting`, `case normalization` and `removing some but not all punctuation`.

So, I would like to see this exercise simplified just a bit by removing those two quotation test cases.

### DIFF
--- a/exercises/word-count/canonical-data.json
+++ b/exercises/word-count/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "word-count",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "comments": [
     "For each word in the input, count the number of times it appears in the",
     "entire sentence."
@@ -85,31 +85,6 @@
       "expected": {
         "go": 3,
         "stop": 2
-      }
-    },
-    {
-      "description": "with apostrophes",
-      "property": "countwords",
-      "input": "First: don't laugh. Then: don't cry.",
-      "expected": {
-        "first": 1,
-        "don't": 2,
-        "laugh": 1,
-        "then": 1,
-        "cry": 1
-      }
-    },
-    {
-      "description": "with quotations",
-      "property": "countwords",
-      "input": "Joe can't tell between 'large' and large.",
-      "expected": {
-        "joe": 1,
-        "can't": 1,
-        "tell": 1,
-        "between": 1,
-        "large": 2,
-        "and": 1
       }
     }
   ]


ErikSchierboom:
So basically the options are:

1. Remove the quotations test cases.
1. Update the documentation.

Personally, I don't really have an opinion either way, but perhaps other people do?

kotp:
This extension happened within the last 6 months, are we finding that there is a decrease in completion, a negative change in the discussions about the exercise, a general failure to solve these tests in solutions submitted?

1) No Change Necessary
2) Document the requirement (arguably this is already done via the tests as a "Did you also consider that words with apostrophes are single words?")
3) Remove the extension

behrtam:
I saw very complex implementations because they didn't know regular expressions. So they wrote a lot of if statements and accessed the the single chars of the string after they already did all the splitting and counting.

mienaikage:
How about keeping quotations but removing apostrophes? i.e. Change `can't` to `cannot`.

kotp:
Contractions are common enough in documents to warrant having this included as part of the problem.  It can be solved without regular expressions.  It happens that regular expressions happen to be a good tool to extract that apostrophe pattern.  But it isn't the only thing.  Many languages support regular expressions either directly or provided through libraries, but many languages do not.

Indeed, even with those using regular expressions, it seems that there are solutions with pretty complex logic.  I am not sure that is a bad thing.  It lets us talk about the choices for the solutions.

What I would be more concerned about is a decrease in the attempt rate on this exercise.  And then finding out why.

ErikSchierboom:
Actually, I quite like the added apostrophe precisely because it makes the exercise just a bit more difficulty. Without this, the exercise would almost be too trivial I feel.

behrtam:
I agree that this change maybe would make it too easy.

So, we have the problems of `splitting` and `counting`. Now one way to make this exercise more challenging is to add (or not remove) a totally new problem domain with normalization and complex punctuation removal.

Or, as most languages have some built-in functions to split on whitespace, we could just make the `splitting` more interesting and also split on other chars like underscore.
```
word_count('hey,my_spacebar_is_broken.') -> {'hey': 1, 'my': 1, 'spacebar': 1, 'is': 1, 'broken': 1}
```

rpottsoh:
I am for not removing....

PR #728, opened 2017-03-17
robot name update

masters3d:
https://github.com/exercism/xecmascript/issues/263

### DIFF
--- a/exercises/robot-name/description.md
+++ b/exercises/robot-name/description.md
@@ -10,4 +10,5 @@ respond with a new random name.
 The names must be random: they should not follow a predictable sequence.
 Random names means a risk of collisions. Your solution should not allow
 the use of the same name twice when avoidable. In some exercism language
-tracks there are tests to ensure that the same name is never used twice.
+tracks there are tests to ensure that the same name is never used twice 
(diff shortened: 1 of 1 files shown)


masters3d:
As a hint on how that can be accomplished? 
I'd rather remove even mentioning the last paragraph but since it is included some tracks have implemented but give no indication on how they are accomplishing it.

ErikSchierboom:
@masters3d Ah, now I see what you are getting at. Although I wouldn't want to add this to the general README, you can have the CLI append text to the README for a specific exercise on a specific language track. This is precisely what you want, as you want to give your users a hint on how to solve this exercise.

To have the CLI append this extra information, just put a `HINTS.md` file in the exercise's directory. As an example, the C# track's `accumulate` exercise has a [HINTS.md](https://github.com/exercism/xcsharp/blob/master/exercises/accumulate/HINTS.md) file. The contents of this file will be append to the README when a user fetches the accumulate exercise in the C# track. 

I think adding a HINTS.md file to your track's version of the `robot-name` exercise is the better option than adding it to the shared x-common version.

masters3d:
Thank you!

ErikSchierboom:
You're welcome!

PR #732, opened 2017-03-21
Simplify triangle description.

stevejb71:
I thought the description for triangle was a little wordy, so I shortened it.
Happy to revert on this if anyone thinks the text is not an improvement, but the line at the top:
> The program should raise an error if the triangle cannot exist.

should be changed, as not all languages would idiomatically raise errors in this case.

### DIFF
--- a/exercises/triangle/description.md
+++ b/exercises/triangle/description.md
@@ -1,28 +1,17 @@
-The program should raise an error if the triangle cannot exist.
+An _equilateral_ triangle has all three sides the same length.<br/>
+An _isoceles_ triangle has at least two sides the same length. (It is sometimes
+specified as having exactly two sides the same length, but for the purposes of
+this exercise we'll say at least two.)<br/>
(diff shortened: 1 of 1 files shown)


kotp:
I do like the changes.

exercises/triangle/description.md
```diff
@@ -1,28 +1,20 @@
-The program should raise an error if the triangle cannot exist.
+Determine whether a triangle is equilateral, isoceles, or scalene.
```
petertseng: because this is the same line as https://github.com/exercism/x-common/blob/master/exercises/triangle/metadata.yml#L2 plus the fact that we prepend the blurb line to the README, this will cause this line to appear twice.

Thus, I would suggest it be removed from README.

stevejb71: I removed that first line.

stevejb71:
I made changes as suggested by @petertseng, is there anything else?

PR #739, opened 2017-03-26
isogram: Add test for input not starting with duplicate letter

hop:
This catches implementations that only check whether the very first
letter of the input is found more than once. Actual example:

    for char in input:
        return (False if input.count(char) > 1 else True)

### DIFF
--- a/exercises/isogram/canonical-data.json
+++ b/exercises/isogram/canonical-data.json
@@ -58,6 +58,12 @@
           "property": "isIsogram",
           "input": "Emily Jung Schwartzkopf",
           "expected": true
+        },
+        {
+          "description": "non isogram with duplicated letter not at the start",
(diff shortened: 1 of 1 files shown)


behrtam:
As this adds a new test case the version needs to be bumped: 1.1.0

petertseng:
Note for reviewers: I think we only need to merge one of #739 and #743, not both

behrtam:
I would argue we don't need this test case. We have two existing test cases that deal with it.

**empty string** `input = ""`

This might be language specific, but if input is empty the for loop is not executed. So, instead of True/False it will return something like None or void. So the example will fail.

**word with duplicated character in mixed case** `input = "Alphabet"`
In this test case, it will count the upper case A and incorrectly assume that the input is an isogram.

PR #740, opened 2017-03-27
Cleanup canonical-data for clock

ErikSchierboom:
This PR does a cleanup of the canonical data of the `clock` exercise. It contains the following changes:

1. The word "Test" is removed from the test group descriptions, as that is superfluous (it is a test group description after all)
1. I've split the add and subtract test into separate groups. This makes things more sensible for test data generators.

I've bumped the PATCH version number, as I believe that according to the README, this qualifies as a patch version bump.

### DIFF
--- a/exercises/clock/canonical-data.json
+++ b/exercises/clock/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "clock",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "comments": [
     "Most languages require constructing a clock with initial values,",
     "adding a positive or negative number of minutes, and testing equality",
@@ -10,7 +10,7 @@
   ],
   "cases": [
     {
-      "description": "Test creating a new clock with an initial time.",
+      "description": "Create a new clock with an initial time",
       "cases": [
         {
           "description": "on the hour",
@@ -148,7 +148,7 @@
       ]
     },
     {
-      "description": "Test adding and subtracting minutes.",
+      "description": "Add minutes",
       "cases": [
         {
           "description": "add minutes",
@@ -213,7 +213,12 @@
           "minute": 1,
           "add": 3500,
           "expected": "11:21"
-        },
+        }
+      ]
+    },
+    {
+      "description": "Subtract minutes",
+      "cases": [
         {
           "description": "subtract minutes",
           "property": "add",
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
Could someone take a look at this PR?

exercises/clock/canonical-data.json
```diff
@@ -213,7 +213,12 @@
...
-        },
+        }
+      ]
+    },
+    {
+      "description": "Subtract minutes",
+      "cases": [
         {
           "description": "subtract minutes",
           "property": "add",
```
NobbZ: This feels a bit funny. We tell we want to subtract, but still test the property add.

Shouldn't we introduce a property `sub` then, or omit subtracting at all?

ErikSchierboom: @NobbZ Hmmm, great point! I don't really have a preference either way.

rpottsoh: Remove minutes instead?

NobbZ: We still had subtract vs add then. Personally I do not have a strong opinion on this either, since subtraction is just adding a negativ number.

But I try to look from a beginners view here, who probably will feel dizzy at first, when we ask to subtract amount x by telling him to add amount -x.

mienaikage: Do `add` and `subtract` need to be separate? How about a single `adjust` property for both?

ErikSchierboom: @rbasso Do you have any suggestion how to handle this?

rbasso: I agree that `add` isn't the best property name, and maybe `adjust` would be a better choice. On the other hand, this is only necessarily visible to track maintainers, as they can choose more descriptive names when implementing the exercise.

If I where to choose, I would merge this the way it is and leave other MAJOR changes for another PR, because maybe it deserves a rewriting. I don't understand what the `create` property is at all...

rbasso: I agree that `add` isn't the best property name, and maybe `adjust` would be a better choice. On the other hand, this is only necessarily visible to track maintainers, as they can choose more descriptive names when implementing the exercise.

If I where to choose, I would merge this the way it is and leave other MAJOR changes for another PR, because maybe it deserves a rewriting. I don't understand what the `create` property is testing at all...

ErikSchierboom:
Okay. Shall we merge this and work on the the add/subtract discussion another time?

PR #741, opened 2017-03-27
Pangram: Add the non-ascii test cases

cogitor:
These are the canonical data changes for additional pangram tests I originally wanted to add to the xscala repo (https://github.com/exercism/xscala/pull/348)

The first non-ascii test already exists in the xscala repo (so this would only sync it up to common), while I wanted to add the second one as well:
> I feel that another useful test would be a string with non-ascii characters, but with a missing character 'x'. This would invalidate solutions that are written to simply pass the previous test (e.g. by making the test for distinct characters be >= 26)

I am not sure if this would make sense for all languages, and what steps would be required to generate the per-language test cases.

Cheers!

### DIFF
--- a/exercises/pangram/canonical-data.json
+++ b/exercises/pangram/canonical-data.json
@@ -64,6 +64,18 @@
           "property": "isPangram",
           "input": "the quick brown fox jumped over the lazy FOX",
           "expected": false
+        },
+        {
+          "description":"pangram with non ascii characters",
+          "property": "isPangram",
+          "input": "Victor jagt zwölf Boxkämpfer quer über den großen Sylter Deich.",
+          "expected": true
+        },
+        {
+          "description":"pangram with non ascii characters missing character 'x'",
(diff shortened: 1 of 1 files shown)


stevejb71:
Non ascii test cases were removed earlier. See

https://github.com/exercism/x-common/issues/441

cogitor:
Figured that could've been the case and the scala tests weren't updated. That's fine, I can close this PR - how do you re-generate tests from the canonical file?

stevejb71:
Looking at the xscala repo, @ricemery has a test generator. If there is an automatic way of generating the tests he might well know, otherwise it would be done by hand.

ricemery:
The tests have to be updated manually. I have submitted a pull request - https://github.com/exercism/xscala/pull/350 . Hopefully is compiles, runs and is merged :)

Thanks

cogitor:
@ricemery awesome! I'm closing this one then, cheers

PR #745, opened 2017-04-04
phone-number: Rewrite description and add tests

behrtam:
Removes the part about how to test, as we have other docs for that. Now it also describes the telephone number system in a bit more detail.

Changes some test that used an invalid number but assumed it was valid.

Adds two tests that check the range of 2-9 for the first digit of the area and the exchange code.

Adds a test with a valid telephone number with country code and punctuation.

### DIFF
--- a/exercises/phone-number/description.md
+++ b/exercises/phone-number/description.md
@@ -1,16 +1,22 @@
-The rules are as follows:
+The **North American Numbering Plan (NANP)** is a telephone numbering system used by many countries in North America like the United States, Canada or Bermuda. All NANP-countries share the same international country code `1`.
 
-- If the phone number is less than 10 digits assume that it is bad
-  number
-- If the phone number is 10 digits assume that it is good
-- If the phone number is 11 digits and the first number is 1, trim the 1
-  and use the last 10 digits
-- If the phone number is 11 digits and the first number is not 1, then
-  it is a bad number
-- If the phone number is more than 11 digits assume that it is a bad
-  number
+NANP numbers are ten-digit numbers consisting of a three-digit Numbering Plan Area code, commonly known as *area code*, followed by a seven-digit local number. The first three digigts of the local number represent the *exchange code*, followed by the unique four-digit number which is the *subscriber number*.
 
-We've provided tests, now make them pass.
 
-Hint: Only make one test pass at a time. Disable the others, then flip
-each on in turn after you get the current failing one to pass.
+The format is usually represented as
+```
+(NXX)-NXX-XXXX
+```
+where N is any digit from 2 through 9 and X is any digit from 0 through 9.
+
+You task is to clean up differently formated telephone numbers by removeing punctuatin and the country code (1) if present.
+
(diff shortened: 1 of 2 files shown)


behrtam:
We could also add new properties, but I think that's maybe to much.

```json
{
  "description": "extract area code",
  "property": "area",
  "phrase": "(223) 456-7890",
  "expected": "223"
},
{
  "description": "extract exchange code",
  "property": "exchange",
  "phrase": "223.456.7890",
  "expected": "456"
},
```

kotp:
I like the tests with the leading number avoiding the 1.  As far as the parts of the number, the names of those parts sometimes comes up in reviews (at least on the Ruby track).

exercises/phone-number/description.md
```diff
@@ -1,16 +1,22 @@
...
 
-Hint: Only make one test pass at a time. Disable the others, then flip
-each on in turn after you get the current failing one to pass.
+The format is usually represented as
+```
+(NXX)-NXX-XXXX
+```
+where N is any digit from 2 through 9 and X is any digit from 0 through 9.
+
+You task is to clean up differently formated telephone numbers by removeing punctuatin and the country code (1) if present.
```
rpottsoh: `You` should be `Your` on line 12.

rpottsoh:
@behrtam LGTM!

behrtam:
So, should I add the additional properties like `area code` or `exchange code` as proposed in the second comment or do we stick with the one existing property?

rpottsoh:
@behrtam I am fine with adding those tests.  However, this [PR](https://github.com/exercism/x-common/pull/477) back in December, touched on the issue of extraction and pretty printing and the decision was made to leave those tests out of the canonical-data.json.

behrtam:
@petertseng wrote:
> Pretty printing and extracting subparts of numbers: Yeah, doesn't seem that important to me, and the README doesn't mention it, so I would not add them to the JSON file. It would make sense if people agreed to add it and it got added to the README and JSON file together. But that can come later.

Well those two additional properties are mentioned in the new description.

rpottsoh:
Good point.  It is later now.  :)


Sent from my android device.

exercises/phone-number/description.md
Insti: and "Punctuation" is misspelled.

Insti: and "Removing"

Insti:
>  those two additional properties are mentioned in the new description.

The conclusion in the previous PR was to not add them anyway.

I vote NO, to adding the new properties.

behrtam:
I wouldn't really call a single post without interaction "conclusion" ...  as I don't have any opinion about the additionl properties I'm fine with not adding them.

I will fix the typos later when I'm back at a real device.

exercises/phone-number/description.md
behrtam: Thx!
Sorry, for the sloppy work. I'll try to remember to use a spell checker next time.

PR #748, opened 2017-04-06
Add Extensions to RobotName

masters3d:
https://github.com/exercism/x-common/issues/731

### DIFF
--- a/exercises/robot-name/description.md
+++ b/exercises/robot-name/description.md
@@ -9,5 +9,8 @@ respond with a new random name.
 
 The names must be random: they should not follow a predictable sequence.
 Random names means a risk of collisions. Your solution should not allow
-the use of the same name twice when avoidable. In some exercism language
-tracks there are tests to ensure that the same name is never used twice.
+the use of the same name twice when avoidable. 
+
+## Extensions
+- Ensure that the same name is never used twice.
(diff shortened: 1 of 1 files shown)


masters3d:
What do u think?

Insti:
The unique robot name part was what made this problem interesting.
I would prefer it remained as a main part of the exercise.

masters3d:
Define unique? Have you gotten a chance to read the linked issue?

Insti:
> Define unique?

From the description:

> Your solution should not allow the use of the same name twice 

I have read the linked issue, do you need me to post my objection there as well?

masters3d:
That part it still part of the description.  I am afraid that most of the tracks have chosen only to test about 10k permutations which is different that's the 600k permutations needed for a truly unique test.

Insti:
> most of the tracks have chosen only to test about 10k permutations which is different that's the 600k permutations needed for a truly unique test.

That's a deficiency of the tracks rather than an issue with the problem.
I believe the problem is good and it is the tests that need updating.

masters3d:
Not when that kind of test was specified like this 

"In some exercism language tracks there are tests to ensure that the same name is never used twice."

That makes it sound optional. That is not the core concept of the exercise from the description.

PR #749, opened 2017-04-06
triangle: Fix spelling of "isosceles" in README

mkantor:


stkent:
Good eye, and thanks for the fix! LGTM :)

PR #751, opened 2017-04-09
book-store: Add missing trailing zero

behrtam:
Sorry, for this micro change. I just couldn't resist. If all the other numbers have up to 2 trailing zeros ...


rpottsoh:
Thanks @behrtam.

PR #752, opened 2017-04-10
Create solution.cpp

Aashu24:


rpottsoh:
@Aashu24, thanks for attempting an exercise from the C++ language track!  It appears to me that you are attempting to submit a solution to an exercise via PR.  The correct way to submit a solution is to utilize the Exercism client application.  You will find more information regarding submitting solutions to exercises [here](http://exercism.io/how-it-works).  Also, you can receive help directly from the client if you execute `exercism help` in a console window.  The `submit` or `s` command is what you should be using to submit your solution for any given exercise.

If you need further assistance please visit gitter.im [here](https://gitter.im/exercism/support).

Issue #746, opened 2017-04-04
Override probot/stale defaults, if necessary


kytrinyx:
Per the discussion in https://github.com/exercism/discussions/issues/128 we
will be installing the probot/stale integration on the Exercism organization on
April 10th, 2017.

By default, probot will comment on issues that are older than 60 days, warning
that they are stale. If there is no movement in 7 days, the bot will close the issue.
By default, anything with the labels `security` or `pinned` will not be closed by
probot.

If you wish to override these settings, create a .github/stale.yml file as described
in https://github.com/probot/stale#usage, and make sure that it is merged
before April 10th.

If the defaults are fine for this repository, then there is nothing further to do.
You may close this issue.

kytrinyx:
I made a mistake.

This was not explicitly documented in the README, but probot/stale **requires the presence of a .github/stale.yml** file.

If you want the default behavior, you need to add an empty .github/stale.yml file to this repo, otherwise the bot will not be active.

petertseng:
Hi! I used https://gist.github.com/petertseng/04edd07d4b35cda2cbc1d8142b6fcc13 for this repo as well. 

My goal was to find the issues in the history of this repo that **would have** been closed by the 60-day inactivity period. Then I wished to answer the question: How many of those issues **should** have been closed?

What does "should" mean, here? It seems to me that there are various kinds of issues:

* Issues where we've decided on a clear plan of action
  * Good examples: the [good first patch issues](https://github.com/exercism/x-common/issues?q=is%3Aissue%20is%3Aclosed%20label%3A%22good%20first%20patch%22%20).
  * These issues should probably stay open until the action is taken, no matter how stale they get.
  * Only exception probably is if they get so stale that the action they recommend is no longer relevant to the repo, but the bot does not have the ability to decide that, so I say the bot should **never** close these issues.
  * For these issues, the bot forces extra work on us, by forcing us to apply a label to exempt it.
* Issues where there isn't a clear plan of action, perhaps discussions or questions.
  * For example, #80 went a long time without activity.
  * For these it seems being stale implies not enough consensus to change the status quo and/or the querent is no longer interested in the question.
  * For these issues, the bot saves us work.

So in this repo's history, 76/231 issues have been open without activity for 60 days at some point in their history (Intentionally without a #. I don't need all those issues to get a link to this issue since they're not actually related to this issue)

```
539 xjulia builds are running long/hanging (?)
502 say: canonical-data needs improvement.
491 Update tests of Luhn
[...]

PR #753, opened 2017-04-11
Probot Stale default configuration

kotp:
closes: #746
Adds "good first patch" as an exception to stale determination, as
discussed [here][].

[here]: #746 (comment)

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,22 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 60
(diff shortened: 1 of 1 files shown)


stkent:
Just a question that came to mind: here you've redefined all defaults as well as the customization regarding labels. Is that a recommended strategy? (For reference, in xjava, we only overrode specific parts of the configuration; I'm not sure whether it's better to adopt defaults where possible, or to be more robust and over-specify). 🤔

kotp:
As per #746 this repository will not use the Probot.

- - - -

@stkent I changed the values based on discussion in the mentioned discussion, but did copy and paste the entirety as reference in case more was to be changed.  I think having the complete set of options that exist being in the file is helpful for quick modification.  I simply did not know if further discussion would have me make changes and wanted the information close.

PR #754, opened 2017-04-12
WIP <Exercise say in C>

srinivasamaringanti:
I will implement exercise say


petertseng:
Thanks for your interest. It seems like it would be best to open this on https://github.com/exercism/xc, would it not? Since the goal is to inform other people who may implement say in **C** that someone else (you) is already doing so, and the information is most effective if placed where contributors to the C track would look.

PR #755, opened 2017-04-17
word-search: add canonical data

### LINKED ISSUE #591: word-search: Implement canonical-data.json

stkent:
Closes #591.

Notes/thoughts:
- Single grid is used for all tests. The likelihood that a practitioner will hard-code the actual search logic in any way is low.
- Grid used for all tests matches the updated grid from the exercise `description.md`; Go track will need to update that to match this spec.
- Test cases for each of the 8 possible word directions are cumulative, which allows better matching of the problem statement ("given a puzzle and a list of words") and a more explicit prompt to "build up" the search capability piece-by-piece. All existing tracks would need to change to accommodate this new input structure, I believe.

Feedback welcome! 👂

### DIFF
--- a/exercises/word-search/canonical-data.json
+++ b/exercises/word-search/canonical-data.json
@@ -1,0 +1,676 @@
+{
+  "exercise": "word-search",
+  "version": "1.0.0",
+  "comments": [
+    "Grid rows and columns are 1-indexed.",
+    "An expected value of -1 indicates that some sort of failure should occur."
+  ],
+  "cases": [
+    {
+      "description": "Should locate words written left to right",
+      "property": "search",
+      "grid": [
+        "jefblpepre",
+        "camdcimgtc",
+        "oivokprjsm",
+        "pbwasqroua",
+        "rixilelhrs",
+        "wolcqlirpc",
+        "screeaumgr",
+        "alxhpburyi",
+        "jalaycalmp",
+        "clojurermt"
+      ],
+      "wordsToSearchFor": [
+        "clojure"
+      ],
+      "expected": {
+        "clojure": {
+          "start": {
+            "column": 1,
+            "row": 10
+          },
+          "end": {
+            "column": 7,
+            "row": 10
+          }
+        }
+      }
+    },
+    {
+      "description": "Should locate words written right to left",
+      "property": "search",
+      "grid": [
+        "jefblpepre",
+        "camdcimgtc",
+        "oivokprjsm",
+        "pbwasqroua",
+        "rixilelhrs",
+        "wolcqlirpc",
+        "screeaumgr",
+        "alxhpburyi",
+        "jalaycalmp",
+        "clojurermt"
+      ],
+      "wordsToSearchFor": [
+        "clojure",
+        "elixir"
+      ],
+      "expected": {
+        "clojure": {
+          "start": {
+            "column": 1,
+            "row": 10
+          },
+          "end": {
+            "column": 7,
+            "row": 10
+          }
+        },
+        "elixir": {
+          "start": {
+            "column": 6,
+            "row": 5
+          },
+          "end": {
+            "column": 1,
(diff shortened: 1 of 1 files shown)


exercises/word-search/canonical-data.json
```diff
@@ -0,0 +1,593 @@
...
+        "jefblpepre",
+        "camdcimgtc",
+        "oivokprjsm",
+        "pbwasqroua",
+        "rixilelhrs",
+        "wolcqlirpc",
+        "screeaumgr",
+        "alxhpburyi",
+        "jalaycalmp",
+        "clojurermt"
```
stkent: As far as I can tell there's no way to share this repeated input between tests.

petertseng: Indeed - before the schema, we had repeated inputs for`tournament` and `pov`. The schema did not support that, so in #699 we discussed and decided not to add the support.

petertseng: EBWODP: It was `grep`, not `tournament`.

stkent: Ah, thanks for the link; given that discussion, I think the duplication here is acceptable.

exercises/word-search/canonical-data.json
```diff
@@ -0,0 +1,594 @@
...
+      ],
+      "wordsToSearchFor": [
+        "clojure",
+        "elixir"
+      ],
+      "expected": [
+        {
```
petertseng: It occurs to me that the only indication that the first element in this array indicates the position of `"clojure"` and the second that of `"elixir"` is convention.

I'm sorry that I cannot speak for all tracks, but I know that in Go the tests at least arrange such that output is a map whose keys are strings and corresponding values are what we have here (start/end column/row).

So I would like to ask about that. It seems unclear to leave it up to convention.

Another interesting item is that it appears the Go track is the only track that asks for multiple words at once. It's true that that matches the description. If it has to be that way (rather than simply changing the description), I would like to make sure that interface is solid before proceeding

exercises/word-search/canonical-data.json
```diff
@@ -0,0 +1,594 @@
...
+        "elixir",
+        "ecmascript",
+        "rust",
+        "java",
+        "lua",
+        "lisp",
+        "ruby",
+        "haskell"
+      ],
+      "expected": -1
```
petertseng: If it becomes the case that the output can allow easy identification of which start/end pair corresponds to which word, I didn't find it too necessary to error on the entire output if just a single word is missing.

stkent:
@petertseng great observations! I'm at a conference right now, but I'll push some updates based on your suggestions over the weekend :o)

stkent:
> Another interesting item is that it appears the Go track is the only track that asks for multiple words at once. It's true that that matches the description. If it has to be that way (rather than simply changing the description), I would like to make sure that interface is solid before proceeding

Not only does this canonical data better match the existing description, but the end result also better reflects what I would consider to be a general word search solver (I've never seen a word search puzzle that asks you to locate a single word 😄). That's the reason I elected to adjust the input structure in the data rather than adjust the problem description. ⚖️

stkent:
Updated!

petertseng:
EBWODP: `require "set"` not necessary, left over from copy from previous problem, sorry. Deleting.

stkent:
> DIRS = [-1, 0, 1].product([-1, 0, 1]).tap { |x| x.delete([0, 0]) }.freeze

😍

Insti:
Thanks @stkent :heart:

PR #756, opened 2017-04-17
saddle-points: add canonical data

### LINKED ISSUE #582: saddle-points: Implement canonical-data.json
exercises in sync.

The Saddle Points exercise can be found in the [./exercises/saddle-points/][exercise-dir] directory.

See http://exercism.io/contribute/canonical-data/saddle-points for the up-to-date list of
language tracks that have an implementation of the Saddle Points exercise.

[exercise-dir]: https://github.com/exercism/x-common/tree/master/exercises/saddle-points

stkent:
Closes #582.

Notes:
- Per https://github.com/exercism/discussions/issues/41, the row/column identification tests found in several tracks represent testing intermediate implementation details and should be removed.
- Larger matrices have been removed from the tests; they did not exercise anything that was not already covered (except possibly for differing matrix dimensions?)
- Readme example is included as the first exercise; the majority of tracks already test this and it's a nice start point.
- Empty example is included; only a few tracks currently test this, but it's a nice boundary case.
- Multiple saddle points test uses the more-common matrix as found in e.g. xcsharp. xpython and xscala will need to update their multiple saddle points tests.
- Example with saddle point in the lower right matrix entry has been added to check for off-by-one errors. This is new for all tracks.
- Example with no saddle points has been expanded from 2x2 matrix to 3x3 matrix to give implementers more of a hint as to how one might go about constructing such a matrix. This is new for all tracks.
- No check for invalid matrix size is included in the canonical suite; can be included if desired, in which case I'll pull the xpython input for that test.

### DIFF
--- a/exercises/saddle-points/canonical-data.json
+++ b/exercises/saddle-points/canonical-data.json
@@ -1,0 +1,87 @@
+{
+  "exercise": "saddle-points",
+  "version": "1.0.0",
+  "comments": [
+    "Matrix rows and columns are 0-indexed."
+  ],
+  "cases": [
+    {
+      "description": "Can identify single saddle point",
+      "comments": [
+        "This is the README example."
+      ],
+      "property": "saddlePoints",
+      "input": [
+        [9, 8, 7],
+        [5, 3, 2],
+        [6, 6, 7]
+      ],
+      "expected": [
+        {
+          "row": 1,
+          "column": 0
+        }
+      ]
+    },
+    {
+      "description": "Can identify that empty matrix has no saddle points",
+      "property": "saddlePoints",
+      "input": [
+        []
+      ],
+      "expected": []
+    },
+    {
+      "description": "Can identify lack of saddle points when there are none",
+      "property": "saddlePoints",
+      "input": [
+        [1, 2, 3],
+        [3, 1, 2],
+        [2, 3, 1]
+      ],
+      "expected": []
+    },
+    {
+      "description": "Can identify multiple saddle points",
+      "property": "saddlePoints",
+      "input": [
+        [4, 5, 4],
+        [3, 5, 5],
+        [1, 5, 4]
+      ],
+      "expected": [
+        {
+          "row": 0,
+          "column": 1
+        },
+        {
+          "row": 1,
+          "column": 1
+        },
+        {
+          "row": 2,
+          "column": 1
+        }
+      ]
+    },
+    {
+      "description": "Can identify saddle point in bottom right corner",
+      "comments": [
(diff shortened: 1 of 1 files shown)


exercises/saddle-points/canonical-data.json
```diff
@@ -0,0 +1,87 @@
...
+        }
+      ]
+    },
+    {
+      "description": "Can identify that empty matrix has no saddle points",
+      "property": "saddlePoints",
+      "input": [
+        []
+      ],
+      "expected": []
```
ErikSchierboom: Shouldn't this be `null` instead of an empty array? The README indicates it does:

> if the input is valid but there is no result for the input, the value at "expected" should be null.

exercises/saddle-points/canonical-data.json
```diff
@@ -0,0 +1,87 @@
...
+    },
+    {
+      "description": "Can identify lack of saddle points when there are none",
+      "property": "saddlePoints",
+      "input": [
+        [1, 2, 3],
+        [3, 1, 2],
+        [2, 3, 1]
+      ],
+      "expected": []
```
ErikSchierboom: Same as above

exercises/saddle-points/canonical-data.json
petertseng: In my interpretation, the empty result is a result, therefore it is not no result. `[]` makes more sense than `null` here.

ErikSchierboom: Okay, understood

stkent: Nice spot, my bad. On reflection, it would seem that this is a case where it makes sense to update the description; I'll push a commit with that change this weekend :o)

stkent: @ErikSchierboom forgive me, but I don't see where the line you quoted exists. The description for this exercise reads as follows:

> A matrix may have zero or more saddle points.
> 
> Your code should be able to provide the (possibly empty) list of all the
saddle points for any given matrix.

Can you please point me to the source of the line you quoted? Thanks!

ErikSchierboom: @stkent It's in this repository's [README](https://github.com/exercism/x-common/blob/master/README.md). However, I think the empty list might after all be better in this case.

stkent: Ah, now I see, thanks! And yes, I think we now have three votes for an empty results list rather than an explicit `null` result :).

Insti:
Thanks for your work on this @stkent

PR #760, opened 2017-04-18
Fixed type of the first plaintextSegments test

icyrockcom:
The test case for the first `plaintextSegments` test case had a wrong type (string vs. array).

Let me know how it looks.


icyrockcom:
Thanks @petertseng!

PR #772, opened 2017-05-05
phone-number: Remove test using malformed input.

Insti:
> Your task is to clean up differently formated telephone numbers by removing punctuation and the country code (1) if present.

Letters are not punctuation and this is not primarily a string parsing exercise.
So remove the test using `1a2b3c4d5e6f7g8h9i0j` as input.


Insti:
Thanks for the reviews. :+1:

Issue #720, opened 2017-03-13
tournament: should reject inputs containing invalid lines


petertseng:
The tournament exercise currently asks the implementation to silently [ignore invalid lines](https://github.com/exercism/x-common/pull/299).

I would say it is not good design for an API to silently ignore invalid inputs, as that may produce surprising results for users of that API.

It seems more helpful to have inputs containing invalid lines to be rejected (in whatever way is appropriate for the target language).

My recommendation is:

* change the [description](https://github.com/exercism/x-common/blob/master/exercises/tournament/description.md) to state that input with invalid lines are rejected.
* change the [cases in the canonical data](https://github.com/exercism/x-common/blob/master/exercises/tournament/canonical-data.json) to expect errors for inputs with invalid lines.
  * I suggest that each test case contain only one invalid line, so that we can be sure that each individual invalid line is rejected. If a test case contained multiple invalid lines we might not be sure that an implementation would reject all invalid lines in that single case.

rbasso:
Considering that we just changed the test data in #717, adding a test case for invalid data, but also removing some more detailed information about the meaning of each type of invalid line, I am linking here the [latest version of the file before the change](https://github.com/exercism/x-common/blob/04905ab9455417b0bfb610350bf4c264494e4326/exercises/tournament/canonical-data.json), for convenience.

The information may be needed to generate the new test cases rejecting invalid data.

Insti:
Alternatively, only have test cases that contain valid lines.

This exercise is about parsing a specified format and converting it into a pretty table, it is not a general string parsing problem.

petertseng:
I thought about it. No complaints here. I'll see if there are any before officially changing the recommendation in the issue title.

To be clear, to me:

"Do not test invalid inputs" is only *slightly* preferred to "Test that invalid inputs are rejected", and both of those are **strongly** preferred to "Test that invalid inputs are ignored".

PR #773, opened 2017-05-05
tournament: Start with simpler tests / remove invald input test.

Insti:
The previous version of these tests "threw you in at the deep end." and expected you to have a fully formed solution immediately.

This PR adds a series of smaller, simpler tests that can be incrementally passed while building up to a full solution.

Included: Removal of test for malformed input data.
This is not a problem about string parsing, so it is OK to expect that all test input will be valid.

### DIFF
--- a/exercises/tournament/canonical-data.json
+++ b/exercises/tournament/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "tournament",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "comments": [
     "The inputs and outputs are represented as arrays of strings to",
     "improve readability in this JSON file.",
@@ -65,25 +65,5 @@
         "Devastating Donkeys            |  3 |  0 |  1 |  2 |  1"
       ]
     },
-    {
-      "comments": [
-        "Invalid input lines in an otherwise-valid game",
-        "still results in valid output."
-      ],
-      "description": "mostly invalid lines",
-      "property": "tally",
-      "input": [
-        "",
-        "Allegoric Alaskans@Blithering Badgers;draw",
-        "Blithering Badgers;Devastating Donkeys;loss",
-        "Devastating Donkeys;Courageous Californians;win;5",
-        "Courageous Californians;Allegoric Alaskans;los"
-      ],
-      "expected": [
-        "Team                           | MP |  W |  D |  L |  P",
-        "Devastating Donkeys            |  1 |  1 |  0 |  0 |  3",
-        "Blithering Badgers             |  1 |  0 |  0 |  1 |  0"
-      ]
-    }
   ]
 }
--- a/exercises/tournament/canonical-data.json
+++ b/exercises/tournament/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "tournament",
-  "version": "1.2.0",
+  "version": "1.3.0",
   "comments": [
     "The inputs and outputs are represented as arrays of strings to",
     "improve readability in this JSON file.",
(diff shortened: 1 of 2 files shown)


Insti:
Resolves: #720

exercises/tournament/description.md
```diff
@@ -60,8 +60,3 @@ Devastating Donkeys;Courageous Californians;draw
 ```
 
 Means that the Devastating Donkeys and Courageous Californians tied.
-
-Your program should only process input lines that follow this format.
```
petertseng: This PR comment is inspired by the fact that one of the questions asked in https://github.com/exercism/xrust/issues/122 is "should I expect invalid lines?" despite the fact that invalid lines were in fact included in the test suite at the time (so the answer to the question was specified in the test suite)

The alternative to removing completely is a sentence saying "You need only process input lines that follow this format, as invalid lines will not be tested."

I very weakly support the addition of such a sentence, because:

One might consider this a useful addition if students will ask this question and it is useful to get an answer without looking at the test suite. Note that it is easier to figure out "there is an invalid input" (you only need to look for any one test case that has it) than "there are no invalid inputs" (you need to look through **all** test cases and confirm that everything is valid).

Wouldn't it be the case that forcing the student to look at every test case to answer that question is going against the one-test-at-a-time metholodogy? Even if we pretended this is a scenario where I am writing the tests from scratch (instead of having them be provided but not run, as various tracks do), that I would want to know whether I would have to deal with invalid tests ahead of time?

Side notes:

* I support removing "program" because #321.
* No opinion expressed on the one-test-at-a-time approach.

Insti: Thanks @petertseng , good points.

> "should I expect invalid lines?"

Isn't "invalid input will not be tested" up to the tracks to decide, since they are free to add or remove tests. 

>  Wouldn't it be the case that forcing the student to look at every test case to answer that question is going against the one-test-at-a-time metholodogy?

Why would they be writing code to handle it without a failing test first? Once they get to the end they might notice that they have not had to handle invalid input.

(I'm not against adding something about test input validity to the description.)

petertseng: > > Wouldn't it be the case that forcing the student to look at every test case to answer that question is going against the one-test-at-a-time metholodogy?
>
> Why would they be writing code to handle it without a failing test first?

I don't think someone doing test-first will write that code without a test. But I don't think this affects whether I would want this question answered (whether I want the sentence to appear in the README) before seeing/writing a test. 

If I pretend that I'm writing tests from scratch, I imagine I am aware of some set of requirements before having written any test and they inform the tests that I will write. I would look to the README for those requirements.

The general question would be stated as "What is the role of an exercise's README? To what extent need it describe requirements and non-requirements?" I would say that https://github.com/exercism/xgo/issues/189 affects how I think about this question (steers me to ask more of a README).

Maybe nobody will worry about this. After all, the proposed sentence states a non-requirement. The consequence of omission is that maybe people code too defensively. That's less severe than the consequence of omitting a requirement.

It also does not really affect me since I generally do look at all the tests beforehand.

> > "should I expect invalid lines?"
>
> Isn't "invalid input will not be tested" up to the tracks to decide, since they are free to add or remove tests.

Oh! So benefit of leaving the sentence off is to give tracks the option. Okay.

I hope we don't run into a situation where all tracks that don't take the option have to each add a HINTS.md saying "we won't test invalid inputs" if we leave the option open **and** students express a preference for answering the question upfront in the README. Of course, we don't need to think about that until it becomes a problem.

---

As I hope is indicated by my prior approval, it does not matter to me whether the proposed sentence is added or not, but I'm happy to continue discussing if it inspires anyone else to have a strong opinion.

petertseng:
(https://github.com/petertseng/x-common/blob/verify/exercises/tournament/verify.rb says the new cases are good)

exercises/tournament/description.md
Insti: My preference is to not add anything to the readme about what is tested.

I'd like to try the new readme that doesn't mention test input validity at all for a while and If it proves to be an issue in the future, someone can submit a PR to add the relevant language.

petertseng: 👍 (replacing this comment with reaction)

exercises/tournament/canonical-data.json
```diff
@@ -1,6 +1,6 @@
 {
   "exercise": "tournament",
-  "version": "1.1.0",
+  "version": "1.3.0",
```
ErikSchierboom: I might have missed it, but shouldn't this be 1.2.0?

exercises/tournament/description.md
ErikSchierboom: I'm with @Insti: omitting that line about validity seems to make the most sense to me.

exercises/tournament/canonical-data.json
Insti: Removing the invalid input requirement was '1.2.0' adding the new initial tests made it '1.3.0', since '1.2.0' was not a separate PR and has never been released I can see an argument that we should stay on '1.2.0'.

I have a slight preference for '1.3.0' since it requires no work to change.

But if you think it should be '1.2.0' I'll change it.

ErikSchierboom: Ah I see. I don't really have a problem with using 1.3.0, I just didn't understand. It's fine to leave it as is

Insti:
Merged. Thanks for the reviews. :+1:

PR #779, opened 2017-05-09
Add a file to keep track of track maintainers.

Insti:
https://github.com/exercism/discussions/issues/97 investigated the status of all Exercism language tracks. This file is intended to continue and maintain that process in a community driven manner.

My intention is for this version to be merged and then further PR's can fill out/update the rest of the information.

See "Ruby" under the "Maintained" section for an example of what information I think there should be for each track.

Ideally a bot could keep it updated, but until someone writes one we will just have to do it manually.

### DIFF
--- a/TRACK_MAINTAINERS.md
+++ b/TRACK_MAINTAINERS.md
@@ -33,73 +33,69 @@ TODO
 ### Orphaned
 
 * Bash
+* Ceylon
 * CoffeeScript
+* ColdFusion
 * D
 * Groovy
+* Haxe
+* Idris
 * Kotlin
+* Nim
 * PL/SQL
+* Pascal
 * Perl 5
 * Pony
+* PowerShell
+* Purescript
 * R
 * Standard ML
+* TECO
+* Tcl
+* The Netwide Assembler
+* VB.NET
+* Vimscript
 
 ### Endangered
 
-* Lua
-* Lisp
-* Erlang
-* Go
-* Factor
-* Perl 6
-* PHP
 * C++
 * Elm
 * Emacs Lisp
+* Erlang
+* Factor
+* Go
+* Lisp
 * Lisp Flavored Erlang
+* Lua
 * MIPS Assembly
+* PHP
+* Perl 6
+* Prolog
 * Racket
 * Scheme
-* Prolog
 
 ### At Risk
 
-* Swift
-* Rust (but very active)
-* OCaml
-* JavaScript
-* ECMAScript
-* Python
-* Scala
-* F#
 * C#
 * Clojure
+* ECMAScript
+* F#
+* JavaScript
+* OCaml
 * Objective-C
+* Python
+* Rust
+* Scala
+* Swift
 
 ### Maintained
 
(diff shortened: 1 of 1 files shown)


TRACK_MAINTAINERS.md
```diff
@@ -0,0 +1,105 @@
...
+* F#
+* C#
+* Clojure
+* Objective-C
+
+### Maintained
+
+* Ruby   
+   Active: @kotp, @insti, @hilary, @tommyschaefer, @bmulvihill  
+   Alumni: @kytrynx, @bernardoamc, @Cohen-Carlisle
```
petertseng: you spelled her name wrong, is kytrinyx (unless there is in fact a secret kytrynx that I am not aware of)

TRACK_MAINTAINERS.md
```diff
@@ -0,0 +1,105 @@
...
+### Inactive Tracks
+
+There has not been enough activity to generate a heartbeat for the following tracks:
+
+* Ceylon
+* ColdFusion
+* Haxe
+* Idris
+* The Netwide Assembler
+* Nim
```
Smarticles101: Should be Vim?

petertseng: Nim is for https://github.com/exercism/xnim (Nim should remain. If Vimscript is to be added, it should be added separately)

Smarticles101: Ah, I see now

TRACK_MAINTAINERS.md
Insti: Thanks. 
How embarrassing :blush:

petertseng:
for the purposes of this document, is there a need to make a distinction between orphaned and inactive?

Insti:
> for the purposes of this document, is there a need to make a distinction between orphaned and inactive?

Probably not. Shall I merge the sections? Should it be under Orphaned or Inactive?

petertseng:
> > for the purposes of this document, is there a need to make a distinction between orphaned and inactive?
> 
> Probably not. Shall I merge the sections? Should it be under Orphaned or Inactive?

I don't suppose I care so much, but if I were to suggest the path of least resistance, then I pick Orphaned, since that is the one that has the definition in the list (orphaned, endangered, etc.)

Theodus:
I would consider the Pony track at risk, but not orphaned.

Insti:
> I would consider the Pony track at risk, but not orphaned.

Can you submit a PR for this once the file is merged.
I'd rather not bog this PR down with making sure all the tracks are correct.

Insti:
We probably should also link to the repositories.

Insti:
> There does not seem to be much point for the @ preceding the names, since they do not get linked. Unless there is some per-user tool that will do it.

I was hoping Github would magically work.

Theodus:
@Insti Sure, no problem

Insti:
@Theodus it's merged now, you can submit your PR.

Theodus:
@Insti Thanks!

PR #790, opened 2017-05-13
atbash-cipher: Fix casing and punctation in example description.

derifatives:


Insti:
Looks reasonable. Thanks.

Issue #789, opened 2017-05-13
difference-of-squares: add 0 test cases


weibeld:
I propose to add the test cases `"square of sum 0"` and `"sum of squares 0"` to the test suite. In this way, the edge case of 0 would be tested properly.

For `"difference of squares"` the 0 input is tested, but not for `"square of sum"` and `"sum of squares"`. So, I think it should be the same for all three cases.

Insti:
If you're not making a PR that already has it, it is useful to link to the thing you are referring to. In this case the difference-of-squares canonical data: https://github.com/exercism/x-common/blob/master/exercises/difference-of-squares/canonical-data.json
and description: https://github.com/exercism/x-common/blob/master/exercises/difference-of-squares/description.md

weibeld:
Ok, I will make a PR 👍

petertseng:
Looks like we decided on 1 instead of 0 in #791.

PR #791, opened 2017-05-14
difference-of-squares: add tests for boundary cases

weibeld:
See #789.

Zero is by some definitions regarded as a natural number, so an implementation should be able to handle this case, as well as the edge case of 1.

Summary of changes:

- Add tests for edge cases 0 and 1
- Remove test for input 10 (there are tests for 5 and 100 and if the implementation works for these numbers, it most likely also works for 10)
- Bump version number of the test suite to 1.1.0

I think the [description](https://github.com/exercism/x-common/blob/master/exercises/difference-of-squares/description.md) is ok, no need to include 0 there as it has anyway no influence on the result.

### DIFF
--- a/exercises/difference-of-squares/canonical-data.json
+++ b/exercises/difference-of-squares/canonical-data.json
@@ -1,21 +1,27 @@
 {
   "exercise": "difference-of-squares",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "cases": [
     {
       "description": "Square the sum of the numbers up to the given number",
       "cases": [
         {
+          "description": "square of sum 0",
+          "property": "squareOfSum",
+          "number": 0,
+          "expected": 0
+        },
+        {
+          "description": "square of sum 1",
+          "property": "squareOfSum",
+          "number": 1,
+          "expected": 1
+        },
+        {
           "description": "square of sum 5",
           "property": "squareOfSum",
           "number": 5,
           "expected": 225
-        },
-        {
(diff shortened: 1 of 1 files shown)


Insti:
Great PR description 👍 :heart:

Insti:
Is the ordering of these tests optimal?
Numeric order is tempting, but my preference would be for ease-of-implementation with the standard cases first followed by the edge cases.

Insti:
Zero is not specified as one of the natural numbers, so I am not sure this is an appropriate test case in the canonical data.

stkent:
+1 I would consider the 1-cases as the boundaries here, and remove the tests for 0, given the problem description.

weibeld:
@Insti @stkent 

- Yes, the edge cases should come after the standard cases. I had the same thought too.
- I agree that it's probably better to take 1 as the edge case and drop 0. It would be more consistent with the problem description which requires a result only for numbers >= 1. (And actually, most solution that I just checked would already handle 0 correctly.)

I'm going to commit these changes.

ErikSchierboom:
I'm also in favor of numerical ordering.

Insti:
I'm fine with numerical ordering now that none of the cases are really exceptional.

stkent:
@weibeld sounds like there's a consensus on one more ordering change (apologies!), and then this will be good to go :)

weibeld:
Done :)

stkent:
Thanks, @weibeld, nice work!

PR #793, opened 2017-05-17
Add canonical-data for poker

### LINKED ISSUE #576: poker: Implement canonical-data.json
exercises in sync.

The Poker exercise can be found in the [./exercises/poker/][exercise-dir] directory.

See http://exercism.io/contribute/canonical-data/poker for the up-to-date list of
language tracks that have an implementation of the Poker exercise.

[exercise-dir]: https://github.com/exercism/x-common/tree/master/exercises/poker

arthurchipdean:
closes #576

### DIFF
--- a/exercises/poker/canonical-data.json
+++ b/exercises/poker/canonical-data.json
@@ -3,315 +3,315 @@
   "version": "1.0.0",
   "cases": [
     {
-      "description": "it should be able to pick the winning poker hand",
-      "cases": [
-        {
-          "description": "single hand always wins",
-          "property": "best_hand",
-          "input": [
-            "4S 5S 7H 8D JC"
-          ],
-          "expected": "4S 5S 7H 8D JC"
-        },
-        {
-          "description": "highest card out of all hands wins",
-          "property": "best_hand",
-          "input": [
-            "4D 5S 6S 8D 3C",
-            "2S 4C 7S 9H 10H",
-            "3S 4S 5D 6H JH"
-          ],
-          "expected": "3S 4S 5D 6H JH"
-        },
-        {
-          "description": "a tie has multiple winners",
-          "property": "best_hand",
-          "input": [
-            "4D 5S 6S 8D 3C",
-            "2S 4C 7S 9H 10H",
-            "3S 4S 5D 6H JH",
-            "3H 4H 5C 6C JD"
-          ],
-          "expected": [
-            "3S 4S 5D 6H JH",
-            "3H 4H 5C 6C JD"
-          ]
-        },
-        {
-          "description": "multiple hands with the same high cards, tie compares next highest ranked, down to last card",
-          "property": "best_hand",
-          "input": [
-            "3S 5H 6S 8D 7H",
-            "2S 5D 6D 8C 7S"
-          ],
-          "expected": [
-            "3S 5H 6S 8D 7H"
-          ]
-        },
-        {
-          "description": "one pair beats high card",
-          "property": "best_hand",
-          "input": [
-            "4S 5H 6C 8D KH",
-            "2S 4H 6S 4D JH"
-          ],
-          "expected": [
-            "2S 4H 6S 4D JH"
-          ]
-        },
-        {
-          "description": "highest pair wins",
-          "property": "best_hand",
-          "input": [
-            "4S 2H 6S 2D JH",
-            "2S 4H 6C 4D JD"
-          ],
-          "expected": [
-            "2S 4H 6C 4D JD"
-          ]
-        },
-        {
-          "description": "two pairs beats one pair",
-          "property": "best_hand",
-          "input": [
-            "2S 8H 6S 8D JH",
-            "4S 5H 4C 8C 5C"
-          ],
-          "expected": [
-            "4S 5H 4C 8C 5C"
-          ]
-        },
-        {
-          "description": "both hands have two pairs, highest ranked pair wins",
-          "property": "best_hand",
-          "input": [
-            "2S 8H 2D 8D 3H",
-            "44S 5H 4C 8S 5D"
-          ],
-          "expected": [
-            "2S 8H 2D 8D 3H"
-          ]
-        },
-        {
-          "description": "both hands have two pairs, with the same highest ranked pair, tie goes to low pair",
-          "property": "best_hand",
-          "input": [
-            "2S QS 2C QD JH",
-            "JD QH JS 8D QC"
-          ],
-          "expected": [
-            "JD QH JS 8D QC"
-          ]
-        },
-        {
-          "description": "both hands have two identically ranked pairs, tie goes to remaining card (kicker)",
-          "property": "best_hand",
-          "input": [
-            "JD QH JS 8D QC",
-            "JS QS JC 2D QD"
-          ],
-          "expected": [
-            "JD QH JS 8D QC"
-          ]
-        },
-        {
-          "description": "three of a kind beats two pair",
-          "property": "best_hand",
-          "input": [
-            "2S 8H 2H 8D JH",
-            "4S 5H 4C 8S 4H"
-          ],
-          "expected": [
-            "4S 5H 4C 8S 4H"
-          ]
-        },
-        {
-          "description": "both hands have three of a kind, tie goes to highest ranked triplet",
-          "property": "best_hand",
-          "input": [
-            "2S 2H 2C 8D JH",
-            "4S AH AS 8C AD"
-          ],
-          "expected": [
-            "4S AH AS 8C AD"
-          ]
-        },
-        {
-          "description": "with multiple decks, two players can have same three of a kind, ties go to highest remaining cards",
-          "property": "best_hand",
-          "input": [
-            "4S AH AS 7C AD",
-            "4S AH AS 8C AD"
-          ],
-          "expected": [
-            "4S AH AS 8C AD"
-          ]
-        },
-        {
-          "description": "a straight beats three of a kind",
-          "property": "best_hand",
-          "input": [
-            "4S 5H 4C 8D 4H",
-            "3S 4D 2S 6D 5C"
-          ],
-          "expected": [
-            "3S 4D 2S 6D 5C"
-          ]
-        },
-        {
-          "description": "aces can be end a straight (10 J Q K A)",
-          "property": "best_hand",
-          "input": [
-            "4S 5H 4C 8D 4H",
-            "10D JH QS KD AC"
-          ],
-          "expected": [
-            "10D JH QS KD AC"
-          ]
-        },
-        {
-          "description": "aces can be start a straight (A 2 3 4 5)",
-          "property": "best_hand",
-          "input": [
-            "4S 5H 4C 8D 4H",
-            "4D AH 3S 2D 5C"
-          ],
-          "expected": [
-            "4D AH 3S 2D 5C"
-          ]
-        },
-        {
-          "description": "both hands with a straight, tie goes to highest ranked card",
-          "property": "best_hand",
-          "input": [
-            "4S 6C 7S 8D 5H",
-            "5S 7H 8S 9D 6H"
-          ],
-          "expected": [
-            "5S 7H 8S 9D 6H"
-          ]
-        },
-        {
-          "description": "even though an ace is usually high, a 5-high straight is the lowest-scoring straight",
-          "property": "best_hand",
-          "input": [
-            "2H 3C 4D 5D 6H",
-            "4S AH 3S 2D 5H"
-          ],
-          "expected": [
-            "2H 3C 4D 5D 6H"
-          ]
-        },
-        {
-          "description": "flush beats a straight",
-          "property": "best_hand",
-          "input": [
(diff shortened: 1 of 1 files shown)


exercises/poker/canonical-data.json
```diff
@@ -0,0 +1,317 @@
...
+          "expected": [
+            "4S 5H 4C 8C 5C"
+          ]
+        },
+        {
+          "description": "both hands have two pairs, highest ranked pair wins",
+          "property": "best_hand",
+          "input": [
+            "2S 8H 2D 8D 3H",
+            "44S 5H 4C 8S 5D"
```
petertseng: 44S

exercises/poker/canonical-data.json
```diff
@@ -0,0 +1,317 @@
...
+    {
+      "description": "it should be able to pick the winning poker hand",
+      "cases": [
+        {
+          "description": "single hand always wins",
+          "property": "best_hand",
+          "input": [
+            "4S 5S 7H 8D JC"
+          ],
+          "expected": "4S 5S 7H 8D JC"
```
petertseng: the value at `expected` should be an array, to match all the rest

exercises/poker/canonical-data.json
```diff
@@ -0,0 +1,317 @@
...
+        },
+        {
+          "description": "highest card out of all hands wins",
+          "property": "best_hand",
+          "input": [
+            "4D 5S 6S 8D 3C",
+            "2S 4C 7S 9H 10H",
+            "3S 4S 5D 6H JH"
+          ],
+          "expected": "3S 4S 5D 6H JH"
```
petertseng: the value at `expected` should be an array, to match all the rest

exercises/poker/canonical-data.json
```diff
@@ -0,0 +1,317 @@
...
+          "input": [
+            "4H 6H 7H 8H 5H",
+            "5S 7S 8S 9S 6S"
+          ],
+          "expected": [
+            "5S 7S 8S 9S 6S"
+          ]
+        }
+      ]
+    }]
```
petertseng: any reason for } and ] on same line?

exercises/poker/canonical-data.json
```diff
@@ -0,0 +1,317 @@
...
+          "input": [
+            "4S 5H 4C 8D 4H",
+            "3S 4D 2S 6D 5C"
+          ],
+          "expected": [
+            "3S 4D 2S 6D 5C"
+          ]
+        },
+        {
+          "description": "aces can be end a straight (10 J Q K A)",
```
petertseng: I think either remove "be" or add "of"

exercises/poker/canonical-data.json
```diff
@@ -0,0 +1,317 @@
...
+          "input": [
+            "4S 5H 4C 8D 4H",
+            "10D JH QS KD AC"
+          ],
+          "expected": [
+            "10D JH QS KD AC"
+          ]
+        },
+        {
+          "description": "aces can be start a straight (A 2 3 4 5)",
```
petertseng: I think either remove "be" or add "of"

exercises/poker/canonical-data.json
```diff
@@ -0,0 +1,317 @@
+{
+  "exercise": "poker",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "it should be able to pick the winning poker hand",
+      "cases": [
+        {
+          "description": "single hand always wins",
+          "property": "best_hand",
```
petertseng: `_` isn't an allowed character in https://github.com/exercism/x-common/blob/master/canonical-schema.json#L117, so I would suggest `bestHand`. Maybe even `bestHands` since there can be more than one.

exercises/poker/canonical-data.json
```diff
@@ -0,0 +1,317 @@
+{
+  "exercise": "poker",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "it should be able to pick the winning poker hand",
+      "cases": [
```
petertseng: I'm not convinced a second level of `cases` nesting is necessary. I suggest just a single level, please, unless there is a specific reason to use two, and it doesn't seem that the extra description here adds any benefit. For example, https://github.com/exercism/x-common/blob/master/exercises/hello-world/canonical-data.json uses just one level.

petertseng:
Of the existing implementations, I'm not going to check which of them have which cases, but I will check for the sake of interest:

* which of them represent the hands as a single string, and which represent them as a list of strings?
* which of them use SDCH as their suits, and which use ♤♢♧♡?
* which of them use 10 as their ten, and which use T?

I don't think any of these need to inform decisions in this PR, but they can if y'all want.

* We probably can't do ♤♢♧♡ since they violate #428.
* I prefer 10 as the ten, but I really don't care.

test | hand type | suits | tens
-----|-----------|-------|-----
https://github.com/exercism/xcsharp/blob/master/exercises/poker/PokerTest.cs | one string | SDCH | T
https://github.com/exercism/xdelphi/blob/master/exercises/poker/uPokerTest.pas | one string | SDCH | T
https://github.com/exercism/xelixir/blob/master/exercises/poker/poker_test.exs | list of strings (`~w`) | SDCH | 10
https://github.com/exercism/xfsharp/blob/master/exercises/poker/PokerTest.fs | one string | SDCH | T
https://github.com/exercism/xgo/blob/master/exercises/poker/poker_test.go | one string | ♤♢♧♡ | 10
https://github.com/exercism/xpython/blob/master/exercises/poker/poker_test.py | list of strings (`split`) | SDCH | T
https://github.com/exercism/xruby/blob/master/exercises/poker/poker_test.rb | list of strings (`%w`) | SDCH | 10
https://github.com/exercism/xswift/blob/master/exercises/poker/Tests/PokerTests/PokerTests.swift | one string | ♤♢♧♡ | 10

(Go and Swift test various unparsable hands)

petertseng:
https://github.com/petertseng/x-common/blob/verify-poker/exercises/poker/verify.rb says all cases here are correct except for the 44S.

arthurchipdean:
I copied the test cases from the elixir exercise. I added an additional level of tests because I thought that was the reason it was failing the travis ci originally. I made all the changes you mentioned. I'm just trying to get this setup so that I can implement it in a standardized way in the ECMAScript exercise track.. Thanks!

petertseng:
https://github.com/petertseng/x-common/tree/verify-poker/exercises/poker has been updated and it says all cases are correct

exercises/poker/canonical-data.json
```diff
@@ -0,0 +1,317 @@
...
+      "input": [
+        "4S 5H 4C 8D 4H",
+        "3S 4D 2S 6D 5C"
+      ],
+      "expected": [
+        "3S 4D 2S 6D 5C"
+      ]
+    },
+    {
+      "description":"aces can end a straight (10 J Q K A)",
```
petertseng: missing a space after the colon (consistency with other cases)

petertseng:
> two hands that each have a pair.

I mean "two hands that each have the same pair", meaning to exercise comparison of the other cards. There is already a case that compares two one-pair hands.

exercises/poker/canonical-data.json
```diff
@@ -0,0 +1,317 @@
...
+      "input": [
+        "2S 2H 2C 8D 2D",
+        "4S 5H 5S 5D 5C"
+      ],
+      "expected": [
+        "4S 5H 5S 5D 5C"
+      ]
+    },
+    {
+      "description": "with multiple decks, both hands with identical four of a kind, tie determined by kicker" ,
```
stevejb71: I've not heard the term kicker before, it may be poker jargon which not everyone will understand.

arthurchipdean: I'd say the term could be easily understood by the context. Look at the test data, the meaning is ridiculously obvious. Plus the word kicker is actually in the dictionary as a poker term so if they opened a dictionary they would also get the answer. If we start dumbing down the description, where do we draw the line? Should I limit my word choices to a certain reading level?

arthurchipdean: @stevejb71 Also the bowling test data contains words like Turkey, Double, etc... Are you saying we should change these to say: 3 strikes in a row, 2 strikes in a row. It seems very silly.

exercises/poker/canonical-data.json
```diff
@@ -0,0 +1,317 @@
...
+      "input": [
+        "2S QS 2C QD JH",
+        "JD QH JS 8D QC"
+      ],
+      "expected": [
+        "JD QH JS 8D QC"
+      ]
+    },
+    {
+      "description": "both hands have two identically ranked pairs, tie goes to remaining card (kicker)",
```
petertseng: it is defined.

exercises/poker/canonical-data.json
petertseng: > the bowling test data contains words like Turkey, Double, etc...

errr? https://github.com/search?q=org%3Aexercism+turkey&type=Code

and just in case that missed something, nothing in https://github.com/exercism/x-common/blob/master/exercises/bowling/canonical-data.json either.

???

confused

---

Do you know though, in this case the problem is solved: "kicker" is defined earlier https://github.com/exercism/x-common/pull/793/files#r117595066

petertseng: Therefore, this description is allowed to rely on the earlier definition!

arthurchipdean: @petertseng  Even if the word turkey is not being used other bowling jargon like Strike and Spare are both being used. My point is still valid....

arthurchipdean: @petertseng  Does that make sense? I see that the term has already been defined so it doesn't matter for this case but let's get it straight for future contributions. Can you use jargon related to the game if the test is related to a game?

petertseng: > Strike and Spare are both being used

Thanks. I see they are defined at https://github.com/exercism/x-common/blob/master/exercises/bowling/description.md .

So in the same way one would ask is it OK to have "straight", "flush", "full house"? (Would anyone have trouble understanding what "pair" means?) The https://github.com/exercism/x-common/blob/master/exercises/poker/description.md incorporates https://en.wikipedia.org/wiki/List_of_poker_hands by reference, can there be any room for doubt there?

petertseng: I would argue we have done what we need to do - the terms are defined in the README or earlier in the file, right?

If there is any case where we failed to do that, in this exercise or any other, it seems good to correct it.

arthurchipdean: I agree. Thanks for clarifying. :)

exercises/poker/canonical-data.json
arthurchipdean: I believe that case is already taken care of by this test: "both hands have two identically ranked pairs, tie goes to remaining card (kicker)".

arthurchipdean:
I have fixed all formatting and other issues. Please let me know if there is anything else I need to do.

Thank you!

exercises/poker/canonical-data.json
petertseng: > > potential case to add: two hands that each have the same pair, to exercise comparison of the other cards
>
> I believe that case is already taken care of by this test: "both hands have two identically ranked pairs, tie goes to remaining card (kicker)".

Interesting. I hadn't thought of it that way, since that was a two-pair hand rather than a one-pair hand, but I can see it working out.

OK, just that extra space

... which you just put in, ok.

Insti:
I strongly prefer `T` for 10

arthurchipdean:
@Insti Since none are blockers --- Thanks for your irrelevant nitpicking.

petertseng:
Looks like it's time to merge, with thanks!

I agree that the non-blocking comments are non-blocking; I'll say what I think about them!

petertseng:
> T for 10

Tough choice, though I admit T has the advantage of taking up the same amount of space... but 10 is what you would see on an actual playing card. No way to please everyone here. Best I can think of is that tracks can convert in their generators as necessary. In the most extreme case, we could represent cards in JSON as `{"rank": 10, "suit": "spades"}` and let the tracks decide how to represent each hand, but it seems a bit too verbose to be my current suggestion.

> 2 hands as the second test rather than jumping straight to 3 hands

This one's interesting, wondering why it wasn't deemed blocking. For me, I'm not the kind of person who likes to write code that would only compare two hands only to have to replace it later; I would just make it work right from the beginning. So I guess that's why I deemed it non-blocking. Any other reason to take into account? Is it easier to debug if there are only two hands? That would be a reason to want that case.

> human readable way that showed the correct ranking from left to right
> Then add further tests at the end that tested that everything worked with unordered hands.

That one is interesting too! It would help readability of the test cases, though perhaps the description already helps here. I would become slightly less confident that the code under test can handle any order... though I guess I have unknown confidence right now since I don't know if there's any rhyme or reason to the orders right now. I'm slightly worried that the number of tests will increase a lot if we have to make sure that every kind of hand can be recognised in an arbitrary order, but I guess it's hard to know without seeing a concrete proposal. I suppose if someone wants to work on it and propose it that person can say so (sorry, not going to be that person)

Issue #335, opened 2016-08-12
[secret-handshake]: Clarify readme


Insti:
In reviewing PR #334 I identified a few areas of the [description](https://github.com/exercism/x-common/blob/master/exercises/secret-handshake/description.md) that need clarification:
1. Be explicit about input types and when decimal/binary conversions are appropriate. (Resolved by #489 - answer is never)
2. Make it clearer in the first example that a decimal to binary conversion has happened. (Resolved by #489)
3. Make it clearer in the second example that a reverse has taken place. (Resolved by #489)
4. **What to do about numbers greater than 2^5 ('11111')** ("will not be tested" is fine)
5. The final line: `The program should consider strings specifying an invalid binary as the value 0.` is not particularly useful. (Resolved by #489)

petertseng:
my suggestion of https://github.com/exercism/x-common/pull/159#discussion_r58829790 is that we do not do binary to decimal conversions

Insti:
So this is a bit matching problem and should only have decimal integers as inputs.

Insti:
By NOT updating the description with what to do about numbers greater than 32 we can leave it up to the tracks about whether they want to test them themselves.

Add a comment to the canonical-data.json that says something like:

Numbers greater than 2^5 ('11111') are not tested in this canonical-data, whether to test numbers greater than 32 is left up to the tracks to decide.

PR #794, opened 2017-05-17
secret-handshake: remove tests for numbers > 31

weibeld:
I saw a solution which just returned an empty list for all numbers >= 32. This is obviously wrong, as the 5 least significant bits should define the sequence of signals, no matter how large the number is. However, the current test suite tests only 32, and so all the tests passed for this solution.

I added therefore a couple of test cases for numbers > 32.

Not sure if the problem description should also be updated to make this clearer.

### DIFF
--- a/exercises/secret-handshake/canonical-data.json
+++ b/exercises/secret-handshake/canonical-data.json
@@ -1,6 +1,19 @@
 {
   "exercise": "secret-handshake",
-  "version": "1.0.0",
+  "version": "1.1.0",
+  "comments": [
+    " In a discussion in https://github.com/exercism/x-common/pull/794 and    ",
+    " https://github.com/exercism/x-common/issues/335 it has been decided to  ",
+    " only include numbers between 0 and 31 (00000 to 11111) in the canonical ",
+    " test data.                                                              ",
+    "                                                                         ",
+    " This is to allow for different implementations in different tracks and  ",
+    " not restrict solutions to bitwise or modulo-based algorithms.           ",
+    "                                                                         ",
+    " Tracks may include additional tests for numbers > 31 in their test      ",
+    " suites. In this case, 32 (100000) should yield the same result as 0,    ",
+    " 33 (100001) should yield the same result as 1, and so on.               "
+  ],
   "cases": [
     {
       "description": "Create a handshake for a number",
@@ -69,12 +82,6 @@
           "description": "do nothing for zero",
           "property": "commands",
           "input": 0,
-          "expected": []
-        },
-        {
-          "description": "do nothing if lower 5 bits not set",
-          "property": "commands",
-          "input": 32,
           "expected": []
         }
       ]


petertseng:
> if the problem description should also be updated to make this clearer.

If it did, then it would close #335.

petertseng:
And I don't know about y'all but I personally ask a lot out of my READMEs. Maybe nobody else cares.

Insti:
While I appreciate the PR, I'm against adding tests for numbers bigger than 32 as it adds a requirement for bitmasking / modulo usage that is not otherwise required.
I like that it is possible\* to solve the current version with a giant case statement or lookup table.

\* Although don't think these are necessarily "best practice" solutions.

petertseng:
An interesting question to be had as well - should 32 be removed too?

weibeld:
Ok, it could make sense to have only numbers up to 2^5 = 31 in the canonical test data and leave the testing of larger numbers to the tracks. After all, I don't know what support for bitwise operation all these languages have, so the exercise could be solved in different ways in different languages.

So in this case I think that also 32 should be removed as it has a bit set beyond the 5 lowest bits. Just for consistency.

Insti:
> 32 should be removed as it has a bit set beyond the 5 lowest bits. 

If you could add a comment about [deliberately only testing 0-31]( https://github.com/exercism/x-common/issues/335#issuecomment-301993641)  to the json file that would be great.

See an example of how this is handled in [triangle](https://github.com/exercism/x-common/blob/master/exercises/triangle/canonical-data.json#L4)

weibeld:
Done 👍

exercises/secret-handshake/canonical-data.json
```diff
@@ -1,6 +1,17 @@
...
+  "comments": [
+    " In a discussion in #794 and #335 it has been decided to only include   ",
+    " numbers between 0 and 31 (00000 to 11111) in the canonical test data.  ",
+    "                                                                        ",
+    " This is to allow for different implementations in different tracks and ",
+    " not restricting implementations to use bitwise or modulo operations.   ",
+    "                                                                        ",
+    " It is up to each track to include numbers > 31 in the concrete test    ",
+    " suites. In this case, 32 (100000) should yield the same result as 0,   ",
+    " 33 (100001) should yield the same result as 1, and so on.              "
```
Insti: This reads to me a bit like tracks are required to add tests for numbers > 31

"Tracks may include additional tests for numbers > 31, in this case..."

exercises/secret-handshake/canonical-data.json
```diff
@@ -1,6 +1,17 @@
 {
   "exercise": "secret-handshake",
   "version": "1.1.0",
+  "comments": [
+    " In a discussion in #794 and #335 it has been decided to only include   ",
+    " numbers between 0 and 31 (00000 to 11111) in the canonical test data.  ",
+    "                                                                        ",
+    " This is to allow for different implementations in different tracks and ",
+    " not restricting implementations to use bitwise or modulo operations.   ",
```
Insti: Suggest: "not restrict solutions to bitwise or modulo based algorithms. "

Insti:
Looks good, 👍 Just some minor wording clarifications. Feel free to disagree with them.

weibeld:
@Insti thanks for the suggestions, I adapted the wording.

weibeld:
Another question: should we also mention something about negative numbers?

A solution using bitwise operations would naturally also work with negative numbers. However, for other types of solutions this would be more difficult and it depends on the used signed number representation.

Or would this make the exercise too complicated?

Insti:
I think it's fine not to mention negative numbers at all.

weibeld:
I also prefer to keep it simple and not introduce negative numbers. If a track really wants to do it, they can still do it, for example, Java in which the signed number representation is defined by the JVM.

Insti:
The comment: 

> In a discussion in #794 and #335 it has been decided to only include numbers between 0 and 31 (00000 to 11111) in the canonical test data.

Already covers not including negative numbers.

The only slight change I would suggest is replacing the `#794` and `#355` with the real links `https://github.com/exercism/x-common/pull/794` and `https://github.com/exercism/x-common/issues/335`

weibeld:
Should I squash the 4 commits into 1?

Insti:
Yes, that sounds like a good idea in this case.

weibeld:
Squashed!

petertseng:
A long time without objections, so let's take it. Thanks.

PR #795, opened 2017-05-17
Add new exercise, the dice

aodpi:


exercises/the-dice/metadata.yml
```diff
@@ -0,0 +1,2 @@
+---
+blurb: "Count number of unique pair of 6s afther rolling the dice"
```
petertseng: have you a source and source_url for this exercise, like other exercises have?

If it is a problem you made up yourself, you could do like https://github.com/exercism/x-common/blob/master/exercises/paasio/metadata.yml#L3 did.

Or if it is acquired from somewhere, you should enter the appropriate values in this file (important so we can know it is OK for us to use it)

aodpi:
Thank you for feedback, actually die is a spelling error, i shall improve the exercise.

exercises/the-dice/metadata.yml
```diff
@@ -1,2 +1,4 @@
 ---
-blurb: "Count number of unique pair of 6s afther rolling the dice"
\ No newline at end of file
+blurb: "Count number of unique pair of 6s afther rolling the dice"
```
petertseng: "afther" should be "after"

exercises/the-dice/canonical-data.json
```diff
@@ -0,0 +1,41 @@
+{
+  "exercise": "the-dice",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "two pairs of 6s",
+      "property": "calculatePairs",
+      "trial": [1, 1, 6, 6, 2, 6, 6],
```
petertseng: this is implicitly also testing the ability to find 6s at the end, which means by the time we get to the actual "6s at the end" case it may not be testing anything new

exercises/the-dice/canonical-data.json
```diff
@@ -0,0 +1,41 @@
...
+      "expected": 1
+    },
+    {
+      "description": "big series",
+      "property": "calculatePairs",
+      "trial": [2, 6, 6, 6, 6, 2, 4, 5, 1, 6, 6, 2, 6, 6, 3, 6, 6, 6],
+      "expected": 2
+    },
+    {
+        "description": "last one",
```
petertseng: indented more than the other cases

petertseng: "last one" doesn't seem very descriptive, though I'm not sure I have anything better. "four pairs"?

exercises/the-dice/canonical-data.json
```diff
@@ -0,0 +1,41 @@
...
+    {
+      "description": "6s at start",
+      "property": "calculatePairs",
+      "trial": [6, 6, 1, 1, 2],
+      "expected": 1
+    },
+    {
+      "description": "6s at end",
+      "property": "calculatePairs",
+      "trial": [1, 1, 2, 3, 6, 4, 6, 6],
```
petertseng: this is also the first case where a single six appears, I wonder if that's worth calling out earlier.

aodpi: The main objective is to find pair of 6s. Pair of 6s at the end of the array might be tricky. It is not worth calling out earlier.

exercises/the-dice/description.md
```diff
@@ -0,0 +1,11 @@
+Let's play dice! 
+
+The traditional dice is a cube. Each of its six faces shows a different number of dots from 1 to 6. 
+
+Your programming task is related to the analysis of sequences produced by rolling a dice.
+Create a program named Dice to solve the following exercise. The input of the program, a so-called trial,
```
petertseng: As #321 has it, please write to not have "program". If it helps, you can see how we dealt with this in #528.

exercises/the-dice/description.md
```diff
@@ -0,0 +1,13 @@
+Let's play dice! 
+
+The traditional dice is a cube. Each of its six faces shows a different number of dots from 1 to 6. 
+
+Your programming task is related to the analysis of sequences produced by rolling a dice.
+Solve the following Dice exercise:
+
+The input of the program, a so-called trial,
```
petertseng: still has "program" here!

exercises/the-dice/canonical-data.json
```diff
@@ -0,0 +1,41 @@
...
+      "property": "calculatePairs",
+      "trial": [2, 6, 6, 6, 6, 2, 4, 5, 1, 6, 6, 2, 6, 6, 3, 6, 6, 6],
+      "expected": 2
+    },
+    {
+      "description": "four pairs",
+      "property": "calculatePairs",
+      "trial": [1, 3, 4, 5, 6, 6, 4, 6, 6, 5, 5, 3, 6, 6, 3, 6, 6],
+      "expected": 4
+    }]
```
petertseng: is it intentional that the } and ] on the same line?

exercises/the-dice/canonical-data.json
```diff
@@ -0,0 +1,41 @@
+{
+  "exercise": "the-dice",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "two pairs of 6s",
```
Insti: Start with some easy test cases.

"No sixes rolled"
"A single six rolled"
"A single pair rolled"

exercises/the-dice/canonical-data.json
```diff
@@ -0,0 +1,41 @@
+{
+  "exercise": "the-dice",
```
Insti: The article "the" can be left out. Just `dice` would be better.

BUT, I think "dice" is not a good name for this problem, because it is not really about dice it is about detecting pairs.

There is an existing dice term two `1`s : [snake eyes](https://en.wikipedia.org/wiki/Snake_eyes)

I'm not aware of a term for two sixes, so I suggest changing the number that should be detected from `6` to `1` and calling the problem `snake-eyes` or `dice-snake-eyes`

stkent: Aside: two sixes is called [Boxcars](http://mathworld.wolfram.com/Boxcars.html), but I don't think that's as recognizable as snake eyes, so +1 to the suggested number swap.

PR #775, opened 2017-05-07
Make description.md self contained.

Insti:
Resolves: #767

The problem description given in description.md should be complete and self contained, and not require the addition of the blurb to make sense.

This patch prepends the blurb from the `metadata.yml` onto the `description.md` together with a newline, matching the format provided by trackler for the README.

Waiting on: Update to trackler so that readmes so that the blurb does not appear to appear twice.

PR #801, opened 2017-05-23
Add period to end of first description line.

Insti:
In #775 the first line of the description (which is/was the blurb.) was sometimes missing a
period.

This PR adds periods in the appropriate places.

This no longer exactly matches the output of the old readme, but makes the
description more grammatically correct.


Insti:
Thanks @ErikSchierboom

Issue #803, opened 2017-05-24
New exercise idea: Collatz Conjecture


kgengler:
An unsolved problem in mathematics 

https://en.wikipedia.org/wiki/3x_%2B_1_problem

Given a positive integer n, you calculate the number of steps required to reach 1 using the conjecture formula. It's an easier but interesting problem, and allows for some fun recursive solutions.

PR #804, opened 2017-05-24
add collatz conjecture exercise

kgengler:
Add the collatz conjecture exercise in #803

### DIFF
--- a/exercises/collatz-conjecture/canonical-data.json
+++ b/exercises/collatz-conjecture/canonical-data.json
@@ -1,0 +1,36 @@
+{
+  "exercise": "collatz-conjecture",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "zero steps for one",
+      "property": "collatz",
+      "number": 1,
+      "expected": 0
+    },
+    {
+      "description": "divide if even",
+      "property": "collatz",
+      "number": 16,
+      "expected": 4
+    },
+    {
+      "description": "even and odd steps",
+      "property": "collatz",
+      "number": 12,
+      "expected": 9
+    },
+    {
+      "description": "zero is an error",
+      "property": "collatz",
+      "number": 0,
+      "expected": -1
+    },
+    {
+      "description": "negative value is an error ",
+      "property": "collatz",
+      "number": -15,
+      "expected": -1
+    }
+  ]
+}
--- a/exercises/collatz-conjecture/canonical-data.json
+++ b/exercises/collatz-conjecture/canonical-data.json
@@ -4,33 +4,33 @@
   "cases": [
     {
       "description": "zero steps for one",
-      "property": "collatz",
+      "property": "steps",
       "number": 1,
       "expected": 0
     },
     {
       "description": "divide if even",
-      "property": "collatz",
+      "property": "steps",
       "number": 16,
       "expected": 4
     },
     {
       "description": "even and odd steps",
-      "property": "collatz",
+      "property": "steps",
       "number": 12,
       "expected": 9
     },
     {
       "description": "zero is an error",
-      "property": "collatz",
+      "property": "steps",
       "number": 0,
-      "expected": -1
+      "expected": { "error": "Only positive numbers are allowed" }
     },
     {
       "description": "negative value is an error ",
-      "property": "collatz",
+      "property": "steps",
       "number": -15,
-      "expected": -1
+      "expected": { "error": "Only positive numbers are allowed" }
     }
   ]
 }
--- a/exercises/collatz-conjecture/description.md
+++ b/exercises/collatz-conjecture/description.md
@@ -1,0 +1,25 @@
+The Collatz Conjecture or 3x+1 problem can be summarized as follows:
+
+Take any positive integer n. If n is even, divide n by 2 to get n / 2. If n is
+odd, multiply n by 3 and add 1 to get 3n + 1. Repeat the process indefinitely.
+The conjecture states that no matter which number you start with, you will
+always reach 1 eventually.
+
+Given a number n, return the number of steps required to reach 1.
+
+## Examples
+Starting with n = 12, the steps would be as follows:
+
+0. 12
+1. 6
+2. 3
+3. 10
+4. 5
+5. 16
+6. 8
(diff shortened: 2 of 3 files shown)


exercises/collatz-conjecture/canonical-data.json
```diff
@@ -0,0 +1,48 @@
+{
+  "exercise": "collatz-conjecture",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "1 is 0 steps",
+      "property": "collatz",
+      "number": 1,
+      "expected": "0"
```
petertseng: I would expect that the `expected` are all integers, rather than strings

exercises/collatz-conjecture/canonical-data.json
```diff
@@ -0,0 +1,48 @@
...
+      "expected": "118"
+    },
+    {
+      "description": "871 is 178 steps",
+      "property": "collatz",
+      "number": 871,
+      "expected": "178"
+    },
+    {
+      "description": "6171 is 261 steps",
```
Insti: why have you chosen these numbers?
Is it possible to pass one of these tests without passing all of them?

exercises/collatz-conjecture/description.md
```diff
@@ -0,0 +1,25 @@
...
+2. 3
+3. 10
+4. 5
+5. 16
+6. 8
+7. 4
+8. 2
+9. 1
+
+Resulting in 9 steps. So for input n = 12, the return value would be 9.
```
Insti: The example should be present in the test cases.

exercises/collatz-conjecture/canonical-data.json
```diff
@@ -0,0 +1,48 @@
...
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "1 is 0 steps",
+      "property": "collatz",
+      "number": 1,
+      "expected": "0"
+    },
+    {
+      "description": "2 is 1 step",
```
Insti: I'd like to see the descriptions describe what is being tested.
"first even number"

Insti: Next test I'd expect would be "3" as "first odd number"

Insti: 4: "Repeated even numbers" (probably should come before the test for 3)

Insti: "odd number followed by power of two"
"multiple odd numbers in the sequence"

exercises/collatz-conjecture/canonical-data.json
kgengler: I initially pulled all the values from the Wikipedia, but I had the same thought and changed the test cases to cover the n/2 and 3n+1 choices somewhat separately. I also added tests to enforce only positive numbers are valid input.

Insti:
This looks good, thanks for making those changes. :heart:

exercises/collatz-conjecture/canonical-data.json
```diff
@@ -0,0 +1,36 @@
...
+      "description": "even and odd steps",
+      "property": "collatz",
+      "number": 12,
+      "expected": 9
+    },
+    {
+      "description": "zero is an error",
+      "property": "collatz",
+      "number": 0,
+      "expected": -1
```
Insti: Would you please update the expected error value:

From: https://github.com/exercism/x-common#test-data-format-canonical-datajson
> if an error is expected (because the input is invalid, or any other reason), the value at "expected" should be an object containing exactly one property, "error", whose value is a string.
>     The string should explain why the error would occur.

exercises/collatz-conjecture/canonical-data.json
```diff
@@ -0,0 +1,36 @@
+{
+  "exercise": "collatz-conjecture",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "zero steps for one",
+      "property": "collatz",
```
Insti: Question: Should the property be "collatz" or "steps" (or something else?)

kgengler: yeah, that name makes a lot more sense.

exercises/collatz-conjecture/canonical-data.json
kgengler: Thanks for the link, I was using the error conditions in the Hamming exercise as an example.

Insti: Was there some documentation that pointed you at that one or did you just happen to choose it?

kgengler: No documentation, It just happened to be the first one I remembered that had error cases. I suppose that means the cononical-data.json file for that exercise could get updated for the error cases.

exercises/collatz-conjecture/metadata.yml
```diff
@@ -0,0 +1,4 @@
+---
+blurb: "Calculate the number of steps to reach 1 given an integer using Collatz Conjecture"
```
Insti: "given an integer" makes this sound a bit confusing to me, what about just leaving it out: 
"Calculate the number of steps to reach 1 using the Collatz conjecture"

petertseng: > "given an integer" makes this sound a bit confusing to me

same, because it makes it sound like the integer is using the Collatz conjecture

rpottsoh:
I like this new exercise and I am reminded of the version that I solved [here](https://projecteuler.net/problem=14) a couple of years ago.  It would be nice to see a test that addresses the unproven aspect of the Collatz Conjecture that the Project Euler exercise touches on.  Just a suggestion, this exercise is fine otherwise.  If there is any interest in this I can provide the highest number, under 1 million, that produces the longest chain; I'd have to review my solution to determine how long is the longest chain.

petertseng:
Thanks for the exercise. Everyone feel free to add on more aspects.

stkent:
+1 this is a nice exercise! I think adding a test for a much higher input (the stated example of the longest chain starting at a number below 1 million would be a great candidate) would be a nice way to cap the suite and really run an implementation through its paces.

PR #811, opened 2017-05-25
difference-of-squares: add tests for input == 0

mhinz:
Although it's probably hard to come up with a solution that works for all cases
but the 0 one, it should be tested for nevertheless.

When testing units that do arithmetic, 0 as input is a special case that
surprisingly often is forgotten about. And in more complicated code this could
matter.

---

For reference, see this conversation: https://github.com/exercism/xvimscript/pull/17#issuecomment-303877482

### DIFF
--- a/exercises/difference-of-squares/canonical-data.json
+++ b/exercises/difference-of-squares/canonical-data.json
@@ -6,6 +6,12 @@
       "description": "Square the sum of the numbers up to the given number",
       "cases": [
         {
+          "description": "square of sum 0",
+          "property": "squareOfSum",
+          "number": 0,
+          "expected": 0
+        },
+        {
           "description": "square of sum 1",
           "property": "squareOfSum",
           "number": 1,
@@ -29,6 +35,12 @@
       "description": "Sum the squares of the numbers up to the given number",
       "cases": [
         {
+          "description": "sum of squares 0",
+          "property": "sumOfSquares",
+          "number": 0,
+          "expected": 0
+        },
+        {
           "description": "sum of squares 1",
           "property": "sumOfSquares",
           "number": 1,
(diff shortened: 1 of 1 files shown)


petertseng:
The counterpoint is https://github.com/exercism/x-common/pull/791#issuecomment-301294437

> Zero is not specified as one of the natural numbers in the description, so I am not sure this is an appropriate test case in the canonical data.

But! I have a response to this! It does not matter whether zero is a natural number! The https://github.com/exercism/x-common/blob/master/exercises/difference-of-squares/description.md says "first N natural numbers", NOT "the natural numbers up to N". I assume that "N" is taken as "the value of `input`". It is valid to say "the first zero natural numbers", so 0 makes sense as an input without contradicting the README.

If this is accepted, **the test version needs to be bumped to 1.2.0**, so please do that.

Don't mind the fact that I'm contradicting decisions made in #791, I really don't care, so all you have to do is let me know whether I convinced you, I won't stand in your way whatever you decide.

Insti:
Hi @mhinz thanks for the PR!

We've recently discussed this in the last PR for this problem and the conclusion was that we should not add a test for zero. Have a read through the discussion starting here: https://github.com/exercism/x-common/pull/791#issuecomment-301294437

After reading that, if you still strongly believe that zero should be included feel free to argue your case.

petertseng:
> I really don't care

I mean "I really don't care what the decision on zero is", not "I don't care that I'm contradicting brashly", which you might think are the same thing, but they convey different attitudes

mhinz:
Interesting! I guess it's all about expectation.

I agree with @petertseng that `0` is a perfectly fine input according to the problem's description.

On the other hand, when all the current tests succeed, it's highly unlikely that the `0` test case would fail.

I opened this PR because I thought that, when testing functions that do arithmetic, testing for `0` would be a good practice in general.

But I'm also a fan of keeping things simple and since this problem expects all input to be treated the same, we can safely omit this edge case here.

Closing!

> If this is accepted, the test version needs to be bumped to 1.2.0, so please do that.

Oops. I'll keep that in mind for the next time!

Insti:
Thanks @mhinz 

Opening this PR was the right thing to do.

Note that you're welcome to add whatever tests you like to the specific language repositories, so if you want to add a test for 0 to xvimscript please do.

PR #822, opened 2017-06-04
collatz-conjecture: add test case for a large number

rpottsoh:
Largest number of steps for number less than one million.  This is loosely based on the Euler Project.  It asks for the number less than 1 million that has the most steps.

### DIFF
--- a/exercises/collatz-conjecture/canonical-data.json
+++ b/exercises/collatz-conjecture/canonical-data.json
@@ -21,10 +21,10 @@
       "expected": 9
     },
     {
-      "description": "Largest number of steps for number less than one million",
+      "description": "Large number of even and odd steps",
       "property": "steps",
-      "number": 837799,
-      "expected": 524
+      "number": 1000000,
+      "expected": 152
     },
     {
       "description": "zero is an error",
--- a/exercises/collatz-conjecture/canonical-data.json
+++ b/exercises/collatz-conjecture/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "collatz-conjecture",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "cases": [
     {
       "description": "zero steps for one",
@@ -19,6 +19,12 @@
       "property": "steps",
       "number": 12,
       "expected": 9
+    },
+    {
+      "description": "Largest number of steps for number less than one million",
+      "property": "steps",
+      "number": 837799,
+      "expected": 524
     },
     {
       "description": "zero is an error",


Insti:
What is this testing? 
What change to the code is required to make this test pass that would not pass the existing tests?

Insti:
(I didn't mean to close sorry. - missclick)

rpottsoh:
@Insti this new case is in response to the conversation that was started at the end of PR [#804](https://github.com/exercism/x-common/pull/804).

Insti:
OK, but I couldn't find the answer to my questions there.

stkent:
> I think adding a test for a much higher input (the stated example of the longest chain starting at a number below 1 million would be a great candidate) would be a nice way to cap the suite and really run an implementation through its paces.

Hopefully no new code would be required in most cases, but with the relatively small number and scale of other tests exercising this code, I can imagine some folk might accidentally chance upon a correct solution that is actually flawed. Exercising an extreme case such as this one reduces the likelihood of that possibility.

rpottsoh:
No new code should be required.  As you state it is meant to be a more realistic use case.  Does this help @insti?  If not I will work on a more detailed response when I can sit down at my computer to think about it further.

Sent from my android device.

Insti:
> No new code should be required.

1. So it's not testing anything?

> I can imagine some folk might accidentally chance upon a correct solution that is actually flawed. Exercising an extreme case such as this one reduces the likelihood of that possibility.

2. If such a solution is discovered, a test can be added that explicitly checks for the thing that causes that problem. It's more helpful to say: "your code doesn't work because x" rather than "your code doesn't work"

3. This test spoils a project Euler question.
Since discovering the longest chain is a different problem to calculating the chain and project Euler only tests for the correct answer, providing this answer as an unnecessary part of the test suite is essentially a spoiler.

Feel free to add this test to your tracks, but I believe it does not belong as part of the canonical data.  I'm interested in hearing your argument for inclusion @rpottsoh :heart:

rpottsoh:
@insti you make a very good point.  The last thing I want to do is create a spoiler to the Project Euler implementation.  My intent here is to essentially stress test the students solution by having it determine the correct number of steps for a very large number.  I thought adding another case was the right way to go about this, obviously my choice of “Number” and “Steps” was ill-conceived.  Is having such a stress test not a good thing for the exercise?  If it is a good thing would it be better to alter let’s say the “even and odd steps” case by using an arbitrarily large even number (like 1 million) or have another case that would be virtually the same as “even and odd steps” but might be worded as “even and odd steps for extremely large number”.

 

Based on your comments I do not like my proposed change to the JSON in its present form.  I appreciate your comments, they are very insightful and helpful.

Insti:
Thanks @rpottsoh

It is better to not spoil the project Euler problem.

>  My intent here is to essentially stress test the students solution by having it determine the correct number of steps for a very large number.

I still don't understand the value of this test. 
What does testing a large number test that a smaller one doesn't?

You say it's a 'stress test' are you trying to test that the student's machine can calculate the steps without overheating? That it doesn't run out of memory? 

Why is 1,000,000 the right number to test? Why not 100 or 100,000,000,000?

It would not surprise me if there are ways to solve this problem that would fail for cases other than those specified in the existing tests, but there's no guarantee that this test will find them either.

We should wait until we know what the flaws are before implementing tests for them. This is where the human based code-reviews on the site are useful.

rpottsoh:
> What does testing a large number test that a smaller one doesn't?

According to the [conjecture](https://en.wikipedia.org/wiki/3x_%2B_1_problem) it should prove nothing, all positive integers should end up at 1.  My original thought was that it may be appealing to the student if at least one of the tests checked a large number, one that would be difficult / or seen as too time consuming to determine the number steps by hand, for example.

> You say it's a 'stress test' are you trying to test that the student's machine can calculate the steps without overheating? That it doesn't run out of memory?

Sorry, I was speaking colloquially.  I was trying to imply that by testing a larger value it could be a more rigorous test of the student's solution to the exercise, forcing their solution to have to iterate more than a handful of times to arrive at the correct answer.  More iterations might reveal weaknesses in the design of the student's solution.
 
> Why is 1,000,000 the right number to test? Why not 100 or 100,000,000,000?

I selected 1 million as it will yield a substantially larger number of steps to reach 1 than was covered in the earlier tests of this exercise.  In fact, I could have picked 97, it has a relatively high number of steps (118) compared to 1 million  (152), however, 1 million would likely be perceived as a large number and might be more satisfying to the student to see their solution compute the correct number of steps.

> We should wait until we know what the flaws are before implementing tests for them. This is where the human based code-reviews on the site are useful.

I am fine with this.

I hope I have done a better job of explaining my intentions and goal of this PR.  Feel free to close this PR if you still feel the issue should be tabled, perhaps for another time, once the exercise has been implemented and possible improvements are discussed.

Insti:
Here's a 'funny' story:

I tried solving this exercise, since the exercise does not exist in the Ruby track yet I just did it  using TDD and without starting from the tests in the canonical-data.json. Once I had a working solution with the minimum test cases, I was done but all the low numbers involved were somehow unsatisfying. So I thought I would try including the big number test.  

And the test failed!  

It seems I had a guard clause to stop infinite looping that I had not removed. Without this test that bug would not have been so easily discovered.

So I have changed my opinion on this test case.

* Although not technically required for the algorithm, I found a large number test satisfying.
* It did detect a useful bug.


Thanks for your kindness and patience @rpottsoh :heart:

rpottsoh:
@Insti don't mention it and thank you for taking the time to try the problem.  I had not considered guarding against an infinite loop when I first worked on this a couple of years ago, smart.  Discussing this has been fun.  :+1:

PR #823, opened 2017-06-05
Update dominoes canonical-data expected key

stevejb71:
It was "can-chain", but most other exercises have "expected" as the expected output of a property being tested.


stevejb71:
This brings the exercise into line with many others, and simplifies the OCaml test generator slightly.

I upgraded the major version based on this line in the readme - to update it when:
"Insertion, deletion or renaming of keys in the test data object."

PR #825, opened 2017-06-07
Remove "ruby" from fenced code block

fxn:
The fenced code block does not contain Ruby.

Matters because some editors highlight the content of fenced code blocks, and you get `true`, `in`, and others colored.


stkent:
Nice catch! LGTM.

PR #829, opened 2017-06-14
fix ToC listing

katrinleinweber:
Hello!

I presume that it's OK in the future for such small, non-content changes to be "Commit[ed] directly to the master branch." Just being careful at the first PR ;-)

Cheers!


kytrinyx:
I actually prefer a PR—even if you then merge it yourself, because then everyone who is following the repo sees what is happening on it.

Thanks for the fix 🌷

katrinleinweber:
Will do! 🌻  Didn't realise that was possible until after submitting the PR & comment.

PR #831, opened 2017-06-17
difference-of-squares: test >32 bit results

stkent:
For an input value of 2000, the square of the sum (4004001000000), sum of the squares (2668667000), and difference of squares (4001332333000) all exceed the largest (signed) value that can be represented using 32 bits (2147483647). The proposed test cases here exercise this case and should guide implementors to learn more about the numerical types available in their language of choice.

[Discussion in xjava](https://github.com/exercism/xjava/issues/331).

Thoughts about including this change in the `x-common` tests? I think `xjava` will retain the tests regardless, but am interested in perspectives based on languages that have alternative and/or unusual numerical types.

### DIFF
--- a/exercises/difference-of-squares/canonical-data.json
+++ b/exercises/difference-of-squares/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "difference-of-squares",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "cases": [
     {
       "description": "Square the sum of the numbers up to the given number",
@@ -22,6 +22,12 @@
           "property": "squareOfSum",
           "number": 100,
           "expected": 25502500
+        },
+        {
+          "description": "square of sum 2000",
+          "property": "squareOfSum",
+          "number": 2000,
+          "expected": 4004001000000
         }
       ]
     },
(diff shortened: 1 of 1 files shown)


Insti:
I am not in favor of adding this test to x-common.

I am in favor of adding a different exercise that illustrates this concept.

exercises/difference-of-squares/canonical-data.json
```diff
@@ -22,6 +22,12 @@
           "property": "squareOfSum",
           "number": 100,
           "expected": 25502500
+        },
+        {
+          "description": "square of sum 2000",
```
Insti: If this **were** to be added, a description that describes why the test is necessary should be used.
"result greater than 2^32" or something similar.

(Note consistency is not a good defence here as all the other tests are also poorly named.)

stkent:
Interesting POV. I guess you could say that `grains` focuses a user on numerical types more directly than this exercise would?

exercises/difference-of-squares/canonical-data.json
stkent: Agreed!

ErikSchierboom:
> I guess you could say that grains focuses a user on numerical types more directly than this exercise would?

I agree. Grains is the exercise where I introduce big integers in two of the tracks I work on (C# and F#). Adding it here too does not add much IMHO.

stkent:
Cool, I'll close this and take the feedback to the Java track. Thanks for weighing in!

PR #838, opened 2017-06-20
rna-transcription: correct abbreviations in user-facing descriptions

katrinleinweber:
Because RNA & DNA are biochemically defined abbreviations, they should not be in lower-case, unless there is a technical reason (like in `property` fields or in slugs).

### DIFF
--- a/exercises/rna-transcription/canonical-data.json
+++ b/exercises/rna-transcription/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "rna-transcription",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "comments": [
     "Language implementations vary on the issue of invalid input data.",
     "A language may elect to simplify this task by only presenting valid",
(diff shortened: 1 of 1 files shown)


kotp:
Thank you @katrinleinweber and good eye!

exercises/rna-transcription/canonical-data.json
```diff
@@ -15,49 +15,49 @@
...
     {
-      "description": "rna complement",
+      "description": "RNA complement",
       "property": "toRna",
       "dna": "ACGTGGTCTTAA",
       "expected": "UGCACCAGAAUU"
     },
     {
-      "description": "dna correctly handles invalid input",
+      "description": "DNA correctly handles invalid input",
```
petertseng: had to double-take here. DNA is correct since the input is being transformed from DNA to RNA.

katrinleinweber: It seems that "DNA there is more of a tag/slug, than a meaningful part of the sentence, or a grammar issue. I'll fix that.

PR #839, opened 2017-06-21
Collatz-conjecture: remove trailing space in test name

kgengler:
Super small change, but I noticed this is screwing up some test generators


stkent:
Nice catch! This should probably be accompanied by a PATCH-level version bump per https://github.com/exercism/x-common#patch-version-changes

kgengler:
ah, thanks @stkent ! I even thought about that when initially looking through this, but completely forgot when making the changes.

PR #840, opened 2017-06-21
spiral-matrix: New exercise proposal.

rbasso:
I just found [this exercise](https://www.reddit.com/r/dailyprogrammer/comments/6i60lr/20170619_challenge_320_easy_spiral_ascension/) in [r/dailyprogrammer](https://www.reddit.com/r/dailyprogrammer/), and it seems to be a good fit for Exercism.

The exercise is really simple, but it allows very distinct solutions:

- Simple matrix filling, element by element.
- Recursive building without rotations.
- Recursive building with rotations.
- Non-recursive building with a recursive element formula.
- Non-recursive building with a non-recursive element formula.

The original problem asked for printing a properly formatted matrix, but I just used lists of lists in the `canonical-data.json` to avoid forcing one more string manipulation exercise. This way each track can decide by itself what would be the proper representation for the matrix, and we keep everything as language neutral as possible.

The first test case checks for the empty matrix, which may or may not make sense for a language, depending on your choice of matrix representation. I included it just for completeness, because it seems to make sense from a [mathematical point of  view](https://en.wikipedia.org/wiki/Matrix_(mathematics)#Empty_matrices), but I don't think it is important here.

### DIFF
--- a/exercises/spiral-matrix/description.md
+++ b/exercises/spiral-matrix/description.md
@@ -1,0 +1,22 @@
+Given the size, return a square matrix of numbers in spiral order.
+
+The matrix should be filled with natural numbers, starting from 1
+in the top-left corner, increasing in a inward, clockwise spiral order,
+like these examples:
+
+###### Spiral matrix of size 3
+
+```plain
+1 2 3
+8 9 4
+7 6 5
+```
+
+###### Spiral matrix of size 4
+
+```plain
+ 1  2  3 4
+12 13 14 5
+11 16 15 6
+10  9  8 7
+```
--- a/exercises/spiral-matrix/canonical-data.json
+++ b/exercises/spiral-matrix/canonical-data.json
@@ -1,0 +1,52 @@
+{
+  "exercise": "spiral-matrix",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "empty spiral",
+      "property": "spiralMatrix",
+      "input": 0,
+      "expected": []
+    },
+    {
+      "description": "trivial spiral",
+      "property": "spiralMatrix",
+      "input": 1,
+      "expected": [ [1] ]
+    },
+    {
+      "description": "spiral of size 2",
+      "property": "spiralMatrix",
+      "input": 2,
+      "expected": [ [1, 2]
+                  , [4, 3] ]
+    },
+    {
+      "description": "spiral of size 3",
+      "property": "spiralMatrix",
+      "input": 3,
+      "expected": [ [1, 2, 3]
+                  , [8, 9, 4]
+                  , [7, 6, 5] ]
+    },
+    {
+      "description": "spiral of size 4",
+      "property": "spiralMatrix",
+      "input": 4,
+      "expected": [ [ 1,  2,  3, 4]
+                  , [12, 13, 14, 5]
+                  , [11, 16, 15, 6]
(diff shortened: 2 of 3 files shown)


ErikSchierboom:
Very nice addition!

rbasso:
Thanks! 😄

exercises/spiral-matrix/description.md
```diff
@@ -0,0 +1,22 @@
+Given a number, return a matrix of numbers in spiral order.
```
stkent: Maybe `Given a square number N, return a matrix filled with the numbers 1 to N arranged in a spiral`?

exercises/spiral-matrix/description.md
```diff
@@ -0,0 +1,22 @@
...
+
+The matrix should be filled with natural numbers, starting from 1
+in the top-left corner, increasing in a inward, clockwise spiral order,
+like these examples:
+
+- Spiral matrix of size 2
+
+    ```plain
```
stkent: Indenting should be unnecessary here since you are using the triple backticks?

stkent: `Spiral matrix of size 2` - where does the 2 come from here? The given matrix is 3x3!

exercises/spiral-matrix/description.md
```diff
@@ -0,0 +1,22 @@
...
+    1 2 3
+    8 9 4
+    7 6 5
+    ```
+
+- Spiral matrix of size 5
```
stkent: `Spiral matrix of size 5` - where does the 5 come from here? The given matrix is 4x4!

exercises/spiral-matrix/description.md
stkent: Same for the blurb, perhaps?

stkent:
Added a few suggestions/questions. Fun exercise idea!

exercises/spiral-matrix/description.md
rbasso: > Indenting should be unnecessary here since you are using the triple backticks?

The indenting was to nest the code block inside each item of the lists, but the result isn't as nice as I expected. I'll change it.

> where does the 2 come from here? The given matrix is 3x3!

Ops! I changed the examples and forgot to update it! 👍

exercises/spiral-matrix/description.md
rbasso: Nice catch!

exercises/spiral-matrix/description.md
rbasso: I reworded it, trying to keep it as short as possible and avoid repeating the next paragraph. I hope it is better now.

exercises/spiral-matrix/description.md
stkent: > The indenting was to nest the code block inside each item of the lists, but the result isn't as nice as I expected. I'll change it.

In that case you might need four spaces (rather than two)? See e.g. https://gist.github.com/clintel/1155906

> Ops! I changed the examples and forgot to update it!

Hah, I figured something like that, thanks for updating!

exercises/spiral-matrix/description.md
stkent: Looks good!

stkent:
@rbasso one additional comment that might help with list formatting, if you're inclined to experiment; LGTM overall!

Issue #827, opened 2017-06-11
hamming: "empty strands" should be first


NobbZ:
Currently the test asks for the distance of two equal strands of length 1. After that different ones, and gets even longer. After the third or fourth tests, most students will have realized that it is best to do a loop or use recursion (or higher order functions).

After realizing this, one gets the empty list for free and will probably not even realize that there was that one test at the end which checks for the empty strands.

Conclusion: At its current position I could remove that test as well and the student wouldn't recognize a difference in the exercise.

Proposal: As an empty list or string is the base case for every recursion over such containers and is also handled as a noop in for-each-like constructs, the most logical position for this test were the first. This would make the student initially aware of this special case and he would not implement it just "by accident". Also a student might realize to use a loop/recursion/higher-order-functions by the second or third test already (so about a test earlier).

Insti:
https://github.com/exercism/x-common/blob/master/exercises/hamming/canonical-data.json

ErikSchierboom:
Good idea! Would you be willing to send a PR to modify the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/hamming/canonical-data.json)?

NobbZ:
I added the `discussion`-label hoping there will be some around this topic, but since there has no-one chimed in so far and vetoed, I think we can remove `discussion` and make it a `good first patch`.

katrinleinweber:
Hello! I presume it doesn't _have_ to be a 1st-time-contributor ;-)

PR #844, opened 2017-06-24
hamming: test for empty input (as proposed in #827)

katrinleinweber:


stkent:
Just a heads up, this should have been a change to 1.1.0 per https://github.com/exercism/problem-specifications#minor-version-changes (my emphasis):

> **MINOR version changes**
> 
> The MINOR version should change when you add functionality in a backwards-compatible manner, make non-breaking changes that alter the meaning of the test suite, make previously passing solutions possibly fail, or failing solutions fail at a different spot.
> 
> There are examples of changes requiring a MINOR version change:
> 
> - Adding or deleting test cases.
> - Changing the test cases inputs and/or outputs.
> - **Changing the test cases ordering.**

katrinleinweber:
Argh, my apologies! Shall I fix my version bump?

I looked through [CONTRIBUTING.md](https://github.com/exercism/problem-specifications/blob/2ed3655/CONTRIBUTING.md): shall I also include the versioning guidelines there?

stkent:
No worries! I think a quick bump now would be good; it's unlikely that anyone has regenerated yet :)

It does seem like the specification should be linked in the contributing guidelines - if you're also able to separately address that it would be awesome!

PR #845, opened 2017-06-24
fix versioning

katrinleinweber:
as explained in https://github.com/exercism/problem-specifications/pull/844#issuecomment-310842590


stkent:
Thanks!

PR #848, opened 2017-06-24
pangram: character set issue, closes #842

### LINKED ISSUE #842: pangram: character set issue
Originally reported as https://github.com/exercism/purescript/issues/53.

[The current description](https://github.com/exercism/problem-specifications/blob/f375051787370bd1af1dd84e8d6c116f07d88ad2/exercises/pangram/description.md) says:

> The alphabet used is ASCII, and case insensitive, from 'a' to 'z' inclusively.

One of [the current canonical test cases](https://github.com/exercism/problem-specifications/blob/f375051787370bd1af1dd84e8d6c116f07d88ad2/exercises/pangram/canonical-data.json) is:

> "the_quick_brown_fox_jumps_over_the_lazy_dog"

It's worth clarifying whether the non-ASCII and non-'a'-to'z' are supposed to be ignored or not at all present in the cases.

Given https://github.com/exercism/problem-specifications/pull/741, I assume that non-ASCII are supposed to be removed altogether

It's not so clear whether that applies to non-'a'-to-'z' and what to do if these are encountered. Filtering out is probably the way to go, but worth clarifying or updating the canonical test cases.

icyrockcom:
An update for https://github.com/exercism/problem-specifications/issues/842. Clarifies that non-ASCII symbols should not be present at all, compared to them being present, but ignored.

### DIFF
--- a/exercises/pangram/description.md
+++ b/exercises/pangram/description.md
@@ -2,5 +2,5 @@ Determine if a sentence is a pangram. A pangram (Greek: παν γράμμα, pan
 "every letter") is a sentence using every letter of the alphabet at least once.
 The best known English pangram is "The quick brown fox jumps over the lazy dog."
 
-The alphabet used is ASCII, and case insensitive, from 'a' to 'z'
-inclusively.
+The alphabet used consists of ASCII letters `a` to `z`, inclusive, and is case
+insensitive. Input will not contain non-ASCII symbols.


exercises/pangram/description.md
```diff
@@ -2,5 +2,6 @@ Determine if a sentence is a pangram. A pangram (Greek: παν γράμμα, pan
 "every letter") is a sentence using every letter of the alphabet at least once.
 The best known English pangram is "The quick brown fox jumps over the lazy dog."
 
-The alphabet used is ASCII, and case insensitive, from 'a' to 'z'
-inclusively.
+The alphabet used consists of ASCII letters `a` to `z`, inclusive, and is case
+insensitive. Per [#741](https://github.com/exercism/problem-specifications/pull/741),
```
stkent: This file is served to users, so including this PR reference feels like we're leaking implementation details. What is the objective of this additional sentence - to convey to users that their solutions should treat any non-ASCII characters as invalidating the input?

icyrockcom: It seemed to me that many users preferred to add non-ASCII test cases. I wanted to give some history. Perhaps I should remove altogether and just retain the objective?

stkent: That would make more sense to me. Does this change also mean we need to remove tests containing anything other than the letters `a`-`z`? What about spaces?

icyrockcom: Sure thing, pull request updated. No tests need to be updated, since the change is to just clarify that tests should be ASCII-clean, but otherwise can contain characters that are not letters (which should be just skipped). That is, current tests already contain only ASCII symbols already.

exercises/pangram/description.md
```diff
@@ -2,5 +2,5 @@ Determine if a sentence is a pangram. A pangram (Greek: παν γράμμα, pan
 "every letter") is a sentence using every letter of the alphabet at least once.
 The best known English pangram is "The quick brown fox jumps over the lazy dog."
 
-The alphabet used is ASCII, and case insensitive, from 'a' to 'z'
-inclusively.
+The alphabet used consists of ASCII letters `a` to `z`, inclusive, and is case
+insensitive. Input should not contain non-ASCII symbols.
```
petertseng: This sentence seems a bit misplaced or misworded, since I think of "should" as either

1. Stating something that is most likely the case, but speaker is not sure
2. Stating a requirement that the reader is expected to conform to

 Were I a student reading this in the README that was given to me via `exercism fetch`, I would think "Should? Didn't you just serve me this exercise and its test cases? You know for sure whether the input contains non-ASCII symbols! Or are you asking me to write my own test cases?"

The sentence seems directed at, say, a contributor to problem-specifications, rather than a student. Does that sound accurate?

stkent: Good call, I couldn't place my finger on this earlier but you captured my niggling feeling well!

petertseng: Perhaps "will not" (rather than "should") will be clear to students. In addition it seems to still serve as sufficient indication to contributors to problem-specifications about restrictions on the input (if that is a desirable trait).

icyrockcom: That sounds good to me, pull request updated so it says "will not" instead.

petertseng:
Well I'm going to trust the "closes #842" declaration, though I personally do not read that this new description addresses one of the two questions raised in #842:

> It's worth clarifying whether the non-ASCII and non-'a'-to'z' are supposed to be ignored or not at all present in the cases.

petertseng:
EBWODP: I meant just the non-'a'-to-'z' part

petertseng:
Thanks

PR #849, opened 2017-06-24
Replace x-common with problem-specifications

mienaikage:


package.json
```diff
@@ -1,9 +1,9 @@
 {
-  "name": "x-common",
+  "name": "problem-specifications",
```
stkent: Does this change have any ramifications? I'm unfamiliar with whatever consumes this JSON file.

stkent:
One question; looks 💯 otherwise!

package.json
mienaikage: I don't believe so. As far as I can tell, this file is only used to check that canonical-data files for all exercises are compatible with the canonical-schema.

PR #852, opened 2017-06-25
fix pangram test for duplicate mixed-case chars

rsslldnphy:
From what I can see, the last test case is inteded to catch incorrect
programs such as this, which don't take into account different-cased
duplicates:

```haskell
isPangram = (== 26) . length . nub . filter isAlpha
```

However the above program passed the final test as the count of distinct
upper and lower case characters is 27, rather than 26.

This commit changes the test text to a non-pangram that has exactly 26
distinct upper and lower case characters, which causes the above,
incorrect program to fail.

Similar incorrect programs that checked if the count of distinct letters was >=, rather than exactly = to 26, would have correctly failed the existing test.

### DIFF
--- a/exercises/pangram/canonical-data.json
+++ b/exercises/pangram/canonical-data.json
@@ -3,7 +3,7 @@
   "comments": [
     "A pangram is a sentence using every letter of the alphabet at least once."
   ],
-  "version": "1.0.0",
+  "version": "1.1.0",
   "cases": [
     {
       "description": "Check if the given string is an pangram",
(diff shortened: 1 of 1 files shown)


rsslldnphy:
Related, in the Haskell track (where I spotted this): https://github.com/exercism/haskell/pull/565

petertseng:
In my review:

* I verified that the proposed case has exactly 26 letters between uppercased and lowercased letters.
* I agree that it is better to have exactly 26 letters, not 27, and thus this case is an improvement over the existing one.
* I verified that the proposed case is not a pangram (it has 24 letters).

Please change the minor version, as https://github.com/exercism/problem-specifications#minor-version-changes

rsslldnphy:
👍, I have added a version bump to the commit.

petertseng:
Very good, thanks!

PR #854, opened 2017-06-25
Add test for Forth that an operator can be overridden.

stevejb71:
I think it's a reasonable edge case that's not covered by the existing tests.

### DIFF
--- a/exercises/forth/canonical-data.json
+++ b/exercises/forth/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "forth",
-  "version": "1.1.0",
+  "version": "1.2.0",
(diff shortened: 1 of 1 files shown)


stevejb71:
(minor) version bumped

petertseng:
> "is this allowed in Forth" 

It is allowed by both https://repl.it/languages/forth and https://skilldrick.github.io/easyforth/, and in the absence of another Forth interpreter of my own I will assume that they are authoritative.

```
JS-Forth 0.5200804171342
http://www.forthfreak.net/jsforth.html
This program is published under the GPL.
   : * + ;
   3 4 *
=> 7
```

PR #862, opened 2017-07-08
Clarify given range defines factors not product

stevensonmt:
Also fix `it's` vs `its` typos.

### DIFF
--- a/exercises/palindrome-products/description.md
+++ b/exercises/palindrome-products/description.md
@@ -15,18 +15,18 @@ of multiple combinations of numbers. For example, the palindrome product `9` has
 the factors `(1, 9)`, `(3, 3)`, and `(9, 1)`.
 
 Write a program that given a range of integers, returns the smallest and largest
-palindromic product within that range, along with all of it's factors.
+palindromic product of factors within that range, along with all the factors in the range for that product.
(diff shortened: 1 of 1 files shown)


petertseng:
The guidelines at https://github.com/exercism/docs/blob/master/contributing-to-language-tracks/pull-request-guidelines.md request that the exercise should go in the first line of the commit message(s).

As evidenced by the fact that this commit message says "also", the commit is doing two completely separate things (1. clarify the range defines the factors 2. it's -> its) and thus it should be two separate commits.

The counterpoint to my argument is that we are highly unlikely to ever bisect this repo nor revert either of the two resulting commits. OK, if you buy the counterargument, I promise I will not complain if it is merged as is. I was simply obligated to point it out.

petertseng:
EBWODP: In addition to my previous comment which suggests things to change, I also say:

This is a good modification to the problem specification. Thanks for doing it.

stevensonmt:
Point taken @petertseng.  I didn't think the `its` vs `it's` corrections warranted a separate PR, though technically perhaps it should have been.

petertseng:
Separate PR not necessarily needed; two commits in same PR would make sense

PR #863, opened 2017-07-08
list-ops: update foldl/foldr tests

### LINKED ISSUE #826: list-ops: add simpler foldl/foldr tests to create gentler difficulty slope?
### Context

- [Prior discussion that touches lightly on the topic at hand](https://github.com/exercism/x-common/pull/521)
- [`kotlin` PR that generated this `x-common` issue](https://github.com/exercism/kotlin/pull/57)

### Description 

The [current canonical data for `list-ops`](https://github.com/exercism/x-common/blob/da6b039ecf8dfc1a9bc8b29759c848cb0c94ab6e/exercises/list-ops/canonical-data.json) includes the following non-degenerate cases for the `foldl` and `foldr` functions:

```json
{
  "description": "division of integers",
  "property": "foldl",
  "list": [1, 2, 3, 4],
  "initial": 24,
  "function": "value / accumulator",
  "expected": 64
}
```

and 

```json
{
  "description": "division of integers",
  "property": "foldr",
  "list": [1, 2, 3, 4],
  "initial": 24,
  "function": "value / accumulator",
  "expected": 9
}
```

The folding captured by these tests is quite tricky to reason about due to

1. the oscillating nature of the accumulating result, and
2. the fact that the input and expected output values are _all_ integers, even though most solutions (at least in statically typed languages) will need to use higher-precision numerical representations.

Since folding is one of the more complex operations we're asking implementors to build in this exercise, I propose (h/t @sdavids13) that we insert _at least one_ simpler test right before each of the division-based tests captured above. In particular, we could follow the example of `xpython` (which is currently not compliant with the canonical data), and have implementors first pass a test that uses a simpler operation (addition, subtraction, multiplication, or all of the above) whose expected result is easier to validate and whose implementation can be integer-only. Examples for `foldl`:

```json
{
  "description": "addition of integers",
  "property": "foldl",
  "list": [1, 2, 3, 4],
  "initial": 5,
  "function": "accumulator + value",
  "expected": 15
}
```

and/or 

```json
{
  "description": "subtraction of integers",
  "property": "foldl",
  "list": [1, 2, 3, 4],
  "initial": 5,
  "function": "accumulator - value",
  "expected": -5
}
```

and/or 

```json
{
  "description": "multiplication of integers",
  "property": "foldl",
  "list": [1, 2, 3, 4],
  "initial": 5,
  "function": "accumulator * value",
  "expected": 120
}
```

Inviting feedback on this idea; happy to handle any changes resulting from discussion :)

stkent:
Closes #826.

Note that I bumped the major version of the tests since this is a breaking change; the representation of the accumulation function used in the fold tests was changed to highlight that the _same_ function is being used in the direction-dependent tests. This should help mitigate confusion regarding what is meant when we declare that the given accumulation function is 'direction-dependent'. See e.g. https://github.com/exercism/java/pull/671#discussion_r125188010 for context.

Questions:

- when writing `x / y`, is there a common way to indicate that this represents integer division in particular? I think that would be useful for implementors.
- would it be useful to include a link to https://en.wikipedia.org/wiki/Fold_(higher-order_function)#Linear_folds to help implementors understand the expected signatures of the fold functions (and in particular, the expected interpretation of the accumulation function parameters)?

### DIFF
--- a/exercises/list-ops/canonical-data.json
+++ b/exercises/list-ops/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "list-ops",
-  "version": "1.0.0",
+  "version": "2.0.0",
   "comments": [
     "Though there are no specifications here for dealing with large lists,",
     "implementers may add tests for handling large lists to ensure that the",
@@ -113,16 +113,24 @@
           "property": "foldl",
           "list": [],
           "initial": 2,
-          "function": "value / accumulator",
+          "function": "(x, y) -> x * y",
           "expected": 2
         },
         {
-          "description": "division of integers",
+          "description": "direction independent function applied to non-empty list",
           "property": "foldl",
           "list": [1, 2, 3, 4],
-          "initial": 24,
-          "function": "value / accumulator",
-          "expected": 64
+          "initial": 5,
+          "function": "(x, y) -> x + y",
+          "expected": 15
+        },
+        {
+          "description": "direction dependent function applied to non-empty list",
+          "property": "foldl",
+          "list": [2, 5],
+          "initial": 5,
+          "function": "(x, y) -> x / y",
+          "expected": 0
         }
       ]
     },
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
Thanks!

PR #864, opened 2017-07-10
rotational-cipher: improve test case order

stkent:
Rotating by 0 seems to be an easier starting point that rotating by 1, and leads to more of a natural progression (rotate by 0 -> 1 -> 26, rather than 1 -> 26 -> 0).


Smarticles101:
I agree.  Rotating by 0 is the simplest case and should therefore be the first

PR #865, opened 2017-07-16
fix grammar error

rpottsoh:
The word `and` is not needed in the sentence.


rpottsoh:
@kytrinyx I stumbled on this minor grammatical error while looking for instructions regarding the uuid.  What I am looking for is instruction on where the uuid value comes from when adding an exercise to config.json.  How is the uuid generated for exercises that may be unique to a single track?  Maybe it isn't documented yet.

kytrinyx:
Thanks @rpottsoh - I haven't documented officially yet (it will go in https://github.com/exercism/docs as soon as I do). For now there are notes in https://github.com/exercism/meta/issues/16 (linked to from the PR I made).

PR #870, opened 2017-07-31
palindrome-products: Fix  description.md

derifatives:
Before this change, the description states (twice) that the palindrome product 9 has the factors (1, 9) , (3, 3), and (9, 1). The tests (at least in both Python and Haskell) will accept (3, 3) and either one of (1, 9) or (9, 1), as pairs are canonicalized going into the test. This problem is plenty hard enough without adding an extra useless hoop of doubling each product, and I think deleting the (9, 1)  more accurately matches the intent of the problem (although I admit it's arguable). I also fixed a typo.


ErikSchierboom:
This PR has conflicts with the master branch. Could you try rebasing?

derifatives:
I'm not sure how to rebase (new to github), but I merged by hand.

PR #875, opened 2017-08-05
Hamming: Updates canonical-data.json to comply with schema

### LINKED ISSUE #812: hamming: Update expected errors in canonical data
From: https://github.com/exercism/x-common#test-data-format-canonical-datajson
> if an error is expected (because the input is invalid, or any other reason), the value at "expected" should be an object containing exactly one property, "error", whose value is a string.
>     The string should explain why the error would occur.

But the [hamming canonical data](https://github.com/exercism/x-common/blob/master/exercises/hamming/canonical-data.json#L113) has an expected value of  `-1`

Todo: 

Update https://github.com/exercism/x-common/blob/master/exercises/hamming/canonical-data.json to use error objects for the expected value in the error cases:  
e.g.  `"expected"   : { "error": "some helpful message" }`

m-a-ge:
Should resolve #812

### DIFF
--- a/exercises/hamming/canonical-data.json
+++ b/exercises/hamming/canonical-data.json
@@ -1,6 +1,6 @@
 {
 "exercise": "hamming",
-"version": "1.1.0",
+"version": "2.0.0",
   "comments": [
     "Language implementations vary on the issue of unequal length strands.",
     "A language may elect to simplify this task by only presenting equal",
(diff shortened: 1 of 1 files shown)


m-a-ge:
@ErikSchierboom, thanks for pointing out to the issue. 
My bad (still new to the project).
I'll update the PR later toady

ErikSchierboom:
@m-a-ge No problem! Thanks for taking the effort to do a PR.

exercises/hamming/canonical-data.json
```diff
@@ -110,14 +110,14 @@
       "property": "distance",
       "strand1": "AATG",
       "strand2": "AAA",
-      "expected": -1
+      "expected": {"error": "leftStrand and rightStrand must be of equal length"}
```
stevejb71: ```leftStrand``` looks a bit like a Java parameter name. How about 'left strand'? Or 'strands are of different lengths'?

m-a-ge: It makes sense, will change it as well

m-a-ge:
@ErikSchierboom, I've made requested changes, please verify 👮

PR #881, opened 2017-08-17
Rename twofer property to name

jpreese:
I'd like to propose to rename the twofer property to instead use name. I noticed this because the C# track really can't follow the canonical data. 

For this exercise, following proper casing and convention, the call becomes TwoFer.TwoFer() which isn't legal as the method matches the name of the class.


m-a-ge:
This make sense, thank you!

jpreese:
No problem!

PR #886, opened 2017-08-22
topics: add "state management"

tleen:
I think there is a place for a topic concerning state
management. Specifically I am thinking of an exercise like
bank-account where you need to maintain account state.

The intention for this topic is to apply to exercises with long-term
state management, spanning more than one function call.


tleen:
Closing for conflicts in existing variation of file. Also seems like this may not be a desired topic. Perhaps another time :)

PR #888, opened 2017-08-30
Simplify control flow topics

robphoenix:
See https://github.com/exercism/meta/issues/83#issuecomment-324048581

From that conversation...

@robphoenix

> Be good to also consider what to do with topics such as Control-flow (conditionals) - control_flow_(conditionals) or control_flow_conditionals?

@ihiD

> For "control_flow_conditionals" can we just have "conditionals" please?

@kytrinyx

> I think it's worth sorting this out at the TOPICS level. Configlet could do it, but it would seem more like a dark art if it's not listed the way we want it up front.

### DIFF
--- a/TOPICS.txt
+++ b/TOPICS.txt
@@ -2,14 +2,14 @@ Basic concepts
 --------------
 Abstract classes
 Callbacks
-Control-flow (conditionals)
-Control-flow (loops)
+Conditionals
 Events
 Equality
 Exception handling
 Generics
 Inheritance
 Interfaces
+Loops
 Pattern matching
 Polymorphism
 Recursion


stkent:
Anyone in the org AFAIK!

Smarticles101:
Technically, it's possible for someone to not be in the track-maintainers team and thus not have access to push to this repo, but idk of anyone who matches that criteria. Anyone who has read access to a repo can approve changes, but it won't fully count unless they have push access

m-a-ge:
My version is https://github.com/m-a-ge/problem-specifications/blob/issue_887_convert_topics_to_snake_case/TOPICS.txt
What do you think?

tleen:
@m-a-ge should be in a separate PR for discussion?

m-a-ge:
@tleen, I can open one if needed

m-a-ge:
I think it's better to save both topics:
```
control_flow_conditionals
control_flow_loops
```
It will be helpful for some exercises

kytrinyx:
> Can anyone in the org approve changes to this repo or are there official maintainers for it?

I've structured the track maintainers using nested teams so that anyone who gets added as a maintainer to a specific Exercism language track automatically has commit on this repository.

I think eventually we might end up with a dedicated team of people who care a lot about the general curriculum without necessarily caring about any specific language track, at which point we can discuss whether another structure would be better, or whether we should just have a specific `problem-specifications` or `curriculum` team who _also_ has commit.

PR #890, opened 2017-08-31
Normalizes topics to snake_case notation

m-a-ge:
This is meant to normalize topics to snake_case notation as discussed earlier.

Also relates to #888 

Your feedback is welcomed

### DIFF
--- a/TOPICS.txt
+++ b/TOPICS.txt
@@ -1,80 +1,80 @@
-Basic concepts
+Basic Concepts
 --------------
-Abstract classes
-Callbacks
-Conditionals
-Events
-Equality
-Exception handling
-Generics
-Inheritance
-Interfaces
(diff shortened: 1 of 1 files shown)


tleen:
Lately it seems to be good practice to keep .txt documents compatible with markdown. As that was the case in the original file, I would recommend keeping the previous header style and limiting this PR to the snake casing.

This will also conflict with #888 (and potentially #886), should probably wait for #888 to be merged in, fix conflicts then this one would be good to go.

m-a-ge:
@tleen, got it, thanks

tleen:
#888 is in :+1: snake_case that into this and we should be good to go!

m-a-ge:
@tleen, PR has been updated accordingly

PR #892, opened 2017-09-07
circular-buffer: Clarify overwrite

shaleh:
This pull request is in two pieces in case the change to the test cases is controversial.

The description change is to make overwrite clear for the reader. To this end another example is added and that there is a difference in behaviour when the buffer is not full.

In the test cases descriptions I exchanged the word "replace" for the word "remove". This should add clarity especially for the readers who are not native English speakers.

### DIFF
--- a/exercises/circular-buffer/canonical-data.json
+++ b/exercises/circular-buffer/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "circular-buffer",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "comments": [
     "In general, these circular buffers are expected to be stateful,",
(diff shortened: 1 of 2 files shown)


shaleh:
Ah, thanks. Sorry about that.

shaleh:
A CI check which complains about such minor things could save some hassle.

tleen:
NP, I fixed it in the commit message. GitHub specific Git takes some getting used to!
I think a CI check could totally false positive too, which could be worse than not having it?

shaleh:
At various employers I have seen it work well. It really depends on if the over head from the commits is high enough to offload the problem on the submitters instead of the committers.

PR #893, opened 2017-09-08
pangram: rework tests (discussion)

Vankog:
Hi,

I'm currently contributing some bits and tits to the JS and ecmascript tracks and while doing so I came across some issues I see with some of the test files. Upon proposing some changes the maintainers asked me to contribute to the canonical track first, before it can be addressed in the particular track. So here I am. :-)

In this PR I'd like to discuss the test spec for the pangram exercise in particular.

First I have some assumptions:

- This is a learning platform, so the 'course material' should be a good example of good practices and (arguably) well designed principles.
- As far as I can tell we utilize TDD with a progression in the tests that evolve the desired implementation one test case at a time. This is called "Fake it till you make it" or "sliming" and is a nice practice to focus on what is essential. (see https://www.youtube.com/watch?v=PhiXo5CWjYU for a short vlog about this concept)
- Because of this, extra special care should be taken in reviewing contributed test cases. Especially, because they are the main reference material for (mostly) inexperienced learners.

The pangram test spec is the first test that I took a closer look at and at least to me there seem to be some things that could be improved.
Here is the current one:
https://github.com/exercism/problem-specifications/blob/fba1aef6b2237f504bfdd0bf575fd49489f12523/exercises/pangram/canonical-data.json

The following things came to me:

- Some test cases basically test the same edge cases. e.g. Two cases test for a missing 'x' specifically. The last two cases check for upper and lower case. The test for special characters are unnecessarily spread upon several cases.
- Some important edge cases are missing. e.g. The null-case or unicode-cases, uppercase-cases. Arguably we should even test for wrong input types like objects, arrays etc. in some languages, which is track specific I guess.
- The order of the cases does not really provide a nice "sliming"-progression. e.g. The second test already includes a special character - space. The last tests check for mixed case, after the tests for special characters.

I am far from an expert in this field. To be fair I'm even a learner who needs practise in TDD. However, I am particularly interested in a good TDD design and the clean code movement. And I think I can add some value from experienced talks and tutorials and videos about testable code and clean code in general to this discussion.

Therefore, I have a suggested version of the pangram test spec you can see in this PR.
Here is the file in a whole:
https://github.com/Vankog/problem-specifications/blob/pangram-test-rework/exercises/pangram/canonical-data.json

I think I provided a good sliming progression, adding one aspect after another. Also I tried to think about the edge cases more carefully and what needs to be tested, adding some and merging others.

Here is a concrete implementation of this proposal in JS that I proposed in an PR over there:
https://github.com/exercism/javascript/pull/406

What do you think?

### DIFF
--- a/exercises/pangram/canonical-data.json
+++ b/exercises/pangram/canonical-data.json
@@ -6,87 +6,75 @@
   "version": "1.2.0",
   "cases": [
     {
-      "description": "Check if the given string is an pangram",
+      "description": "checks if the given string is a pangram.",
       "comments": [
         "Output should be a boolean denoting if the string is a pangram or not."
       ],
       "cases": [
         {
-          "description": "can handle an empty sentence",
+          "description": "returns false for an empty message.",
           "property": "isPangram",
           "input": "",
           "expected": false
         },
         {
-          "description": "handles an undefined message as empty message",
+          "description": "returns false for an undefined/null message as argument.",
           "property": "isPangram",
-          "input": "null",
+          "input": null,
           "expected": false
         },
         {
-          "description": "recognizes a lower case pangram",
+          "description": "recognizes a perfect lower case pangram.",
           "property": "isPangram",
-          "input": "thequickbrownfoxjumpsoverthelazydog",
+          "input": "abcdefghijklmnopqrstuvwxyz",
           "expected": true
         },
         {
-          "description": "recognizes a missing character 'x'",
+          "description": "recognizes missing characters, e.g. 'x'.",
           "property": "isPangram",
           "input": "aquickmovementoftheenemywilljeopardizefivegunboats",
           "expected": false
         },
         {
-          "description": "recognizes a missing character 'h'",
+          "description": "recognizes missing characters, e.g. 'h'.",
           "property": "isPangram",
           "input": "fiveboxingwizardsjumpquicklyatit",
           "expected": false
         },
         {
-          "description": "recognizes an upper case pangram",
+          "description": "recognizes an upper-case pangram.",
           "property": "isPangram",
           "input": "THEQUICKBROWNFOXJUMPSOVERTHELAZYDOG",
           "expected": true
         },
         {
-          "description": "recognizes a missing character in upper case",
+          "description": "recognizes a missing character in upper-case, e.g. 'G'.",
           "property": "isPangram",
           "input": "THEQUICKBROWNFOXJUMPSOVERTHELAZYDOLL",
           "expected": false
         },
         {
-          "description": "recognizes a mixed case pangram",
+          "description": "recognizes a mixed-case pangram.",
           "property": "isPangram",
           "input": "ThEFiVebOxInGWiZaRdSJuMpqUiCkLy",
           "expected": true
         },
         {
-          "description": "recognizes a missing character in mixed case",
+          "description": "recognizes a missing character in mixed-case, e.g. 'Y'.",
           "property": "isPangram",
           "input": "ThEFiVebOxInGWiZaRdSJuMpqUiCkL",
           "expected": false
         },
         {
-          "description": "ignores other characters in correct pangram",
+          "description": "recognizes a pangram even if additional characters other than a-z are present.",
           "property": "isPangram",
           "input": "Victor_jagt-zwölf.B0xkämpfer >qu3r<über;den'großen\" Sylter#Deich!",
           "expected": true
         },
         {
-          "description": "ignores other characters in incorrect pangram",
+          "description": "recognizes missing characters (e.g. 'd') even if additional characters other than a-z are present.",
           "property": "isPangram",
           "input": "Victor_jagt-zwölf.B0xkämpfer >qu3r< über;einen'großen\" Sylter#Teich!",
-          "expected": false
-        },
-        {
-          "description": "ignores non-ANSI characters in correct pangram",
-          "property": "isPangram",
-          "input": "Few quips galvanized the ① mock 😮 jury box 🗃️.",
-          "expected": true
-        },
-        {
-          "description": "ignores non-ANSI characters in incorrect pangram",
-          "property": "isPangram",
-          "input": "Few quips gal℣anized the ① mock 😮 jury box 🗃️.",
           "expected": false
         }
       ]
--- a/exercises/pangram/canonical-data.json
+++ b/exercises/pangram/canonical-data.json
@@ -1,7 +1,14 @@
 {
   "exercise": "pangram",
   "comments": [
-    "A pangram is a sentence using every letter of the alphabet at least once."
+    "A pangram is a sentence using every letter of the alphabet at least once.",
+    "",
+    "It is the track's responsibility to add further language specific tests, like:",
+    "input error handling (null/undefined parameters etc.)",
+    "referential transparency (i.e. evaluating, a function/method gives the same value for same arguments every time.)",
+    "etc.",
+    "",
+    "'error' and 'null' should be treated according to the languages' specifics and possibilities."
   ],
   "version": "1.2.0",
   "cases": [
@@ -18,63 +25,63 @@
           "expected": false
         },
         {
-          "description": "returns false for an undefined/null message as argument.",
-          "property": "isPangram",
-          "input": null,
-          "expected": false
-        },
-        {
           "description": "recognizes a perfect lower case pangram.",
           "property": "isPangram",
           "input": "abcdefghijklmnopqrstuvwxyz",
           "expected": true
         },
         {
+          "description": "recognizes lower case pangram with spaces.",
+          "property": "isPangram",
+          "input": "few quips galvanized the mock jury box",
+          "expected": true
+        },
+        {
           "description": "recognizes missing characters, e.g. 'x'.",
           "property": "isPangram",
-          "input": "aquickmovementoftheenemywilljeopardizefivegunboats",
+          "input": "a quick movement of the enemy will jeopardize five gunboats",
           "expected": false
         },
         {
           "description": "recognizes missing characters, e.g. 'h'.",
           "property": "isPangram",
-          "input": "fiveboxingwizardsjumpquicklyatit",
+          "input": "five boxing wizards jump quickly at it",
           "expected": false
         },
         {
           "description": "recognizes an upper-case pangram.",
           "property": "isPangram",
-          "input": "THEQUICKBROWNFOXJUMPSOVERTHELAZYDOG",
+          "input": "THE QUICK BROWN FOX JUMPS OVER THE LAZY DOG",
           "expected": true
         },
         {
           "description": "recognizes a missing character in upper-case, e.g. 'G'.",
           "property": "isPangram",
-          "input": "THEQUICKBROWNFOXJUMPSOVERTHELAZYDOLL",
+          "input": "THE QUICK BROWN FOX JUMPS OVER THE LAZY DOLL",
           "expected": false
         },
         {
           "description": "recognizes a mixed-case pangram.",
           "property": "isPangram",
-          "input": "ThEFiVebOxInGWiZaRdSJuMpqUiCkLy",
+          "input": "ThE FiVe bOxInG WiZaRdS JuMp qUiCkLy",
           "expected": true
         },
         {
           "description": "recognizes a missing character in mixed-case, e.g. 'Y'.",
           "property": "isPangram",
-          "input": "ThEFiVebOxInGWiZaRdSJuMpqUiCkL",
+          "input": "ThE FiVe bOxInG WiZaRdS JuMp qUiCkL",
           "expected": false
         },
         {
-          "description": "recognizes a pangram even if additional characters other than a-z are present.",
+          "description": "recognizes a pangram with special characters.",
           "property": "isPangram",
-          "input": "Victor_jagt-zwölf.B0xkämpfer >qu3r<über;den'großen\" Sylter#Deich!",
+          "input": "Sphinx_of-'black\" .quartz, >judge< my; #v0w!",
           "expected": true
         },
         {
-          "description": "recognizes missing characters (e.g. 'd') even if additional characters other than a-z are present.",
+          "description": "recognizes missing characters (e.g. 'o') with special characters present.",
           "property": "isPangram",
-          "input": "Victor_jagt-zwölf.B0xkämpfer >qu3r< über;einen'großen\" Sylter#Teich!",
+          "input": "Sphinx_0f-'black\" .quartz, >judge< my; #v0w!",
           "expected": false
         }
       ]
--- a/exercises/pangram/canonical-data.json
+++ b/exercises/pangram/canonical-data.json
@@ -3,7 +3,7 @@
   "comments": [
     "A pangram is a sentence using every letter of the alphabet at least once."
   ],
-  "version": "1.1.0",
+  "version": "1.2.0",
   "cases": [
     {
       "description": "Check if the given string is an pangram",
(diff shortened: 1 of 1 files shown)


exercises/pangram/canonical-data.json
```diff
@@ -12,57 +12,81 @@
       ],
```
m-a-ge: Don't forget to change the version accordingly, the current one is (semantic versioning):
```
"version": "1.1.0",
```

Vankog: Funny you brought this up right in the moment I amended this change ^^

stkent: Here are the guidelines governing test versioning: https://github.com/exercism/problem-specifications#test-data-versioning

Since it appears that (correct me if I'm wrong):

- there's no new property
- the existing property has not been renamed
- there are no new keys
- no key types have been changed

we don't need a major version bump here, and v1.2.0 should suffice (minor vs patch because inputs/outputs were changed).

Vankog: thanks for the guide. Changed it to 1.2

exercises/pangram/canonical-data.json
```diff
@@ -12,57 +12,81 @@
...
           "property": "isPangram",
           "input": "",
           "expected": false
         },
         {
-          "description": "pangram with only lower case",
+          "description": "handles an undefined message as empty message",
           "property": "isPangram",
-          "input": "the quick brown fox jumps over the lazy dog",
+          "input": "null",
```
Insti: The description is misleading and the input being the string "null" here is confusing.
What are you trying to test?

exercises/pangram/canonical-data.json
```diff
@@ -12,57 +12,81 @@
...
+          "expected": false
+        },
+        {
+          "description": "recognizes a lower case pangram",
+          "property": "isPangram",
+          "input": "thequickbrownfoxjumpsoverthelazydog",
```
Insti: Why not just test against the complete alphabet?

exercises/pangram/canonical-data.json
```diff
@@ -12,57 +12,81 @@
...
+          "description": "recognizes a missing character 'x'",
+          "property": "isPangram",
+          "input": "aquickmovementoftheenemywilljeopardizefivegunboats",
+          "expected": false
+        },
+        {
+          "description": "recognizes a missing character 'h'",
           "property": "isPangram",
-          "input": "a quick movement of the enemy will jeopardize five gunboats",
+          "input": "fiveboxingwizardsjumpquicklyatit",
```
Insti: What is this testing? Why is 'h' different from 'x' and more important than 'y'?

exercises/pangram/canonical-data.json
```diff
@@ -12,57 +12,81 @@
...
-          "input": "the 1 quick brown fox jumps over the 2 lazy dogs",
+          "input": "Victor_jagt-zwölf.B0xkämpfer >qu3r<über;den'großen\" Sylter#Deich!",
           "expected": true
         },
         {
-          "description": "missing letters replaced by numbers",
+          "description": "ignores other characters in incorrect pangram",
           "property": "isPangram",
-          "input": "7h3 qu1ck brown fox jumps ov3r 7h3 lazy dog",
+          "input": "Victor_jagt-zwölf.B0xkämpfer >qu3r< über;einen'großen\" Sylter#Teich!",
```
Insti: This has the same description as the previous test.

Non ASCII test cases should not be included, see: https://github.com/exercism/problem-specifications/issues/428

exercises/pangram/canonical-data.json
```diff
@@ -12,57 +12,81 @@
...
         },
         {
-          "description": "pangram with numbers",
+          "description": "recognizes a missing character in mixed case",
+          "property": "isPangram",
+          "input": "ThEFiVebOxInGWiZaRdSJuMpqUiCkL",
+          "expected": false
+        },
+        {
+          "description": "ignores other characters in correct pangram",
```
stevejb71: "correct" is not the right word here - it's a pangram or a non-pangram.

exercises/pangram/canonical-data.json
Vankog: I need to change this to ```null``` instead.
My reasoning is the same as I wrote here:
https://github.com/exercism/problem-specifications/pull/895#discussion_r139141823

For everyone else, here's the quote to prevent jumping around:
I see what you mean. 
But as far as I experienced it, test suits should check for edge cases. null checks are especially important (at least in almost all languages). 
In every educational piece of material the advice is not to only check for usual inputs, but for all possible (equivalent kinds of) inputs. null checks, empty checks and others are always on top of the example lists given. Just take any arbitrary course about QA and this will be part of it.

I think this is especially important, because we are in an educational context. The users should learn from the getgo what it means to write (arguably good) tests (first), because it is still a major problem, even upon seasoned programmers. 
From my own business experience I know that new programmers are completely lost when it comes to writing tests. But also experienced programmers fall in the traps of "I'll do it later", "This is a trivial method", "I can't test this", "There are too many dependencies" etc. Just take any talk or tutorial about writing tests or even TDD and you will see the same old counter-arguments or questions again and again from the comments or the audience. This has a reason: *Writing good tests is hard! Very hard to be exact and it is a skill that needs exercise.* And where better to begin than right from the start?

exercises/pangram/canonical-data.json
Vankog: Good point, never occurred to me. ^^
The input was already there so I reused it.

exercises/pangram/canonical-data.json
Vankog: This is the core functionality of this exercise, so I'd argue that we should test it with more than just one possible failing input. 
The original version of the test had the following structure:
```
        {
          "description": "missing character 'x'",
          "property": "isPangram",
          "input": "a quick movement of the enemy will jeopardize five gunboats",
          "expected": false
        },
        {
          "description": "another missing character 'x'",
          "property": "isPangram",
          "input": "the quick brown fish jumps over the lazy dog",
          "expected": false
        },
```
So I guess, checking for 'h' instead of 'x' twice is already an improvement.

I'd rather like to add even more test cases, maybe about 3-5 in total. However, because the schema does not allow for multiple "expected" entries, I decided to keep it at two.

exercises/pangram/canonical-data.json
Vankog: OK, I'll remove non-ASCII tests.

exercises/pangram/canonical-data.json
Vankog: I agree, but because of the context I wrote it extra clear to make it distinct to the complementary non-pangram input. Sometimes I feel it might be better to be redundantly clear instead.

Vankog:
@Insti 
Thank you for your feedback.
Regarding one of your suggestions:
> Add the non-alphabet character exclusion test(s) early so you can use spaces in the subsequent tests.

I see what you want to achieve here. It would increase the readability of  subsequent test cases. 
However, in my eyes the exercise user should focus on the core functionality first. Sure, one could argue that the upper-cases and mixed-cases are special cases as well. But, when viewed from an implementation progress standpoint it is a much smaller step to add an ```toUpperCase()``` or something alike than checking for irrelevant characters. It seems smoother to me.

exercises/palindrome-products/canonical-data.json
```diff
@@ -0,0 +1,138 @@
+{
```
Vankog: oh boy... it includes the rebase....

exercises/word-search/canonical-data.json
```diff
@@ -1,13 +1,161 @@
 {
```
Vankog: oh boy, it includes the rebase....

Insti:
Hi @Vankog, thanks for taking on board the feedback and arguing for the cases you disagree with. :heart:

The main requirement from the [description](https://github.com/exercism/problem-specifications/blob/master/exercises/pangram/description.md) is: 
> "Determine if a sentence is a pangram."

It would be nice to use sentence in the examples we're testing.

Otherwise why not just sort everything alphabetically so the student can more easily see which letters are missing? (This is also undesirable.)

"the quick brown fox jumps over the lazy dog" is clearly a sentence.
"thequickbrownfoxjumpsoverthelazydog" only is if you put spaces in the right places.
"abcdeeefghhijklmnoooopqrrsttuuvwxyz" no amount of spaces will turn this into a sentence, but it's easier for a human to work out if it's a pangram.

> Be very wary of removing/changing existing tests

It also makes your PR harder to review and agree to merge.
For example, If there is agreement with all the additions, that would be easy to review and merge.
If everything is mixed together it is harder to review to see what has changed, and if there is disagreement about the removals, it holds up the otherwise uncontroversial additions.

Vankog:
The sentence-argument was also something that came to me and I considered it. In a way it is hard to trade-off. Either using a single long word all the time or introducing spaces early. 
For a smooth implementation progression my suggestion might be a better fit, but from a readability point of view your stance is also valid. I'll reconsider it and commit a new suggestion.

> For example, If there is agreement with all the additions, that would be easy to review and merge.
If everything is mixed together it is harder to review to see what has changed, and if there is disagreement about the removals, it holds up the otherwise uncontroversial additions.

I understand. Well, there is a reason why I worded the title of this discussion "rework" and the nucleotide one "refactor".
Looking at the concrete pangram test in the JS track, I considered it a mess. But not only me, also the track maintainer agreed. Looking at the canonical data is not nearly that bad, but as I already wrote in my opening post, there is a lot to be desired. Now, because the test data is a relatively small piece of code, a rewrite made sense to me. This includes deletion, merges, restructuring and additions. So yes, it might be hard to review when just looking at the diff. 
However, that is the reason why I linked to the whole files in my first post. They are meant to be viewed as a whole, because it is a rewrite. :-)

exercises/pangram/canonical-data.json
```diff
@@ -3,66 +3,78 @@
...
           "property": "isPangram",
           "input": "",
           "expected": false
         },
         {
-          "description": "pangram with only lower case",
+          "description": "returns false for an undefined/null message as argument.",
           "property": "isPangram",
-          "input": "the quick brown fox jumps over the lazy dog",
+          "input": null,
```
petertseng: If `null` is tested as an input, I am thinking it should be an error, rather than to return false. The reason for this belief is that if we consider that the function under test must only accept strings, well `null` is not a string, so this is in error.

(Of course, I imagine languages for which strings are non-nullable will simply skip this test)

This comment must not be read as an endorsement or rejection of the statement "`null` should be tested as an input".

exercises/pangram/canonical-data.json
```diff
@@ -3,66 +3,78 @@
...
-          "input": "7h3 qu1ck brown fox jumps ov3r 7h3 lazy dog",
+          "input": "ThEFiVebOxInGWiZaRdSJuMpqUiCkL",
           "expected": false
         },
         {
-          "description": "pangram with mixed case and punctuation",
+          "description": "recognizes a pangram even if additional characters other than a-z are present.",
           "property": "isPangram",
-          "input": "\"Five quacking Zephyrs jolt my wax bed.\"",
+          "input": "Victor_jagt-zwölf.B0xkämpfer >qu3r<über;den'großen\" Sylter#Deich!",
```
petertseng: can you confirm whether or not ö ä ü and ß are in ASCII, given the declaration of intent in https://github.com/exercism/problem-specifications/pull/893#discussion_r139159361 of

> OK, I'll remove non-ASCII tests.

exercises/pangram/canonical-data.json
```diff
@@ -3,66 +3,78 @@
...
           "expected": true
         },
         {
-          "description": "upper and lower case versions of the same character should not be counted separately",
+          "description": "recognizes missing characters (e.g. 'd') even if additional characters other than a-z are present.",
           "property": "isPangram",
-          "input": "the quick brown fox jumps over with lazy FX",
+          "input": "Victor_jagt-zwölf.B0xkämpfer >qu3r< über;einen'großen\" Sylter#Teich!",
```
petertseng: can you confirm whether or not ö ä ü and ß are in ASCII, given the declaration of intent in https://github.com/exercism/problem-specifications/pull/893#discussion_r139159361 of

> OK, I'll remove non-ASCII tests.

Insti:
> I made some replies above.

Are there any that need a response that I've not responded to?

PRs get complicated when they contain many different conversation threads and Github doesn't help when it starts hiding conversations.

(Another reason for small focused PRs.)

exercises/pangram/canonical-data.json
Vankog: Yeah, I agree. I had this same thought but had not changed it, yet, because of the ongoing discussion. I will probably change it to expect a throw in an upcoming commit.

Vankog:
@Insti hm... besides the input validation discussion I think this one is the last:
https://github.com/exercism/problem-specifications/pull/893/commits/8d9b0a15d65c011b3c7e0f547879df6ad3049af7#r137843296
About testing for a missing x and h

exercises/pangram/canonical-data.json
Insti: What if it doesn't catch a missing 'a' ?
Either the 'h' test is not necessary or you're missing another 24 tests.

Vankog: Exactly! That is my problem I'm trying to balance. 
But it does not help to check for x in two cases like the original suite does atm.
If you want, then I'll do 26 test cases for every character.

But just one is simply not a reasonable option for testing the core functionality.

Insti:
@Vankog are you interested in continuing to work on these pull requests?
I think they have achieved their goals of stimulating discussion, but still require further work to be considered ready to merge.

(If no further progress is made, I will close this on or after the 8th October.)

Vankog:
I do. I was just waiting for the sub-discussions to conclude so we can finally decide what should or shouldn't be done.

Vankog:
OK, new suggestion.

- converted the controversial case into a comment to support the track maintainers with suggestions.
- added a space-test early to work with proper sentences later on.
- removed ASCII chars > 126.
- tweaked some descriptions

exercises/pangram/canonical-data.json
```diff
@@ -1,68 +1,87 @@
...
   "comments": [
-    "A pangram is a sentence using every letter of the alphabet at least once."
+    "A pangram is a sentence using every letter of the alphabet at least once.",
+    "",
+    "It is the track's responsibility to add further language specific tests, like:",
+    "input error handling (null/undefined parameters etc.)",
+    "referential transparency (i.e. evaluating, a function/method gives the same value for same arguments every time.)",
+    "etc.",
+    "",
+    "'error' and 'null' should be treated according to the languages' specifics and possibilities."
```
petertseng: this is not strictly necessary, since there are no longer any `error` or `null` used in this file (unless I missed one, sorry!)

exercises/pangram/canonical-data.json
```diff
@@ -1,68 +1,87 @@
...
           "property": "isPangram",
-          "input": "the 1 quick brown fox jumps over the 2 lazy dogs",
+          "input": "ThE FiVe bOxInG WiZaRdS JuMp qUiCkLy",
           "expected": true
         },
         {
-          "description": "missing letters replaced by numbers",
+          "description": "recognizes a missing character in mixed-case, e.g. 'Y'.",
           "property": "isPangram",
-          "input": "7h3 qu1ck brown fox jumps ov3r 7h3 lazy dog",
```
petertseng: one interesting thing about this input is that the number of unique non-space characters in it is exactly 26, which is the intent of https://github.com/exercism/problem-specifications/pull/852 . 

Now it looks like we are testing mixed-case before non-alphabetics, all right, so we wouldn't have the letters -> numbers case, that seems fine.

If you believe that is still a useful bug to catch, then it would be useful to have a case like "Aabcdefghijklmnopqrstuvwxy"

petertseng: The criterion is not met by "ThE FiVe bOxInG WiZaRdS JuMp qUiCkL" because the number of unique non-space characters in it is 28.

Insti:
As a reviewer this PR has gotten to a stage where reviewing it is very difficult.
I still agree with some of the changes and disagree with others, but reviewing all the different threads and working out what has changed in amongst all the changes and comments will take more time than I have available to review, so I do not expect I will change my my review status from changes requested.

I recommend this PR be closed and new separate PRs be opened for the different changes.

Vankog:
first split:
- [x] adding case with "perfect pangram" #952 
- [ ] changing case "another missing character 'x'" 
to "recognizes missing characters, e.g. 'h'."
to prevent testing for 'x' twice.
- [ ] replacing cases "pangram with underscores", "pangram with numbers" and "missing letters replaced by numbers" with 
"recognizes a pangram with special characters." 
"recognizes missing characters (e.g. 'o') with special characters present."
Because, testing just underscores and numbers and number substitution makes not much sense in the overall TDD progression. You only need a (positive and negative) test focusing on non-a-z letters.
- [ ] replacing case "pangram with mixed case and punctuation" with
"recognizes an upper-case pangram."
"recognizes a missing character in upper-case, e.g. 'G'."
"recognizes a mixed-case pangram."
"recognizes a missing character in mixed-case, e.g. 'Y'."
Because of mixed concerns, incomplete coverage and better TDD progression.
- [ ] changing some inputs for more pangram variety, instead of using the same sentence over and over again
- [ ] improving case ordering after previous changes took place
- [ ] add suite comments (incl. suggestions for track maintainers)
- [ ] improve case descriptions

exercises/pangram/canonical-data.json
Vankog: @petertseng Thanks for this hint. 
However, I am not sure I understand the intention of this bugfix or the changed testcase. Is it to test for mixed-case, but with 26 distinct chars? 
If so, I think we should add a testcase with exactly that intent to prevent future misunderstandings/changes.

petertseng: > Is it to test for mixed-case, but with 26 distinct chars?

Yes. The test case that got replaced is: lowercase + numbers, with 26 distinct non-space chars.

I was thinking it would get replaced with: lowercase + uppercase, with 26 distinct non-space chars.

I think you understood correctly.

Insti:
Closing this one too.
You're still welcome to keep posting and tracking the todos in these threads, but I don't think the PR needs to remain open for that.
Thanks :heart:

PR #894, opened 2017-09-08
Add canonical data for palindrome products

stevejb71:


Insti:
I have no recent experience with this problem, but this file looks reasonable. 👍

exercises/palindrome-products/canonical-data.json
```diff
@@ -0,0 +1,126 @@
...
+                    [1, 9],
+                    [3, 3]
+                ]
+            }
+        },
+        {
+            "description": "find the largest palindrome from double digit factors",
+            "property": "largest",
+            "input_min": 10,
+            "input_max": 99,
```
Insti: Missing test for the smallest in this range?
It is used as an example in the description, so I would expect to find it here.

stevejb71: I added this test.

petertseng:
Please say that this closes
https://github.com/exercism/problem-specifications/issues/573

PR #895, opened 2017-09-08
nucleotide-count: refactoring tests (discussion)

Vankog:
Hi,

In the same vain as the other PR  #893 here is the discussion about nucleotide-count.

-------------------
### Background:
I'm currently contributing some bits and tits to the JS and ecmascript tracks and while doing so I came across some issues I see with some of the test files. Upon proposing some changes the maintainers asked me to contribute to the canonical track first, before it can be addressed in the particular track. So here I am. :-)

In this PR I'd like to discuss the test spec for the pangram exercise in particular.

First I have some assumptions:

- This is a learning platform, so the 'course material' should be a good example of good practices and (arguably) well designed principles.
- As far as I can tell we utilize TDD with a progression in the tests that evolve the desired implementation one test case at a time. This is called "Fake it till you make it" or "sliming" and is a nice practice to focus on what is essential. (see https://www.youtube.com/watch?v=PhiXo5CWjYU for a short vlog about this concept)
- Because of this, extra special care should be taken in reviewing contributed test cases. Especially, because they are the main reference material for (mostly) inexperienced learners.

----------------------
### Analysis:
The nucleotide-count test spec is another test that I took a closer look at and at least to me there seem to be some things that could be improved.
Here is the current one:
https://github.com/exercism/problem-specifications/blob/fba1aef6b2237f504bfdd0bf575fd49489f12523/exercises/nucleotide-count/canonical-data.json

The following things came to me:

- Some important edge cases are missing. e.g. The null-case or lowercase-cases. Arguably we should even test for wrong input types like objects, arrays etc. in some languages, which is track specific I guess.
- The order of the cases seems to be nice enough for a "sliming"-progression, but could be improved with some steps inbetween. 

-----------------
### Suggestion:
I am far from an expert in this field. To be fair I'm even a learner who needs practise in TDD. However, I am particularly interested in a good TDD design and the clean code movement. And I think I can add some value from experienced talks and tutorials and videos about testable code and clean code in general to this discussion.

Therefore, I have a suggested version of the nucleotide-count test spec you can see in this PR.
Here is the file in a whole:
https://github.com/Vankog/problem-specifications/blob/refactor-nucleotide-count-test/exercises/nucleotide-count/canonical-data.json

I think I provided a good sliming progression, adding one aspect after another. This mandated some intermediate test cases, like testing a single char input before testing a sequence. 
Also I tried to think about the edge cases more carefully and what needs to be tested, adding some and improving the descriptions.

What do you think?

### DIFF
--- a/exercises/nucleotide-count/canonical-data.json
+++ b/exercises/nucleotide-count/canonical-data.json
@@ -1,12 +1,19 @@
 {
   "exercise": "nucleotide-count",
+  "comments": [
+    "Given a string of nucleotides, the count of each nucleotide is returned."
+  ],
   "version": "1.1.0",
   "cases": [
     {
-      "description": "nucleotideCounts counts all DNA-nucleotides in the input strand and returns an object with properties A, C, G and T holding their respective amounts.",
+      "description": "counts all nucleotides in a provided strand.",
+      "comments": [
+        "nucleotideCounts counts all DNA-nucleotides in the input strand and returns an object with properties",
+        "A, C, G and T holding their respective amounts."
+      ],
       "cases": [
         {
-          "description": "counts an empty DNA strand as 0.",
+          "description": "returns zeroes in case of an empty input strand.",
           "property": "nucleotideCounts",
           "strand": "",
           "expected": {
@@ -17,8 +24,10 @@
           }
         },
         {
-          "comments" : ["Should check for null-references and assume an empty input."],
-          "description": "counts an undefined DNA strand as empty strand.",
+          "comments" : [
+            "Should check for null-references and assume an empty input."
+          ],
+          "description": "returns zeroes for an undefined/null strand as argument.",
           "property": "nucleotideCounts",
           "strand": null,
           "expected": {
@@ -29,7 +38,7 @@
           }
         },
         {
-          "description": "counts a single nucleotide correctly.",
+          "description": "returns the counts of nucleotides in a single-character input.",
           "property": "nucleotideCounts",
           "strand": "G",
           "expected": {
@@ -40,7 +49,7 @@
           }
         },
         {
-          "description": "counts a single repetitive nucleotide correctly.",
+          "description": "returns the counts of nucleotides in a strand containing one repeated nucleotide.",
           "property": "nucleotideCounts",
           "strand": "GGGGGGG",
           "expected": {
@@ -51,7 +60,7 @@
           }
         },
         {
-          "description": "counts all nucleotides correctly.",
+          "description": "returns the count of nucleotides in a strand containing all nucleotides.",
           "property": "nucleotideCounts",
           "strand": "AGCTTTTCATTCTGACTGCAACGGGCAATATGTCTCTGTGTGGATTAAAAAAAGAGTGTCTGATAGCAGC",
           "expected": {
@@ -62,7 +71,7 @@
           }
         },
         {
-          "description": "counts all nucleotides correctly in mixed case.",
+          "description": "returns the count of nucleotides in a strand containing mixed-case nucleotides.",
           "property": "nucleotideCounts",
           "strand": "aGcTtTtCaTtCtGaCtGcAaCgGgCaAtAtGtCtCtGtGtGgAtTaAaAaAaGaGtGtCtGaTaGcAgC",
           "expected": {
@@ -83,10 +92,10 @@
           "expected" : true
         },
         {
-          "description": "handle incorrect inputs",
+          "description": "handles incorrect inputs.",
           "cases": [
             {
-              "description": "checks if DNA input does not include invalid characters.",
+              "description": "throws error if DNA input does not include anything but ACGT-characters.",
               "property": "nucleotideCounts",
               "strand": "AGUUÄ1+ :'A'}CT",
               "expected": {
--- a/exercises/nucleotide-count/canonical-data.json
+++ b/exercises/nucleotide-count/canonical-data.json
@@ -1,12 +1,12 @@
 {
   "exercise": "nucleotide-count",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "cases": [
     {
-      "description": "count all nucleotides in a strand",
+      "description": "nucleotideCounts counts all DNA-nucleotides in the input strand and returns an object with properties A, C, G and T holding their respective amounts.",
       "cases": [
         {
-          "description": "empty strand",
+          "description": "counts an empty DNA strand as 0.",
           "property": "nucleotideCounts",
           "strand": "",
           "expected": {
@@ -17,7 +17,30 @@
           }
         },
         {
-          "description": "strand with repeated nucleotide",
+          "comments" : ["Should check for null-references and assume an empty input."],
+          "description": "counts an undefined DNA strand as empty strand.",
+          "property": "nucleotideCounts",
+          "strand": null,
+          "expected": {
+            "A": 0,
+            "C": 0,
+            "G": 0,
+            "T": 0
+          }
+        },
+        {
+          "description": "counts a single nucleotide correctly.",
+          "property": "nucleotideCounts",
+          "strand": "G",
+          "expected": {
+            "A": 0,
+            "C": 0,
+            "G": 1,
+            "T": 0
+          }
+        },
+        {
+          "description": "counts a single repetitive nucleotide correctly.",
           "property": "nucleotideCounts",
           "strand": "GGGGGGG",
           "expected": {
@@ -28,7 +51,7 @@
           }
         },
         {
-          "description": "strand with multiple nucleotides",
+          "description": "counts all nucleotides correctly.",
           "property": "nucleotideCounts",
           "strand": "AGCTTTTCATTCTGACTGCAACGGGCAATATGTCTCTGTGTGGATTAAAAAAAGAGTGTCTGATAGCAGC",
           "expected": {
@@ -39,12 +62,38 @@
           }
         },
         {
-          "description": "strand with invalid nucleotides",
+          "description": "counts all nucleotides correctly in mixed case.",
           "property": "nucleotideCounts",
-          "strand": "AGXXACT",
+          "strand": "aGcTtTtCaTtCtGaCtGcAaCgGgCaAtAtGtCtCtGtGtGgAtTaAaAaAaGaGtGtCtGaTaGcAgC",
           "expected": {
-            "error": "Invalid nucleotide in strand"
+            "A": 20,
+            "C": 12,
+            "G": 17,
+            "T": 21
           }
+        },
+        {
+          "comments" : [
+            "This should call the property/function at least two times and check if it returns the same",
+            "result. This prevents errors where the call aggregates (sums up) the counts of each call."
+          ],
+          "description": "returns the same amounts for consecutive nucleotideCounts calls.",
+          "property": "nucleotideCounts",
+          "strand": "AGCTTTTCATTCTGACTGCAACGGGCAATATGTCTCTGTGTGGATTAAAAAAAGAGTGTCTGATAGCAGC",
+          "expected" : true
+        },
+        {
+          "description": "handle incorrect inputs",
+          "cases": [
+            {
+              "description": "checks if DNA input does not include invalid characters.",
+              "property": "nucleotideCounts",
+              "strand": "AGUUÄ1+ :'A'}CT",
+              "expected": {
+                "error": "Invalid nucleotide in strand"
+              }
+            }
+          ]
         }
       ]
     }
--- a/exercises/nucleotide-count/canonical-data.json
+++ b/exercises/nucleotide-count/canonical-data.json
@@ -1,7 +1,13 @@
 {
   "exercise": "nucleotide-count",
   "comments": [
-    "Given a string of nucleotides, the count of each nucleotide is returned."
+    "Given a string of nucleotides, the count of each nucleotide is returned.",
+    "",
+    "It is the track's responsibility to add further language specific tests like:",
+    "input error handling (null/undefined parameters etc.)",
+    "referential transparency (i.e. evaluating, a function/method gives the same value for same arguments every time.)",
+    "",
+    "'error' and 'null' should be treated according to the languages' specifics and possibilities."
   ],
   "version": "1.1.0",
   "cases": [
@@ -24,21 +30,7 @@
           }
         },
         {
-          "comments" : [
-            "Should check for null-references and assume an empty input."
-          ],
-          "description": "returns zeroes for an undefined/null strand as argument.",
-          "property": "nucleotideCounts",
-          "strand": null,
-          "expected": {
-            "A": 0,
-            "C": 0,
-            "G": 0,
-            "T": 0
-          }
-        },
-        {
-          "description": "returns the counts of nucleotides in a single-character input.",
+          "description": "can count one nucleotide in single-character input.",
           "property": "nucleotideCounts",
           "strand": "G",
           "expected": {
@@ -49,7 +41,7 @@
           }
         },
         {
-          "description": "returns the counts of nucleotides in a strand containing one repeated nucleotide.",
+          "description": "can count one nucleotide in multi-character input.",
           "property": "nucleotideCounts",
           "strand": "GGGGGGG",
           "expected": {
@@ -60,7 +52,7 @@
           }
         },
         {
-          "description": "returns the count of nucleotides in a strand containing all nucleotides.",
+          "description": "can count all nucleotides in a strand containing all nucleotides.",
           "property": "nucleotideCounts",
           "strand": "AGCTTTTCATTCTGACTGCAACGGGCAATATGTCTCTGTGTGGATTAAAAAAAGAGTGTCTGATAGCAGC",
           "expected": {
@@ -71,7 +63,7 @@
           }
         },
         {
-          "description": "returns the count of nucleotides in a strand containing mixed-case nucleotides.",
+          "description": "ignores case in a strand of mixed-case nucleotides.",
           "property": "nucleotideCounts",
           "strand": "aGcTtTtCaTtCtGaCtGcAaCgGgCaAtAtGtCtCtGtGtGgAtTaAaAaAaGaGtGtCtGaTaGcAgC",
           "expected": {
@@ -82,22 +74,12 @@
           }
         },
         {
-          "comments" : [
-            "This should call the property/function at least two times and check if it returns the same",
-            "result. This prevents errors where the call aggregates (sums up) the counts of each call."
-          ],
-          "description": "returns the same amounts for consecutive nucleotideCounts calls.",
-          "property": "nucleotideCounts",
-          "strand": "AGCTTTTCATTCTGACTGCAACGGGCAATATGTCTCTGTGTGGATTAAAAAAAGAGTGTCTGATAGCAGC",
-          "expected" : true
-        },
-        {
           "description": "handles incorrect inputs.",
           "cases": [
             {
-              "description": "throws error if DNA input does not include anything but ACGT-characters.",
+              "description": "treats an input with anything else than ACGT-characters as error.",
               "property": "nucleotideCounts",
-              "strand": "AGUUÄ1+ :'A'}CT",
+              "strand": "AGUU1+ :'A'}CT",
               "expected": {
                 "error": "Invalid nucleotide in strand"
               }


exercises/nucleotide-count/canonical-data.json
```diff
@@ -39,12 +62,38 @@
...
+          "expected" : true
+        },
+        {
+          "description": "handle incorrect inputs",
+          "cases": [
+            {
+              "description": "checks if DNA input does not include invalid characters.",
+              "property": "nucleotideCounts",
+              "strand": "AGUUÄ1+ :'A'}CT",
+              "expected": {
```
Vankog: check for some (ASCII compliant) invalid chars.

exercises/nucleotide-count/canonical-data.json
```diff
@@ -17,7 +17,30 @@
           }
         },
         {
-          "description": "strand with repeated nucleotide",
+          "comments" : ["Should check for null-references and assume an empty input."],
+          "description": "counts an undefined DNA strand as empty strand.",
+          "property": "nucleotideCounts",
+          "strand": null,
+          "expected": {
```
Vankog: do i have to write ```null``` or ```"null"```?

exercises/nucleotide-count/canonical-data.json
```diff
@@ -17,7 +17,30 @@
...
+        {
+          "description": "counts a single nucleotide correctly.",
+          "property": "nucleotideCounts",
+          "strand": "G",
+          "expected": {
+            "A": 0,
+            "C": 0,
+            "G": 1,
+            "T": 0
+          }
```
Vankog: even tough this case might already be included in the next case, this test is for a nice sliming progression. Developers should have quick success by implementing a subset of the spec first.

exercises/nucleotide-count/canonical-data.json
```diff
@@ -39,12 +62,38 @@
...
+        {
+          "comments" : [
+            "This should call the property/function at least two times and check if it returns the same",
+            "result. This prevents errors where the call aggregates (sums up) the counts of each call."
+          ],
+          "description": "returns the same amounts for consecutive nucleotideCounts calls.",
+          "property": "nucleotideCounts",
+          "strand": "AGCTTTTCATTCTGACTGCAACGGGCAATATGTCTCTGTGTGGATTAAAAAAAGAGTGTCTGATAGCAGC",
+          "expected" : true
+        },
```
Vankog: This is a case that I actually took over from the concrete JS implementation. I thought it is an interesting thought and it actually might occur.

exercises/nucleotide-count/canonical-data.json
```diff
@@ -39,12 +62,38 @@
...
+        {
+          "description": "handle incorrect inputs",
+          "cases": [
```
Vankog: added a separate case suite for error handling.

exercises/nucleotide-count/canonical-data.json
```diff
@@ -1,12 +1,12 @@
 {
   "exercise": "nucleotide-count",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "cases": [
     {
-      "description": "count all nucleotides in a strand",
+      "description": "nucleotideCounts counts all DNA-nucleotides in the input strand and returns an object with properties A, C, G and T holding their respective amounts.",
```
Insti: This does not need to be the description of the problem, that is provided in a separate file.

Several tracks generate the test name based on the test descriptions in this file so having a concise description is a good thing.

exercises/nucleotide-count/canonical-data.json
```diff
@@ -1,12 +1,12 @@
...
       "cases": [
         {
-          "description": "empty strand",
+          "description": "counts an empty DNA strand as 0.",
```
Insti: The original name here was fine, you are adding redundant information here.

exercises/nucleotide-count/canonical-data.json
Insti: I don't think null input handling is appropriate to test here.
The point of this exercise is string parsing, not having a string at all defeats the whole purpose of the exercise.

exercises/nucleotide-count/canonical-data.json
Insti: Why is returning a hardcoded 1 different from returning a hardcoded 7?

exercises/nucleotide-count/canonical-data.json
Insti: See comments on error case handling on your panagram PR.

exercises/nucleotide-count/canonical-data.json
Insti: This is a difficult thing to specify in this json file format.
I think it's better left out of the canonical data and implemented at the track level.

exercises/nucleotide-count/canonical-data.json
```diff
@@ -39,12 +62,38 @@
           }
         },
         {
-          "description": "strand with invalid nucleotides",
+          "description": "counts all nucleotides correctly in mixed case.",
           "property": "nucleotideCounts",
-          "strand": "AGXXACT",
+          "strand": "aGcTtTtCaTtCtGaCtGcAaCgGgCaAtAtGtCtCtGtGtGgAtTaAaAaAaGaGtGtCtGaTaGcAgC",
```
Insti: This exercise is not about case comparisons. We should be able to trust that all the input characters are valid.

exercises/nucleotide-count/canonical-data.json
Insti: To answer the question you asked: `null`

Vankog: I see what you mean. 
But as far as I experienced it, test suits should check for edge cases. null checks are especially important (at least in almost all languages). 
In every educational piece of material the advice is not to only check for usual inputs, but for all possible (equivalent kinds of) inputs. null checks, empty checks and others are always on top of the example lists given. Just take any arbitrary course about QA and this will be part of it.

I think this is especially important, because we are in an educational context. The users should learn from the getgo what it means to write (arguably good) tests (first), because it is still a major problem, even upon seasoned programmers. 
From my own business experience I know that new programmers are completely lost when it comes to writing tests. But also experienced programmers fall in the traps of "I'll do it later", "This is a trivial method", "I can't test this", "There are too many dependencies" etc. Just take any talk or tutorial about writing tests or even TDD and you will see the same old counter-arguments or questions again and again from the comments or the audience. This has a reason: *Writing good tests is hard! Very hard to be exact and it is a skill that needs exercise.* And where better to begin than right from the start.

exercises/nucleotide-count/canonical-data.json
Vankog: This is a good point. 

I agree that this is not necessary. However, why should anyone hardcode two or more different test cases? Programming means using your brain to accomplish a tasks. The TDD cycle is red-green-refactor(?) and sure, you could hardcode the handling of a one-length input and a n-length input to make it green, but then there is still the third phase: To check if there is any need to improve the code. If someone just hardcodes staements to pass several tests, then I dare to say this person might not want to program in the first place.

My intent with the two test cases was, that it is easier to handle a single character input than a sequence. I think the handling of a sequence is an extension of the single char input. Small steps with fast achievements (in form of another green test) seem to be a big motivator to me, especially for inexperienced developers who have to think hard for every step they take.

exercises/nucleotide-count/canonical-data.json
Vankog: I agree with you. But how do you make sure it gets implemented in the tracks?

Limiting the possibilities of testing, just to fit into a constrained model or framework is generally a bad idea, though. 
In this case I suggest we should think about a solution to this constraint in the first place (in another discussion of course). In the meantime, as a walkaround, what arguments speak against utilizing descriptions and comments to specify the case? How many tracks use test generators and how many do so blindly?

exercises/nucleotide-count/canonical-data.json
Vankog: I see, thanks. I'll change this.

exercises/nucleotide-count/canonical-data.json
Vankog: You are right. Though, my intent was to implement well-formed descriptions in the informal style of 
´´´{it / the test} does xyz in the case of abc.```
What I've experienced so far, this is not even enough. A test case (in TDD especially) acts as a specification. There are best practices (in some x-driven-development movements even rules) that suggest a test description should have several parts:
- in case of x
- then y
- reasoning
- some also want a precondition (like BDDs GIVEN WHEN THEN)

Just stating "empty string" as description tells a developer absolutely nothing about what should happen (or even why). I know you care about the reasoning part the same way I do, so I am not sure why you want to omit it here.

So in my eyes my suggested description might even be phrased like:
"Given an empty input string, it returns zeroes for all nucleotides."
While the reasoning for this behaviour results from the description of the test suite, because there is no need to state this dogmatically in every description.

exercises/nucleotide-count/canonical-data.json
Vankog: I respectfully disagree. See my reasoning above (https://github.com/exercism/problem-specifications/pull/895#discussion_r139141823).

We educate people about programming and teaching them what it means to test for all equivalence classes should be an integral part of this progress right from the beginning.

Why should we assume that all inputs are valid? This works to a certain degree in one-man projects and quick (and/or dirty) implementations, but this is nothing we should promote. In reality murphy's law rules.

I know that several test cases seem tedious to write, but why bother here? The users don't even write the tests, we do. But we should be a good example.

stevejb71: In many cases when developing, it is perfectly safe to assume that an input is valid. You may for example have integration test coverage proving that invalid input never reaches your code under unit test. 

Validations may be redundant, and cost both programmer time and CPU cycles.

Vankog: I agree, we should avoid redundant tests.
However, we don't have such (integration-) tests in place.

exercises/nucleotide-count/canonical-data.json
Vankog: Yes. I only included error handling of ASCII chars.

Vankog:
@Insti Thanks for your feedback.
I have replied to them individually above.
For now I changed the descriptions again.

exercises/nucleotide-count/canonical-data.json
```diff
@@ -1,12 +1,19 @@
 {
   "exercise": "nucleotide-count",
-  "version": "1.0.0",
+  "comments": [
+    "Given a string of nucleotides, the count of each nucleotide is returned."
+  ],
+  "version": "1.1.0",
   "cases": [
     {
```
Vankog: I moved the longer description into a comment like others do it.

exercises/nucleotide-count/canonical-data.json
```diff
@@ -1,12 +1,19 @@
...
-      "description": "count all nucleotides in a strand",
+      "description": "counts all nucleotides in a provided strand.",
+      "comments": [
+        "nucleotideCounts counts all DNA-nucleotides in the input strand and returns an object with properties",
+        "A, C, G and T holding their respective amounts."
+      ],
       "cases": [
```
Vankog: I moved the longer description into a comment like others do it.

exercises/nucleotide-count/canonical-data.json
```diff
@@ -1,12 +1,19 @@
...
+  ],
+  "version": "1.1.0",
   "cases": [
     {
-      "description": "count all nucleotides in a strand",
+      "description": "counts all nucleotides in a provided strand.",
+      "comments": [
+        "nucleotideCounts counts all DNA-nucleotides in the input strand and returns an object with properties",
+        "A, C, G and T holding their respective amounts."
+      ],
```
Vankog: I moved the longer description into a comment like others do it.

exercises/nucleotide-count/canonical-data.json
stevejb71: You might want to add a comment at the top of the file, suggesting a check for referential transparency. But I don't think it should have a separate test case.

Insti: >  how do you make sure it gets implemented in the tracks?

That is not something we need to or should worry about. 

The people maintaining tracks are experienced developers with knowledge of their language, who have spent time reviewing many of the different solutions to exercises, when they identify solutions that are deficient in some way test cases can be added to the track. If these changes are thought to be more generally applicable they can then be added to this general repository.

The test case you are proposing is very implementation dependent and makes no sense in languages with immutable data. So for that reason I do not believe it is necessary to include it in the canonical data.

exercises/nucleotide-count/canonical-data.json
Insti: > "counts an empty DNA strand as 0."

'counts an' is redundant because all function calls will "count"

'DNA' is redundant because all input is DNA strands.

'as 0.' is information already present in the expected output.

Which leaves us with "empty strand"

Which means: "Returns the correct value when provided with an empty strand as input"

> my intent was to implement well-formed descriptions in the informal style of
{it / the test} does xyz in the case of abc.

If a description can be generated from the data within the test it adds no value.
A descriptions should describe the "why" of the test. What does this test test that all the others do not?

exercises/nucleotide-count/canonical-data.json
Insti: I can't see the code context for this comment any more. So I can't comment further on that.

> The TDD cycle is red-green-refactor(?) and sure, you could hardcode the handling of a one-length input and an n-length input to make it green, but then there is still the third phase: To check if there is any need to improve the code. If someone just hardcodes statements to pass several tests, then I dare to say this person might not want to program in the first place.

I disagree with this assessment. 

If hardcoded values make the tests pass, then hardcode the values. The refactoring that needs to happen is improving the design of the existing code without altering its behaviour. If after "refactoring" your code can handle more cases than it could previously, you've strayed from the path. 

The solution is to add more tests to test the cases that your code does not handle, and then write the code to pass these tests, and then to refactor to improve the design, again without altering its behaviour.

exercises/nucleotide-count/canonical-data.json
Insti: There are also no tests that the keyboard is plugged in or the screen is turned on.

exercises/nucleotide-count/canonical-data.json
Insti: I think this concept deserves wider discussion so I've replied as a new issue: https://github.com/exercism/problem-specifications/issues/902

exercises/nucleotide-count/canonical-data.json
Vankog: @Insti  Oh, the context disappeared even after committing normally. You can see the contest when viewing the commit directly:

exercises/nucleotide-count/canonical-data.json
Vankog: OK, I agree with this point about immutable languages. 
(One could argue now that some cases might be included in the canonical data as suggestions, kind of optional and meant to be mnemonic. Maybe in forms of comments. But this is another topic.)

exercises/nucleotide-count/canonical-data.json
Vankog: I see it this way:
- When running a test suite, you don't see the data, but only the names. So if for example something fails, then the test name itself should already give a clue about what might be the error or the constraints of the error. 
- When a user implements code inside an exercise track from a test spec or does bug fixing, he/she sees the data of the spec. But this only helps with trivial test logic. Say you have some more complex test logic, even though its intention can be described rather specifically in a sentence. Then why not being specific in the first place?
- When a track maintainer implements a test from the canonical data, he/she sees the name and the data. But to create a meaningful test case he/she has to be able to know exactly what the test is about. Inferring things implicitly leads more often than not to misunderstandings. ("Oh, you meant this? I thought you meat that?") I think every programmer knows what it means when the specifications are unclear or when the customer has a different understanding than the developers. That's the reason why whole methodologies like Domain Driven Design evolved in the first place. I can't think of any requirement engineer who ever said: "Meh, no need to be redundant with this spec, this is obvious."

When you just say "empty strand", noone could infer 
> "Returns the correct value when provided with an empty strand as input"
without context. It might as well mean that an empty strand should be returned. 
Yes, in this particular context here it might be obvious, but e.g. in a bit more complex production code it ain't. 

So the question that comes into my mind is: 
Why is redundancy here in the canonical data considered a bad thing?
Don't get me wrong, being concise is nice and logical and good practice indeed. But is it pragmatic from the users or implementers perspective? Is it readable and most importantly understandable and unambiguous? Does a beginning learner already has the knowledge/skill to infer the meaning from the context?
For example: I just recently saw someone learning the very basics of programming from a nicely done beginners website. And he struggled with the easiest and basic tasks where everyone else with a bit more understanding would knew what was going on in an instant. But he was not fully aware of the context that was given and therefore could not infer it easily.

So the TL;DR ^^:
Why do you consider redundancy in a spec file (more so even meta-spec file) a bad thing if it helps making the matter more clear and unambiguous?

exercises/nucleotide-count/canonical-data.json
Vankog: Now you are dragging the matter into ridiculousness. Sorry, but this is not an argument. ;-)

exercises/nucleotide-count/canonical-data.json
```diff
@@ -39,12 +71,38 @@
...
+          "strand": "AGCTTTTCATTCTGACTGCAACGGGCAATATGTCTCTGTGTGGATTAAAAAAAGAGTGTCTGATAGCAGC",
+          "expected" : true
+        },
+        {
+          "description": "handles incorrect inputs.",
+          "cases": [
+            {
+              "description": "throws error if DNA input does not include anything but ACGT-characters.",
+              "property": "nucleotideCounts",
+              "strand": "AGUUÄ1+ :'A'}CT",
```
petertseng: can you confirm whether or not Ä is in ASCII, given (I think) a declaration of intent in https://github.com/exercism/problem-specifications/pull/895#discussion_r139198420

> I only included error handling of ASCII chars.

(Let me know if I simply misinterpreted the declaration)

exercises/nucleotide-count/canonical-data.json
```diff
@@ -17,7 +24,32 @@
           }
         },
         {
-          "description": "strand with repeated nucleotide",
+          "comments" : [
+            "Should check for null-references and assume an empty input."
+          ],
+          "description": "returns zeroes for an undefined/null strand as argument.",
+          "property": "nucleotideCounts",
+          "strand": null,
```
petertseng: If `null` is tested as an input, I am thinking it should be an error, rather than to be treated as if `""` was the input. 

(Of course, I imagine languages for which strings are non-nullable will simply skip this test)

This comment must not be read as an endorsement or rejection of the statement "`null` should be tested as an input".

petertseng: EBWODP:

> If `null` is tested as an input, I am thinking it should be an error, rather than to be treated as if "" was the input.

The reason for this belief is that if we consider that the function under test must only accept strings, well `null` is not a string, so this is in error.

Vankog: Yeah, I agree. I had this same thought but had not changed it, yet, because of the ongoing discussion. I will probably change it to expect a throw in an upcoming commit.

exercises/nucleotide-count/canonical-data.json
Insti: > Now you are dragging the matter into ridiculousness.

Guilty. :blush:

> However, we don't have such (integration-) tests in place.

Is this not guilty of the same thing?
That we don't have integration tests is irrelevant because we do not need them, we can trust that the test input is correct and valid.

exercises/nucleotide-count/canonical-data.json
Vankog: Yep, `Ä` is ASCII: :-)
http://www.theasciicode.com.ar/extended-ascii-code/letter-a-umlaut-diaeresis-a-umlaut-lowercase-ascii-code-132.html

(Btw: `ß` too, wich was part of the Unicode-discussion ;-) )

exercises/nucleotide-count/canonical-data.json
Vankog: To a degree, yes.
But I think we should continue this thread over at the discussion thread.

exercises/nucleotide-count/canonical-data.json
Insti: > Adding a test for a single-character input
> and following up a test for a multi-char input.

This is a good idea.

exercises/nucleotide-count/canonical-data.json
Insti: > When running a test suite, you don't see the data, but only the names.

You need a better test runner. I see the expected input and output values and highlighted diffs of the results.

>  this only helps with trivial test logic. Say you have some more complex test logic

I'm not sure what your point is in this paragraph, but if your test logic is complex you're doing something wrong.


>  to create a meaningful test case he/she has to be able to know exactly what the test is about

Several tracks implement test generators that automatically generate the tests without the track maintainer needing to look at any of the tests, so I would argue this statement is false.

> Why do you consider redundancy in a spec file (more so even meta-spec file) a bad thing if it helps making the matter more clear and unambiguous?

Redundancy in a description is often a sign that the description is not telling me why this test is important, but is just repeating what the inputs and outputs are.

If it really is making things clear an unambiguous is it really redundant?
(see: no true Scotsman)

rpottsoh:
I believe that character is not in the lower 128 ASCII character table.  Are we restricting characters to lower 128?

Sent from my android device.

exercises/nucleotide-count/canonical-data.json
rpottsoh: `Ä` is an upper ASCII character.  Is the intent to restrict to printable characters?  `Ä` (142) is not considered a printable character.

Vankog: @rpottsoh Printable in what context?
In ASCII everything is considered printable from 32-255 (excluding 127).

In the context of this exercise only aA,cC,gG,tT is considered valid.
In another exercise (e.g. pangram) it might be everything from a-zA-Z.
In other cases even everything technically printable, might be valid. It really depends.

The issue discussed here was just, whether it is a Unicode-exclusive character like `℣` or `💩`, because there is consensus not to check them.

rpottsoh: Chars 32 through 126 are considered printable.  128 through 255 are considered extended.

petertseng: > Yep, `Ä` is ASCII: :-)
>
> http://www.theasciicode.com.ar/extended-ascii-code/letter-a-umlaut-diaeresis-a-umlaut-lowercase-ascii-code-132.html

In this file I see the two bytes 0xc3 0x84, which is in accordance with what you say since 0x84 is 132. Do you know what the 0xc3 is for? I couldn't figure it out, hoping you can help

rpottsoh: I don't see how this change is any better than the original test case.  In this exercise only ACGT are valid.  How is including extended ASCII characters to `"strand"` improving this test case?

Vankog: @petertseng im a bit confused. Where do you see characters 0x84 and 0xc3?
0x84 = 132 is the lowercase ä. In the file you should see the uppercase Ä = 142 instead.
And 0xc3 = 195 is `├`. Which I don't see here.

@rpottsoh The intention was to provide some random invalid characters from the ASCII space instead of just an X. But for this I'd like to refer you to the separate discussion thread at #902.

petertseng: > I'm a bit confused. Where do you see the characters 0x84 and 0xc3?

```
$ xxd canonical-data.json | grep AGUU
00000cf0: 6e64 223a 2022 4147 5555 c384 312b 203a  nd": "AGUU..1+ :
```

Let me know if you get a different result from doing the same on this file

NobbZ: That's utf8 for unicode u00c4 which is capital umlaut a

Vankog: Jep. You are assuming ASCII encoding of the file, but the standard is UTF8. So it's reasonable why you see these char numbers. They are still ASCII compliant tough, because the usual functions like ```charAt()``` and others usually convert to ASCII numbers by default.

petertseng: OK, and this is as you intended, correct?

petertseng: OK, and this (the fact that the file containing 0xc3 0x84) is as you intended, correct? As you can see, all I needed was to be sure it matches your intent.

petertseng: It was strange to me because I thought https://github.com/exercism/problem-specifications/pull/895#discussion_r139296590

> In the file you should see the uppercase Ä = 142 instead.

meant that one would expect to see byte 142 (0x8e) in the file, but that byte does not appear.

Also note that because the bytes are 0x84 and 0xc3, they would appears in the search I used in https://github.com/exercism/problem-specifications/issues/428#issuecomment-258366030

> `git grep -Pnl "[\x80-\xFF]"`

This may mean that I misunderstood the correct definitions somewhere. As you can see, I'm far from being an expert on this matter, so it's quite possible someone needs to set me straight.

exercises/nucleotide-count/canonical-data.json
```diff
@@ -39,12 +71,38 @@
...
+          "description": "returns the same amounts for consecutive nucleotideCounts calls.",
+          "property": "nucleotideCounts",
+          "strand": "AGCTTTTCATTCTGACTGCAACGGGCAATATGTCTCTGTGTGGATTAAAAAAAGAGTGTCTGATAGCAGC",
+          "expected" : true
+        },
+        {
+          "description": "handles incorrect inputs.",
+          "cases": [
+            {
+              "description": "throws error if DNA input does not include anything but ACGT-characters.",
```
stevejb71: Some languages do not have exceptions, or would return an error type for invalid inputs.

exercises/nucleotide-count/canonical-data.json
Vankog: I am no expert as well, but I think you might have a tripped upon a fallacy in encoding: You can't grep UTF8 encoded files assuming ASCII char numbers. You have to grep them with Unicode UTF8 encoding in mind.

Let me explain:
The json file here is encoded in Unicode UTF8. Which means that the symbol `Ä` will be encoded to bytecode as `0xC3 0x84`.

What you have assumed is that 0xC3 and 0x84 are ASCII-encoded characters. This means you interpreted the bytecode as ASCII, which translates to 0xc3 = 195 = `├` and 0x84 = 132 = `ä`. 
Obviously this is not what the file shows, just look at the file in your browser or any text editor.

Another way you could have interpret the bytecode wrongly is by assuming ANSI encoding. This would translate to `Ã` and `„`. Or Mac encoding: `√` and `Ñ`. Or EBDCID encoding: `C` and `d`.

Now let me show you how the bytecode looks, if I encode the file itself in ASCII. 
This is what you just xxd'ed with the UTF8 encoded file:
`6E 64 22 3A 20 22 41 47 55 55 C3 84 31 2B 20 3A`
This is what you xxd with the file encoded in ASCII:
`6E 64 22 3A 20 22 41 47 55 55 8E 31 2B 20 3A 27` 
This is what you xxd with the file encoded in ANSI:
`6E 64 22 3A 20 22 41 47 55 55 C4 31 2B 20 3A 27` 

Notice that most codes are the same. This is due to UTF8s backwards compatibility with ASCII. However, `Ä` is greater than the ASCII representation of 127, so it needs two bits in UTF8:
```
6E 64 22 3A 20 22 41 47 55 55 C3 84 31 2B 20 3A      // UTF8
6E 64 22 3A 20 22 41 47 55 55 8E    31 2B 20 3A 27   // ASCII
6E 64 22 3A 20 22 41 47 55 55 C4    31 2B 20 3A 27   // ANSI
``` 


The encoding is very important. It is the basics upon how the bytecode will be interpreted. In UTF8 case here, C3 and 84 are actually interpreted as one group of bytes to `Ä`. But this does not mean you can just grep for two-byte values now. UTF8 is actually even more complicated, because depending on what symbol is encoded, the number of bytes varies. See here for a nice explanation:
http://www.fileformat.info/info/unicode/utf8.htm
To use xxd, you actually have to couple it with an UTF8-aware text viewer, like this post explains:
https://askubuntu.com/a/848696

exercises/nucleotide-count/canonical-data.json
Vankog: But this is not an issue with this canonical data.
"error" is explicitly described in the readme:
https://github.com/exercism/problem-specifications#test-data-format-canonical-datajson
and also in the schema.

Or do you just mind the wording of the description: "throws error"?
Not sure how to describe this instead.

exercises/nucleotide-count/canonical-data.json
NobbZ: > The encoding is very important.

Yupp, and since we don't even know if a track will **encode** the data as a string or something else (especially functional languages may tend to wrap it in an ADT), we shouldn't even think about creating invalid input that might not be even representable.

Also there might be languages which require your source code in certain encodings, which can't even represent `Ä`.

Or transcoding the UTF-8 JSON string into the encoding required by the targetted compiler might be problematic.

So we have a plenty of reasons to stick to 7bit ASCII, an encoding which is even a direct subset of **all** today important encodings.

petertseng: > You can't grep UTF8 encoded files assuming ASCII char numbers

I'm sorry, I don't understand. The specific grep I performed (`grep AGUU`) doesn't affect the bytes in the files; it simply makes it so that the only line that xxd displays is the line with the bytes we are discussing, since I grepped for the characters immediately before. If I need to use grep a different way, can I ask if you know any proper flags I need to pass to grep?

Alternatively, if you were referring to the other grep I mentioned (`git grep -Pnl "[\x80-\xFF]"`) then I acknowledge that the specific characters found by it will be dependent on encoding; if we need a different pattern, I will discuss and find in #428.

> This is what you xxd with the file encoded in ANSI
> To use xxd, you actually have to couple it with an UTF8-aware text viewer

I'm sorry, I don't understand either of these two assertions. I thought xxd simply shows me the bytes in the file. Can I ask you do one of the two following?

1. Show me different flags xxd on canonical-data.json which demonstrates the three possibilities you report (reproduced below)
2. OR show me any other way of reading the exact bytes in canonical-data.json that demonstrates that the byte(s) that you desire exist in the file?

```
6E 64 22 3A 20 22 41 47 55 55 C3 84 31 2B 20 3A      // UTF8
6E 64 22 3A 20 22 41 47 55 55 8E    31 2B 20 3A 27   // ASCII
6E 64 22 3A 20 22 41 47 55 55 C4    31 2B 20 3A 27   // ANSI
```

What I thought I did was to report to you the bytes as they lie in this file on my hard drive as they were delivered to me by Git. Is that not what I did?

petertseng: > Show me
> Show me

Why am I asking to be shown? Simply so that I know that you are OK with the bytes that exist in this file, and they match your intention.

Vankog: Sorry for the grep confusion. I thought you xxd'ed and grep'ed afterwards. I am on Windows, so I am not that firm with unix commands.

> Alternatively, if you were referring to the other grep I mentioned (git grep -Pnl "[\x80-\xFF]") then I acknowledge that the specific characters found by it will be dependent on encoding; if we need a different pattern, I will discuss and find in #428.
Jup, that's what I mean.

Let me say it in other words:
When you view the bytecode of the file now as-is from git, it will show you what you already reported:
`6E 64 22 3A 20 22 41 47 55 55 -> C3 84 <- 31 2B 20 3A`
However, the file has no separate ASCII characters C3 and 84 in it, but has a single character C384, because it is UTF8-encoded.

If I would have encoded the same text content of the file in ASCII, though, and sent it to you, you *would* see the following byte code:
`6E 64 22 3A 20 22 41 47 55 55 -> 8E <- 31 2B 20 3A 27`
This was just included by me for demonstration purposes.

petertseng: > However, the file has no separate ASCII characters C3 and 84 in it, but has a single character C384, because it is UTF8-encoded.

I agree with this statement. The file does not contain the characters 0xC3 and 0x84. It contains the **bytes** 0xC3 and 0x84, which together represent Unicode codepoint 196 (Feel free to check my math: 0xc3 is 110**00011** and 0x84 is 10**000100** so we concatenate 00011 and 000100 to get 11000100 which is 196). So we seem to be on the same page with this, good.

> If I would have encoded the same text content of the file in ASCII, though, and sent it to you, you *would* see the following byte code:
> `6E 64 22 3A 20 22 41 47 55 55 -> 8E <- 31 2B 20 3A 27`

Got it.

Vankog: Yep, we are on the same page. :-)

petertseng: I didn't see that you had edited your post.

> > All test cases should only use ASCII characters
> 
> In my interpretation this meant all ASCII, including the extended ASCII (8-bit).
> In contrast, I think you interpreted it as the 7-bit ASCII only, didn't you?

It is true that given that I grepped for bytes 0x80 to 0xFF that I looked for files that had characters outside of 7-bit ASCII. I wouldn't know for sure what the original text meant, but I know I got no objections at the time, so that is moderate evidence in favour of 7-bit ASCII. If it does in fact mean 7-bit ASCII, it would imply needing to remove the Ä. I'll leave that decision to the interested parties.

exercises/nucleotide-count/canonical-data.json
stevejb71: I simply object to the use of the word "throws", because it is not language general enough.  Instead, you might write "gives an error if.." or "xxx input is invalid".

Vankog: @stevejb71 thanks!
If you don't mind, there is also a discussion about this over here: #905

Insti:
@Vankog are you interested in continuing to work on these pull requests?
I think they have achieved their goals of stimulating discussion, but still require further work to be considered ready to merge.

(If no further progress is made, I will close this on or after the 8th October.)

Vankog:
I do. I was just waiting for the sub-discussions to conclude so we can finally decide what should or shouldn't be done.

Vankog:
OK, new suggestion.

- converted the controversial cases into comments to support the track maintainers with this task.
- shortened/tweaked some test descriptions a bit.
- removed the ASCII >126 character.

exercises/nucleotide-count/canonical-data.json
```diff
@@ -1,12 +1,25 @@
...
   "exercise": "nucleotide-count",
-  "version": "1.0.0",
+  "comments": [
+    "Given a string of nucleotides, the count of each nucleotide is returned.",
+    "",
+    "It is the track's responsibility to add further language specific tests like:",
+    "input error handling (null/undefined parameters etc.)",
+    "referential transparency (i.e. evaluating, a function/method gives the same value for same arguments every time.)",
+    "",
+    "'error' and 'null' should be treated according to the languages' specifics and possibilities."
```
petertseng: since there is no more `null` that I see (I could have missed), the part about `null` wouldn't be necessary

exercises/nucleotide-count/canonical-data.json
```diff
@@ -39,12 +63,28 @@
           }
         },
         {
-          "description": "strand with invalid nucleotides",
+          "description": "ignores case in a strand of mixed-case nucleotides.",
           "property": "nucleotideCounts",
-          "strand": "AGXXACT",
+          "strand": "aGcTtTtCaTtCtGaCtGcAaCgGgCaAtAtGtCtCtGtGtGgAtTaAaAaAaGaGtGtCtGaTaGcAgC",
```
petertseng: I'm not convinced it's necessary to add the extra functionality (of ignoring case), but I also don't care.

exercises/nucleotide-count/canonical-data.json
```diff
@@ -39,12 +63,28 @@
...
+            "T": 21
           }
+        },
+        {
+          "description": "handles incorrect inputs.",
+          "cases": [
+            {
+              "description": "treats an input with anything else than ACGT-characters as error.",
+              "property": "nucleotideCounts",
+              "strand": "AGUU1+ :'A'}CT",
```
petertseng: I don't think this case is any better or worse than "AGXXACT" at ensuring the description is true. I might even call it worse just for clarity. I would keep AGXXACT. Strength of opinion 2/10.

Insti:
I agree with @petertseng 

As a reviewer this PR has gotten to a stage where reviewing it is very difficult.
I still agree with some of the changes and disagree with others, but reviewing all the different threads and working out what has changed in amongst all the changes and comments will take more time than I have available to review, so I do not expect I will change my my review status from changes requested.

I recommend this PR be closed and new separate PRs be opened for the different changes.

Vankog:
first split:
- [x] adding single char input case #951 
- [ ] adding char-case-agnostic case
- [ ] extend "strand with invalid nucleotides" case
- [ ] add suite comments (incl. suggestions for track maintainers)
- [ ] improve case descriptions

Insti:
Since this is being split into new separate PRs I think we can close this one.
Thanks for starting the discussion @Vankog :heart:

PR #901, opened 2017-09-15
Add titles to exercises with non-default capitalisation/punctuation

petertseng:
In generated READMEs, the rule for converting an exercise slug to a
human-readable form is to capitalise only the first letter of each
hyphen-separated word. For example, `lorem-ipsum` will become "Lorem
Ipsum". This is true of both the Trackler and Configlet README
generators.

Trackler's README generator (see `name` function):
https://github.com/exercism/trackler/blob/master/lib/trackler/specification.rb

Configlet's README generator (see `Name` function):
https://github.com/exercism/configlet/blob/master/track/problem_specification.go

For some exercises this is not the best possible title. For example see
http://exercism.io/exercises/haskell/rna-transcription/readme and
observe that the title is "Rna Transcription", whereas "RNA
Transcription" seems more faithful to the rules (RNA, as an initialism,
should be in all caps).

If we would like the README generator to generate these non-default
titles, it seems to make the most sense to add it in
problem-specifications. The alternative of adding the exceptions to
configlet has the disadvantage of requiring an update to configlet for
any new exercise that desires a non-default title.

The proposal is a `title` field in metadata.yml that overrides the
default capitalisation or punctuation. For those exercises for which the
default title is appropriate, it is recommend that this field be
omitted, to avoid unnecessary configuration.

Please receive confirmation from
https://github.com/exercism/configlet/issues/76 that this proposal is
appropriate on the configlet side before merging.

Please prepare a corresponding update to
https://github.com/exercism/docs/blob/master/language-tracks/exercises/anatomy/readmes.md
before merging.

### DIFF
--- a/exercises/dot-dsl/metadata.yml
+++ b/exercises/dot-dsl/metadata.yml
@@ -1,2 +1,3 @@
 ---
+title: "DOT DSL"
 blurb: "Write a Domain Specific Language similar to the Graphviz dot language"
--- a/exercises/etl/metadata.yml
+++ b/exercises/etl/metadata.yml
@@ -1,4 +1,5 @@
 ---
+title: "ETL"
 blurb: "We are going to do the `Transform` step of an Extract-Transform-Load."
 source: "The Jumpstart Lab team"
 source_url: "http://jumpstartlab.com"
--- a/exercises/ocr-numbers/metadata.yml
+++ b/exercises/ocr-numbers/metadata.yml
@@ -1,4 +1,5 @@
 ---
+title: "OCR Numbers"
 blurb: "Given a 3 x 4 grid of pipes, underscores, and spaces, determine which number is represented, or whether it is garbled."
 source: "Inspired by the Bank OCR kata"
 source_url: "http://codingdojo.org/cgi-bin/wiki.pl?KataBankOCR"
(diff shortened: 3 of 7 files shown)


petertseng:
EBWODP: The commit message has been amended to include the following:

For those exercises for which the
    default title is appropriate, it is recommend that this field be
    omitted, to avoid unnecessary configuration.

petertseng:
Another item that I want to check before merging:

Does anything else translate slugs into human titles and thus would benefit from a code change? Or if I ask very specifically: Does the website need to change?

petertseng:
> Does the website need to change?

v2: Titles are displayed as-is https://github.com/exercism/prototype/blob/master/app/views/exercises/show.html.haml#L21 They are a field in the database https://github.com/exercism/prototype/blob/master/db/schema.rb#L71 . The code that imports the exercises into the database would benefit from being aware of `title`.

petertseng:
v1: https://github.com/exercism/exercism.io/blob/master/app/views/languages/_exercises.erb#L10 which I think comes from trackler. Nice to have but not necessary if we assume v1 doesn't stay around for long.

Insti:
Is there any reason why we could not merge this as it is?
The extra information should be ignored by the tools until they are updated to use it, but once they are it will be already there.

petertseng:
I guess it can be merged as soon as I got agreement that the `title` key is acceptable to possible consumers. I guess I got that agreement in https://github.com/exercism/configlet/issues/76.

However, as a rule I do not allow myself to merge this until I write the doc for it. If someone else merges it because their rules != my rules, I guess I can't complain, if they're in a hurry.

PR #903, opened 2017-09-16
Correct `test-one` instructions in README

matthewmorgan:
`npm test` will work as described, because it's an alias for `npm run test`, but `test-one` requires the user to type `npm run test-one`

I realized this when (after making my last PR) I read the docs for the repo  ☺️


jpreese:
Thanks @matthewmorgan 🎉

Issue #824, opened 2017-06-07
isogram: spaces and hyphens not described.


Insti:
The [isogram description](https://github.com/exercism/x-common/blob/master/exercises/isogram/description.md) does not mention anything about spaces or hyphens, however the [canoncial-data](https://github.com/exercism/x-common/blob/master/exercises/isogram/canonical-data.json#L51) includes tests with spaces and hyphens.

Either: 
a) Remove the tests with hyphens/spaces.
or
b) Add something to the description about ignoring spaces and hyphens.

ErikSchierboom:
Either option is fine with me.

tejasbubane:
Closed via #912.

PR #912, opened 2017-09-28
isogram: Update spec to include info about spaces/hyphens.

sjwarner-bp:
Fix of issue #824 - isogram: spaces and hyphens not described.

Add details of how spaces and hyphens are acceptable, and an example of an isogram including hyphens.


Insti:
Closes: #824 

It would it be a good idea to add a test for 'six-year-old' to the canonical data?

Insti:
Thanks @sjwarner-bp :heart:

PR #915, opened 2017-09-30
all: minor description.md formatting fixes

navossoc:
Just some minor fixes and some changes to the formatting style.

Most of the changes are ignored while rendering, so, nothing fancy here...

As discussed here:
https://github.com/exercism/go/pull/870

Fell free to squash all the changes.

PS: This is the first PR, the second PR will go to `exercism/go` later to regenerate the content and fix one header on a specific go template.


Insti:
Thanks for doing all these as separate atomic commits, makes it much easier to review :heart:

These all look OK to me.

I've not checked the generated Readmes but don’t expect there will be any issues.
(And if there are any they can easily be fixed by another PR.)

Issue #810, opened 2017-05-25
protein-translation: fact mistake readme


roberthopman:
Right now (incorrect):
`There are also four terminating codons (also known as 'STOP' codons);` 

Change to:
`There are also three terminating codons (also known as 'STOP' codons);` 

Facts: [(https://en.wikipedia.org/wiki/Stop_codon)](https://en.wikipedia.org/wiki/Stop_codon)

Insti:
Great, thanks. Do you want to edit the description file and submit a pull request?

https://github.com/exercism/x-common/blob/master/exercises/protein-translation/description.md

roberthopman:
@Insti I'll keep adding more like these while finishing the elixir track. When I'm done, I might decide to fix them all at once. Is this the right documentation to follow while submitting a pr: https://github.com/exercism/docs/blob/master/maintaining-a-track/reviewing-a-pull-request.md

Insti:
> Is this the right documentation to follow while submitting a pr

Probably not, but I've not yet been able to find the correct documentation either :(

Insti:
Maybe: https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#pull-request-guidelines

Insti:
Hmm github didn't auto close this after all 😕

petertseng:
Use the presence or absence of the exclamation mark to determine ahead of time whether it would be auto closed

petertseng:
???? I can't find the exclamation mark anymore. My suggestion is invalid...?!?!!?!?

petertseng:
Sigh, fine. Use the presence or absence of the tooltip to determine ahead of time whether it would be auto closed

![image](https://user-images.githubusercontent.com/346181/31092540-9afe5f18-a763-11e7-9d64-2af65a3a95c4.png)

PR #916, opened 2017-10-01
protein-translation: Update description.md with correct fact as in #810

Selina-Jane:
Corrected "four stop codons" to "three stop codons" as described in issue #810


Insti:
Awesome! 
Great title, Accurate and helpful description, linked to the original issue 👍 
Thanks :heart:

PR #913, opened 2017-09-28
nucleotide-count: Refactoring the description

Dysp:
I have tried to be more precise, specific and accurate.
I am not sure what you are trying to accomplish with this description, so please let me know if you need any further help if the description expands.

PR #918, opened 2017-10-01
nucleotide-count: Remove reference to RNA from description

Insti:
In #913 the description of DNA was made more scientifically accurate.

However the reference to RNA is not relevant to this problem and is likely to cause confusion, so I have removed it from the description.


petertseng:
EBWODP

> want to remove

want to *also* remove (to make clear I'm not asking you to *instead* remove)

Insti:
Thanks for the reviews! :heart:

Issue #750, opened 2017-04-07
crypto-square: Remove test of intermediate functions


behrtam:
So, I was just looking at the canonical test data for `crypto-square` and I think it was the first time I saw test data for an exercise that also tested the intermediate steps of the problem like *normalization* or *concatenation of segments*.

Shouldn't those intermediate tests be removed, as it was agreed on in: https://github.com/exercism/discussions/issues/41

They originated from this PR: #250

Insti:
Link to [canonical-data.json](https://github.com/exercism/x-common/blob/master/exercises/crypto-square/canonical-data.json)

petertseng:
Indeed, the Haskell track discussed in https://github.com/exercism/xhaskell/issues/259 and decided not to test intermediate functions, diverging from x-common in https://github.com/exercism/xhaskell/pull/391.

Not saying that all have to take the exact approach the Haskell track took or that any other track should agree with the exact same reasons, but if this data helps make the decision, we are glad to provide it.

petertseng:
Since crypto-square's introduction into the Go track at https://github.com/exercism/xgo/pull/68 the Go track has never tested the intermediate functions. So if x-common goes this route, xgo would become closer to x-common.

nholden:
https://github.com/exercism/problem-specifications/pull/922 was merged, so I think we can close out this issue.

PR #922, opened 2017-10-04
crypto-square: Don't test intermediate functions

nholden:
We want to remove tests of intermediate functions in crypto-square that pry into implementation details per these issues:
- https://github.com/exercism/discussions/issues/41
- https://github.com/exercism/ruby/issues/422

This rewrites the canonical data for crypto-square so that we're only testing the expected value of the final "ciphertext" property and none of the intermediate properties.

### DIFF
--- a/exercises/crypto-square/canonical-data.json
+++ b/exercises/crypto-square/canonical-data.json
@@ -1,82 +1,27 @@
 {
   "exercise": "crypto-square",
-  "version": "2.0.0",
+  "version": "3.0.0",
   "cases": [
     {
       "description": "the spaces and punctuation are removed from the English text and the message is downcased",
       "cases": [
         {
           "description": "Lowercase",
-          "property": "normalizedPlaintext",
-          "plaintext": "Hello",
-          "expected": "hello"
+          "property": "ciphertext",
+          "plaintext": "A",
+          "expected": "a"
         },
         {
           "description": "Remove spaces",
-          "property": "normalizedPlaintext",
-          "plaintext": "Hi there",
-          "expected": "hithere"
+          "property": "ciphertext",
+          "plaintext": "  b ",
+          "expected": "b"
         },
         {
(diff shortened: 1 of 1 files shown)


nholden:
Here's one more relevant issue: https://github.com/exercism/problem-specifications/issues/750

exercises/crypto-square/canonical-data.json
```diff
@@ -1,82 +1,27 @@
 {
   "exercise": "crypto-square",
-  "version": "2.0.0",
+  "version": "3.0.0",
   "cases": [
     {
       "description": "the spaces and punctuation are removed from the English text and the message is downcased",
       "cases": [
```
Insti: Should we still be using sub `cases` keys if we are only testing one property?

Insti:
Removing the intermediate functions is good 👍 

There are probably more test cases that need to be added, but they can be included by a separate PR once we work out what they are.

exercises/crypto-square/canonical-data.json
nholden: Hmm, good question! I see [in the README](https://github.com/exercism/problem-specifications/blob/master/README.md#test-data-format-canonical-datajson):

> Test cases can be arbitrarily grouped with a description to make organization easier.

These `cases` all have to do with "normalizing" text, while the next group of `cases` deal with "chunking" and ordering letters. I'm not sure whether or not that "makes organization easier." I'm happy to dump the sub `cases` or save that re-organization for when we add new tests.

Insti: I'm not sure what the answer is either, I'm OK with leaving it as it is for now as the logical separation is still there even if the property name has changed.
If it ends up breaking someone's test generator I'm sure they'll submit a PR to fix it.

ErikSchierboom:
Thanks @nholden !

PR #924, opened 2017-10-04
Add isbn-verifier exercise

christophschlosser:
This was an starter exercise in my CS class.

The exercise is really simple, depending on how much your language supports converting types and string manipulation but it can be blown up into a full grown exercise with the bonus tasks.

This teaches you some simple string manipulation combined with basic numerical calculations.
You also get to use the three datatypes of number, string and bool.

### DIFF
--- a/exercises/isbn-verifier/canonical-data.json
+++ b/exercises/isbn-verifier/canonical-data.json
@@ -79,7 +79,7 @@
             "expected": false
         },
         {
-            "description": "check digit should be 0 but is X to prevent solutions to use X as 0",
+            "description": "check digit of X should not be used for 0",
             "property": "isbn",
             "input": "3-598-21515-X",
             "expected": false
--- a/exercises/isbn-verifier/canonical-data.json
+++ b/exercises/isbn-verifier/canonical-data.json
@@ -1,0 +1,82 @@
+{
+    "exercise": "isbn-verifier",
+    "version": "1.0.0",
+    "comments": [
+        "An expected value of true indicates a valid ISBN-10, ",
+        "whereas false means the ISBN-10 is invalid."
+    ],
+    "cases": [
+        {
+            "description": "valid isbn number",
+            "property": "isbn",
+            "input": "3-598-21508-8",
+            "expected": true
+        },
+        {
+            "description": "invalid isbn check digit",
+            "property": "isbn",
+            "input": "3-598-21508-9",
+            "expected": false
+        },
+        {
+            "description": "valid isbn number with a check digit of 10",
+            "property": "isbn",
+            "input": "3-598-21507-X",
+            "expected": true
+        },
+        {
+            "description": "check digit is a character other than X",
+            "property": "isbn",
+            "input": "3-598-21507-A",
+            "expected": false
+        },
+        {
+            "description": "invalid character in isbn",
+            "property": "isbn",
+            "input": "3-598-2K507-0",
+            "expected": false
+        },
+        {
+            "description": "valid character in isbn",
+            "property": "isbn",
+            "input": "3-598-2X507-0",
+            "expected": false
+        },
+        {
+            "description": "valid isbn without separating dashes",
+            "property": "isbn",
+            "input": "3598215088",
+            "expected": true
+        },
+        {
+            "description": "isbn without separating dashes",
+            "property": "isbn",
+            "input": "359821507X",
+            "expected": true
+        },
+        {
+            "description": "isbn without check digit",
+            "property": "isbn",
+            "input": "359821507",
+            "expected": false
+        },
+        {
+            "description": "too long isbn",
+            "property": "isbn",
+            "input": "3598215078X",
+            "expected": false
+        },
+        {
+            "description": "isbn without check digit",
+            "property": "isbn",
+            "input": "3-598-21507",
+            "expected": false
+        },
+        {
+            "description": "too long isbn",
+            "property": "isbn",
+            "input": "3-598-21507-XA",
+            "expected": false
+        }
+    ]
+}
--- a/exercises/isbn-verifier/canonical-data.json
+++ b/exercises/isbn-verifier/canonical-data.json
@@ -37,7 +37,7 @@
             "expected": false
         },
         {
-            "description": "valid character in isbn",
+            "description": "X is only valid as a check digit",
             "property": "isbn",
             "input": "3-598-2X507-0",
             "expected": false
@@ -49,19 +49,19 @@
             "expected": true
         },
         {
-            "description": "isbn without separating dashes",
+            "description": "isbn without separating dashes and X as check digit",
             "property": "isbn",
             "input": "359821507X",
             "expected": true
         },
         {
-            "description": "isbn without check digit",
+            "description": "isbn without check digit and dashes",
             "property": "isbn",
             "input": "359821507",
             "expected": false
         },
         {
-            "description": "too long isbn",
+            "description": "too long isbn and no dashes",
             "property": "isbn",
             "input": "3598215078X",
             "expected": false
@@ -76,6 +76,12 @@
             "description": "too long isbn",
             "property": "isbn",
             "input": "3-598-21507-XA",
+            "expected": false
+        },
+        {
+            "description": "check digit should be 0 but is X to prevent solutions to use X as 0",
+            "property": "isbn",
+            "input": "3-598-21515-X",
             "expected": false
         }
     ]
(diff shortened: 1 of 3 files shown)


NobbZ:
There should be at least a rough description of how a ISBN-10 is structured and how the checksum is calculated.

exercises/isbn-verifier/canonical-data.json
```diff
@@ -0,0 +1,82 @@
...
+            "input": "3-598-21507",
+            "expected": false
+        },
+        {
+            "description": "too long isbn",
+            "property": "isbn",
+            "input": "3-598-21507-XA",
+            "expected": false
+        }
+    ]
```
Insti: Test suggestion: Valid ISBN for which the check digit should be `0` but is `X` to catch solutions that convert `X` to `0`

exercises/isbn-verifier/canonical-data.json
```diff
@@ -0,0 +1,82 @@
...
+            "expected": false
+        },
+        {
+            "description": "invalid character in isbn",
+            "property": "isbn",
+            "input": "3-598-2K507-0",
+            "expected": false
+        },
+        {
+            "description": "valid character in isbn",
```
Insti: This description is confusing. 

Maybe: "X is only valid as a checkdigit."

Also make sure that this ISBN would be valid if 10 was used in place of X in the calculation.

exercises/isbn-verifier/canonical-data.json
```diff
@@ -0,0 +1,82 @@
...
+            "expected": false
+        },
+        {
+            "description": "valid isbn without separating dashes",
+            "property": "isbn",
+            "input": "3598215088",
+            "expected": true
+        },
+        {
+            "description": "isbn without separating dashes",
```
Insti: This description is the same as the last, what is this test testing differently?

exercises/isbn-verifier/canonical-data.json
```diff
@@ -0,0 +1,82 @@
...
+            "expected": false
+        },
+        {
+            "description": "too long isbn",
+            "property": "isbn",
+            "input": "3598215078X",
+            "expected": false
+        },
+        {
+            "description": "isbn without check digit",
```
Insti: duplicate description

exercises/isbn-verifier/canonical-data.json
```diff
@@ -0,0 +1,82 @@
...
+            "property": "isbn",
+            "input": "3-598-21507",
+            "expected": false
+        },
+        {
+            "description": "too long isbn",
```
Insti: duplicate description

Insti:
Good exercise. 👍 
Needs a few tweaks but is heading in the right direction.

exercises/isbn-verifier/canonical-data.json
christophschlosser: The X is the checkdigit

exercises/isbn-verifier/canonical-data.json
christophschlosser: Good idea, thanks!

exercises/isbn-verifier/canonical-data.json
christophschlosser: Updated.

exercises/isbn-verifier/canonical-data.json
christophschlosser: Updated

exercises/isbn-verifier/canonical-data.json
christophschlosser: Great idea, added!

Insti:
Thanks for those changes.
I'm sure it will evolve as it gets implemented and people start solving the exercise, but this seems like a reasonable starting point.

exercises/isbn-verifier/description.md
```diff
@@ -0,0 +1,35 @@
...
+Which proves that the ISBN is valid.
+
+## Caveats
+
+Converting from string to number can be tricky in certain languages.
+It's getting even trickier since the check-digit of an ISBN-10 can be 'X'.
+
+## Bonus tasks
+
+* Generate an valid ISBN-13 from the input ISBN-10 (and maybe verify it again with an derived verifier)
```
rpottsoh: Change `an` to `a`.

exercises/isbn-verifier/canonical-data.json
```diff
@@ -0,0 +1,88 @@
...
+            "property": "isbn",
+            "input": "3-598-21507-XA",
+            "expected": false
+        },
+        {
+            "description": "check digit should be 0 but is X to prevent solutions to use X as 0",
```
rpottsoh: How about?:
```text
"description": "check digit of X should not be used for 0",
```
This test is essentially the same as:
```text
"description": "invalid isbn check digit",
```

christophschlosser: Thanks for the improved description.

Yes indeed but X is in a bit of a special position for ISBN so I think it doesn't hurt to add some extra tests for this.

exercises/isbn-verifier/description.md
christophschlosser: Done

exercises/isbn-verifier/description.md
```diff
@@ -15,7 +15,7 @@ The following number block is to identify the publisher. Since this is a three d
 The last digit in the ISBN is the check digit which is used to detect read errors.
 
 The first 9 digits in the ISBN have to be between 0 and 9.
-The check digit can additionally be an 'X' to allow 10 to be a valid check digit as well.
+The check digit can additionally be a 'X' to allow 10 to be a valid check digit as well.
```
rpottsoh: @christophschlosser my apologies.  I was not clear on when `an` should be `a`.  In this instance `an` is OK.  An is used when the next word starts with a vowel or starts with a vowel sound.  In this instance 'X' starts with a vowel sound, `e` in this instance.

exercises/isbn-verifier/canonical-data.json
```diff
@@ -79,7 +79,7 @@
             "expected": false
         },
         {
-            "description": "check digit should be 0 but is X to prevent solutions to use X as 0",
+            "description": "check digit of X should not be used for 0",
```
rpottsoh: :+1:   Thanks!

exercises/isbn-verifier/description.md
```diff
@@ -30,6 +30,6 @@ It's getting even trickier since the check-digit of an ISBN-10 can be 'X'.
 
 ## Bonus tasks
 
-* Generate an valid ISBN-13 from the input ISBN-10 (and maybe verify it again with an derived verifier)
+* Generate a valid ISBN-13 from the input ISBN-10 (and maybe verify it again with a derived verifier)
```
rpottsoh: :+1:  Thanks!

exercises/isbn-verifier/description.md
christophschlosser: No problem @rpottsoh . TIL. I uploaded a new commit where this is fixed.

exercises/isbn-verifier/canonical-data.json
```diff
@@ -0,0 +1,88 @@
...
+        {
+            "description": "invalid character in isbn",
+            "property": "isbn",
+            "input": "3-598-2K507-0",
+            "expected": false
+        },
+        {
+            "description": "X is only valid as a check digit",
+            "property": "isbn",
+            "input": "3-598-2X507-0",
```
petertseng: An implementation that unconditionally treats X as 10 will still reject this ISBN:

3 * 10 + 5 * 9 + 9 * 8 + 8 * 7 + 2 * 6 + 10 * 5 + 5 * 4 + 0 * 3 + 7 * 2 + 0 * 1 = 299, and 299 is not divisible by 11.

**If** the intention of this test is to ensure that an implementation only treats X as 10 in the check digit position, then a case such as "3-598-2X507-9" is needed.

exercises/isbn-verifier/canonical-data.json
```diff
@@ -0,0 +1,88 @@
...
+        {
+            "description": "isbn without check digit",
+            "property": "isbn",
+            "input": "3-598-21507",
+            "expected": false
+        },
+        {
+            "description": "too long isbn",
+            "property": "isbn",
+            "input": "3-598-21507-XA",
```
petertseng: An implementation that would potentially attempt to checksum ISBNs that are too long would still reject this input because it would see that the "A" is invalid.

**If** the intention of this case is to ensure that an implementation does in fact reject ISBNs that are too long, at the very least a valid character should be used instead of A.

If one wanted to be thorough, one might have to include three different cases here, respectively checking for implementations that try to checksum:

* all 11 resulting digits (with a coefficient of 11 on the leftmost one, naturally)
* the 10 rightmost digits
* the 10 leftmost digits

by providing a too-long ISBN that has a valid check digit under each respective scheme.

PR #928, opened 2017-10-05
change: add archived link to exercise source

robphoenix:


exercises/change/metadata.yml
```diff
@@ -1,2 +1,4 @@
 ---
 blurb: "Correctly determine change to be given using the least number of coins"
+source: "Software Craftsmanship - Kata-logue"
```
rpottsoh: Maybe snip off "Kata-logue"?  I don't see those words mentioned anywhere.

robphoenix: I've changed them to the title on the page

Issue #356, opened 2016-08-28
crypto-square: "n empty spaces... should be distributed evenly across the last n rows" is unclear


matthewfischer:
The crypto-square examples and directions do not mention that trailing spaces should be stripped. It instead requires perfect squares which should include trailing spaces. This should be fixed in the README for example, the splunk example should be:

"s-u- "
"p-n- "
"l-k- "

However the output requires no space after the lk pair. This issue is repeated in other ones as well.

Originally filed against the Golang README as https://github.com/exercism/xgo/issues/348

kytrinyx:
@ErikSchierboom @NobbZ @IanWhitney @petertseng @kotp @Insti you're the first ones to come to mind with respect to details on curriculum. Would you mind having a look/think about this one?

Insti:
All spaces are stripped from the input string. From the Readme:
> First, the input is normalized: the spaces and punctuation are removed ...

Then the output is generated as code blocks separated by a single space there should be no trailing space.
> ... output in r-length chunks separated by spaces.

I believe the Go example solution is correct.

ErikSchierboom:
I agree with @Insti, this is mentioned in the README and thus the example is correct.

matthewfischer:
If you remove the trailing spaces it's no longer a perfect square which is
also a requirement right?

petertseng:
> perfect square which is also a requirement right?

[Readme](https://github.com/exercism/x-common/blob/master/crypto-square.md) asks for a rectangle, not necessarily a square. It says `c - r <= 1` so it doesn't have to be a square (it won't be a square if `c - r = 1`)

NobbZ:
As far as I do understand the README, there is no way to have anything in the encoded text that is not matching `/[[:alnum:]]/`, since whitespace and punctuation has to be stripped (personally I do strip all non-printables as whitespace).

I haven't looked at the xgo issue though, I have some problems to read go and currently I do not have the time to try it ;)

matthewfischer:
Stripping the input is not relevant here. The question is on adding back
the filler spaces at the end.

Insti:
There are no filler spaces.
@matthewfischer can you quote the section of the readme that leads you to believe there are?

matthewfischer:
Absolutely. There will always be spaces in the output from what I see. The text says:

> Phrases that fill perfect squares
> `(r X r)` should be output in `r`-length chunks separated by spaces.
[...]

PR #929, opened 2017-10-07
crypto-square: Fix description copy and example

nholden:
The final example in the crypto-square description is inconsistent with the source of the exercise, the canonical data, and itself. This change does a few things to correct that:

- While the encoding method is called a "square code," we're really dealing with all sorts of rectangles, squares included. The description gets this right up until the copy before the last example, and this change clarifies that.

- In both the [source of the exercise](http://users.csc.calpoly.edu/~jdalbey/103/Projects/ProgrammingPractice.html) and [the canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/crypto-square/canonical-data.json#L47), when a "chunk" is shorter than the length of a row, a blank space is added to the end of that chunk. This change fixes that in the final example.

- The copy before the final example says that "spaces should be distributed evenly across the last `n` rows," but it doesn't explain how. This change specifies that the spaces should be added to the end of each chunk.

Here's a related discussion: https://github.com/exercism/problem-specifications/issues/356

### DIFF
--- a/exercises/crypto-square/description.md
+++ b/exercises/crypto-square/description.md
@@ -43,12 +43,14 @@ The message above is coded as:
 imtgdvsfearwermayoogoanouuiontnnlvtwttddesaohghnsseoau
 ```
 
-Output the encoded text in chunks.  Phrases that fill perfect squares
-`(r X r)` should be output in `r`-length chunks separated by spaces.
-Imperfect squares will have `n` empty spaces.  Those spaces should be distributed evenly across the last `n` rows.
+Output the encoded text in chunks.  Phrases that fill perfect rectangles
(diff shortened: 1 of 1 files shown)


nholden:
Is there anything I can do to help get this and https://github.com/exercism/problem-specifications/pull/937 merged? I'd like to resolve https://github.com/exercism/ruby/issues/422, and it makes sense to have these two PRs merged before submitting a PR in the Ruby repo.

Insti:
2 approvals, no objections, I'll merge this.

Insti:
Done, thanks for the bump.
And thanks for doing the changes :heart:

nholden:
Thanks, @Insti! 😄

PR #934, opened 2017-10-09
zipper: add canonical data

### LINKED ISSUE #593: zipper: Implement canonical-data.json
exercises in sync.

The Zipper exercise can be found in the [./exercises/zipper/][exercise-dir] directory.

See http://exercism.io/contribute/canonical-data/zipper for the up-to-date list of
language tracks that have an implementation of the Zipper exercise.

[exercise-dir]: https://github.com/exercism/x-common/tree/master/exercises/zipper

HarrisonMc555:
Closes #593.

Of the five tracks that had already implemented this exercise, five had the exact same test cases, and the other two had the same test cases plus a few more (identical) test cases. I opted to simply list all of the test cases I found in the order they were in.

I ran the test against the `canonical-schema.json` and it said the file is valid.

### DIFF
--- a/exercises/zipper/canonical-data.json
+++ b/exercises/zipper/canonical-data.json
@@ -1,0 +1,671 @@
+{
+  "exercise": "zipper",
+  "version": "1.0.0",
+  "comments": [
+    " The test cases for this exercise include an initial tree and a     ",
+    " series of operations to perform on the initial tree.               ",
+    "                                                                    ",
+    " Trees are encoded as nested objects. Each node in the tree has     ",
+    " three members: 'value', 'left', and 'right'. Each value is a       ",
+    " number (for simplicity). Left and right are trees. An empty node   ",
+    " is encoded as null.                                                ",
+    "                                                                    ",
+    " Each operation in the operations list is an object. The function   ",
+    " name is listed under 'operation'. If the function requires         ",
+    " arguments, the argument is listed under 'item'. Some functions     ",
+    " require values (i.e.  numbers), while others require trees.        ",
+    " Comments are always optional and can be used almost anywhere.      "
+  ],
+  "cases": [
+    {
+      "description": "data is retained",
+      "property": "expectedValue",
+      "initialTree": {
+        "value": 1,
+        "left": {
+          "value": 2,
+          "left": null,
+          "right": {
+            "value": 3,
+            "left": null,
+            "right": null
+          }
+        },
+        "right": {
+          "value": 4,
+          "left": null,
+          "right": null
+        }
+      },
+      "operations": [
+        {
+          "operation": "to_tree"
+        }
+      ],
+      "expected": {
+        "type": "tree",
+        "value": {
+          "value": 1,
+          "left": {
+            "value": 2,
+            "left": null,
+            "right": {
+              "value": 3,
+              "left": null,
+              "right": null
+            }
+          },
+          "right": {
+            "value": 4,
+            "left": null,
+            "right": null
+          }
+        }
+      }
+    },
+    {
+      "description": "left, right and value",
+      "property": "expectedValue",
+      "initialTree": {
+        "value": 1,
+        "left": {
+          "value": 2,
+          "left": null,
+          "right": {
+            "value": 3,
+            "left": null,
+            "right": null
+          }
+        },
+        "right": {
+          "value": 4,
+          "left": null,
+          "right": null
+        }
+      },
+      "operations": [
+        {
+          "operation": "left"
+        },
+        {
+          "operation": "right"
+        },
+        {
+          "operation": "value"
+        }
+      ],
+      "expected": {
+        "type": "int",
+        "value": 3
+      }
+    },
+    {
+      "description": "dead end",
+      "property": "expectedValue",
+      "initialTree": {
+        "value": 1,
+        "left": {
+          "value": 2,
+          "left": null,
+          "right": {
+            "value": 3,
+            "left": null,
+            "right": null
+          }
+        },
+        "right": {
+          "value": 4,
+          "left": null,
+          "right": null
+        }
+      },
+      "operations": [
+        {
+          "operation": "left"
+        },
+        {
+          "operation": "left"
+        }
+      ],
+      "expected": {
+        "type": "zipper",
+        "value": null
+      }
+    },
+    {
+      "description": "tree from deep focus",
+      "property": "expectedValue",
+      "initialTree": {
+        "value": 1,
+        "left": {
+          "value": 2,
+          "left": null,
+          "right": {
+            "value": 3,
+            "left": null,
+            "right": null
+          }
+        },
+        "right": {
+          "value": 4,
+          "left": null,
+          "right": null
+        }
+      },
+      "operations": [
+        {
+          "operation": "left"
+        },
+        {
+          "operation": "right"
+        },
+        {
+          "operation": "to_tree"
+        }
+      ],
+      "expected": {
+        "type": "tree",
+        "value": {
+          "value": 1,
+          "left": {
+            "value": 2,
+            "left": null,
+            "right": {
+              "value": 3,
+              "left": null,
+              "right": null
+            }
+          },
+          "right": {
+            "value": 4,
+            "left": null,
+            "right": null
+          }
+        }
+      }
+    },
+    {
+      "description": "traversing up from top",
+      "property": "expectedValue",
+      "initialTree": {
+        "value": 1,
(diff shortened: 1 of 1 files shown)


HarrisonMc555:
For history, these are the tracks that had implemented this exercise when I created this branch:
  - C# (common)
  - F# (common)
  - Haskell (common)
  - OCaml (common)
  - Scala (common)
  - Erlang (augmented)
  - Elixir (augmented)

By "common", I mean that the track had the same test cases as every other track. By "augmented", I mean the track had additional test cases. Both "augmented" tracks had the same additional test cases.

NobbZ:
> * Erlang (augmented)
> * Elixir (augmented)
>
> [...] Both "augmented" tracks had the same additional test cases.

I used the elixir tests as base when I created the erlang version. It was the most logical thing for me, because both languages share the same virtual machine and therefore the same base types.

Can you tell which cases are the additional ones?

HarrisonMc555:
Here you go:

## Categorizing Test Cases

#### Common
  - data is retained
  - left, right and value
  - dead end
  - tree from deep focus
  - setValue
  - setLeft with Just
  - setRight with Nothing
  - different paths to same zipper


#### Augmented
  - data is retained
  - left, right and value
  - dead end
  - tree from deep focus
  - traversing up from top
  - left, right, and up
  - set_value
  - set_value after traversing up
  - set_left with leaf
  - set_right with nil
  - set_right with subtree
  - set_value on deep focus


#### Additional
  - traversing up from top
  - left, right, and up
  - set_value after traversing up
  - set_right with subtree
  - set_value on deep focus


#### Missed
  - different paths to same zipper

## Work to be done

I didn't realize until now that there was one test case that was included in the "common" set of tracks but not in the "augmented" tracks, and that is "different paths to same zipper".

In my current version of `canonical_data.json`, the "expected" value can either be a value (`int`) or a tree. However, there are test cases that expect a zipper. The only one I've included so far expected the zipper to be `null`, so it wasn't really an issue. However, this last test case poses two challenges:
  1. How do we encode a zipper, which includes data for both the tree it is traversing and the current position inside the tree?
  2. How do we encode an expected value that is not just a static value (whether `int`, tree, or zipper) but instead a sequence of operations to perform, just like what we're testing against?

OCaml specifically differentiates between comparing `int`s, trees, and zippers. Should I make the expected value an object, with a `"type"` parameter specifying whether it is an `int`, tree, or zipper?

Also, should the "expected" value be encoded with either just a static value OR an initial tree and a sequence of operations?

I would love any input.

## Example

The way I see it, explicit type encodings would be the most consistent and useful. The only part about this that I don't totally feel comfortable with is the difference between value based encoding of zippers and operation based encoding of zippers...is the fact that one has a "value" member and one has "initialTree" and "operations" members make them different enough to make it easy to parse?

#### Tree
```json
"expected": {
  "type": "tree",
  "value": {
    "value": 1,
    "left": null,
    "right": null
  }
}
```

#### Int
```json
"expected": {
  "type": "int",
  "value": 3
}
```

#### Zipper

##### Value based encoding
```json
"expected": {
  "type": "zipper",
  "value": null
}
```

##### Operations based encoding
"expected": {
  "type": "zipper",
  "initialTree": {
    "value": 1,
    "left": null,
    "right": null
  },
  "operations": [
    {
      "operation": "from_tree"
    },
    {
      "operation": "right"
    }
  ]
}
```

**Comment by Insti** [comment]
I'll recuse myself from this one due to lack of familiarity with the exercise, but I'd like to say a big thankyou to you @HarrisonMc555 for tackling this! :star: :heart: ⭐️

File: exercises/zipper/canonical-data.json [hunk 1]
```diff
@@ -0,0 +1,715 @@
...
+    " require values (i.e.  numbers), while others require trees.        ",
+    " Comments are always optional and can be used almost anywhere.      "
+  ],
+  "cases": [
+    {
+      "description": "Creating a zipper for a binary tree",
+      "cases": [
+        {
+          "description": "data is retained",
+          "property": "property",
```
petertseng: I understand that there is no real good name for this property, but I hope there is at least a way to sneak a little bit of usefulness in here rather than leaving it as "property".

What of the suggestion of https://github.com/exercism/problem-specifications/blob/master/exercises/circular-buffer/canonical-data.json ("run") or https://github.com/exercism/problem-specifications/blob/master/exercises/react/canonical-data.json ("react")? Are they suitable for use here?

exercises/zipper/canonical-data.json
```diff
@@ -0,0 +1,715 @@
...
+    " Each operation in the operations list is an object. The function   ",
+    " name is listed under 'operation'. If the function requires         ",
+    " arguments, the argument is listed under 'item'. Some functions     ",
+    " require values (i.e.  numbers), while others require trees.        ",
+    " Comments are always optional and can be used almost anywhere.      "
+  ],
+  "cases": [
+    {
+      "description": "Creating a zipper for a binary tree",
+      "cases": [
```
petertseng: far as I can tell the only reason for this second level of nesting under `cases` is to attach a description to this entire group? I feel like we could do without, and just one level of nesting under `cases`. Yes? No?

petertseng:
Alternatively, given that `from_tree` is the first operation of of all the lists of `operations`:

Leave `initialTree` in place but remove `from_tree`. All other operations need a `zipper` to act on, so it's implicit that that zipper is the zipper resulting from using `from_tree` on `initialTree`.

exercises/zipper/canonical-data.json
```diff
@@ -0,0 +1,715 @@
...
+              "right": {
+                "value": 4,
+                "left": null,
+                "right": null
+              }
+            }
+          }
+        },
+        {
+          "description": "set_right with subtree",
```
petertseng: "set_right with subtree" case currently needs a `to_tree` as its last operation, otherwise the `operations` result in a zipper, but the expected value is a tree.

petertseng:
OK, I managed to finish https://github.com/petertseng/exercism-problem-specifications/blob/verify-zipper/exercises/zipper/verify.rb . I have reported all problems it found.

petertseng:
> I have reported all problems it found.

One test is missing a `to_tree` is the only problem. Other suggestions might help make things clearer, but don't affect correctness.

exercises/zipper/canonical-data.json
HarrisonMc555: I meant to decide on a good name but forgot :sweat_smile:
Do you have any suggestions?

Would we possibly use the property to categorize what kind of test we're running, i.e. `function(input) == output` vs. `function(input) == function(input)`?

exercises/zipper/canonical-data.json
HarrisonMc555: This...was a mistake. I think I copied another exercise as a starting point and didn't realize this was happening. I just changed it.

HarrisonMc555:
> Leave initialTree in place but remove from_tree. All other operations need a zipper to act on, so it's implicit that that zipper is the zipper resulting from using from_tree on initialTree.
I think I like this the best. It's true that the rest of the operations implicitly are using the zipper produced by the previous operation, so this would make it more consistent.

HarrisonMc555:
I'm still trying to decide on what to use for the `"property"` key. I can either use an arbitrary string (not much better than `"property": "property"` or try to use it to encode what kind of test I'm running. Would something like `expectedValue` and `sameResultFromOperations` be any good? The idea being that the first is `function(input) == output` and the latter is `function1(input1) == function2(input2)`? Or is that not what the `"property"` key is for?

petertseng:
`git show -w b21648f56841289a94d22edb93123118eb999edd` looks as I expect it to.

> Would something like `expectedValue` and `sameResultFromOperations` be any good? The idea being that the first is `function(input) == output` and the latter is `function1(input1) == function2(input2)`? Or is that not what the `"property"` key is for?

Looks like an acceptable use of `property`. To me differing `property` means the meaning of the keys is different, and this is what you are proposing.

petertseng:
https://github.com/petertseng/exercism-problem-specifications/blob/verify-zipper/exercises/zipper/verify.rb has been updated (remove nesting, and use `initialTree` as the initial zipper, without the `from_tree` operation) and tests pass.

HarrisonMc555:
Ok I changed the property values. I think I'm good to go. Thanks @petertseng!

HarrisonMc555:
Done, ready to merge.

petertseng:
Great, thanks.

I hope I am not remiss in merging. I figure that if anyone wanted to chime in, they've had 72 hours to do so.

Issue #722, opened 2017-03-13
robot-simulator: should represent positions as array or object, not string


petertseng:
Currently, positions are represented as strings like `"(11, 5)"` in the [canonical data for robot-simulator](https://github.com/exercism/x-common/blob/master/exercises/robot-simulator/canonical-data.json).

For languages where `(11, 5)` is valid syntax, this seems fine - the test generator will simply be able to interpolate this string wherever it is necessary.

For those languages where it is not valid syntax, it seems like an unusual burden to force test generators for that language to parse the x and y coordinate from the string `"(11, 5)"`.

Consider changing this to an array containing two integers such as `[11, 5]` or an object with explicit keys such as `{"x": 11, "y": 5}`.

rbasso:
I think that using strings is definitely not a good way to represent tuples in JSON. Number should always be represented as numbers.

👍  for an array or an object.

ErikSchierboom:
Great point! I think having an object would be best as it is the most explicit.

rbasso:
###### Encoding tuples as arrays

```json
    {
      "description": "A robot is created with a position and a direction",
      "cases": [
        {
          "description": "Robots are created with a position and direction",
          "property": "create",
          "robot": {
            "position": [0, 0],
            "direction": "north"
          },
          "expected": {
            "position": [0, 0],
            "direction": "north"
          }
    },
```

###### Encoding tuples as objects

```json
    {
      "description": "A robot is created with a position and a direction",
      "cases": [
        {
          "description": "Robots are created with a position and direction",
          "property": "create",
          "robot": {
            "position": {
              "x": 0,
              "y": 0
            },
            "direction": "north"
          },
          "expected": {
            "position": {
              "x": 0,
              "y": 0
            },
            "direction": "north"
          }
    },
```

The Google JSON Style Guide usually recommends avoiding nesting structures
in favor of flatter ones, but that is not a rule:

> Data elements should be "flattened" in the JSON representation. Data should
> not be arbitrarily grouped for convenience.
>
> In some cases, such as a collection of properties that represents a single
> structure, it may make sense to keep the structured hierarchy. These cases
> should be carefully considered, and only used if it makes semantic sense.
> For example, an address could be represented two ways, but the structured
> way probably makes more sense for developers:
> (...)

**Question:** Why the input data is in a `robot` object and the expected result are not? Isn't the opposite one would expect from a `create` function?

Anyway, every alternative is better than representing tuples with strings. 😄

[...]

PR #935, opened 2017-10-09
queen-attack: represent positions as objects

### LINKED ISSUE #721: queen-attack: should represent positions as array or object, not string
Currently, positions are represented as strings like `"(2, 2)"` in the [canonical data for queen-attack](https://github.com/exercism/x-common/blob/master/exercises/queen-attack/canonical-data.json).

For languages where `(2, 2)` is valid syntax, this seems fine - the test generator will simply be able to interpolate this string wherever it is necessary.

For those languages where it is not valid syntax, it seems like an unusual burden to force test generators for that language to parse the rank and file from the string `"(2, 2)"`.

Consider changing this to an array containing two integers such as `[2, 2]` or an object with explicit keys such as `{"rank": 2, "file": 2}`.

Note that we don't currently decide to use true algebraic notation for these coordinates: #446, so both the rank and file should remain as numbers.

HarrisonMc555:
Closes #721.

I decided to use objects, since it seems like most people were agreeing it was more explicit. I will also be using objects to encode the position for #722, which is deals with a similar issue.

### DIFF
--- a/exercises/queen-attack/canonical-data.json
+++ b/exercises/queen-attack/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "queen-attack",
-  "version": "1.0.0",
+  "version": "2.0.0",
   "comments": [
     "Testing invalid positions will vary by language. The expected",
     "value of -1 is there to indicate some sort of failure should",
@@ -18,39 +18,54 @@
           "description": "queen with a valid position",
           "property": "create",
           "queen": {
-            "position": "(2,2)"
+            "position": {
+              "row": 2,
+              "column": 2
+            }
           },
           "expected": 0
         },
         {
-          "description": "queen must have positive rank",
+          "description": "queen must have positive row",
           "property": "create",
           "queen": {
-            "position": "(-2,2)"
+            "position": {
+              "row": -2,
+              "column": 2
+            }
           },
           "expected": -1
         },
         {
-          "description": "queen must have rank on board",
+          "description": "queen must have row on board",
           "property": "create",
           "queen": {
-            "position": "(8,4)"
+            "position": {
+              "row": 8,
+              "column": 4
+            }
           },
           "expected": -1
         },
         {
-          "description": "queen must have positive file",
+          "description": "queen must have positive column",
           "property": "create",
           "queen": {
-            "position": "(2,-2)"
+            "position": {
+              "row": 2,
+              "column": -2
+            }
           },
           "expected": -1
         },
         {
-          "description": "queen must have file on board",
+          "description": "queen must have column on board",
           "property": "create",
           "queen": {
-            "position": "(4,8)"
+            "position": {
+              "row": 4,
+              "column": 8
+            }
           },
           "expected": -1
         }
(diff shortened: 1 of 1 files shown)


stkent:
Nice change! Please also bump the canonical data version as appropriate: https://github.com/exercism/problem-specifications#test-data-versioning :)

HarrisonMc555:
@stkent Thanks for the reminder, totally forgot about that! Since we're changing the type of one of the data keys I assume this is a major bump, correct?

HarrisonMc555:
Thanks, bumped major number.

exercises/queen-attack/canonical-data.json
```diff
@@ -18,39 +18,54 @@
           "description": "queen with a valid position",
           "property": "create",
           "queen": {
-            "position": "(2,2)"
+            "position": {
+              "file": 2,
+              "rank": 2
```
Insti: The [description](https://github.com/exercism/problem-specifications/blob/master/exercises/queen-attack/description.md) uses the terms `row` and `column` I think we should stick with those.

NobbZ:
It seems as if I had to revoke my request for changes, as per the version numbering documentation, we have to bump MAJOR when we break test generators, and this changes propably will:

> MAJOR changes should be expected to break even well-behaved test generators.

I previously always understood the version as a version of the exercise, to understand evolvement of exercises, but it seems as if it were only meant to check for compatibility in testgenerators, nothing more, nothing less.

exercises/queen-attack/canonical-data.json
ErikSchierboom: Agree with @Insti, we should follow what is in the description (or else change the description).

Insti: I'd argue against changing the description.

HarrisonMc555:
I'm totally fine with row/column, let's use that. Just submitted the patch. Anything else?

exercises/queen-attack/canonical-data.json
HarrisonMc555: I agree, row/column works great for me. Made the change.

exercises/queen-attack/canonical-data.json
```diff
@@ -18,39 +18,54 @@
...
-            "position": "(2,2)"
+            "position": {
+              "row": 2,
+              "column": 2
+            }
           },
           "expected": 0
         },
         {
           "description": "queen must have positive rank",
```
petertseng: this is very awkward that the description (and thus the keys in the JSON file) use "row" yet the description uses "rank". We'll want to fix that up later (or now??)

exercises/queen-attack/canonical-data.json
```diff
@@ -18,39 +18,54 @@
           "description": "queen with a valid position",
           "property": "create",
           "queen": {
-            "position": "(2,2)"
+            "position": {
+              "row": 2,
+              "column": 2
+            }
           },
           "expected": 0
```
petertseng: a little awkward that these are 0/-1 instead of just true/false. Probably worth fixing later on.

exercises/queen-attack/canonical-data.json
HarrisonMc555: I'm willing to change it now.

exercises/queen-attack/canonical-data.json
HarrisonMc555: Right now it's using 0/1 as false/true and -1 as error. It would probably be better to use the accepted `"error"` methodology, but that's probably worth its own pull request. Should one of us open an issue?

HarrisonMc555:
@petertseng I made the changes in the test case descriptions like you mentioned.
@insti Do things look good? I think I'm ready to merge, I'll go ahead and squash the commits now.

exercises/queen-attack/canonical-data.json
Insti: I agree that a separate PR for this is a good idea.

PR #936, opened 2017-10-09
robot-simulator: encode position as object instead of string

### LINKED ISSUE #722: robot-simulator: should represent positions as array or object, not string

HarrisonMc555:
Closes #722.

I decided to use objects. Some people preferred arrays and some people preferred objects, but it seems like in other problems objects were preferred because they were more explicit (see #721). I don't believe we have a policy yet on whether to use arrays or objects for things like this, but I believe that objects would be a better decision. There are some situations in which it doesn't really matter and some people would prefer arrays, but there are some situations where it really would be better to have objects. Since it sometimes is desirable to have objects and only sometimes slightly preferred to have arrays, I think objects are a better default policy.

### DIFF
--- a/exercises/robot-simulator/canonical-data.json
+++ b/exercises/robot-simulator/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "robot-simulator",
-  "version": "1.0.0",
+  "version": "2.0.0",
   "comments": [
     "Some tests have two expectations: one for the position, one for the direction",
     "Optionally, you can also test",
--- a/exercises/robot-simulator/canonical-data.json
+++ b/exercises/robot-simulator/canonical-data.json
@@ -16,11 +16,17 @@
           "description": "Robots are created with a position and direction",
           "property": "create",
           "robot": {
-            "position": "(0,0)",
+            "position": {
+              "x": 0,
+              "y": 0
+            },
             "direction": "north"
           },
           "expected": {
-            "position": "(0,0)",
+            "position": {
+              "x": 0,
+              "y": 0
+            },
             "direction": "north"
           }
         },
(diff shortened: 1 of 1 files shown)


stkent:
Nice change! Please also bump the canonical data version as appropriate: https://github.com/exercism/problem-specifications#test-data-versioning :)

HarrisonMc555:
Thanks, bumped major number.

NobbZ:
Since I can't alter my review… Of course I like the overal change, since it is much more clear how the coordinates have to be read.

Also I do clearly prefer objects over arrays/lists here for the representation of the testdata, since `{"x": 1, "y": 2}` is so much more obvoius than `[1, 2]`. To be honest, there were not much of a semantic difference between `[1, 2]` ans `"(1, 2)"`…

Insti:
> So version should in my opinion only get bumped to 1.0.1.

I think I agree with this.

Insti:
@NobbZ do you want to make an issue for this discussion? I think it's worth having but we seem to be having this discussion across 3 different PRs. It would be good to have it all in one place we can tag as `policy` at some point.

NobbZ:
Yeah, I was going to do so, but first I had to remove false negatives to not block the process of merging.

HarrisonMc555:
Hey sorry I missed the party. I think I was supposed to squash but didn't get around to it in time. Thanks everyone!

Issue #673, opened 2017-03-09
Semantic Versioning of canonical data.


rbasso:
As questioned in #671, we still don't have defined how to version the `canonical-data.json` files in the new schema. I think we should strive to get something like [SemVer](http://semver.org/).

---

**Edit:** Below we well keep updated the proposed semantic versioning resulting from the discussion. 

Current proposal
-------------------------------
Adapting from SemVer 2.0.0 ...

> Given a version number MAJOR.MINOR.PATCH, increment the:
>
> - MAJOR version when you make incompatible API changes

Only changes in the test suite in a fundamentally incompatible way should
change the MAJOR version:

  - A new "property" (test type).
  - Renaming a "property".
  - Insertion, deletion or renaming of keys in the test data object.
  - Changing the type of one of the test data keys.

MAJOR changes would break even well-behaved test generators, that are not affected by test grouping/nesting.

> - MINOR version when you add functionality in a backwards-compatible manner, and

Non-breaking changes that alter the *meaning of the test suite*, making previously
passing solutions possibly fail, or failing solutions fail at a different spot, should bump the MINOR version:

  - Adding or deleting test cases.
  - Changing the test cases inputs and/or outputs.
  - Changing the test cases ordering.

MINOR changes would never break well-designed test generators, because the test-generation logic remains exactly the same.

> - PATCH version when you make backwards-compatible bug fixes.

Anything else that do not change the meaning of the tests should bump the
PATCH version:

  - Regrouping/"Renesting" test cases without changing test cases ordering.
  - Changing descriptions or comments.
  - Changing keys' ordering or formatting (would result in an equivalent JSON file).

behrtam:
_Maybe .._

Given a version number MAJOR.MINOR.PATCH, increment the:
**MAJOR** version when you add new test cases or change input/output values,
**MINOR** version when you change descriptions or other descriptive fields, and
**PATCH** version when you fix small typos or other trivial things.

rbasso:
The **MAJOR* number could be reserved for changes in the structure of the tests, things that would probably break any well design test generator.

behrtam:
Good point. Than we move the new test cases and input/output changes to _MINOR_ and put the change (rewrite/typos/etc) of descriptive fields into *PATCH* .

rbasso:
I'm writing a proposal so that we can discuss the details. There are a lot of corner cases...

rbasso:
[...]

PR #937, opened 2017-10-09
crypto-square: Improve test data

nholden:
This cleans up a few things about crypto-square's canonical data that I thought were worth addressing before I generated new tests in the Ruby repo to address https://github.com/exercism/ruby/issues/422.

- Removes unnecessary sub-groupings of test cases ([related discussion](https://github.com/exercism/problem-specifications/pull/922#discussion_r142604966))
- Moves the blank string test case to the top since it feels like the simplest one to get passing using TDD
- Adds a new test case

### DIFF
--- a/exercises/crypto-square/canonical-data.json
+++ b/exercises/crypto-square/canonical-data.json
@@ -1,52 +1,42 @@
 {
   "exercise": "crypto-square",
-  "version": "3.0.0",
+  "version": "3.1.0",
   "cases": [
     {
-      "description": "the spaces and punctuation are removed from the English text and the message is downcased",
-      "cases": [
-        {
-          "description": "Lowercase",
-          "property": "ciphertext",
-          "plaintext": "A",
-          "expected": "a"
-        },
-        {
-          "description": "Remove spaces",
-          "property": "ciphertext",
-          "plaintext": "  b ",
-          "expected": "b"
-        },
-        {
-          "description": "Remove punctuation",
-          "property": "ciphertext",
-          "plaintext": "@1,%!",
-          "expected": "1"
-        }
-      ]
+      "description": "Lowercase",
+      "property": "ciphertext",
+      "plaintext": "A",
+      "expected": "a"
     },
     {
-      "description": "Output the encoded text in chunks.  Phrases that fill perfect squares `(r X r)` should be output in `r`-length chunks separated by spaces.  Imperfect squares will have `n` empty spaces.  Those spaces should be distributed evenly across the last `n` rows.",
-      "cases": [
-        {
-          "description": "empty plaintext results in an empty ciphertext",
-          "property": "ciphertext",
-          "plaintext": "",
-          "expected": ""
-        },
-        {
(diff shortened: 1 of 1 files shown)


exercises/crypto-square/canonical-data.json
```diff
@@ -1,52 +1,48 @@
...
+      "property": "ciphertext",
+      "plaintext": "This is fun!",
+      "expected": "tsf hiu isn"
+    },
+    {
+      "description": "8 character plaintext results in 3 chunks, the last one with a trailing space",
+      "property": "ciphertext",
+      "plaintext": "Chill out.",
+      "expected": "clu hlt io "
+    },
```
nholden: This is the new test case, which I borrowed from [the source of the problem](http://users.csc.calpoly.edu/~jdalbey/103/Projects/ProgrammingPractice.html). I think it's a useful bridge between the previous test case, which tests a perfect rectangle, and the next test case, which like this one tests an imperfect rectangle but is quite a bit longer.

Insti:
I'm not sure the version needs to go to `4.0`, maybe `3.1` is enough?
Does the removal of test groupings makes a semver relevant difference?

See also: https://github.com/exercism/problem-specifications/pull/936#pullrequestreview-68097414

NobbZ:
I now took a look into the version numbering documentation…

This sentence:

> MAJOR changes should be expected to break even well-behaved test generators.

Makes this changes (as well as the other two I've requested changes in) worth a major bump.

But what exactly is “well behaved”? To be honest… I do expect about every change, that goes beyond renaming or adding/removing a test to break the generator.

But especially adding and removing tests, are often responsible for a major bump as well, since they can change the exercises path, the way to solve it, even the resulting program a lot…

So I conclude: just omit minor and patch, do only major in the future…

But that is probably better discussed elsewere…

nholden:
Thanks for the quick review!

> I'm not sure the version needs to go to 4.0, maybe 3.1 is enough?
Does the removal of test groupings makes a semver relevant difference?

Good question. After I read [the section in the README about minor version changes](https://github.com/exercism/problem-specifications#minor-version-changes), I thought your point seemed like a good one and revised the original commit to make it a minor version change.

However, I think @NobbZ has a good point as well:

> But what exactly is “well behaved”? To be honest… I do expect about every change, that goes beyond renaming or adding/removing a test to break the generator.

I'm not familiar with how generators work across all languages, so maybe with the removal of test groupings, it would be safer to make this a major version change.

Does anyone have strong feelings one way or the other?

NobbZ:
Even worse…

> Regrouping/"Renesting" test cases without changing test case ordering.

But I expect this to break the generators, since the cases aren't anymore were the generator expects them…

nholden:
> > [MINOR when] Regrouping/"Renesting" test cases without changing test case ordering.
>
> But I expect this to break the generators, since the cases aren't anymore were the generator expects them…

I think the bit you quoted makes it clear that this should be a minor version bump given the current policies. If regrouping/renesting test cases does break generators, I think it's worth re-opening https://github.com/exercism/problem-specifications/issues/673 or starting a new one.

PR #939, opened 2017-10-10
complex-numbers: add detailed description

### LINKED ISSUE #933: complex-numbers: add detailed description
In discussion under PR exercism/ruby#733 a problem with README has been suggested.

The description does not explicitly say what needs to be done (what operations need to be implemented).
I also believe the README should contain definitions of conjugate and exponentiation of a complex number. 

This issue is common to all tracks so I think updating description.md in problem-specifications first would be the best option.

m1kal:
Adding detailed description as some users might not be familiar with complex numbers.
New content includes definitions of operations on complex numbers and lists the operations that need to be implemented.

Closes #933.

### DIFF
--- a/exercises/complex-numbers/description.md
+++ b/exercises/complex-numbers/description.md
@@ -1,3 +1,30 @@
 A complex number is a number in the form `a + b * i` where `a` and `b` are real and `i` satisfies `i^2 = -1`.
 
+`a` is called the real part and `b` is called the imaginary part of `z`.
+The conjugate of the number `a + b * i` is the number `a - b * i`.
+The absolute value of a complex number `z = a + b * i` is a real number `|z| = sqrt(a^2 + b^2)`. The square of the absolute value `|z|^2` is the result of multiplication of `z` by its complex conjugate.
+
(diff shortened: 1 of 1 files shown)


exercises/complex-numbers/description.md
```diff
@@ -1,3 +1,30 @@
 A complex number is a number in the form `a + b * i` where `a` and `b` are real and `i` satisfies `i^2 = -1`.
 
+`a` is called the real part and `b` is called the imaginary part of `z`.
+The conjugate of the number `a + b * i` is the number `a - b * i`.
+The absolute value of a complex number `z = a + b * i` is a real number `|z| = sqrt(a^2 + b^2)`. The square of the absolute value `|z|^2` is the result of multiplication of `z` by its complex conjugate.
+
+The sum / difference of two complex numbers involves adding / subtracting their real and imaginary parts separately:
```
ErikSchierboom: Why the spaces between `sum / difference` and `adding / subtracting`? I think I would prefer it without the spaces, to better visually represent that they belong together.

m1kal: I agree. I removed the spaces.

ErikSchierboom:
Thanks a lot @m1kal !

Issue #925, opened 2017-10-05
Why have null as input in pascals-triangle exercise


ErikSchierboom:
In the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/pascals-triangle/canonical-data.json) for the pascals-triangle exercise, the last test case checks for null/missing input: https://github.com/exercism/problem-specifications/blob/master/exercises/pascals-triangle/canonical-data.json#L50 However, to me this does not make sense at all, as you have to specify a number to calculate pascal's triangle. Having no input does not make any sense, unless you want to calculate a default, but the test case for the null input states that passing null means a failure should occur. 

My suggestion is to remove this null error case, as it doesn't really add anything (there alreadt is a negative number case that will enforce error handling).

snahor:
It doesn't make any sense either to return -1.

ErikSchierboom:
@snahor Returning -1 is just a general way of saying that an error has occured, see the [header](https://github.com/exercism/problem-specifications/blob/master/exercises/pascals-triangle/canonical-data.json#L7): 

> An expectation of -1 indicates some sort of failure should occur

snahor:
I'm aware of it, but I think that shouldn't be explicit. Some exercises are already using `{"error": "..."}`.

petertseng:
I would refer to #902 here, though I'm not convinced a conclusion arose. Nevertheless, I do not think testing `null` makes a lot of sense.

In most languages that have static type checking, even if they have nullable references, integers are not reference types, so if the function under test is defined to take an integer, null is not a possible input.

For languages without static type checking, `null` would be a possible input, but I also generally do not see functions that expect to take integers explicitly check for `null` (or language equivalent); let me know if I'm wrong. So I don't think the tests need to make students write it.

So, removing `null` seems fine.

Separately, indeed, that `"expected": -1` could change to the error object.

ErikSchierboom:
I've created a [PR](#926) to fix this issue. I've not updated the expected return value, as that should be a different PR I think.

Issue #902, opened 2017-09-16
Discussion: Testing 'invalid' input.


Insti:
Background: [nucleotide-count description](https://github.com/exercism/problem-specifications/blob/master/exercises/nucleotide-count/description.md)

"Given a DNA string, compute how many times each nucleotide occurs in the string."

Part of the PR "nucleotide-count: refactoring tests (discussion)"  (#895) proposed adding tests for various "error" input cases, `null` input, strings that contain non ascii characters and other tests that involve strings containing various other non-DNA characters.

In the discussion that arose around that,  @Vankog wrote: https://github.com/exercism/problem-specifications/pull/895#discussion_r139141823

> But as far as I experienced it, test suits should check for edge cases. null checks are especially important (at least in almost all languages).
In every educational piece of material the advice is not to only check for usual inputs, but for all possible (equivalent kinds of) inputs. null checks, empty checks and others are always on top of the example lists given. Just take any arbitrary course about QA and this will be part of it.

> I think this is especially important, because we are in an educational context. The users should learn from the getgo what it means to write (arguably good) tests (first), because it is still a major problem, even upon seasoned programmers.
From my own business experience I know that new programmers are completely lost when it comes to writing tests. But also experienced programmers fall in the traps of "I'll do it later", "This is a trivial method", "I can't test this", "There are too many dependencies" etc. Just take any talk or tutorial about writing tests or even TDD and you will see the same old counter-arguments or questions again and again from the comments or the audience. This has a reason: Writing good tests is hard! Very hard to be exact and it is a skill that needs exercise. And where better to begin than right from the start?

I agree that writing good tests is hard. One of the traps that is often fallen in to, is testing too much, and the skill is in knowing where to draw the line between what is important to test and what is not important to test.

Part of the challenge of programming is designing clean and self contained abstractions so that you have to worry about as few things as possible at any one time.

The exercises on Exercism help by taking away a lot of the ambiguity and uncertainty by providing small, self contained and well defined exercises, for which the 'interesting' part of the problem is implementing the algorithm to solve a specific problem, rather than working out what the problem is and splitting it up.

In this case, where the problem is "counting nucleotides in a DNA string" it is OK to assume that all the input to the function will be valid DNA strings, and testing against things that are NOT valid DNA strings is inappropriate.

Think of it at it as there was another step prior to every problem that insures that the input is valid. In this case:
```
Unvalidated input -> DNA Parser -> nucleotide-count
```

[...]

PR #940, opened 2017-10-10
binary-search-tree: implement canonical-data.json

### LINKED ISSUE #555: binary-search-tree: Implement canonical-data.json
exercises in sync.

The Binary Search Tree exercise can be found in the [./exercises/binary-search-tree/][exercise-dir] directory.

See http://exercism.io/contribute/canonical-data/binary-search-tree for the up-to-date list of
language tracks that have an implementation of the Binary Search Tree exercise.

[exercise-dir]: https://github.com/exercism/x-common/tree/master/exercises/binary-search-tree

kabiir:
Closes #555 

Edit to add: [Description](https://github.com/exercism/problem-specifications/blob/master/exercises/binary-search-tree/description.md)

### DIFF
--- a/exercises/binary-search-tree/canonical-data.json
+++ b/exercises/binary-search-tree/canonical-data.json
@@ -2,67 +2,153 @@
     "exercise": "binary-search-tree",
     "version": "1.0.0",
     "comments": [
-        "For simplicity only positive integer is considered as valid data,",
-        "null should be replaced by the respective language equivalent or -1",
-        "if the language does not have any.",
-
-        "The purpose of ths exercise is to create a data structure which",
-        "makes sure the data is always sorted, hence the exercise should",
-        "not make use of any sort() functions or sorting algorithms."
+        "Each test case assumes an empty/new tree.",
+        "The key 'tree_data' represents an array of numbers the data should be ",
+        "inserted/added to the tree as it appears in the array from left to right.",
+        "e.g. tree_data: ['2', '1', '3', '6', '7', '5']",
+        "Insert 2. Insert 1. Insert 3. Insert 6, so on...",
+        "The key 'expected' represents tree state as JSON object of schema :",
+        "{",
+        "    'title':'nodeObject',",
+        "    'type':'object',",
+        "    'properties':{",
+        "        'data':{",
+        "            'type':'string'",
+        "        },",
+        "        'left':{",
+        "            'type':'nodeObject'",
+        "        },",
+        "        'right':{",
+        "            'type':'nodeObject'",
+        "        }",
+        "    },",
+        "    'required':['data', 'left', 'right']",
+        "}"
     ],
     "cases": [{
-            "description": "insert invalid/null data into the tree",
-            "property": "insert",
-            "value": null,
-            "expected": false
+            "description": "data is retained",
+            "property": "data",
+            "tree_data": ["4"],
+            "expected": {
+                "data": "4",
+                "left": null,
+                "right": null
+            }
         },
         {
-            "description": "insert a number into the tree",
-            "property": "insert",
-            "value": 6,
-            "expected": true
+            "description": "insert data at proper node",
+            "cases": [{
+                    "description": "smaller number at left node",
+                    "property": "data",
+                    "tree_data": ["4", "2"],
+                    "expected": {
+                        "data": "4",
+                        "left": {
+                            "data": "2",
+                            "left": null,
+                            "right": null
+                        },
+                        "right": null
+                    }
+                },
+                {
+                    "description": "same number at left node",
+                    "property": "data",
+                    "node": "root->left",
+                    "tree_data": ["4", "4"],
+                    "expected": {
+                        "data": "4",
+                        "left": {
+                            "data": "4",
+                            "left": null,
+                            "right": null
+                        },
+                        "right": null
+                    }
+                },
+                {
+                    "description": "greater number at right node",
+                    "property": "data",
+                    "tree_data": ["4", "5"],
+                    "expected": {
+                        "data": "4",
+                        "left": null,
+                        "right": {
+                            "data": "5",
+                            "left": null,
+                            "right": null
+                        }
+                    }
+                }
+            ]
         },
         {
-            "description": "insert a number that already exist",
-            "property": "insert",
-            "value": 6,
-            "expected": false
+            "description": "can create complex tree",
+            "property": "data",
+            "tree_data": ["4", "2", "6", "1", "3", "5", "7"],
+            "expected": {
+                "data": "4",
+                "left": {
+                    "data": "2",
+                    "left": {
+                        "data": "1",
+                        "left": null,
+                        "right": null
+                    },
+                    "right": {
+                        "data": "3",
+                        "left": null,
+                        "right": null
+                    }
+                },
+                "right": {
+                    "data": "6",
+                    "left": {
+                        "data": "5",
+                        "left": null,
+                        "right": null
+                    },
+                    "right": {
+                        "data": "7",
+                        "left": null,
+                        "right": null
+                    }
+                }
+            }
         },
         {
-            "description": "search invalid/null data",
-            "property": "search",
-            "value": null,
-            "expected": false
-        },
-        {
-            "description": "search a number that exists",
-            "property": "search",
-            "value": 6,
-            "expected": true
-        },
-        {
-            "description": "search a number that does not exist",
-            "property": "search",
-            "value": 11,
-            "expected": false
-        },
-        {
-            "description": "traverse the data in ascending order",
-            "property": "ascendingOrder",
-            "comments": [
-                "Assuming the data is entered in this order :",
-                "6 -> 1 -> 3 -> 2 -> 5"
-            ],
-            "expected": [1, 2, 3, 5, 6]
-        },
-        {
-            "description": "traverse the data in descending order",
-            "property": "descendingOrder",
-            "comments": [
-                "Assuming the data is entered in this order :",
-                "6 -> 1 -> 3 -> 2 -> 5"
-            ],
-            "expected": [6, 5, 3, 2, 1]
+            "description": "can sort data",
+            "cases": [{
+                    "description": "can sort single number",
+                    "property": "sortedData",
+                    "tree_data": ["2"],
+                    "expected": ["2"]
+                },
+                {
+                    "description": "can sort if second number is smaller than first",
+                    "property": "sortedData",
+                    "tree_data": ["2", "1"],
+                    "expected": ["1", "2"]
+                },
+                {
+                    "description": "can sort if second number is same as first",
+                    "property": "sortedData",
+                    "tree_data": ["2", "2"],
+                    "expected": ["2", "2"]
+                },
+                {
+                    "description": "can sort if second number is greater than first",
+                    "property": "sortedData",
+                    "tree_data": ["2", "3"],
+                    "expected": ["2", "3"]
+                },
+                {
+                    "description": "can sort complex tree",
+                    "property": "sortedData",
+                    "tree_data": ["2", "1", "3", "6", "7", "5"],
+                    "expected": ["1", "2", "3", "5", "6", "7"]
+                }
+            ]
         }
     ]
 }
--- a/exercises/binary-search-tree/canonical-data.json
+++ b/exercises/binary-search-tree/canonical-data.json
@@ -62,7 +62,6 @@
                 {
                     "description": "same number at left node",
                     "property": "data",
-                    "node": "root->left",
                     "input": {
                         "tree_data": ["4", "4"]
                     },
--- a/exercises/binary-search-tree/canonical-data.json
+++ b/exercises/binary-search-tree/canonical-data.json
@@ -1,0 +1,68 @@
+{
+    "exercise": "binary-search-tree",
+    "version": "1.0.0",
+    "comments": [
+        "For simplicity only positive integer is considered as valid data,",
+        "null should be replaced by the respective language equivalent or -1",
+        "if the language does not have any.",
+
+        "The purpose of ths exercise is to create a data structure which",
+        "makes sure the data is always sorted, hence the exercise should",
+        "not make use of any sort() functions or sorting algorithms."
+    ],
+    "cases": [{
+            "description": "insert invalid/null data into the tree",
+            "property": "insert",
+            "value": null,
+            "expected": false
+        },
+        {
+            "description": "insert a number into the tree",
+            "property": "insert",
+            "value": 6,
+            "expected": true
+        },
+        {
+            "description": "insert a number that already exist",
+            "property": "insert",
+            "value": 6,
+            "expected": false
+        },
+        {
+            "description": "search invalid/null data",
+            "property": "search",
+            "value": null,
+            "expected": false
+        },
+        {
+            "description": "search a number that exists",
+            "property": "search",
+            "value": 6,
+            "expected": true
+        },
+        {
+            "description": "search a number that does not exist",
+            "property": "search",
+            "value": 11,
+            "expected": false
+        },
+        {
+            "description": "traverse the data in ascending order",
+            "property": "ascendingOrder",
+            "comments": [
+                "Assuming the data is entered in this order :",
+                "6 -> 1 -> 3 -> 2 -> 5"
+            ],
+            "expected": [1, 2, 3, 5, 6]
+        },
+        {
+            "description": "traverse the data in descending order",
+            "property": "descendingOrder",
+            "comments": [
+                "Assuming the data is entered in this order :",
+                "6 -> 1 -> 3 -> 2 -> 5"
+            ],
+            "expected": [6, 5, 3, 2, 1]
+        }
+    ]
+}
--- a/exercises/binary-search-tree/canonical-data.json
+++ b/exercises/binary-search-tree/canonical-data.json
@@ -3,10 +3,14 @@
     "version": "1.0.0",
     "comments": [
         "Each test case assumes an empty/new tree.",
-        "The key 'tree_data' represents an array of numbers the data should be ",
+        "As per exercism/problem-specifications#996 key 'tree_data' counts as an input",
+        "to test generators.",
+        "The key 'tree_data' represents an array of numbers for which the data should be ",
         "inserted/added to the tree as it appears in the array from left to right.",
         "e.g. tree_data: ['2', '1', '3', '6', '7', '5']",
         "Insert 2. Insert 1. Insert 3. Insert 6, so on...",
+        "This canonical-data does not restrict the data type of array elements to either",
+        "strings or integers.",
         "The key 'expected' represents tree state as JSON object of schema :",
         "{",
         "    'title':'nodeObject',",
@@ -28,7 +32,9 @@
     "cases": [{
             "description": "data is retained",
             "property": "data",
-            "tree_data": ["4"],
+            "input": {
+                "tree_data": ["4"]
+            },
             "expected": {
                 "data": "4",
                 "left": null,
@@ -40,7 +46,9 @@
             "cases": [{
                     "description": "smaller number at left node",
                     "property": "data",
-                    "tree_data": ["4", "2"],
+                    "input": {
+                        "tree_data": ["4", "2"]
+                    },
                     "expected": {
                         "data": "4",
                         "left": {
@@ -55,7 +63,9 @@
                     "description": "same number at left node",
                     "property": "data",
                     "node": "root->left",
-                    "tree_data": ["4", "4"],
+                    "input": {
+                        "tree_data": ["4", "4"]
+                    },
                     "expected": {
                         "data": "4",
                         "left": {
@@ -69,7 +79,9 @@
                 {
                     "description": "greater number at right node",
                     "property": "data",
-                    "tree_data": ["4", "5"],
+                    "input": {
+                        "tree_data": ["4", "5"]
+                    },
                     "expected": {
                         "data": "4",
                         "left": null,
@@ -85,7 +97,9 @@
         {
             "description": "can create complex tree",
             "property": "data",
-            "tree_data": ["4", "2", "6", "1", "3", "5", "7"],
+            "input": {
+                "tree_data": ["4", "2", "6", "1", "3", "5", "7"]
+            },
             "expected": {
                 "data": "4",
                 "left": {
@@ -121,31 +135,41 @@
             "cases": [{
                     "description": "can sort single number",
                     "property": "sortedData",
-                    "tree_data": ["2"],
+                    "input": {
+                        "tree_data": ["2"]
+                    },
                     "expected": ["2"]
                 },
                 {
                     "description": "can sort if second number is smaller than first",
                     "property": "sortedData",
-                    "tree_data": ["2", "1"],
+                    "input": {
+                        "tree_data": ["2", "1"]
+                    },
                     "expected": ["1", "2"]
                 },
                 {
                     "description": "can sort if second number is same as first",
                     "property": "sortedData",
-                    "tree_data": ["2", "2"],
+                    "input": {
+                        "tree_data": ["2", "2"]
+                    },
                     "expected": ["2", "2"]
                 },
                 {
                     "description": "can sort if second number is greater than first",
                     "property": "sortedData",
-                    "tree_data": ["2", "3"],
+                    "input": {
+                        "tree_data": ["2", "3"]
+                    },
                     "expected": ["2", "3"]
                 },
                 {
                     "description": "can sort complex tree",
                     "property": "sortedData",
-                    "tree_data": ["2", "1", "3", "6", "7", "5"],
+                    "input": {
+                        "tree_data": ["2", "1", "3", "6", "7", "5"]
+                    },
                     "expected": ["1", "2", "3", "5", "6", "7"]
                 }
             ]
--- a/exercises/binary-search-tree/canonical-data.json
+++ b/exercises/binary-search-tree/canonical-data.json
@@ -3,11 +3,11 @@
     "version": "1.0.0",
     "comments": [
         "Each test case assumes an empty/new tree.",
-        "As per exercism/problem-specifications#996 key 'tree_data' counts as an input",
+        "As per exercism/problem-specifications#996 key 'treeData' counts as an input",
         "to test generators.",
-        "The key 'tree_data' represents an array of numbers for which the data should be ",
+        "The key 'treeData' represents an array of numbers for which the data should be ",
         "inserted/added to the tree as it appears in the array from left to right.",
-        "e.g. tree_data: ['2', '1', '3', '6', '7', '5']",
+        "e.g. treeData: ['2', '1', '3', '6', '7', '5']",
         "Insert 2. Insert 1. Insert 3. Insert 6, so on...",
         "This canonical-data does not restrict the data type of array elements to either",
         "strings or integers.",
@@ -33,7 +33,7 @@
             "description": "data is retained",
             "property": "data",
             "input": {
-                "tree_data": ["4"]
+                "treeData": ["4"]
             },
             "expected": {
                 "data": "4",
@@ -47,7 +47,7 @@
                     "description": "smaller number at left node",
                     "property": "data",
                     "input": {
-                        "tree_data": ["4", "2"]
+                        "treeData": ["4", "2"]
                     },
                     "expected": {
                         "data": "4",
@@ -63,7 +63,7 @@
                     "description": "same number at left node",
                     "property": "data",
                     "input": {
-                        "tree_data": ["4", "4"]
+                        "treeData": ["4", "4"]
                     },
                     "expected": {
                         "data": "4",
@@ -79,7 +79,7 @@
                     "description": "greater number at right node",
                     "property": "data",
                     "input": {
-                        "tree_data": ["4", "5"]
+                        "treeData": ["4", "5"]
                     },
                     "expected": {
                         "data": "4",
@@ -97,7 +97,7 @@
             "description": "can create complex tree",
             "property": "data",
             "input": {
-                "tree_data": ["4", "2", "6", "1", "3", "5", "7"]
+                "treeData": ["4", "2", "6", "1", "3", "5", "7"]
             },
             "expected": {
                 "data": "4",
@@ -135,7 +135,7 @@
                     "description": "can sort single number",
                     "property": "sortedData",
                     "input": {
-                        "tree_data": ["2"]
+                        "treeData": ["2"]
                     },
                     "expected": ["2"]
                 },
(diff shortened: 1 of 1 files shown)


kabiir:
Since I don't know who to ask for review @kytrinyx I'll leave that to you.

exercises/binary-search-tree/canonical-data.json
```diff
@@ -0,0 +1,68 @@
...
+        "if the language does not have any.",
+
+        "The purpose of ths exercise is to create a data structure which",
+        "makes sure the data is always sorted, hence the exercise should",
+        "not make use of any sort() functions or sorting algorithms."
+    ],
+    "cases": [{
+            "description": "insert invalid/null data into the tree",
+            "property": "insert",
+            "value": null,
```
petertseng: Similar reasoning as #925: I don't think it is good to have this case.

> In most languages that have static type checking, even if they have nullable references, integers are not reference types, so if the function under test is defined to take an integer, `null` is not a possible input.
> 
> For languages without static type checking, `null` would be a possible input, but I also generally do not see functions that expect to take integers explicitly check for `null` (or language equivalent); let me know if I'm wrong. So I don't think the tests need to make students write it.

exercises/binary-search-tree/canonical-data.json
```diff
@@ -0,0 +1,68 @@
...
+        {
+            "description": "insert a number that already exist",
+            "property": "insert",
+            "value": 6,
+            "expected": false
+        },
+        {
+            "description": "search invalid/null data",
+            "property": "search",
+            "value": null,
```
petertseng: Similar reasoning as #925: I don't think it is good to have this case.

> In most languages that have static type checking, even if they have nullable references, integers are not reference types, so if the function under test is defined to take an integer, `null` is not a possible input.
> 
> For languages without static type checking, `null` would be a possible input, but I also generally do not see functions that expect to take integers explicitly check for `null` (or language equivalent); let me know if I'm wrong. So I don't think the tests need to make students write it.

exercises/binary-search-tree/canonical-data.json
```diff
@@ -0,0 +1,68 @@
...
+        {
+            "description": "insert a number into the tree",
+            "property": "insert",
+            "value": 6,
+            "expected": true
+        },
+        {
+            "description": "insert a number that already exist",
+            "property": "insert",
+            "value": 6,
```
petertseng: I didn't see how we can understand that the number already exists, using the information provided in this case. If we suppose that we are inserting into a tree that already had `6` inserted into it at some point, that needs to be specified in some way here. Otherwise, compare this case with the case immediately above it: the only difference is the `description` and `expected` values, so there doesn't seem to be any apparent reason why one would expect true and the other would expect false.

exercises/binary-search-tree/canonical-data.json
```diff
@@ -0,0 +1,68 @@
...
+            "description": "search invalid/null data",
+            "property": "search",
+            "value": null,
+            "expected": false
+        },
+        {
+            "description": "search a number that exists",
+            "property": "search",
+            "value": 6,
+            "expected": true
```
petertseng: How do we know that 6 exists in the tree? What tree is this search being performed on?

exercises/binary-search-tree/canonical-data.json
```diff
@@ -0,0 +1,68 @@
...
+        },
+        {
+            "description": "search a number that does not exist",
+            "property": "search",
+            "value": 11,
+            "expected": false
```
petertseng: How do we know that 11 does exist in the tree? What tree is this search being performed on?

exercises/binary-search-tree/canonical-data.json
```diff
@@ -0,0 +1,68 @@
...
+        },
+        {
+            "description": "traverse the data in ascending order",
+            "property": "ascendingOrder",
+            "comments": [
+                "Assuming the data is entered in this order :",
+                "6 -> 1 -> 3 -> 2 -> 5"
```
petertseng: it looks like these values should be an array of numbers, rather than forcing people to parse this string, splitting by `->` to find the numbers, etc

exercises/binary-search-tree/canonical-data.json
```diff
@@ -0,0 +1,68 @@
...
+            ],
+            "expected": [1, 2, 3, 5, 6]
+        },
+        {
+            "description": "traverse the data in descending order",
+            "property": "descendingOrder",
+            "comments": [
+                "Assuming the data is entered in this order :",
+                "6 -> 1 -> 3 -> 2 -> 5"
```
petertseng: it looks like these values should be an array of numbers, rather than forcing people to parse this string, splitting by `->` to find the numbers, etc

exercises/binary-search-tree/canonical-data.json
petertseng: EBWODP:

> 11 does exist

I meant "11 does not exist"

exercises/binary-search-tree/canonical-data.json
kabiir: I thought about it but i wasn't sure if i should add a new key called "state" which represents the data entered in the tree at that case in that specific order. This new key "state" will also prove useful in previous cases where its not clear what data the tree currently has.

exercises/binary-search-tree/canonical-data.json
```diff
@@ -0,0 +1,68 @@
+{
+    "exercise": "binary-search-tree",
+    "version": "1.0.0",
+    "comments": [
+        "For simplicity only positive integer is considered as valid data,",
+        "null should be replaced by the respective language equivalent or -1",
+        "if the language does not have any.",
+
+        "The purpose of ths exercise is to create a data structure which",
```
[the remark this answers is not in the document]
kabiir: Well as i understand this exercise from [description.md](https://github.com/exercism/problem-specifications/blob/master/exercises/binary-search-tree/description.md) the purpose is to create a data structure which represent sorted data and specifically binary search tree. Usage of any sort() function or algorithm will defeat its purpose don't you think?

exercises/binary-search-tree/canonical-data.json
kabiir: As I stated in the comments : if the specific track does not have any equivalent  to `null` they should use `-1`. The purpose of this case is not limited to `null` but actually invalid data. I knew about #925 when I was writing this case and that's why I added the above comment.

exercises/binary-search-tree/canonical-data.json
petertseng: I think there has to be such a key. You could call it `previously_inserted_values` instead of `state` if needing a more descriptive name.

kabiir: Um. maybe `current_tree_data` because then i can also have this key in all cases and make all cases consistent.

exercises/binary-search-tree/canonical-data.json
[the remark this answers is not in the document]
petertseng: > Usage of any sort() function or algorithm will defeat its purpose don't you think?

Agree.

I don't think it is likely that anyone will think to use a sorting function, because they should already be at the point where they're inserting nodes into a tree.

Therefore, I don't think it is useful to have this text here, and I will find it confusing becuase I will read it and think "why are you telling me this?"

It could be that I am biased and this text does help those who see it from a different point of view.

exercises/binary-search-tree/canonical-data.json
petertseng: > if the specific track does not have any equivalent to null they should use -1. The purpose of this case is not limited to null but actually `invalid` data

I don't understand why -1 is invalid for any reason other than "the comments of this file say so".

I think it is a strange restriction to only allow positive numbers. It seems like the restriction is there for the sole purpose of classifying some numbers to be invalid, allowing there to be a test for inserting invalid values. If that is the sole purpose of the restriction, I would think it just makes more sense to not have the restriction and not have this test.

Further, if negative values are to be deemed invalid values for the tree being tested, that seems like information that is very important to be added in description.md, where students will see it, not canonical-data.json (where students will not see it!).

exercises/binary-search-tree/canonical-data.json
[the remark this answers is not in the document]
kabiir: - It's possible that in future exercism adds a track in which the obvious choice would be to use the sort() function or a sorting  algorithm. In this case this exercise might become invalid for that track.
- It is also possible that the person contributing the exercise is new and binary search tree is a new concept to them.

exercises/binary-search-tree/canonical-data.json
kabiir: Well the exercise is supposed to teach how to implement a data structure. And a data structure should know the difference between valid and invalid data before inserting it. Now some languages do not have static type checking whereas some don't have equivalent of null.  But we still need to draw a line between two. And I agree the description.md might need an update.

kabiir: I think this case should exist, because if we draw a line between valid and invalid data and specify in the description we should also have a test case to check against it.

petertseng: > And a data structure should know the difference between valid and invalid data before inserting it.

I expect that for a BST, by contract, the criterion for validity is a standard "are two elements comparable against each other?". I therefore also expect that this is the criterion that the tests will use, rather than a different standard ("is the number positive?", etc.)

So then, in a language with static type checking, this would get enforced by the type system. In a language without one, now it gets interesting. We could consider a test of adding in "hello" to a tree that already contains a 6, for example. This would fail. However, is it the student's responsibility to write the code that checks for that? The failure might happen without the student having to write any additional code. If it would, then there is no value in having this test. 

> the exercise is supposed to teach how to implement a data structure.

This makes it useful to keep the standard criterion for validity. If the exercise adds in an additional criterion, students may wonder why it exists and/or whether they have to add such an additional restriction in any future BSTs they write.

kabiir: So I guess the case for invalid data should exist but anything other than an integer should be considered as invalid.

exercises/binary-search-tree/canonical-data.json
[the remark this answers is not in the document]
NobbZ: Therefore current implementations of the testsuite do test for usually internal/private functions of such a datastructure.

They ask for the left or right subtree of the root node and assert if its value is correct given the insertion order.

The do roughly the following:

```
root = empty |> insert(1) |> insert(3) |> insert(2)
three = left_of(root)
two = root |> left_of |> right_of
assert value_of(root) = 1
assert value_of(three) = 3
assert value_of(two= 2
```

So not only testing if the tree contains the values, but also if it is built correctly is not possible by just using a `sort` function, since original order of insertion dictates the structure of the tree.

So I do not think that bullet 1 is a concern if you are testing for internals in this canonical data. If you do not, please add those tests.

----

A person that is new to the BST, will see the README.md which explains the exercise to solve. The comments in the canonical data are meant as a guidance/help for the implementor of a testsuite/a generator, not to describe the actual exercise.

----

Because of this two reasons, I am with @petertseng and think this part of the comment should be removed,

exercises/binary-search-tree/canonical-data.json
petertseng: > anything other than an integer should be considered as invalid

Right.

> the case for invalid data should exist

Only if it is useful. I haven't been convinced it is useful because in most cases the student doesn't need to write any code to pass the test. Okay, I can name at least one language where `6 < "hello"` is not a type error (whether that be at run time or compile time), so in that language it could be useful.

I can't tell whether https://github.com/exercism/problem-specifications/issues/902 is telling us "yes, put such cases in the canonical-data.json and let languages that don't need the test ignore it" or "no, don't put such cases in and let languages that need the tests add them". I am thinking the number of languages that benefit from the case is lower than the number of languages that don't, if that helps decide.

I still say no. Remember that it's not necessary to convince me, it's only necessary to convince those who approve the pull request.

NobbZ: A BST is about comperability, and I do not see why string and integer should not be comparable to each other...

In the erlang VM (and therefore in at least 3 languages on exercism) there is a total order defined over all values possible. `6 =< "foo"` is true in erlang, lfe and elixir (in their corresponding syntaxes).

Also after having read the introductionary comment, where you state that we only use positiv numbers as valid input, I'd use an unsigned int in statically typed languages.

But you have specified `-1` as an error indicator, but the only thing that can cause an error, is invalid input (which can't happen in accordingly typed languages) or someone else mangles with the tree from the outside, which we suppose not to happen.

So independently how I look at it, I do not feel that this test is necessary in any way.

exercises/binary-search-tree/canonical-data.json
NobbZ: I'm not even sure how to read this...

> 6, then 1, then 3, then...

or is it

> 6 after, 1 after 3, after...

or

> 6 has child 1, has child 3, has child (but which child then?)

Some implementations ask for implementing a function that creates a BST from a list, growing the tree from the first item to the last.

Roughly like this:

```hs
foldl Leaf (\ x t -> insert t x) list
```

In the introductionary comment we could explain that a property `"value"` could be used to create a singleton tree, or inserting a single value to a given tree, while `"values"` or `"tree"` implies to create a tree from a list using this function.

kabiir:
So in conclusion I should 
- remove the *invalid data*` test 
- add a `current_tree_data ` key which is an array.
- add a comment which explains the order of the data in `curernt_tree_data` 
   e.g. [6,5,3,1] means insert 6 then insert 5 then insert 3 then insert 1.

I'd like your opinions on the order of tests too and if I should add another test. @petertseng @NobbZ

petertseng:
> So in conclusion I should 
> - remove the *invalid data*` test 
> - add a `current_tree_data ` key to every test which is an array.
> - add a comment which explains the order of the data in `curernt_tree_data` 
>    e.g. [6,5,3,1] means insert 6 then insert 5 then insert 3 then insert 1.

I agree with that

> I'd like your opinions on the order of tests too and if I should add another test

I defer to all tracks that have implemented this exercise.

`$ curl http://x.exercism.io/v3/problems/data-todos | jq '.[] | map(select(.["slug"] == "binary-search-tree"))'`

```json
[
  {
    "slug": "binary-search-tree",
    "blurb": "Insert and search for numbers in a binary tree.",
    "readme_url": "https://github.com/exercism/problem-specifications/blob/master/exercises/binary-search-tree/description.md",
    "data": null,
    "implementations": [
      {
        "track_id": "csharp",
        "url": "https://github.com/exercism/csharp/tree/master/exercises/binary-search-tree"
      },
      {
        "track_id": "clojure",
        "url": "https://github.com/exercism/clojure/tree/master/exercises/binary-search-tree"
      },
      {
        "track_id": "coffeescript",
        "url": "https://github.com/exercism/coffeescript/tree/master/exercises/binary-search-tree"
      },
      {
        "track_id": "crystal",
        "url": "https://github.com/exercism/crystal/tree/master/exercises/binary-search-tree"
      },
      {
        "track_id": "ecmascript",
        "url": "https://github.com/exercism/ecmascript/tree/master/exercises/binary-search-tree"
      },
      {
        "track_id": "elixir",
        "url": "https://github.com/exercism/elixir/tree/master/exercises/binary-search-tree"
      },
      {
        "track_id": "fsharp",
        "url": "https://github.com/exercism/fsharp/tree/master/exercises/binary-search-tree"
      },
      {
        "track_id": "go",
        "url": "https://github.com/exercism/go/tree/master/exercises/binary-search-tree"
      },
      {
        "track_id": "haskell",
        "url": "https://github.com/exercism/haskell/tree/master/exercises/binary-search-tree"
      },
      {
        "track_id": "java",
        "url": "https://github.com/exercism/java/tree/master/exercises/binary-search-tree"
      },
      {
        "track_id": "javascript",
        "url": "https://github.com/exercism/javascript/tree/master/exercises/binary-search-tree"
      },
      {
        "track_id": "lua",
        "url": "https://github.com/exercism/lua/tree/master/exercises/binary-search-tree"
      },
      {
        "track_id": "perl5",
        "url": "https://github.com/exercism/perl5/tree/master/exercises/binary-search-tree"
      },
      {
        "track_id": "ruby",
        "url": "https://github.com/exercism/ruby/tree/master/exercises/binary-search-tree"
      },
      {
        "track_id": "scala",
        "url": "https://github.com/exercism/scala/tree/master/exercises/binary-search-tree"
      },
      {
        "track_id": "swift",
        "url": "https://github.com/exercism/swift/tree/master/exercises/binary-search-tree"
      },
      {
        "track_id": "typescript",
        "url": "https://github.com/exercism/typescript/tree/master/exercises/binary-search-tree"
      }
    ]
  }
]
```

The suggestions in https://github.com/exercism/docs/blob/master/you-can-help/improve-exercise-metadata.md#extracting-canonical-test-data say to consult the implementations to see what tests to add. I am presently unable to do this on your behalf, so I trust you to do it.

Insti:
@kabiir are you still working on this?

kabiir:
@Insti Yes!, I just didn't get enough time. I've been going through tracks that have implemented tests, they all seem to follow a trend it's like there is a hidden canonical-data.json. Anyway i'm working on it and i'll try to follow the trend too.

ErikSchierboom:
> they all seem to follow a trend it's like there is a hidden canonical-data.json

Which is probably entirely due to the fact that people often copy an existing set of tests and modify them to fit their track :) Great to hear that you are working on this by the way!

kabiir:
So for test case "insert smaller number on left node" and data [4, 2], to check 
``` tree.root.left.data = 2```  which means left node of root contains 2. How should I represent this test case?
More importantly how should I represent `root.left` node. Should I create a new key called `node: "root.left"` and set `property:"data"`
e.g. 
```json
               {
                    "description": "smaller number at left node",
                    "property": "data",
                    "node": "root.left",
                    "tree_data": [4, 2],
                    "expected": 2
                }
```
Or maybe I could add a `comments` key and explain in plain english for every test case
e.g.
```json
               {
                    "description": "smaller number at left node",
                    "property": "data",
                    "comments": "Left child of node '4' should be node '2'",
                    "tree_data": [4, 2],
                    "expected": 2
                }
```
I'm inclined to use comments. Thoughts? @petertseng @NobbZ @Insti

NobbZ:
My generator were not able to parse those comments. Therefore I think we need to find a way to do it in a formalized way, but I'M undercaffeinated, so I do not have a clue how right now…

petertseng:
You may use a series of operations as in https://github.com/exercism/problem-specifications/blob/master/exercises/zipper/canonical-data.json#L86-L96 if you desire.

This statement must not be taken as an endorsement or rejection of using a series of operations.

kabiir:
Since most tests are expecting the tree to be in a certain form. Adding an `expectedTree` key which represents the tree, will the following representation do?
```json
"comments": [
        "The key expectedTree represents a tree as a node object starting from root node",
        "a node object consists of 3 keys 'data', 'left' and 'right'",
        "The key 'data' contains data of the said node, 'left' contains",
        "a node object of left-child node and 'right' contains a node object",
        "of right-child node."
]
{
    "description": "greater number at right node",
    "property": "insert",
    "tree_data": [4, 5],
    "expectedTree": {
        "data": 4,
        "left": null,
        "right": {
            "data": 5,
            "left": null,
            "right": null
        }
    }
}
```

petertseng:
I agree that our goal should be to check that the tree resulting from inserting a given sequence of numbers has a shape matching the expected shape.

I advise that if the trees are implemented as abstract data types that an outside caller such as a test suite may not be able to directly construct a tree that can be compared against the observed tree (otherwise it would force the implementation), so they may be restricted to using the interface to test the shape of the tree.

This advice doesn't prevent the proposed representation, therefore it is not a rejection of the proposal.

The schema will require that `expectedTree` instead be `expected`.

coriolinus:
One advantage of the `"node": "root.left"` proposal is that it succinctly describes the traversal path. For certain tree implementations and languages, the test runner may be able to literally `eval` the `node` item and compare it to the expected value. It's a less verbose variant of the sequence of operations idea.

A disadvantage is that it doesn't represent the entire tree, which increases the number of tests required to determine that the whole tree is in the state expected. 

I'm moderately in favor of the traversal path suggestion even so, because I suspect that for most tests, we don't need to test the entire tree state to determine whether or not the student has done the right thing; we just need to ensure that the result of the tested operation is as expected.

Insti:
> that should be a simple matter of replacing all instances of `tree_data` with `input`.

You'll need to wrap it rather than just replace it.

```
 "input": ["4", "2"],
```
Would still not match the specification.
It will need to be:

```
"input": { "tree_data": ["4", "2"] },
```

kabiir:
@Insti I agree, but I still don't understand how this particular canonical-data can be parsed by test generators and produce a usable output. `tree_data` is not actually an input! It's just a convenient way of saying  
_I want the tree to have these values inserted in this order before checking for this test_  
I used property `data` because I didn't know how to convey my test case otherwise. I would appreciate some input on this.

coriolinus:
@kabiir I've got exercise generators on my mind because I recently wrote the one the Rust track uses. I can't speak for other tracks' exercise generators, but what the Rust track's generator will do is group the tests by the tested `property`. For this set of canonical data, there are two tested properties: `data` and `sortedData`. For each tested property, it builds a test stub, like this:

```rust
fn process_data_case<I, O>(input: I, expected: O) {
    unimplemented!()
}

fn process_sorteddata_case<I, O>(input: I, expected: O) {
    unimplemented!()
}
```

It then generates a bunch of actual tests written in terms of those functions. Crucially, if the case in the canonical data includes `input`, it uses the actual input value. If it doesn't, it collects all the properties, excludes the ones it knows about, sorts them by key name, and sends the batch as a tuple to the test_sorteddata_property. 

So, in cases such as [`book-store`](https://github.com/exercism/problem-specifications/blob/master/exercises/book-store/canonical-data.json#L13-L19) which don't have an `input` property, extraneous data may be included. Those cases include a `targetgrouping` property which is intended as an output hint, but the test generator can't know that. Instead, it is duly passed into the `process_total_case` function and then [ignored forever](https://github.com/exercism/rust/blob/master/exercises/book-store/tests/book-store.rs#L16-L22). 

Back to `binary-search-tree`, if `input` is a literal list, then type `I` in both `process_*_case` functions will be `Vec<&'static str>`; if it's an object literal, type `I` will be `HashMap<&'static str, Vec<&'static str>>`<sup>1</sup>. It's the responsibility of the person running the exercise generator to read the comments and description and implement the `process_*_case` functions appropriately. Assuming we stick with the standard and use object literals, it'll look something like this:

```rust
// assuming BST is the name of the student's struct
fn process_data_case(input: HashMap<&'static str, Vec<&'static str>>, expected: HashMap<&'static str, Option<&'static str>>) {
    let mut student_bst = BST::new();
    for item in input.get("tree_data").unwrap() {
        student_bst.insert(item.parse());
    }
    unimplemented!() // TODO: build the expected tree here, which might be kind of tedious
    assert_eq!(student_bst, expected_bst);
}
```

The exercise generator does the busywork of generating a whole bunch of tests which look like this<sup>2</sup>:

```rust
#[test]
#[ignore]
fn test_data_is_retained() {
    process_data_case(hashmap!("tree_data" => vec!["4"]), hashmap!("data" => "4", "left" => None, "right" => None));
}
```

However, it's _not_ the responsibility of the exercise generator to get a complete exercise ready to merge without human input. It's just there to streamline and simplify the work of generating an exercise. 

---
1. Why are you passing these numbers as strings, by the way? Numeric literals would make perfect sense here. It's not a blocker for me because strings also compare, but it seems odd.

2. If you're fluent in Rust, you'll notice that the generated tests won't actually compile at all in this case due to a type error in the `expected` hashmaps. This is a known limitation of the test generator; it can't currently tell when it should assume things are actually hashmaps, and when they're heterogenous objects. I'm working on it, but as a proper solution would likely involve full type inference, a proper solution is likely to be some time away.

Insti:
> `tree_data` is not actually an input!

It is a precondition to the test thus it counts as input to the test generator even if it's not going to be an argument to the function being tested.

kabiir:
@coriolinus I saw that some of the tracks were using strings to represent numbers so I made the choice of using strings because the person who will implement the test whether using test generators or not will know that data is not restricted to integers only. Since depending upon the test the student will model there program and its not described in the Readme which data type input to BST is going to be. So the tests should not restrict it to numbers or strings but I had to make a choice so strings.  

-------
@Insti Should I mention that on the comment?

Insti:
> I can try to follow all of the standard but at some point some test cases just won't be parsed by test generators

Why do you think this would be the case? 
Test generators are customised to the particular exercise they are generating tests for.
Just ensure that the format is consistent and logical and there won't be a problem.

> Should I mention that on the comment?

Probably.

kabiir:
@Insti 
>Why do you think this would be the case?
Test generators are customised to the particular exercise they are generating tests for.
Just ensure that the format is consistent and logical and there won't be a problem.  

I thought test generators were general purpose for all exercises

exercises/binary-search-tree/canonical-data.json
```diff
@@ -0,0 +1,178 @@
...
+                            "left": null,
+                            "right": null
+                        },
+                        "right": null
+                    }
+                },
+                {
+                    "description": "same number at left node",
+                    "property": "data",
+                    "node": "root->left",
```
petertseng: I think in this file you have taken your suggestion of https://github.com/exercism/problem-specifications/pull/940#issuecomment-341854357 to use `expected` to represent the entire tree, and not your suggestion of https://github.com/exercism/problem-specifications/pull/940#issuecomment-341592212 to list one specific node being traversed to. If you confirm this is true, I think this `node` line should be removed.

I don't think `node` is still necessary, right? Now that `expected` shows the full tree.

kabiir: Yes!, good catch.

exercises/binary-search-tree/canonical-data.json
```diff
@@ -0,0 +1,177 @@
...
+        "        }",
+        "    },",
+        "    'required':['data', 'left', 'right']",
+        "}"
+    ],
+    "cases": [{
+            "description": "data is retained",
+            "property": "data",
+            "input": {
+                "tree_data": ["4"]
```
kabiir: @coriolinus @Insti 
I just saw issue #987 
Should the input key `tree_data` be changed to lowerCamelCase

kabiir: From what I understand #996 is not allowing additional properties. so does #987 make sense?

coriolinus: I have a mild preference for `snake_case` over `lowerCamelCase`, but TBH I didn't even think about the applicability of #987 here. 

As I understand it, the driving intent behind that issue was to avoid using keys which require special handling because they contain spaces or otherwise are not permissible as variable names. As that's handled by either naming convention, I don't have a preference one way or the other about whether you should update your cases here.

ErikSchierboom:
@rpottsoh @petertseng Is this PR ready to be merged?

exercises/binary-search-tree/canonical-data.json
```diff
@@ -0,0 +1,177 @@
...
+                "right": null
+            }
+        },
+        {
+            "description": "insert data at proper node",
+            "cases": [{
+                    "description": "smaller number at left node",
+                    "property": "data",
+                    "input": {
+                        "tree_data": ["4", "2"]
```
rpottsoh: I think this should be lowerCamelCase in accordance with #987.

exercises/binary-search-tree/canonical-data.json
```diff
@@ -0,0 +1,177 @@
...
+                            "right": null
+                        },
+                        "right": null
+                    }
+                },
+                {
+                    "description": "same number at left node",
+                    "property": "data",
+                    "input": {
+                        "tree_data": ["4", "4"]
```
rpottsoh: I think this should be lowerCamelCase in accordance with #987.

exercises/binary-search-tree/canonical-data.json
```diff
@@ -0,0 +1,177 @@
...
+                            "right": null
+                        },
+                        "right": null
+                    }
+                },
+                {
+                    "description": "greater number at right node",
+                    "property": "data",
+                    "input": {
+                        "tree_data": ["4", "5"]
```
rpottsoh: I think this should be lowerCamelCase in accordance with #987.

exercises/binary-search-tree/canonical-data.json
```diff
@@ -0,0 +1,177 @@
...
+                        }
+                    }
+                }
+            ]
+        },
+        {
+            "description": "can create complex tree",
+            "property": "data",
+            "input": {
+                "tree_data": ["4", "2", "6", "1", "3", "5", "7"]
```
rpottsoh: I think this should be lowerCamelCase in accordance with #987.

exercises/binary-search-tree/canonical-data.json
```diff
@@ -0,0 +1,177 @@
...
+                }
+            }
+        },
+        {
+            "description": "can sort data",
+            "cases": [{
+                    "description": "can sort single number",
+                    "property": "sortedData",
+                    "input": {
+                        "tree_data": ["2"]
```
rpottsoh: I think this should be lowerCamelCase in accordance with #987.

exercises/binary-search-tree/canonical-data.json
```diff
@@ -0,0 +1,177 @@
...
+                    },
+                    "expected": ["2"]
+                },
+                {
+                    "description": "can sort if second number is smaller than first",
+                    "property": "sortedData",
+                    "input": {
+                        "tree_data": ["2", "1"]
```
rpottsoh: I think this should be lowerCamelCase in accordance with #987.

exercises/binary-search-tree/canonical-data.json
```diff
@@ -0,0 +1,177 @@
...
+                    },
+                    "expected": ["1", "2"]
+                },
+                {
+                    "description": "can sort if second number is same as first",
+                    "property": "sortedData",
+                    "input": {
+                        "tree_data": ["2", "2"]
```
rpottsoh: I think this should be lowerCamelCase in accordance with #987.

exercises/binary-search-tree/canonical-data.json
```diff
@@ -0,0 +1,177 @@
...
+                    },
+                    "expected": ["2", "2"]
+                },
+                {
+                    "description": "can sort if second number is greater than first",
+                    "property": "sortedData",
+                    "input": {
+                        "tree_data": ["2", "3"]
```
rpottsoh: I think this should be lowerCamelCase in accordance with #987.

exercises/binary-search-tree/canonical-data.json
```diff
@@ -0,0 +1,177 @@
...
+                    },
+                    "expected": ["2", "3"]
+                },
+                {
+                    "description": "can sort complex tree",
+                    "property": "sortedData",
+                    "input": {
+                        "tree_data": ["2", "1", "3", "6", "7", "5"]
```
rpottsoh: I think this should be lowerCamelCase in accordance with #987.

rpottsoh:
I think this is now ready to be merged.  Not sure if it should be squashed.

ErikSchierboom:
I've decided to go for squashing. Thanks @kabiir @rpottsoh ! 🎉

Stargator:
Way to go @kabiir! That was a lot of work over a period of time.

Issue #938, opened 2017-10-09
Versioning guidelines


NobbZ:
Today 3 PRs have been made (perhaps more that slipped my RADAR), which all bumped MAJOR version, because of trivial changes.

In all 3 PRs voices rose that wanted a less drastic version bump:

* #935 and #936: It was requested that only PATCH was bumped, since the actual testdata didn't change, only its source representation. No track had to change its language specific representation of the testdata to actually comply to the new testdata.
* #937: It was requested that only MINOR was bumped, because the tests stayed the same, but were only slightly re-orderd, a single one added and also de-nested.

But the current versioning and bumping guidelines tell us this:

> * MAJOR changes should be expected to break even well-behaved test generators.
> * MINOR changes would never break well-designed test generators, because the test-generation logic remains exactly the same.
> * PATCH changes would never break well-designed test generators, because the test data remains exactly the same

[enumeration mine, detailed descriptions omitted for now]

After reading this TL;DRs of the various kinds of version bumps, some question arise.

* What is a “well-behaved” testgenator?
* What changes except for renaming, adding, removing or reordering tests, will not break a generator?

Even worse, on a further read of the more thorough description of each bump I found this:

> [bump PATCH when] Regrouping/"Renesting" test cases without changing test case ordering.

But how is changing the nesting, not changing the logic of the testgenerator? It can't find the tests anymore where it expects them.

Also I do feel, that adding and removing testcases, may end in a completely different way to solve the exercise, which makes the whole testsuite incompatible with its previous version, so it should be a MAJOR bump here, instead of a MINOR as in the document:

> [bump MINOR when] Adding or deleting test cases.

Nearly the same reasoning applies in my eyes for the other two cases where a MINOR were sufficient, but I'd do a MAJOR because it changes the way to solve the exercise drastically:

> * Changing the test cases inputs and/or outputs.
> * Changing the test cases ordering.

---

When we introduced the versioning scheme, I've understood it as versioning the exercise, today I learnt it's meant to version the generator.

I do not think this is the way to go, but I will not try to change that anymore, I think I'm just to late to the party to change this, but what I can try, is to simplify the current bumping guide and remove ambiguity:

* Bump MAJOR when you do not bump MINOR
* Bump MINOR on fixing typos in comments and descriptions or you are reformatting/reordering the JSON document without changing its semantics.
* Remove PATCH from versioning scheme

[...]

PR #941, opened 2017-10-10
two-bucket: fix incorrect test case

### LINKED ISSUE #911: two-bucket: incorrect test case in canonical data?
According to the [README] for the exercise, valid moves include:
>pouring from one bucket to another
>emptying one bucket and doing nothing to the other
>filling one bucket and doing nothing to the other

The only other restriction stated is
>That means, when starting with the larger bucket full, you are NOT allowed at any point to have the smaller bucket full and the larger bucket empty (aka, the opposite starting point)

In other words, if one bucket is full, and the other empty, the full bucket must be the starting bucket.

The last test case description says ["Measure using bucket one of size 2 and bucket two of size 3 - start with bucket one and end with bucket two"](https://github.com/exercism/problem-specifications/blob/df076b2bd75e06e42a2ad0cb7eb0d5a8bdccd9f6/exercises/two-bucket/canonical-data.json#L70-L84) 

With the above rules, what disqualifies the following from being a valid solution for the above test case?
(2 moves): `(fill first bucket, as directed) [2, 0] -> (fill second bucket) [2, 3]`

cmccandless:
Resolves #911

### DIFF
--- a/exercises/two-bucket/canonical-data.json
+++ b/exercises/two-bucket/canonical-data.json
@@ -75,9 +75,9 @@
       "goal": 3,
       "start_bucket": "one",
       "expected": {
-        "moves": 4,
+        "moves": 2,
         "goal_bucket": "two",
-        "other_bucket": 1
+        "other_bucket": 2
       }
     }
   ]
--- a/exercises/two-bucket/canonical-data.json
+++ b/exercises/two-bucket/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "two-bucket",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "cases": [
     {
       "description": "Measure using bucket one of size 3 and bucket two of size 5 - start with bucket one",


Insti:
1.2.0 ?

The result of a test has changed, this may mean solutions that once passed now fail.
1.1.1 would be fixing a typo in the test name or similar.

(plus or minus whatever #938 decides.)

rpottsoh:
>1.2.0 ?

Works for me.  @cmccandless please change `"version":` to equal 1.2.0.

cmccandless:
Version updated to 1.2.0

rpottsoh:
This should be squashed when it is merged.

cmccandless:
@rpottsoh Would it be easier for me to go ahead and squash, or is there a simple way to do that from GitHub? I don't have much experience merging pull requests from command line, and that's mostly what I use.

petertseng:
> Would it be easier for me to go ahead and squash, or is there a simple way to do that from GitHub?

You may squash via the command line if you like. Or you may leave it, and the one who *merges* the PR will squash, since GitHub provides the way to do so. I haven't yet seen a way that GitHub provides the *submitter* of the PR to squash via web, so if you want to do it you will be doing it via the command line.

You are welcome do to so, or not, as you please.

cmccandless:
If there's an easy way for the merger to squash, I'll leave it alone.

⁣


Corey McCandless​

Insti:
Squashed and merged, thanks @cmccandless :heart:

PR #944, opened 2017-10-11
diffie-hellman: Add canonical data

dkinzer:
REF exercism/problem-specifications#557


Insti:
How are you going with this @dkinzer ?

dkinzer:
@Insti hey, sorry, I'm planning to get to this tonight or tomorrow.  The week was busier than I expected :)

Insti:
There seems to be something wrong with the json:

https://travis-ci.org/exercism/problem-specifications/builds/288096848#L514

exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,126 @@
...
+    , " ask the user to implement a private key generator.     "
+    , "                                                        "
+    , " The other major difference is that both javascript and "
+    , " ecmascript test boundaries and verify that their       "
+    , " implementations throw errors for specific cases.       "
+    , " The other implementations do no do this (perhaps       "
+    , " because they are all typed?).                          "
+    , "                                                        "
+    , " One last difference is that the go implementation uses "
+    , " hexadecimal notation for the bit Ints."
```
Insti: These comments are probably better put in the git commit message and/or the Github pull request description.

exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,126 @@
...
+          , "                                                               "
+          , " Alice's private key = 6                                       "
+          , " Alice's public key = 8                                        "
+          , "                                                               "
+          , " Bob's private key = 15                                        "
+          , " Bob's public key = 19                                         "
+          ]
+      , "description": "Boundary tests and exceptions."
+      , "cases":
+          [ { "description": "Throws if the constructor arguments are out of range."
```
Insti: These descriptions are saying what it does and will not make sense in all languages.
They also seem to be testing multiple things per test.

exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,126 @@
...
+      , "property"   : "generateSharedSecret"
+      , "p": "23"
+      , "bob_public_key": "19"
+      , "alice_private_key": "6"
+      , "expected": "2"
+      }
+    , { "description": "Shared secret generated by Alice and Bob should be equal."
+      , "property"   : "generateSharedSecret"
+      }
+    , { "description": "Tests with very large primes."
```
Insti: Are these tests necessary?
How are very large primes different from small primes?

exercises/diffie-hellman/canonical-data.json
dkinzer: Right, the implementations are not consistent currently.  I don't think it may be possible to do really bit primes in some languages.  Maybe there can be an optional set of tests.

exercises/diffie-hellman/canonical-data.json
dkinzer: I personally don't think they are not needed.  But since so many of the current implementers (including golang) are doing this, I didn't want to presume.  Maybe they should go into a "optional" section?

exercises/diffie-hellman/canonical-data.json
dkinzer: Oops I think, I answered the wrong question.  Maybe I can try describing it as "fails when" instead of "throws"?

exercises/diffie-hellman/canonical-data.json
dkinzer: Got it (though it's not clear from the template that these types are not what goes here).

dkinzer:
@Insti I didn't realized that the JSON was going to be validated against a schema (good thing).  Earlier I had pushed up a valid JSON but it was not validating against the provided schema.

I made some changes per your suggestions.

dkinzer:
@petertseng I can go ahead and do that for the non large prime cases.

exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,100 @@
...
+    , { "description": "Shared secret generated by Alice and Bob should be equal."
+      , "property"   : "generateSharedSecret"
+      }
+    , { "description": "Boundary tests."
+      , "cases":
+          [ { "description": "Fails when the constructor arguments are out of range."
+            , "property"   : "new"
+            , "p"  : 0
+            , "g"   : 9999
+            , "expected"   : { "error": "Constructor arguments are out of range or non-prime!" }
```
Insti: > out of range **or** non-prime

Which is it?
We should be able to know exactly why the error is occurring.

Insti:
I'm strongly against including the huge prime tests in the canonical data.

* The algorithm requires the numbers be prime but they need not be large.
* In some languages implementing support for large primes (or even numbers) is non-trivial.
* The description does not mention large numbers.

Does anyone want to convince me that including them is a good idea?

dkinzer:
@Insti I removed the the optional big prime cases (like I said, I'm not partial to them either, I only added them because so many of the current implementations already have them.)

Also, I made the error message more specific as you requested.

exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,78 @@
...
+      , "expected": 8
+      }
+    , { "description": "Shared secret is correctly calculated."
+      , "property"   : "generateSharedSecret"
+      , "p": 23
+      , "bob_public_key": 19
+      , "alice_private_key": 6
+      , "expected": 2
+      }
+    , { "description": "Shared secret generated by Alice and Bob should be equal."
```
petertseng: I would have expected this to define `alice_public_key` and `bob_private_key` and `expected`, otherwise it is not clear what values we are testing.

If anyone reads my suggestion and is concerned that the same property `generateSharedSecret` may display two sets of input keys (`[alice_public_key, bob_private_key]` vs `[bob_public_key, alice_private_key]`) may I instead suggest just `public_key` and `private_key`?

If anyone reads that suggestion and now objects that it's not clear that the public and private key belong to different people, what about `their_public_key` and `my_private_key` or something to that effect?

exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,78 @@
...
+      , "property"   : "generateSharedSecret"
+      }
+    , { "description": "Boundary tests."
+      , "cases":
+          [ { "description": "Fails when the constructor arguments are out of range."
+            , "property"   : "new"
+            , "p"  : 0
+            , "g"   : 9999
+            , "expected"   : { "error": "Constructor arguments are out of range." }
```
petertseng: For my understanding: `p` is out of range because it is 0, right?

For the sake of understanding of implementors of this test: I would think that the error here should tell us which one is out of range.

exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,78 @@
...
+      , "bob_public_key": 19
+      , "alice_private_key": 6
+      , "expected": 2
+      }
+    , { "description": "Shared secret generated by Alice and Bob should be equal."
+      , "property"   : "generateSharedSecret"
+      }
+    , { "description": "Boundary tests."
+      , "cases":
+          [ { "description": "Fails when the constructor arguments are out of range."
```
petertseng: I wasn't sure, but if `p` is the only argument is out of range, then the description cannot claim to be testing the arguments being out of range. It can only claim to be testing that `p` is out of range. The description should reflect it.

I do not pass judgment on whether there should be an additional test added to test what happens when `g` is out of range.

exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,78 @@
...
+            , "property"   : "new"
+            , "p"  : 0
+            , "g"   : 9999
```
petertseng: note that `g` is also non-prime, therefore this may also trigger the non-prime error. To make the test unambiguous, I would expect that `g` should be prime, so that one who sees the test fail can be confident that it is being checked.

exercises/diffie-hellman/canonical-data.json
petertseng: Should not mention "constructor arguments" because that constrains the implementation. I would just say "primes", since that is what they are.

exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,78 @@
...
+            , "expected"   : { "error": "Constructor arguments are out of range." }
+            }
+          , { "description": "Fails when the constructor arguments are not prime."
```
petertseng: > Should not mention "constructor arguments" because that constrains the implementation. I would just say "primes", since that is what they are.

same

petertseng: > I wasn't sure, but if `p` is the only argument is out of range, then the description cannot claim to be testing the arguments being out of range. It can only claim to be testing that `p` is out of range. The description should reflect it.

> I do not pass judgment on whether there should be an additional test added to test what happens when `g` is out of range.

Same, except now I'm talking about primality.

exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,78 @@
+{ "exercise": "diffie-hellman"
+, "version" : "1.0.0"
+, "comments":
+    [ " In the following descriptions, p and g represent two          "
+    , " prime numbers.                                                "
+    , "                                                               "
+    , " Unless specified otherwise, tests are setup with              "
+    , " new DiffieHellman object using p = 23; g = 5.                 "
```
petertseng: I do not think this sentence is necessary since all the cases in this file seem to be providing `p` and `g` whenever they are necessary... or if they do not, I think they should! If we leave it implicit, it may become unclear which values are needed for which tests. For example, `g` is not necessary in the test of generating a private key.

exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,78 @@
...
+    , "                                                               "
+    , " Alice's private key = 6                                       "
```
petertseng: I do not think this is necessary since this is specified in tests that need it.

exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,78 @@
...
+    , " Alice's public key = 8                                        "
```
petertseng: >  I do not think this is necessary since this is specified in tests that need it.

same

exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,78 @@
...
+    , "                                                               "
+    , " Bob's private key = 15                                        "
```
petertseng: >  I do not think this is necessary since this is specified in tests that need it.

same

exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,78 @@
...
+    , " Bob's public key = 19                                         "
```
petertseng: >  I do not think this is necessary since this is specified in tests that need it.

same

exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,78 @@
...
+    ]
+, "cases":
+    [ { "description": "Private keys are generated in Range (1..p)"
+      , "property"   : "generatePrivateKey"
+      , "p": 23
+      }
+    , { "description": "generatePrivateKey's output is random."
+      , "property"   : "generatePrivateKey"
```
petertseng: this is difficult to test, but since `property` tells us how we should interpret the other keys in the same object, and we have two cases with same `property` but testing different things, perhaps they should be combined into one, OR this one should have some `property` like `generatePrivateKeyIsRandom`.

Note that there is no precedent for this, since `robot-name` and `simple-cipher` are to my knowledge the only other exercises that need randomness. the former has no JSON file but the latter is https://github.com/exercism/problem-specifications/pull/966, so perhaps the same approach may be used for both.

exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,78 @@
...
+            , "expected"   : { "error": "Private key must be greater than one but less than modulus parameter p!" }
+            }
+          , { "description": "Fails when private key is 1."
+            , "property"   : "getPublicKeyFromPrivateKey"
+            , "private_key" : 1
+            , "expected"   : { "error": "Private key must be greater than one but less than modulus parameter p!" }
+            }
+          , { "description": "Fails when private key is greater than the modulus parameter p."
+            , "property"   : "getPublicKeyFromPrivateKey"
+            , "private_key" : "p + 1"
```
petertseng: I would provide `"p": 23` and `"private_key": 24` here, because I don't think I would encourage test generators to insert this string as a parameter, but maybe I'm wrong about that.

petertseng: note that `p` should also be considered invalid, so `p + 1` is not on a boundary.

exercises/diffie-hellman/canonical-data.json
petertseng: EBWODP:

> so that one who sees the test fail can be confident that **it** is being checked.

"it" needs to be clarified: range

dkinzer:
@petertseng  thanks for the input, but at this time I'm not sure how to proceed.  The changes I added come directly from the current implementation of this problem from various tracks.  If there are issues with how this problem has been implemented they should probably be addressed at that level.

However, at the very least there should be some indication which one of the various implementations should be taken as the exemplary one so that whoever takes this issue on next does not need to know the details of this problem in order to take on this issue.

 I had assumed (incorrectly) that this issue was merely a mapping of an implementation into a standard JSON format.  But clearly a more in depth knowledge is required of the problem at this time (a knowledge I currently  I do not have, and also do not have the time to gain).

In that light, I think it makes sense for me to bow out and give someone else a chance at implementing this JSON correctly.

Apologies for taking peoples time.

Insti:
Thanks for your work on this @dkinzer :heart:

Over time, as exercises in different tracks are developed, the description changes, the focus of the exercise may change and the Exercism problem philosophy evolves.

Part of the process of creating the canonical data for a problem is looking at all the existing implementations and seeing what the different tracks do, then having a conversation to try and establish what the best common set of tests is.

The `canonical-data.json` pull request is usually the first time that this discussion happens, so it can be that it takes a while to focus in on the core tests for the exercise. Once they have been decided the canonical data *becomes* the exemplary case. These changes then propagate out to the tracks the next time they choose to update their test suites.

We're sorry if this process has not been communicated via the documentation, could you suggest a place where we could put it to help set the expectations for future contributors?

PR #946, opened 2017-10-11
space-age: Revert "Note expected accuracy in space-age description"

petertseng:
This reverts commit 7eb0dc966d843d74329c6d6475ea90a14fff440b.

In https://github.com/exercism/problem-specifications/pull/880 we
discussed that tracks may need to use a tolerance in their floating
point comparisons.

In light of this, asking students to round the answer seems strangely
arbitrary and theoretically doesn't prevent the need for having a
tolerance. Despite this, I do not currently have an example where
lacking a tolerance would have a false negative or false positive if a
floating point literal of two decimal places is compared to a value that
was rounded to two decimal places. So, then, the primary motivation is
to avoid adding requirements whose reasons are unclear.

Thus, remove the sentence completely. Tracks should add it if it is
relevant for their track.

Closes https://github.com/exercism/problem-specifications/pull/945 by
mutual exclusion.

The tracks that currently implement the exercise, and whether their
equality tests have tolerances:

Please excuse me if I miscategorised your track because of my
unfamiliarity with your language (and then correct me).

Exactly equal:

* https://github.com/exercism/csharp/blob/master/exercises/space-age/SpaceAgeTest.cs
* https://github.com/exercism/cfml/blob/master/exercises/space-age/SpaceAgeTest.cfc
* https://github.com/exercism/ecmascript/blob/master/exercises/space-age/space-age.spec.js
* https://github.com/exercism/fsharp/blob/master/exercises/space-age/SpaceAgeTest.fs
* https://github.com/exercism/javascript/blob/master/exercises/space-age/space-age.spec.js
* https://github.com/exercism/kotlin/blob/master/exercises/space-age/src/test/kotlin/SpaceAgeTest.kt
* https://github.com/exercism/lua/blob/master/exercises/space-age/space-age_spec.lua
* https://github.com/exercism/perl5/blob/master/exercises/space-age/space.t
* https://github.com/exercism/perl6/blob/master/exercises/space-age/space-age.t
* https://github.com/exercism/python/blob/master/exercises/space-age/space_age_test.py
* https://github.com/exercism/r/blob/master/exercises/space-age/test_space-age.R
* https://github.com/exercism/scala/blob/master/exercises/space-age/src/test/scala/SpaceAgeTest.scala
* https://github.com/exercism/swift/blob/master/exercises/space-age/Tests/SpaceAgeTests/SpaceAgeTests.swift
* https://github.com/exercism/typescript/blob/master/exercises/space-age/space-age.test.ts

Fixed tolerance 0.5 in either direction (so that the answer rounds to the desired value if rounding to closest integer)

* https://github.com/exercism/elm/blob/master/exercises/space-age/tests/Tests.elm

Fixed tolerance 0.01 in either direction

* https://github.com/exercism/java/blob/master/exercises/space-age/src/test/java/SpaceAgeTest.java
* https://github.com/exercism/objective-c/blob/master/exercises/space-age/SpaceAgeTest.m
* https://github.com/exercism/php/blob/master/exercises/space-age/space-age_test.php
* https://github.com/exercism/ruby/blob/master/exercises/space-age/space_age_test.rb
* https://github.com/exercism/rust/blob/master/exercises/space-age/tests/space-age.rs

Fixed tolerance 0.005 in either direction (so that the answer rounds to the desired value if rounding to two decimal places)

* https://github.com/exercism/cpp/blob/master/exercises/space-age/space_age_test.cpp
* https://github.com/exercism/clojure/blob/master/exercises/space-age/test/space_age_test.clj
* https://github.com/exercism/common-lisp/blob/master/exercises/space-age/space-age-test.lisp
* https://github.com/exercism/elixir/blob/master/exercises/space-age/space_age_test.exs
* https://github.com/exercism/erlang/blob/master/exercises/space-age/test/space_age_tests.erl
* https://github.com/exercism/haskell/blob/master/exercises/space-age/test/Tests.hs
* https://github.com/exercism/lfe/blob/master/exercises/space-age/test/space-age-tests.lfe
* https://github.com/exercism/ocaml/blob/master/exercises/space-age/test.ml
* https://github.com/exercism/prolog/blob/master/exercises/space-age/space_age_tests.plt

Variable tolerances (decided ahead-of-time):

* https://github.com/exercism/c/blob/master/exercises/space-age/test/test_space_age.c

### DIFF
--- a/exercises/space-age/description.md
+++ b/exercises/space-age/description.md
@@ -10,8 +10,7 @@ Given an age in seconds, calculate how old someone would be on:
    - Neptune: orbital period 164.79132 Earth years
 
 So if you were told someone were 1,000,000,000 seconds old, you should
-be able to say that they're 31.69 Earth-years old. Round all ages to
-the nearest hundredth of a year.
+be able to say that they're 31.69 Earth-years old.
 
 If you're wondering why Pluto didn't make the cut, go watch [this
 youtube video](http://www.youtube.com/watch?v=Z_2gbGXzFbs).


petertseng:
It's only possible to merge one of #945 and #946 (this PR). Reviewers, you should indicate your preference.

exercises/space-age/description.md
```diff
@@ -10,8 +10,7 @@ Given an age in seconds, calculate how old someone would be on:
    - Neptune: orbital period 164.79132 Earth years
 
 So if you were told someone were 1,000,000,000 seconds old, you should
-be able to say that they're 31.69 Earth-years old. Round all ages to
-the nearest hundredth of a year.
+be able to say that they're 31.69 Earth-years old.
```
rpottsoh: Unlike in #945 the burden of getting the answer properly rounded is shared equally by the maintainers and students via implication.  It is implied that the student will return `31.69 Earth-years`  not 31.689.  But at the same time the maintainers may opt to also be a little easy going and decide to round the students' answers to two decimal places, or otherwise check the answer against a tolerance amount before deciding if it right or wrong.

I think I agree with the assessment that how (and or when) rounding is handled should decided at the track level unless one of the topics envisioned for the exercise is rounding....  So I am in favor of this PR.

petertseng: Whoever writes the test suite is going to have to be comparing two floating-point values. The students do not need to compare; they only need to provide floating-point values to be compared against.

> It is implied that the student will return 31.69 Earth-years not 31.689.

Ah, I see. Would it be helpful to add the word "approximately" to the description, then?

> unless one of the topics envisioned for the exercise is rounding

As I understand it, it is not. I want to find a source to back up my understanding, but it is difficult.

> how (and or when) rounding is handled should decided at the track level

If one of the choices is wrong for some reason or other and therefore should not be shown to students, problem-specifications would not be remiss to specify the correct choice.
My question, then, is whether either choice is wrong. I do not have enough experience with floating-point numbers to answer that question, please help.

petertseng: > > unless one of the topics envisioned for the exercise is rounding
>
> As I understand it, it is not. I want to find a source to back up my understanding, but it is difficult.

Found it. https://github.com/exercism/go/issues/3#issuecomment-45052043

Space Age is meant as a metaprogramming exercise.

rpottsoh: > Would it be helpful to add the word "approximately" to the description, then?

The feeling I get from this is that the student will not need to do anything further to their answer.

In the exercises that I have worked with that involved floats, I don't think I bothered with rounding.  I was happy if the students answer was within a tolerance (of my design).

I don't think either #945 or #946 is wrong.  I favor #946 because it is more concise.  #945 feels more wishy washy to me.

petertseng:
I unintentionally lied. This is not, strictly speaking, a revert of the named commit, since it leaves the 31.69 in place instead of just 31. I think it's better to say 31.69. I'm inclined to still leave the commit message as it currently is.

PR #950, opened 2017-10-13
markdown: Do not add <p> to list items

petertseng:
Compare the case of "unordered lists" vs "With a little bit of
everything":

In the former, each list item is surrounded by `<p>` tags.

In the latter, they are not, instead being surrounded by other tags that
they otherwise had.

This could plausibly be explained by the rule "If there are any tags on
list items, don't add any, otherwise do add `<p>`" but it seems weird
because we don't usually see `<p>` for list items.

On a survey of two arbitrarily chosen Markdown parsers:

* http://parsedown.org/demo
* https://daringfireball.net/projects/markdown/

Neither emits `<p>` for the list items.
It seems they were a mistake and should be removed, unless information
to the contrary is reported at
https://github.com/exercism/problem-specifications/pull/262#discussion_r144485967

### DIFF
--- a/exercises/markdown/canonical-data.json
+++ b/exercises/markdown/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "markdown",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "comments": [
     "Markdown is a shorthand for creating HTML from text strings."
   ],
@@ -51,7 +51,7 @@
       "description": "unordered lists",
       "property": "parse",
       "input": "* Item 1\n* Item 2",
-      "expected": "<ul><li><p>Item 1</p></li><li><p>Item 2</p></li></ul>"
+      "expected": "<ul><li>Item 1</li><li>Item 2</li></ul>"
     },
     {
       "description": "With a little bit of everything",


Insti:
This change looks reasonable. 

I'm not sure putting the version in the PR subject is a good idea.

petertseng:
Honestly, no idea either.

I have a suspicion I got the idea from Linux kernel development, like https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?h=v4.13&id=569dbb88e80deb68974ef6fdd6a13edb9d686261. But in those cases, the version is the entirety of the commit message and the commit is only to change the version number. In that case, I'll just have to reason using reasons applicable to this project.

Only reason I thought of is to be able to tell when the version changed by looking only at the commit message. Does anyone need that?

Insti:
> Does anyone need that?

No, and anyone who does can `git log -G version`

petertseng:
> anyone who does can `git log -G version`

I would say I've successfully been convinced that version number in subject is unnecessary and does not need to be there.

At this point, I should invoke "perfection is attained not when you have nothing more to add, but nothing more to take away".

But I'll ask anyway. Next step is to convince that it is a bad idea to keep the version number there.

Don't worry, none of this discussion is unduly delaying the merging of this PR; I was wanting either someone who actually knows markdown to tell me I have not done something terribly wrong, or I guess if that never comes I'll look at what's going on in the tracks that have this exercise to see if someone knew at some point. So I'm just trying to pass the time until one of those two things can happen.

petertseng:
> someone who actually knows markdown to tell me I have not done something terribly wrong

It is possible that one of the two people who have looked at this PR does actually know markdown but just didn't say it explicitly enough. This sentence shouldn't be read as a claim that "psh, but neither of you *actually* know markdown"

Insti:
The version in the PR subject bugs me, so I took it out :)

petertseng:
* C#: Has never had `<p>` since initial https://github.com/exercism/csharp/pull/152
* CFML: Has `<p>` since initial https://github.com/exercism/cfml/pull/63
* Elixir: Has never had `<p>` since initial https://github.com/exercism/elixir/pull/249
* F#: Originally had `<p>` since initial https://github.com/exercism/fsharp/pull/167 but removed after https://github.com/exercism/fsharp/pull/258
* PHP: Has `<p>` since initial https://github.com/exercism/php/pull/87

By a failure of process, https://github.com/exercism/fsharp/pull/258 did not make it into problem-specifications as it should have.

Given the information I have, I believe this PR is correct.

PR #952, opened 2017-10-13
pangram: add a case with a perfect pangram for better TDD progression

Vankog:
Splitting #893 into single PRs.

Adding case with a "perfect pangram" for better TDD progression.
As suggested by ~~@petertseng (https://github.com/exercism/problem-specifications/pull/893/files#r144452355) and~~ @Insti (https://github.com/exercism/problem-specifications/pull/893#discussion_r137843067)

Edit: The suggestion from @petertseng above does not count towards this, because I misunderstood it for another issue.

### DIFF
--- a/exercises/pangram/canonical-data.json
+++ b/exercises/pangram/canonical-data.json
@@ -3,7 +3,7 @@
   "comments": [
     "A pangram is a sentence using every letter of the alphabet at least once."
   ],
-  "version": "1.1.0",
+  "version": "1.2.0",
   "cases": [
     {
       "description": "Check if the given string is an pangram",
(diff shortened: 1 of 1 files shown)


exercises/pangram/canonical-data.json
```diff
@@ -18,6 +18,12 @@
           "expected": false
         },
         {
+          "description": "recognizes a perfect lower case pangram.",
```
petertseng: presently the only description that ends in punctuation, until other descriptions are changed. This means that inevitably this will cause to discuss:

* whether the descriptions in this file should, in general, end in punctuation
* is the (temporary, right?) inconsistency OK

stevejb71: The ocaml test generator uses the description as the name of the test. Not a big deal, but I'd prefer no full stop.

Vankog:
done and squashed while rebasing.

Insti:
Great. Looks good.

PR #953, opened 2017-10-14
Simplify two tests so the difference is easily spotted.

Stargator:
I feel like the cases `"non-unique character in first strand"` and `"non-unique character in second strand"` are not clear in how the respective strand has a non-unique character.

So I suggest simplifying it. Otherwise, I would be fine with switching them as seen below. Where it's clear by the pattern of the other strand that the strand with the non-unique character doesn't follow the expected pattern.

```json
{
      "description": "non-unique character in first strand",
      "property": "distance",
      "strand1": "AGG",
      "strand2": "AGA",
      "expected": 1
    },
    {
      "description": "non-unique character in second strand",
      "property": "distance",
      "strand1": "AGA",
      "strand2": "AGG",
      "expected": 1
    },```

### DIFF
--- a/exercises/hamming/canonical-data.json
+++ b/exercises/hamming/canonical-data.json
@@ -1,6 +1,6 @@
 {
 "exercise": "hamming",
-"version": "2.0.0",
+"version": "2.0.1",
   "comments": [
     "Language implementations vary on the issue of unequal length strands.",
     "A language may elect to simplify this task by only presenting equal",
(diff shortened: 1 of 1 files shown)


petertseng:
I agree that minor version increment is correct because test input was changed, and it is a case listed in https://github.com/exercism/problem-specifications/blob/master/README.md#minor-version-changes.

I would also like to point out that this PR did not increase minor version, but increased patch version.

This doesn't matter for any track I maintain, so I don't need anything to change. If it matters for your track, you should change the minor version.

Insti:
> this PR did not increase minor version, but increased patch version.

I think that the change that happened was the correct one.

So it seems I disagree with the documentation.
See also #938

PR #954, opened 2017-10-14
twelve-days: Implement canonical-data.json

### LINKED ISSUE #590: twelve-days: Implement canonical-data.json
exercises in sync.

The Twelve Days exercise can be found in the [./exercises/twelve-days/][exercise-dir] directory.

See http://exercism.io/contribute/canonical-data/twelve-days for the up-to-date list of
language tracks that have an implementation of the Twelve Days exercise.

[exercise-dir]: https://github.com/exercism/x-common/tree/master/exercises/twelve-days

alpakhomov:
------
Resolves #590 // m-a-age

### DIFF
--- a/exercises/twelve-days/canonical-data.json
+++ b/exercises/twelve-days/canonical-data.json
@@ -1,0 +1,171 @@
+{
+  "exercise": "twelve-days",
+  "version": "1.0.0",
+  "comments": [
+    "JSON doesn't allow for multi-line strings, so all verses are presented ",
+    "here as arrays of strings. It's up to the test generator to join the ",
+    "lines together with line breaks."
+  ],
+  "cases": [
+    {
+      "description": "verse",
+      "cases": [
+        {
+          "description": "first day a partridge in a pear tree",
+          "property": "recite",
+          "startVerse": 1,
+          "endVerse": 1,
+          "expected": [
+            "On the first day of Christmas my true love gave to me, a Partridge in a Pear Tree."
+          ]
+        },
+        {
+          "description": "second day two turtle doves",
+          "property": "recite",
+          "startVerse": 2,
+          "endVerse": 2,
+          "expected": [
+            "On the second day of Christmas my true love gave to me, two Turtle Doves, and a Partridge in a Pear Tree."
+          ]
+        },
+        {
+          "description": "third day three french hens",
+          "property": "recite",
+          "startVerse": 3,
+          "endVerse": 3,
+          "expected": [
+            "On the third day of Christmas my true love gave to me, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree."
+          ]
+        },
+        {
+          "description": "fourth day four calling birds",
+          "property": "recite",
+          "startVerse": 4,
+          "endVerse": 4,
+          "expected": [
+            "On the fourth day of Christmas my true love gave to me, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree."
+          ]
+        },
+        {
+          "description": "fifth day five gold rings",
+          "property": "recite",
+          "startVerse": 5,
+          "endVerse": 5,
+          "expected": [
+            "On the fifth day of Christmas my true love gave to me, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree."
+          ]
+        },
+        {
+          "description": "sixth day six geese-a-laying",
+          "property": "recite",
+          "startVerse": 6,
+          "endVerse": 6,
+          "expected": [
+            "On the sixth day of Christmas my true love gave to me, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree."
+          ]
+        },
+        {
+          "description": "seventh day seven swans-a-swimming",
+          "property": "recite",
+          "startVerse": 7,
+          "endVerse": 7,
+          "expected": [
+            "On the seventh day of Christmas my true love gave to me, seven Swans-a-Swimming, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree."
+          ]
+        },
+        {
+          "description": "eighth day eight maids-a-milking",
+          "property": "recite",
+          "startVerse": 8,
+          "endVerse": 8,
+          "expected": [
+            "On the eighth day of Christmas my true love gave to me, eight Maids-a-Milking, seven Swans-a-Swimming, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree."
+          ]
+        },
+        {
+          "description": "ninth day nine ladies dancing",
+          "property": "recite",
+          "startVerse": 9,
+          "endVerse": 9,
+          "expected": [
+            "On the ninth day of Christmas my true love gave to me, nine Ladies Dancing, eight Maids-a-Milking, seven Swans-a-Swimming, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree."
+          ]
+        },
+        {
+          "description": "tenth day ten lords-a-leaping",
+          "property": "recite",
+          "startVerse": 10,
+          "endVerse": 10,
+          "expected": [
+            "On the tenth day of Christmas my true love gave to me, ten Lords-a-Leaping, nine Ladies Dancing, eight Maids-a-Milking, seven Swans-a-Swimming, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree."
+          ]
+        },
+        {
+          "description": "eleventh day eleven pipers piping",
+          "property": "recite",
+          "startVerse": 11,
+          "endVerse": 11,
+          "expected": [
+            "On the eleventh day of Christmas my true love gave to me, eleven Pipers Piping, ten Lords-a-Leaping, nine Ladies Dancing, eight Maids-a-Milking, seven Swans-a-Swimming, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree."
+          ]
+        },
+        {
+          "description": "twelfth day twelve drummers drumming",
+          "property": "recite",
+          "startVerse": 12,
+          "endVerse": 12,
+          "expected": [
+            "On the twelfth day of Christmas my true love gave to me, twelve Drummers Drumming, eleven Pipers Piping, ten Lords-a-Leaping, nine Ladies Dancing, eight Maids-a-Milking, seven Swans-a-Swimming, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree."
+          ]
+        }
+      ]
+    },
+    {
+      "description": "lyrics",
+      "cases": [
+        {
+          "description": "recites first three verses of the song",
+          "property": "recite",
+          "startVerse": 1,
+          "endVerse": 3,
+          "expected": [
+            "On the first day of Christmas my true love gave to me, a Partridge in a Pear Tree.",
+            "On the second day of Christmas my true love gave to me, two Turtle Doves, and a Partridge in a Pear Tree.",
+            "On the third day of Christmas my true love gave to me, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree."
+          ]
+        },
+        {
+          "description": "recites three verses from the middle of the song",
+          "property": "recite",
+          "startVerse": 4,
+          "endVerse": 6,
+          "expected": [
+            "On the fourth day of Christmas my true love gave to me, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree.",
+            "On the fifth day of Christmas my true love gave to me, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree.",
+            "On the sixth day of Christmas my true love gave to me, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree."
+          ]
+        },
+        {
+          "description": "recites the whole song",
+          "property": "recite",
+          "startVerse": 1,
+          "endVerse": 12,
+          "expected": [
+            "On the first day of Christmas my true love gave to me, a Partridge in a Pear Tree.",
+            "On the second day of Christmas my true love gave to me, two Turtle Doves, and a Partridge in a Pear Tree.",
+            "On the third day of Christmas my true love gave to me, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree.",
+            "On the fourth day of Christmas my true love gave to me, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree.",
+            "On the fifth day of Christmas my true love gave to me, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree.",
+            "On the sixth day of Christmas my true love gave to me, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree.",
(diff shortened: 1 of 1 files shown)


alpakhomov:
I'm not sure about my implementation of canonical-data.json. The problem is that a lot of language tracks test every single line of the song - like 1, 2, 3 and so on till 12. I don't understand the reason behind it - it looks redundant to me. So, I think it's important to test the first and the last lines (checking the offset if some array-like data structure is used). Then, we check that all other parts of the song are there when we will run tests for intervals (for example, from first to third line). So, I don't really know what should be a priority - sticking to test cases that already are in many language tracks, or creating a suite of essential tests.
And maybe I'm wrong about this "test redundancy" in general.

Insti:
With only 12 items it doesn't seem unreasonable to test every verse individually.

alpakhomov:
Added test cases for every verse and changed their descriptions accordingly

exercises/twelve-days/canonical-data.json
```diff
@@ -0,0 +1,96 @@
+{
+  "exercise": "twelve-days",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "returns verse number 1",
```
Insti: I'd like to see more descriptive descriptions. This does not tell me anything that cannot be automatically generated from the input.

How about:
"The first day's gift is a partridge in a pair tree"
"The second day's gift is two turtle doves"
etc.

exercises/twelve-days/canonical-data.json
```diff
@@ -0,0 +1,96 @@
...
+    {
+      "description": "returns verse number 12",
+      "property": "verse",
+      "input": [12],
+      "expected": "On the twelfth day of Christmas my true love gave to me, twelve Drummers Drumming, eleven Pipers Piping, ten Lords-a-Leaping, nine Ladies Dancing, eight Maids-a-Milking, seven Swans-a-Swimming, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree.\n"
+    },
+    {
+      "description": "finds multiple verses, which are a fragment of the song",
+      "property": "verses",
+      "input": [1, 3],
```
Insti: There is no test for verses that doesn't start at 1
what should happen in that case?

alpakhomov:
Thanks @Insti, I've changed test descriptions and added test for interval from the middle of the song (4-6). Also changed interval test description accordingly.

I have also rebased commits into one.

alpakhomov:
Thank you @petertseng 
About this newlines - I just copied them from current implementations, it looks like most people included them in their tests. But I don't know if there is any logic behind it.

ErikSchierboom:
> I am wondering whether the trailing newlines on all expected outputs are necessary. Naively I might assume they aren't, but maybe there is something I don't know.

I agree. It makes sense for the verses to be separated by newlines, but for an individual verse it doesn't IMHO. Maybe this is the point where we should remove those extra newlines?

alpakhomov:
Okay, I a bit confused now, so, just to be clear - we are talking about last newline symbol at the end of each verse/block of verses, right?

exercises/twelve-days/canonical-data.json
```diff
@@ -0,0 +1,102 @@
+{
+  "exercise": "twelve-days",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "first day a partridge in a pear tree",
+      "property": "verse",
+      "input": 1,
+      "expected": "On the first day of Christmas my true love gave to me, a Partridge in a Pear Tree.\n"
```
Insti: I believe the suggestion is that this line should not have the `\n` at the end.

exercises/twelve-days/canonical-data.json
```diff
@@ -0,0 +1,102 @@
...
+      "description": "twelfth day twelve drummers drumming",
+      "property": "verse",
+      "input": 12,
+      "expected": "On the twelfth day of Christmas my true love gave to me, twelve Drummers Drumming, eleven Pipers Piping, ten Lords-a-Leaping, nine Ladies Dancing, eight Maids-a-Milking, seven Swans-a-Swimming, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree.\n"
+    },
+    {
+      "description": "finds multiple verses, which are a fragment from the beggining of the song",
+      "property": "verses",
+      "input": [1, 3],
+      "expected": "On the first day of Christmas my true love gave to me, a Partridge in a Pear Tree.\n\nOn the second day of Christmas my true love gave to me, two Turtle Doves, and a Partridge in a Pear Tree.\n\nOn the third day of Christmas my true love gave to me, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree.\n"
```
Insti: But whenever there are multiple verses they should be separated by `\n\n`
But no trailing newline.

exercises/twelve-days/canonical-data.json
ErikSchierboom: Correct

exercises/twelve-days/canonical-data.json
ErikSchierboom: But it only makes sense in the context of multiple verses, so why add it to a single verse?

Insti: > Maybe the trailing single newline should be there.

In this case I am wondering about how many newlines should be at the end of the completed `verses` string. 0, 1 or 2? Currently there is 1.

exercises/twelve-days/canonical-data.json
alpakhomov: Removed trailing newline for single verse tests

exercises/twelve-days/canonical-data.json
ErikSchierboom: I would expect the following to hold:

1. A single verse would have no newline at the end
1. Verses (more than one) are separated by an empty line
1. Verses don't have trailing newlines

alpakhomov:
> A single verse would have no newline at the end
Verses (more than one) are separated by an empty line
Verses don't have trailing newlines (so that means 0)

Fixed

Insti:
Do you have an automated checker for this exercise @petertseng ?

exercises/twelve-days/canonical-data.json
```diff
@@ -0,0 +1,102 @@
...
+    {
+      "description": "twelfth day twelve drummers drumming",
+      "property": "verse",
+      "input": 12,
+      "expected": "On the twelfth day of Christmas my true love gave to me, twelve Drummers Drumming, eleven Pipers Piping, ten Lords-a-Leaping, nine Ladies Dancing, eight Maids-a-Milking, seven Swans-a-Swimming, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree."
+    },
+    {
+      "description": "finds multiple verses, which are a fragment from the beggining of the song",
+      "property": "verses",
+      "input": [1, 3],
```
petertseng: hmm, sorry to swoop in with another  thing again, but I just remembered... what about assigning meaning to the two elements of this array? See how `beer song` does it at https://github.com/exercism/problem-specifications/blob/master/exercises/beer-song/canonical-data.json#L54-L55 . Any desire to do the same here?

alpakhomov: Added two elements instead of one array; named them `beginning` and `end` respectively. Not sure if it will be a good idea to split all cases into two groups, like "beer-song" does (for single and multiple verse test cases).

Insti: should `beginning` and `end` be subkeys of input?
```
"input": {"beginning": 1, "end": 3}
```

alpakhomov: Should it? Wouldn't it be confused with some object or map data structure?

exercises/twelve-days/canonical-data.json
```diff
@@ -0,0 +1,105 @@
...
+      "expected": "On the twelfth day of Christmas my true love gave to me, twelve Drummers Drumming, eleven Pipers Piping, ten Lords-a-Leaping, nine Ladies Dancing, eight Maids-a-Milking, seven Swans-a-Swimming, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree."
+    },
+    {
+      "description": "finds multiple verses, which are a fragment from the beggining of the song",
+      "property": "verses",
+      "beginning": 1,
+      "end": 3,
+      "expected": "On the first day of Christmas my true love gave to me, a Partridge in a Pear Tree.\n\nOn the second day of Christmas my true love gave to me, two Turtle Doves, and a Partridge in a Pear Tree.\n\nOn the third day of Christmas my true love gave to me, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree."
+    },
+
```
rpottsoh: Line 84 should be removed.

exercises/twelve-days/canonical-data.json
```diff
@@ -0,0 +1,105 @@
...
+    {
+      "description": "eleventh day eleven pipers piping",
+      "property": "verse",
+      "input": 11,
+      "expected": "On the eleventh day of Christmas my true love gave to me, eleven Pipers Piping, ten Lords-a-Leaping, nine Ladies Dancing, eight Maids-a-Milking, seven Swans-a-Swimming, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree."
+    },
+    {
+      "description": "twelfth day twelve drummers drumming",
+      "property": "verse",
+      "input": 12,
```
rpottsoh: Instead of `"input"` how about `"number"` since it is the verse number being asked for.

rpottsoh:
If you would like to make this more `beer-song` like, you could group the `Verse` cases under a description of "Verse" and all the other cases under a description of "Lyrics".

alpakhomov:
Thanks @rpottsoh 
I've splitted cases into two groups (`verse` and `lyrics`), renamed `input` to `number`, and removed unnecessary empty line.

exercises/twelve-days/canonical-data.json
petertseng: > Should `beginning` and `end` be subkeys of input?

That would be an interesting change to make for the JSON Files all around. Then, instead of having to vary the input keys per exercise, test generators would know that all inputs can be found in `"input"`. That could be a useful change for the schema! I think the track maintainer who wishes that to happen for all JSON files should make the issue requesting it.

petertseng:
> Do you have an automated checker for this exercise

OK, guess that wasn't so much to write.

```ruby
require 'json'

DAYS = {
  twelfth: 'twelve Drummers Drumming',
  eleventh: 'eleven Pipers Piping',
  tenth: 'ten Lords-a-Leaping',
  ninth: 'nine Ladies Dancing',
  eighth: 'eight Maids-a-Milking',
  seventh: 'seven Swans-a-Swimming',
  sixth: 'six Geese-a-Laying',
  fifth: 'five Gold Rings',
  fourth: 'four Calling Birds',
  third: 'three French Hens',
  second: 'two Turtle Doves',
  first: 'a Partridge in a Pear Tree',
}.each_value(&:freeze).freeze

def list(things)
  things.size == 1 ? things[0] : (things[0...-1] << "and #{things[-1]}").join(', ')
end

def verse(n)
  ordinal = DAYS.keys[-n]
  gifts = DAYS.values.last(n)
  "On the #{ordinal} day of Christmas my true love gave to me, #{list(gifts)}."
end

def verses(a, b)
  (a..b).map { |n| verse(n) }.join("\n\n")
end

raise "There should be exactly two cases, not #{json['cases'].size}" if json['cases'].size != 2

verify(json['cases'][0]['cases'], property: 'verse') { |c|
  verse(c['number'])
}

subcases = by_property(json['cases'][1]['cases'], %w(verses sing))

verify(subcases['verses'], property: 'verses') { |c|
  verses(c['beginning'], c['end'])
}

verify(subcases['sing'], property: 'sing') { |c|
  verses(1, DAYS.size)
}
```

As written, it's fine.

Note that it may be a bit strange that both `verses` and `sing` are in the same group. As you can see though it can be dealt with.

Insti:
Should `verses` and `sing` have their own sections?

Insti:
Looking at the [description](https://github.com/exercism/problem-specifications/blob/master/exercises/twelve-days/description.md) and the policy that we should not be testing intermediate values it would seem that there should only be one test for the complete song.

alpakhomov:
I changed properties and descriptions of tests in the `lyrics` block - now there is only one test for the whole song, plus we can still have only two groups of test cases (`verse` and `lyrics`). I think it looks better now - but I can revert chages back if needed. 
About changes in exercise [description](https://github.com/exercism/problem-specifications/blob/master/exercises/twelve-days/description.md) - should it be done? If yes, do I need a separate pull request?

petertseng:
If we take this PR as is, this exercise is on par with:

* beer-song
  * whose https://github.com/exercism/problem-specifications/blob/master/exercises/beer-song/description.md only says to "produce the lyrics"
  * but whose https://github.com/exercism/problem-specifications/blob/master/exercises/beer-song/canonical-data.json checks individual verses
* house
  * whose https://github.com/exercism/problem-specifications/blob/master/exercises/house/description.md only says to "output the nursery rhyme"
  * but whose https://github.com/exercism/problem-specifications/blob/master/exercises/house/canonical-data.json checks individual verses
* food-chain
  * whose https://github.com/exercism/problem-specifications/blob/master/exercises/food-chain/description.md only says to "generate the lyrics"
  * but whose https://github.com/exercism/problem-specifications/blob/master/exercises/food-chain/canonical-data.json checks individual verses

Whatever treatment is given to this should also be given to all the others.

Insti:
Hi @alpakhomov, are you willing to make some more changes to this to match what we're planning on doing to food-chain in #973

`property`: was `verse` and `sing`, both become `recite`

`number`,`beginning`, `end`:  become `start verse` and `end verse`, each test case should have both. Where there was only one number they should `start verse` and `end verse` should be the same.

alpakhomov:
Changed property and parameter names, everything else is the same

ErikSchierboom:
Great work @alpakhomov ! What I was wondering, is if we should encoding the expected value as an array of strings instead of a single string, similar to what happens in the [food-chain canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/food-chain/canonical-data.json)? My personal preference would be to use an array. @Insti, what are your thoughts?

Insti:
> encoding the expected value as an array of strings

I have no strong opinion, an array of strings seems fine, so long as there is a comment about it for the test implementer: https://github.com/exercism/problem-specifications/blob/master/exercises/food-chain/canonical-data.json#L4-L7

exercises/twelve-days/canonical-data.json
```diff
@@ -0,0 +1,121 @@
...
+  "exercise": "twelve-days",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "verse",
+      "cases": [
+        {
+          "description": "first day a partridge in a pear tree",
+          "property": "recite",
+          "start verse": 1,
```
ErikSchierboom: We have now come up with an updated format, which requires the properties to be lowerCamelCase (see #983). Could you update the property to `startVerse`? And do the same for the other "start verse" properties?

exercises/twelve-days/canonical-data.json
```diff
@@ -0,0 +1,121 @@
...
+          "end verse": 1,
```
ErikSchierboom: We have now come up with an updated format, which requires the properties to be lowerCamelCase (see #983). Could you update the property to `endVerse`? And do the same for the other "end verse" properties?

exercises/twelve-days/canonical-data.json
```diff
@@ -0,0 +1,121 @@
...
+          "expected": "On the first day of Christmas my true love gave to me, a Partridge in a Pear Tree."
+        },
+        {
+          "description": "second day two turtle doves",
+          "property": "recite",
+          "start verse": 2,
+          "end verse": 2,
+          "expected": "On the second day of Christmas my true love gave to me, two Turtle Doves, and a Partridge in a Pear Tree."
```
ErikSchierboom: Could you encode the expected value as an array of strings? See #983 and #973. And do the same for the other "expected" properties?

ErikSchierboom:
@alpakhomov Sorry about the churn in the changing requirements! It has taken us a little while to come up with uniform specifications, but the final result is in #983 and it should now be stable.

alpakhomov:
Renamed parameters with camelCase, changed all expected values to arrays (even single line ones), added comment about expected values same as here https://github.com/exercism/problem-specifications/blob/master/exercises/food-chain/canonical-data.json#L4-L7

ErikSchierboom:
I think this is ready to merge. @Insti, do you agree?

Insti:
Yes it looks good to me.

ErikSchierboom:
Merged! Thanks a lot @alpakhomov!

Insti:
Thanks for all your work on this @alpakhomov and your patience with the constantly changing requirements :heart:

PR #955, opened 2017-10-15
leap: Improve test suite to cover corner case.

paulrex:
Currently, the test suite has a loophole where it is possible to pass
all the tests with a function that simply checks if the year is evenly
divisble by 16.

This commit changes one of the test cases so that is no longer possible.


Insti:
Thanks @paulrex :heart:

PR #951, opened 2017-10-13
nucleotide-count: Adding a single character case

Vankog:
Splitting #895 into single PRs.

Adding a single char input case for better TDD progression as agreed upon by @petertseng (https://github.com/exercism/problem-specifications/pull/895#pullrequestreview-69117017) and @Insti (https://github.com/exercism/problem-specifications/pull/895#discussion_r139292551)

PR #957, opened 2017-10-16
nucleotide-count: fix wrong order introduced in #951

Vankog:
For whatever reason, in #951 I added the single-char-input case after the multi-char-input case.

Obviously this is bonkers and needs to be the other way around.


Insti:
Thanks @Vankog

PR #958, opened 2017-10-16
pangram: fix duplicative case for a missing 'x'

Vankog:
Splitting #893 into multiple PRs.

Currently two cases check for a missing 'x'. This does not make much sense and the second case was changed to test for a missing 'h' instead, while 'h' was chosen randomly.

This improves the TDD progression and forces the user to check for more than just 'x'.


Vankog:
@Insti well, the big ones are discussions after all with provided proposals to discuss. I still think it was the right attempt to at least talk about it in the whole context.

PR #959, opened 2017-10-17
nucleotide-count: add case-agnostic test

Vankog:
Splitting #895 into multiple PRs.

This adds a test case that ensures that lower- and uppercase inputs do not matter and don't interfere with each other. You can use both interchangeably.

Reasoning:
1. "Postel's law" (a.k.a. "robustness principle")
https://en.wikipedia.org/wiki/Robustness_principle
1. Also, in everyday life, we got used to ignore cases. I see no reasoning why lower-case inputs should be in any way incorrect for this exercise.
1. The test is located near the end of the spec to cover some generic special cases as part of the finishing touches of the exercise.

### DIFF
--- a/exercises/nucleotide-count/canonical-data.json
+++ b/exercises/nucleotide-count/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "nucleotide-count",
-  "version": "1.2.0",
+  "version": "1.3.0",
   "cases": [
     {
       "description": "count all nucleotides in a strand",
@@ -42,6 +42,17 @@
           "description": "strand with multiple nucleotides",
           "property": "nucleotideCounts",
           "strand": "AGCTTTTCATTCTGACTGCAACGGGCAATATGTCTCTGTGTGGATTAAAAAAAGAGTGTCTGATAGCAGC",
+          "expected": {
+            "A": 20,
+            "C": 12,
+            "G": 17,
+            "T": 21
+          }
+        },
+        {
+          "description": "is case-agnostic against a strand of mixed-case nucleotides",
+          "property": "nucleotideCounts",
+          "strand": "aGcTtTtCaTtCtGaCtGcAaCgGgCaAtAtGtCtCtGtGtGgAtTaAaAaAaGaGtGtCtGaTaGcAgC",
           "expected": {
             "A": 20,
             "C": 12,


exercises/nucleotide-count/canonical-data.json
```diff
@@ -50,6 +50,17 @@
           }
         },
         {
+          "description": "is case-agnostic against a strand of mixed-case nucleotides",
+          "property": "nucleotideCounts",
+          "strand": "aGcTtTtCaTtCtGaCtGcAaCgGgCaAtAtGtCtCtGtGtGgAtTaAaAaAaGaGtGtCtGaTaGcAgC",
```
Insti: Would `AaCcGgTt` also be valid for this test?

Is there a reason why this needs to be so long and mixed up?
If I'm debugging my program I'd like to be able to manually verify the expected inputs and outputs.

NobbZ:
I've never seen those in lowercase when talking about nucleotides, therefore I'm against that change

exercises/nucleotide-count/canonical-data.json
Vankog: @Insti
The input was taken directly from the original case above ("strand with multiple nucleotides").
I wanted to keep it consistent.
Any other suggested input is welcomed as well.

rpottsoh:
I agree with @NobbZ unless a mixed case example can be sited I would expect there would be a screening process ahead of the solution to screen for bad input.

Vankog:
@NobbZ @rpottsoh 
Well, this brings us back to our discussion here: #902

NobbZ:
Well, only halfway… If it is a valid input, we need to test it, if it is invalid input then we need to discuss if it is worth testing as an additional invalid case or not compared to those invalids that currently exist…

Since I do not have domain knowledge about biology or biochemistry, I have to assume, that those written representations of DNA that I've seen so far were correct and canonical. Among those I've never seen any lower case letter. Therefore I also assume given inputs with a lowercase character as invalid.

If my assumption is right, we need to decide wether we should test invalids or not, but if we do, we should explicitely for lowercase nucleotides because of

> […] in everyday life, we got used to ignore cases.

rpottsoh:
[This](https://en.wikipedia.org/wiki/Nucleic_acid_sequence#/media/File:AMY1gene.png) is the best example I could quickly find that shows mixed case.

Vankog:
I found this:
http://www.sbcs.qmul.ac.uk/iubmb/misc/naseq.html

> As the present recommendations present unique alphabetic symbols for each nucleotide combination, the use of upper- and lower-case letters as equivalent does not lead to confusion. However, such use may cause confusion between r (ribo-) and R (purine), and care must be taken in those rare cases where the various symbols are used in combination. In general, it should be emphasised (i) that upper-case symbols are advocated. and (ii) that the present recommendations are not intended to prejudice any possible future use of contrasting upper- and lower-case letters for specific purposes.

> There is, however, a problem of discriminating between the upper-case letters G and C on poorly copied sequences. Nevertheless, the use of alternative symbols for G (such as a barred-G, ) is not recommended. Discrimination between the lower-case letters is much clearer.

exercises/nucleotide-count/canonical-data.json
Insti: I also object to the case above, but wasn't asking you to change that one.

Vankog:
@Insti Sorry, am I mistaken? @NobbZ said we should check if mix-case is a valid use-case for this domain and act accordingly. And @rpottsoh found a valid case, so it seems he might agree with this assessment. Further research shows that mixed case is indeed a common use-case for DNA-strands.

I am not sure what to make out of this (besides of choosing a better input).

Insti:
@Vankog what is this exercise about?
What is the main aspect of this exercise that is important?

Vankog:
To me the main aspect is to count the number of occurrences of certain nucleotides. 

How these nucleotides are represented is part of the domain specific language. With that in mind it turned out the nucleotides in question are usually presented inside sequences composed of A, C, G, T in upper- as well as lower-case and even mixed-case.

That's why I want to make sure this domain specific requirement does not shake up the user's solution. Hence the test. Further, I think we should also (or: "at least") update the description to match what we will agree on in the end.

Insti:
> the main aspect is to count the number of occurrences of certain nucleotides.

I agree with this.

> the nucleotides in question are usually presented inside sequences composed of A, C, G, T 

And this.

> as well as lower-case and even mixed-case.

This is where we diverge. The description specifies the characters A, C, G, T. These are only ones we need to use in our test inputs, and testing for anything else is an unnecessary distraction.

Vankog:
that is why I suggest to adjust the description as well.

Dysp:
In all papers I've read and in text books, uppercase is used, but as it is pointed out above, there is no rule against using lower case. But I honestly do not think it is of great importance whether it's lower or uppercase from a bio scientific point of view.

Insti:
@Vankog if you feel that the description could be improved, feel free to create a separate PR that addresses the issue.

I do not think there is any problem with the description. It clearly specifies the 4 symbols to be used.

Vankog:
Im on vacation for the next two weeks. I'll take care of his then.

Insti:
I'll close this PR for now, we can re-open it again later if the PR to the description makes it relevant again in the future.

Thanks for your contributions :heart:

Have a good holiday.

PR #963, opened 2017-10-18
pascals-triangle: Added test prototypes and more README info.

hekrause:
In relation to [this](https://github.com/exercism/problem-specifications/pull/362) pull request, where it was mentioned that fewer test greater 4 rows would not be needed.

To see the structure of the algorithm I wrote myself a few additional tests because I had my difficulties with getting a grip on the "pascals-triangle" exercise.

Maybe they are not necessary to solve this exercise but they do no harm.
I implemented the tests in [this](https://github.com/exercism/rust/pull/366) pull request and now made the corresponding changes in this repository.


rpottsoh:
@hekrause I haven't looked this all over, but it appears that you have an extra file in the commit that doesn't belong.

hekrause:
@rpottsoh You are right! Some left overs from another issue. :-)

exercises/pascals-triangle/description.md
```diff
@@ -4,10 +4,11 @@ In Pascal's Triangle each number is computed by adding the numbers to
...
-   1 1
-  1 2 1
- 1 3 3 1
-1 4 6 4 1
+     1
+    1 1
+   1 2 1
+  1 3 3 1
+ 1 4 6 4 1
+1 5 10 10 5 1
```
petertseng: the alignment between this row and the one above is no longer effective because 10 has two digits

Insti:
Thanks for writing a good description of the PR :heart:

exercises/pascals-triangle/description.md
```diff
@@ -4,10 +4,11 @@ In Pascal's Triangle each number is computed by adding the numbers to
...
-   1 1
-  1 2 1
- 1 3 3 1
-1 4 6 4 1
+      1
+     1 1
+    1 2 1
+   1 3 3 1
+  1 4 6 4 1
+1 5 10 10 5 1
```
Insti: Why do you want to add this row?

I think the first 5 rows demonstrate the triangle sufficently, and adding 2 digit numbers makes the formatting difficult. The current formatting is not correctly triangular.

Insti:
I go back and forth on whether adding test cases that can be passed without changing any code is a good idea or not, but in this case adding the extra cases 5,6,10 doesn't seem too bad.

petertseng:
reminder: version 1.2.0

I am kind of thinking the way I would test pascal's triangle would be to test that `∀n in [0..20] pascals_triangle(n) == precomputed_pascals_triangle[0..n)` or something. I would probably tire of writing all of the cases in the JSON file but I am thinking that's how I would write the test file anyway. Shrug. So I guess I don't care about whether this is here.

exercises/pascals-triangle/description.md
hekrause: Sorry for not replying on your comment.
Ah i wrote this line for me because i wanted simply more example data and then thought it would make no big difference because somehow the alignment is nearly symmetric.
@Insti So you think it is better to leave it unchanged?

Insti: Yes, unchanged is better.

petertseng: when there were only single-digit numbers, it was the case that no number is directly above another.

but right now, there is a 1 directly above each 5, and there is a 4 directly above each 10

so this example is not currently showing the nature of the triangle

It is possible that this could be fixed. If it were, that would be better.

some possible solutions involve an unequal number of spaces in each gap in the same row. the resulting inconsistency raises questions about whether the relationship is different simply because of the difference in representation.

If a sixth row is to be added, such concerns should be addressed

ErikSchierboom: I agree that adding the line makes reading the triangle harder, while not providing much benefit I feel. @hekrause would you be opposed to removing the extra line?

hekrause: Ok, I reset the description in the exercism/problem-specification repository. Can we change the description in this repository automatic or do i have to make the same changes in this repository by hand?

hekrause:
@Insti I reset the description.

Insti:
Thanks @hekrause :heart:

PR #969, opened 2017-10-23
nth-prime: Use object instead of bool for invalid value

ErikSchierboom:
This is more in line with how other exercises indicate error cases.

### DIFF
--- a/exercises/nth-prime/canonical-data.json
+++ b/exercises/nth-prime/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "nth-prime",
-  "version": "2.0.0",
+  "version": "1.1.0",
   "cases": [
     {
       "description": "first prime",
--- a/exercises/nth-prime/canonical-data.json
+++ b/exercises/nth-prime/canonical-data.json
@@ -1,11 +1,6 @@
 {
   "exercise": "nth-prime",
-  "version": "1.0.0",
-  "comments": [
-    "The tests that expect 'false' should be implemented to raise",
-    "an error, or indicate a failure. Implement this in a way that",
-    "makes sense for your language."
-  ],
+  "version": "2.0.0",
   "cases": [
     {
       "description": "first prime",
@@ -35,7 +30,7 @@
       "description": "there is no zeroth prime",
       "property": "prime",
       "input": 0,
-      "expected": false
+      "expected": { "error": "there is no zeroth prime" }
     }
   ]
 }
--- a/exercises/nth-prime/canonical-data.json
+++ b/exercises/nth-prime/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "nth-prime",
-  "version": "1.1.0",
+  "version": "2.0.0",
   "cases": [
     {
       "description": "first prime",


exercises/nth-prime/canonical-data.json
```diff
@@ -1,11 +1,6 @@
 {
   "exercise": "nth-prime",
-  "version": "1.0.0",
-  "comments": [
-    "The tests that expect 'false' should be implemented to raise",
-    "an error, or indicate a failure. Implement this in a way that",
-    "makes sense for your language."
-  ],
+  "version": "2.0.0",
```
Insti: This change should not cause the version to change to `2.0.0`.
(This opinion may differ from the documented procedure, see also: #938

Insti:
The actual change is good 👍

exercises/nth-prime/canonical-data.json
ErikSchierboom: @Insti Ah, I missed that discussion. I'll make it a minor version update then.

ErikSchierboom: @Insti Updated.

stkent: Should we not actually update the documented procedure before introducing deviations on a per-exercise basis?

petertseng: I'm going to pretend that this practice (of changing version numbers in contravention of the written meanings) disturbs me deeply and write this comment accordingly.

If we follow the written rules, even if we disagree with them, we at least know what each version change means.

Now that we are openly flaunting the rules, now we have no idea what each version change means. (It is possible they **do** have a consistent meaning, but the meaning is not written down)

I cannot* stand by this behaviour because I need my version numbers to have meaning. With these meaningless version numbers, my burden as a maintainer has increased unacceptably.

*: I actually don't care, since I don't have a track for which it matters in what precise way the tests were updated. Right now the only property I need to hold true is that the version changes in some way when the file changes... which I guess I could have achieved using commit shas.

ErikSchierboom: @petertseng @stkent @Insti What should I do with this versioning problem?

stkent: Even on a track that uses not-generated tests, I don't really care about the difference between a major and a minor change, so I say move forward with this as-is and maybe separately open an issue to begin discussion of the current versioning guidelines?

petertseng: I guess I should have made clear that you don't need to ask me what to do before merging, but OK.

As I explained, I cannot accept any version change that does not correspond to a written versioning scheme that explains what a change in each component means. Either use 2.0.0 as the current versioning scheme would have it, or explain in writing the new versioning scheme.

petertseng: EBWODP:

> I guess I should have made clear that you don't need to ask me what to do before merging

I will clarify that the following sentence in a previous comment was the attempt to make that clear:

>  I actually don't care

petertseng: > does not correspond to a written versioning scheme

Sorry, is the versioning scheme used here that written in https://github.com/exercism/problem-specifications/issues/938#issue-264031108? It wasn't clear to me because that proposal says to remove the third component, which was not removed, so I rejected the possibility that it was that scheme.

Insti:
Discussion about versioning and updating the documented guidelines can continue in #938.

PR #970, opened 2017-10-23
saddle-points: Remove extraneous 'the' from description

N-Parsons:
I noticed a small typo in the description for `saddle-points`. This PR fixes the error.

### DIFF
--- a/exercises/saddle-points/description.md
+++ b/exercises/saddle-points/description.md
@@ -13,7 +13,7 @@ So say you have a matrix like so:
 It has a saddle point at (1, 0).
 
 It's called a "saddle point" because it is greater than or equal to
-every element in its row and the less than or equal to every element in
+every element in its row and less than or equal to every element in
 its column.
 
 A matrix may have zero or more saddle points.


exercises/saddle-points/description.md
```diff
@@ -13,7 +13,7 @@ So say you have a matrix like so:
 It has a saddle point at (1, 0).
 
 It's called a "saddle point" because it is greater than or equal to
-every element in its row and the less than or equal to every element in
+every element in its row and less than or equal to every element in
```
rpottsoh: I think this is OK, or you could replace `the` with `is` as well.

N-Parsons: I feel like while adding an `is` would be fine, it's not really necessary and it changes the flow of the sentence. I'm happy to change it if people think that it would be better, though.

stkent: I like it without the is also :)

rpottsoh: My reasoning behind perhaps considering adding `is` because then the pattern structure of the sentence is the same as on line 15.  It is a minor detail, I don't think it is wrong to leave the change as is as well..

jpreese:
Looks good to me as well, thanks @N-Parsons !

PR #971, opened 2017-10-24
leap: Fix loophole in unit tests

cruxicheiros:
Currently, it's possible to pass all the unit tests by checking if `(year % 100) == (year % 400)`.

This commit changes one of the test cases so that this is no longer possible. Note that this is a separate issue to #955 -- however, 1996, like 2020, is not divisible by 16, so that loophole will remain closed.

### DIFF
--- a/exercises/leap/canonical-data.json
+++ b/exercises/leap/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "leap",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "cases": [
     {
       "description": "year not divisible by 4: common year",
@@ -11,7 +11,7 @@
     {
       "description": "year divisible by 4, not divisible by 100: leap year",
       "property": "leapYear",
-      "input": 2020,
+      "input": 1996,
       "expected": true
     },
     {


petertseng:
> it's possible to pass all the unit tests by checking if `(year % 100) == (year % 400)`

I was unable to verify that this statement is true.

The first case is 2015.

`2015 % 100 == 2015 % 400` is `false == false` is true, whereas the expected answer is false. So a solution that checks whether `(year % 100) == (year % 400)` will fail, as it should.

Did you mean to cite a different failing solution?

petertseng:
EBWODP

> a different failing solution?

I mean a different improperly-passing solution

petertseng:
oh wait hold on

petertseng:
EBWODP:

> 2015 % 100 == 2015 % 400 is false == false is true, whereas the expected answer is false. So a solution that checks whether (year % 100) == (year % 400) will fail, as it should.

This entire sentence is garbage. It's supposed to be:

2015 % 100 == 2015 % 400 is 15 == 15 is true, whereas the expected answer is false. So a solution that checks whether (year % 100) == (year % 400) will fail, as it should.

cruxicheiros:
@petertseng [PR #979 in exercism/python](https://github.com/exercism/python/pull/979)

petertseng:
So, the solution in question also checks `year % 4 == 0` so as to be able to correctly reject 2015, which was not stated in the PR.

This means that the commit message was incorrect. We all need to be more careful that commit messages are correct.

I considered rewriting master to amend the commit message, but I suppose it's too discourteous to ask everyone who already acquired that commit to reset it away.

I considered recommending that the commit be reverted, then reapplied with a correct message, but perhaps as long as GitHub continues to exist, this written record is sufficient, but now needs an internet connection to be able to see it.

cruxicheiros:
I've edited the body of the pull request so that it contains the correct information for anyone who checks on GitHub.

In future PRs I will triple-check that what I am writing is correct. Again, I apologise.

petertseng:
👍

Insti:
> In future PRs I will triple-check that what I am writing is correct. Again, I apologise.

We all miss things sometimes. @stkent, @cmccandless and @rpottsoh also missed this.

It's the responsibility of the reviewers to help catch these issues so they can be fixed before the pull request is merged. That's why we have reviews.

PR #972, opened 2017-10-25
dominoes: improve description readability

cmccandless:
Rework domino representation to avoid confusion with 2-digit numbers `23` vs. a domino `(2, 3)`.

### DIFF
--- a/exercises/dominoes/description.md
+++ b/exercises/dominoes/description.md
@@ -5,9 +5,9 @@ correct domino chain (the dots on one half of a stone match the dots on the
 neighbouring half of an adjacent stone) and that dots on the halfs of the stones
 which don't have a neighbour (the first and last stone) match each other.
 
-For example given the stones `(2, 1)`, `(2, 3)` and `(1, 3)` you should compute something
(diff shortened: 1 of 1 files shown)


exercises/dominoes/description.md
```diff
@@ -5,9 +5,9 @@ correct domino chain (the dots on one half of a stone match the dots on the
 neighbouring half of an adjacent stone) and that dots on the halfs of the stones
 which don't have a neighbour (the first and last stone) match each other.
 
-For example given the stones `21`, `23` and `13` you should compute something
-like `12 23 31` or `32 21 13` or `13 32 21` etc, where the first and last numbers are the same.
+For example given the stones `(2, 1)`, `(2, 3)` and `(1, 3)` you should compute something
```
Insti: Would square brackets be more appropriate here?
And is "comma space" the best separator?

Possible alternatives:
`[1, 1]` `[2,2]` `[3|3]` `[4:4]`

Insti:
Moving away from 2 digit numbers is an improvement 👍

cmccandless:
I like [3|3].

"Tuple" format was chosen because many languages can implement this will naturally using tuples, so there representation is quite clear.

Insti:
> "Tuple" format was chosen because many languages can implement this will naturally using tuples, so there representation is quite clear.

That makes sense but we don't necessarily want the description to be prompting specific implementation details.

PR #973, opened 2017-10-26
food-chain: Make properties less ambiguous

ErikSchierboom:
Currently, the canonical data uses the same property to get a single verse and to get multiple verses. I think it would make sense to instead use a different property for multiple verses, aptly named `verses` (plural). 

Furthermore, the single verse cases used `"start verse"`  as the input property, but to me that suggests also having an `"end verse"` property, which it doesn't. I've therefore renamed the property for the single verse cases to just `"verse"`.

### DIFF
--- a/exercises/food-chain/canonical-data.json
+++ b/exercises/food-chain/canonical-data.json
@@ -17,9 +17,9 @@
       "cases": [
         {
           "description": "fly",
-          "property": "verses",
-          "start verse": 1,
-          "end verse": 1,
+          "property": "recite",
+          "startVerse": 1,
+          "endVerse": 1,
           "expected": [
             "I know an old lady who swallowed a fly.",
             "I don't know why she swallowed the fly. Perhaps she'll die."
@@ -27,9 +27,9 @@
         },
         {
           "description": "spider",
-          "property": "verses",
-          "start verse": 2,
-          "end verse": 2,
+          "property": "recite",
+          "startVerse": 2,
+          "endVerse": 2,
           "expected": [
             "I know an old lady who swallowed a spider.",
             "It wriggled and jiggled and tickled inside her.",
@@ -39,9 +39,9 @@
         },
         {
           "description": "bird",
-          "property": "verses",
-          "start verse": 3,
-          "end verse": 3,
+          "property": "recite",
+          "startVerse": 3,
+          "endVerse": 3,
           "expected": [
             "I know an old lady who swallowed a bird.",
             "How absurd to swallow a bird!",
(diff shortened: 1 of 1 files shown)


Insti:
What about only having a single `verses` property and requesting verses(from n to n) when you only want one?

See also: https://github.com/exercism/problem-specifications/pull/954#issuecomment-337577142 (and down.)

ErikSchierboom:
> What about only having a single verses property and requesting verses(from n to n) when you only want one?

Ah, so that means single verse case should just have the start and end verse set to the same value then? I like it.

ErikSchierboom:
@Insti I've updated the PR.

Insti:
In this case I think the version bump to 2.0.0 is appropriate, due to the change in the properties that are being tested.

rpottsoh:
It seems strange to me to have a method that implies it wants to return more than 1 thing only returning one thing for most of the tests.  This exercise is essentially the same as `beer-song` and `twelve-days`.  I am trying to think of a better name than `verses` for the property name that would be better suited for both returning a single verse or multiple verses.  I got nothing though.  :frowning_face: 

Why does this exercise have to be different than the two I mentioned above?  Should they also be changed?  Potential upside, if these changes are merged, then at least different solutions will probably come out of this exercise than the other two.

I am not against the idea behind the changes, and my negative feeling about the property naming is minor, but I see no harm in throwing my thoughts out there to see what might happen.

Insti:
> I am trying to think of a better name than verses for the property name that would be better suited for both returning a single verse or multiple verses. 

`sing`? `song`? `lyrics`?

> `beer-song` ... `twelve-days` ...  [`house`] Should they also be changed?

Probably. [@petertseng seems to think so](https://github.com/exercism/problem-specifications/pull/954#issuecomment-338585738).

 The thinking about what makes a "good" exercise has evolved since they were originally written. The current thinking seems to be to keep exercises more focused on single methods/functions/properties and to avoid steering the implementation too much by testing intermediate methods.

rpottsoh:
`sing` :+1:

Insti:
I like `recite`

ErikSchierboom:
I like `recite` too. Any objections to `recite`?

Insti:
> It seems strange to me to have a method that implies it wants to return more than 1 thing only returning one thing for most of the tests

By being explicit in the canonical data we can maintain a single consistent unambiguous interface for all tests.

If a track wants to implement tests for a function that takes differing numbers of arguments or call different methods if there is only one argument they can still do this using the provided properties.  There is no requirement for a track to implement a 1-1 mapping from canonical data properties to test inputs.

ErikSchierboom:
I've updated the PR to conform to #983.

Insti:
Thanks @ErikSchierboom :heart:

PR #975, opened 2017-10-27
two-fer: remove the section on test-driven development

### LINKED ISSUE #949: Don't assume two-fer is the second exercise.
Ref. exercism/bash#86

Don't assume two-fer is the second exercise. Remove the section on test driven development and maybe put it somewhere else? What is the consensus on this?

Smarticles101:
closes #949


Smarticles101:
@rpottsoh I think you're right, it might be good to add it somewhere else

PR #979, opened 2017-10-29
forth: Add tests for case-insensitivity

N-Parsons:
This PR add tests for case insensitivity to `forth` to better match the problem description. Previously there was some testing for case-insensitivity, but it was limited to `dup` - I have removed this in favour of a block of tests for each word and type of word at the end of the canonical data.

I'm not sure whether "copies a value on the stack" on line 152 is really necessary - I'm happy to remove it if so desired.

The updated version number assumes the merge of #976 (removal of the empty test case `[]`) before this one.

### DIFF
--- a/exercises/forth/canonical-data.json
+++ b/exercises/forth/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "forth",
-  "version": "1.3.0",
+  "version": "1.4.0",
   "comments": [
     "The cases are split into multiple sections, all with the same structure.",
     "In all cases, the `expected` key is the resulting stack",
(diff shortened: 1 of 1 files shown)


exercises/forth/canonical-data.json
```diff
@@ -311,6 +311,53 @@
...
+            "1 FOO Foo foo"
+          ],
+          "expected": [1, 1, 1, 1]
+        },
+        {
+          "description": "overridden words are case-insensitive",
+          "property": "evaluate",
+          "input": [
+            ": swap dup ;",
+            "1 SWAP Swap swap"
```
petertseng: I find it unlikely that any implementation will fail this test if it's the case that they check that user-defined words are case-insensitive, since at this point swap is a user-defined word, but I'm happy to take your word for it.

I would consider something that checks that definition bodies are case-insensitive, I suppose? `: swap DUP Dup dup ;`, etc.

N-Parsons: @petertseng, that's a good point - would you be happy with the below alternative? I could envisage an implementation treating definitions separately and forgetting to normalise the case, and I think that this should catch any such oversight.

```
: SWAP DUP Dup dup ;
1 swap
```

petertseng: 👍

Insti:
> The updated version number assumes the merge of #976 (removal of the empty test case []) before this one.

There appears to be a merge conflict due to the version number change.

ErikSchierboom:
@N-Parsons Would you be willing to try and fix the merge conflict?

N-Parsons:
@ErikSchierboom, should be fixed now. Sorry for the delay - I've had a busy week.

ErikSchierboom:
Merged. Thanks @N-Parsons!

Issue #983, opened 2017-10-30
recitation exercises: Standardise canonical data interface.


Insti:
There are 4 exercises that involve reciting songs/poems.
`beer-song`, `food-chain`, `house` and `twelve-days`.

Currently these are tested using differing combinations of `verse` and `verses`.

There has been some discussion in recent PRs (#973, #954) about standardising the interfaces specified in the canonical data. With the current thinking that:

---

The property we are testing should be `recite`
There should be 2 other properties present: `startVerse` and `endVerse`
Both `startVerse` and `endVerse` must be specified in the canonical-data.json file, even if only one verse is desired.

---

By being explicit about `startVerse` and `endVerse` in the canonical data we can maintain a single consistent unambiguous interface for all tests.

There is no requirement for a language track to implement a one-to-one mapping from canonical data properties to test inputs.

If a track wants to implement tests for a function that takes differing numbers of arguments or call different methods if there is only one argument, or use a name other than `recite` they can still do this by applying minor transformations to the provided canonical properties. 

**To do:** 

Standardise the canonical data for the following exercises:

- [x] beer-song (Complete #984)
  * [description](https://github.com/exercism/problem-specifications/blob/master/exercises/beer-song/description.md) says to "produce the lyrics"
  * [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/beer-song/canonical-data.json) checks `verse` and `verses`
- [x] house (Complete #985)
  * [description](https://github.com/exercism/problem-specifications/blob/master/exercises/house/description.md) says to "output the nursery rhyme"
  * [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/house/canonical-data.json) checks `verse` and `verses`
- [x] food-chain  (Complete #973)
  * [description](https://github.com/exercism/problem-specifications/blob/master/exercises/food-chain/description.md) says to "generate the lyrics"
  * [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/food-chain/canonical-data.json) checks `verse` sometimes with one argument, sometimes with two
- [x] twelve-days (Complete #954)
  * [description](https://github.com/exercism/problem-specifications/blob/master/exercises/twelve-days/description.md) says to "Output the lyrics"
  * canonical-data.json does not exist yet. 

Additional recitation exercises:
- [x] Proverb (Complete #994)
[...]

PR #984, opened 2017-10-30
beer-song: Update to test property recite

robkeim:
Addresses #983

Note:
- The end verse value is numerically less than the start verse for multi-verse tests which may seem counter-intuitive but makes sense given the song counts down from 99 to 0

@Insti FYI


robkeim:
After re-reading the issue I need to update to `"start verse"` and `"end verse"` instead of `"startVerse"` and `"endVerse"`

Insti:
Would `start bottles` and `end bottles` make more sense for this exercise?

rpottsoh:
> Would start bottles and end bottles make more sense for this exercise?

I think that it would.  I reviewed my solution to this problem. (I should have done that before I submitted my review.)

My implementation passes in the number of bottles and not the verse number, and thus effectively passes in start bottles and end bottles.

I will update my review of this PR.

robkeim:
`start bottles` and `end bottles` would make more sense although it's not consistent with the other three verse exercises then. It's probably a better approach than having to do math to figure out which verse corresponds to how many bottles.

exercises/beer-song/canonical-data.json
```diff
@@ -10,32 +10,37 @@
           "cases": [
             {
               "description": "first generic verse",
-              "property": "verse",
-              "number": 99,
+              "property": "recite",
+              "start verse": 99,
+              "end verse": 99,
```
rpottsoh: Consider `startBottles` and `endBottles` instead.  Same for all occurrences later in this files.

exercises/beer-song/description.md
```diff
@@ -1,4 +1,4 @@
-Produce the lyrics to that beloved classic, that field-trip favorite: 99 Bottles of Beer on the Wall.
```
rpottsoh: `lyrics` was OK.

robkeim:
@Insti and @rpottsoh updated to `startBottles` and `endBottles`.

exercises/beer-song/canonical-data.json
```diff
@@ -11,36 +11,36 @@
...
               "description": "last generic verse",
               "property": "recite",
-              "start verse": 3,
-              "end verse": 3,
+              "startBottles": 3,
+              "endBottles": 3,
               "expected": "3 bottles of beer on the wall, 3 bottles of beer.\nTake one down and pass it around, 2 bottles of beer on the wall.\n"
             },
             {
               "description": "verse 2",
```
rpottsoh: Since we are not strictly talking verses anymore, perhaps this should be:
```text
"description": "2 bottles of beer",
```
or something to the affect.

Similar changes would be needed for the other cases.

robkeim: Updated!

exercises/beer-song/description.md
Insti: > #983 calls for standardizing the descriptions as well as the property name 

Hmm I didn't notice that :)
I like lyrics better than verses here tbh.

exercises/beer-song/description.md
```diff
@@ -1,4 +1,4 @@
...
+Recite the verses to that beloved classic, that field-trip favorite: 99 Bottles of Beer on the Wall.
```
Insti: The change to `Recite` is good.

"Recite the lyrics" is better (and more consistent) than "Recite the verses"

robkeim: Done

exercises/beer-song/canonical-data.json
```diff
@@ -10,32 +10,37 @@
           "cases": [
             {
               "description": "first generic verse",
-              "property": "verse",
-              "number": 99,
+              "property": "recite",
+              "startBottles": 99,
+              "endBottles": 99,
```
Insti: There are 98 bottles left on the wall at the end...

exercises/beer-song/canonical-data.json
```diff
@@ -10,32 +10,37 @@
...
               "expected": "1 bottle of beer on the wall, 1 bottle of beer.\nTake it down and pass it around, no more bottles of beer on the wall.\n"
             },
             {
-              "description": "verse 0",
-              "property": "verse",
-              "number": 0,
+              "description": "verse with 0 bottles",
+              "property": "recite",
+              "startBottles": 0,
+              "endBottles": 0,
```
Insti: There are 99 bottles on the wall at the end.

Maybe startBottles and endBottles doesn't make sense for this problem.

startBottles and takeDowns? perhaps.

(Trying to work out what's best here, not advocating changes just yet.)

Insti: `takeBottles`? `bottlesToTake`?

Insti: Whatever it is we're diverging from the common specification, but I think that is OK as this is a separate exercise with its own nuances.

exercises/beer-song/description.md
Insti: thanks.

exercises/beer-song/canonical-data.json
robkeim: For the first two verses are you suggesting:
```
"startBottles": 99,
"takeDowns": 2,
```

or
```
"startBottles": 99,
"takeDowns": 98,
```

The first makes more sense to me but that's moving away in consistency with the other exercises unfortunately. If we're going to head down that direction, I think `numVerses` would be more explicit.

rpottsoh:
Grabbing lunch....  you bring up a good point.  TakeDown would be less disruptive than endBottles. 



Sent from my android device.

exercises/beer-song/canonical-data.json
robkeim: Woops... just read your other comments you're one step ahead of me :)

rpottsoh:
> TakeDown would be less disruptive than endBottles.

I have to correct myself. The above is not a true statement.  The way the exercise is implemented is the start number of bottles and the resulting number of bottles are passed in as parameters.  So I say stick with `startBottles` and `endBottles`.

exercises/beer-song/canonical-data.json
rpottsoh: Should the words in the lyrics dictate that naming of the properties?

When we get down to zero beers on the wall, the only reason there ends up being 99 bottles of beer on the wall is because the lyrics say to go buy more beer.

To trigger a trip to the store, `endBottles` has to be zero.

Insti: I'm actually suggesting: 

```
"startBottles": 99,
"takeDowns": 1,
```

Insti:
> The way the exercise is implemented is the start number of bottles and the resulting number of bottles are passed in as parameters. So I say stick with startBottles and endBottles.

That is an unnecessary implementation detail.

The test needs to specify, the number of bottles that begin on the wall `startBottles` and the number of verses to output. There are various ways to specify this, but using the language of the song seems like the most appropriate, therefore `take` `n` (down).

This also works in the zero wrap around case where we can start with 0 and take 1 down and end at 99 again. If we were to take more than 1 we could still expect the right answer.

If you REALLY wanted to maintain consistency (which you shouldn't) the thing to do would be to update the other exercises to take `startVerses` and `numberOfVerses` parameters.
(But don't do this.)

In this case breaking consistency with the way the other exercises are implemented is a good thing.

rpottsoh:
@Insti I am coming around to your way of seeing this exercise.  These changes accentuate the perpetual nature behind this song.....  It never ends.  :smile:

Insti:
With `takeDown` one could also specify a test that was 
```
"startBottles": 99,
"takeDown": 200,
```
And ensure that it looped through twice. This would not be possible with the start/end pair.

(I don't recommend adding this test either.)

robkeim:
@Insti and @rpottsoh I've updated to `takeDown`. Please double check I didn't make a silly off-by-one error :)

Insti:
Looks good. 👍 
One last thing..
In [food-chain](https://github.com/exercism/problem-specifications/pull/973) and [twelve-days](https://github.com/exercism/problem-specifications/pull/954) we return an array of strings rather than a single newline separated string. Can you add another commit that converts the expected values to that format please?

Insti:
@robkeim, Just waiting on this one now. If you don't have time we can merge this PR and do the line adjustments as a new PR.

robkeim:
@Insti I should be able to get to this on Sunday. You can either merge this as is or wait for that update.

robkeim:
@Insti I couldn't sleep so I took advantage of the extra time to wrap-up this PR :) It should be ready for you to merge.

Insti:
Thanks for all your work on this @robkeim :heart:

robkeim:
You're welcome @Insti, thanks for the feedback!

PR #988, opened 2017-10-30
zipper: fix word order in description

jgr:
The adjective phrase "purely functional" should go before the noun "way" in this sentence.


petertseng:
I didn't see a justification given in https://github.com/exercism/exercism.io/pull/964 why it has to stay the way it currently is, so this means it's less likely that anyone missed anything.

petertseng:
Thanks.

PR #991, opened 2017-10-31
custom-set: add test for inequality of subset

cmccandless:
Consider the following pseudocode
```
equals(self, other)
    if self.empty
        return other.empty
    else
        for x in self
            if !other.contains(x)
                return false
        return true
```

This solution would return a false positive for [1, 2, 3] == [1, 2, 3, 4], but would pass all other tests for `equals`.

### DIFF
--- a/exercises/custom-set/canonical-data.json
+++ b/exercises/custom-set/canonical-data.json
@@ -179,6 +179,13 @@
           "set1": [1, 2, 3],
           "set2": [1, 2, 4],
           "expected": false
+        },
+        {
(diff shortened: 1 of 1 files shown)


Insti:
Good title 👍 
Good issue description 👍 
Good test addition 👍 
Good test description 👍 

Remember to update the version number at the top of the file.

cmccandless:
@Insti With all the discussion lately about versioning, I'm not sure whether to bump minor or build...?

Insti:
Me either ¯\_(ツ)_/¯

I'd bump the middle one.

PR #994, opened 2017-11-03
proverb: Add canonical data.

### LINKED ISSUE #578: proverb: Implement canonical-data.json
exercises in sync.

The Proverb exercise can be found in the [./exercises/proverb/][exercise-dir] directory.

See http://exercism.io/contribute/canonical-data/proverb for the up-to-date list of
language tracks that have an implementation of the Proverb exercise.

[exercise-dir]: https://github.com/exercism/x-common/tree/master/exercises/proverb

coriolinus:
This is based on the Rust tests as implemented in exercism/rust#380.
The lack of canonical data had resulted in odd custom tests being
designed, which was reported in exercism/rust#379.

Closes #578.

### DIFF
--- a/exercises/proverb/canonical-data.json
+++ b/exercises/proverb/canonical-data.json
@@ -1,0 +1,69 @@
+{
+    "cases": [
+        {
+            "cases": [
+                {
+                    "description": "zero pieces",
+                    "expected": "",
+                    "input": [],
+                    "property": "construct"
+                },
+                {
+                    "description": "one piece",
+                    "expected": "And all for the want of a nail.",
+                    "input": [
+                        "nail"
+                    ],
+                    "property": "construct"
+                },
+                {
+                    "description": "two pieces",
+                    "expected": "For want of a nail the shoe was lost.\nAnd all for the want of a nail.",
+                    "input": [
+                        "nail",
+                        "shoe"
+                    ],
+                    "property": "construct"
+                },
+                {
+                    "description": "three pieces",
+                    "expected": "For want of a nail the shoe was lost.\nFor want of a shoe the horse was lost.\nAnd all for the want of a nail.",
+                    "input": [
+                        "nail",
+                        "shoe",
+                        "horse"
+                    ],
+                    "property": "construct"
+                },
+                {
+                    "description": "full proverb",
+                    "expected": "For want of a nail the shoe was lost.\nFor want of a shoe the horse was lost.\nFor want of a horse the rider was lost.\nFor want of a rider the message was lost.\nFor want of a message the battle was lost.\nFor want of a battle the kingdom was lost.\nAnd all for the want of a nail.",
+                    "input": [
+                        "nail",
+                        "shoe",
+                        "horse",
+                        "rider",
+                        "message",
+                        "battle",
+                        "kingdom"
+                    ],
+                    "property": "construct"
+                },
+                {
+                    "description": "three pieces modernized",
+                    "expected": "For want of a pin the gun was lost.\nFor want of a gun the soldier was lost.\nFor want of a soldier the battle was lost.\nAnd all for the want of a pin.",
+                    "input": [
+                        "pin",
+                        "gun",
+                        "soldier",
+                        "battle"
+                    ],
+                    "property": "construct"
+                }
+            ],
+            "description": "Construct a proverb of the form 'for want of an X, the Y was lost'"
+        }
+    ],
+    "exercise": "proverb",
+    "version": "1.0.0"
+}
--- a/exercises/proverb/canonical-data.json
+++ b/exercises/proverb/canonical-data.json
@@ -2,97 +2,93 @@
     "exercise": "proverb",
     "version": "1.0.0",
     "comments": [
-       "JSON doesn't allow for multi-line strings, so all expected outputs are presented ",
-       "here as arrays of strings. It's up to the test generator to join the ",
-       "lines together with line breaks."
+        "JSON doesn't allow for multi-line strings, so all expected outputs are presented ",
+        "here as arrays of strings. It's up to the test generator to join the ",
+        "lines together with line breaks. ",
+        "All cases expect the student to construct a proverb of the form ",
+        "'For want of an X, the Y was lost.', terminating with the stock phrase ",
+        "'And all for the want of a Z.'"
     ],
     "cases": [
         {
-            "cases": [
-                {
-                    "description": "zero pieces",
-                    "expected": [
-                        ""
-                    ],
-                    "input": [],
-                    "property": "recite"
-                },
-                {
-                    "description": "one piece",
-                    "expected": [
-                        "And all for the want of a nail."
-                    ],
-                    "input": [
-                        "nail"
-                    ],
-                    "property": "recite"
-                },
-                {
-                    "description": "two pieces",
-                    "expected": [
-                        "For want of a nail the shoe was lost.",
-                        "And all for the want of a nail."
-                    ],
-                    "input": [
-                        "nail",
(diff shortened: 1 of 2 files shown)


Insti:
A couple of things that came up during https://github.com/exercism/problem-specifications/issues/983 might be appropriate here:

* Name the method `recite`
* The expected output should be an array of strings without newlines.
* Make sure the exercise and version properties are at the top.

See [the canonical-data.json for food-chain](https://github.com/exercism/problem-specifications/blob/master/exercises/food-chain/canonical-data.json) as an example.

exercises/proverb/canonical-data.json
```diff
@@ -0,0 +1,69 @@
...
+                        "horse",
+                        "rider",
+                        "message",
+                        "battle",
+                        "kingdom"
+                    ],
+                    "property": "construct"
+                },
+                {
+                    "description": "three pieces modernized",
```
Insti: There seem to be four pieces in the array.

Insti:
It's useful to link to the description: https://github.com/exercism/problem-specifications/blob/master/exercises/proverb/description.md

exercises/proverb/canonical-data.json
```diff
@@ -0,0 +1,69 @@
...
+                        "nail",
+                        "shoe",
+                        "horse"
+                    ],
+                    "property": "construct"
+                },
+                {
+                    "description": "full proverb",
+                    "expected": "For want of a nail the shoe was lost.\nFor want of a shoe the horse was lost.\nFor want of a horse the rider was lost.\nFor want of a rider the message was lost.\nFor want of a message the battle was lost.\nFor want of a battle the kingdom was lost.\nAnd all for the want of a nail.",
+                    "input": [
```
Insti: This method of input is not really specified in the description and is an implementation detail.

We either need to: 
* Update the description to specify that you will be provided with a list of items.
* Change the tests so we're just testing for the specified verses. (see also food-chain.)

I think in order to differentiate this exercise from `food-chain` it might be good to use the list of things version.

cc: @ErikSchierboom

ErikSchierboom: @Insti I agree that having a different input here makes for a more interesting exercise. The existing exercises are quite similar, so having some variety is nice.

Insti: So the description should also be updated to clarify the problem.

coriolinus:
@Insti Good ideas all; I've incorporated all of them.

exercises/proverb/canonical-data.json
```diff
@@ -0,0 +1,98 @@
...
+       "JSON doesn't allow for multi-line strings, so all expected outputs are presented ",
+       "here as arrays of strings. It's up to the test generator to join the ",
+       "lines together with line breaks."
+    ],
+    "cases": [
+        {
+            "cases": [
+                {
+                    "description": "zero pieces",
+                    "expected": [
```
petertseng: I was considering to ask that this should have zero elements, but given that this is converted to a string by joining all elements with newlines, then `[""]` and `[]` are both representations of the empty string, so do whatever you want.

exercises/proverb/canonical-data.json
```diff
@@ -0,0 +1,98 @@
...
+                    "input": [
+                        "pin",
+                        "gun",
+                        "soldier",
+                        "battle"
+                    ],
+                    "property": "recite"
+                }
+            ],
+            "description": "Construct a proverb of the form 'for want of an X, the Y was lost'"
```
petertseng: I wonder if someone will miss this at the crucial time because it's on the bottom. But I don't think it's necessary at all. I would remove it and remove one level of `cases` nesting.

rpottsoh: The description does say that the lines of the proverb could vary, only the "input".

exercises/proverb/canonical-data.json
```diff
@@ -2,97 +2,93 @@
...
+                "gun",
+                "soldier",
+                "battle"
             ],
-            "description": "Construct a proverb of the form 'for want of an X, the Y was lost'"
+            "expected": [
+                "For want of a pin the gun was lost.",
+                "For want of a gun the soldier was lost.",
+                "For want of a soldier the battle was lost.",
+                "And all for the want of a pin."
```
rpottsoh: Is it to be presumed that the proverb is not set in stone?  The description only refers to a single proverb and makes not mention that the `"input"` provided might refer to a different proverb.  I don't think `"four pieces modernized"` should be included unless the description is also updated.

rpottsoh: @coriolinus my apologies!  I didn't read closely enough.  You are passing the expected proverb.

coriolinus: @rpottsoh I did update the description [for that](https://github.com/exercism/problem-specifications/pull/994/files#diff-a82423102a1295ada2179ae4ce091d03R15): "your solution should be able to handle lists of arbitrary length and content". 

In short, the idea is to ensure that students won't just hard-code the entire poem as a list, and then return its last N items, where N is the length of the input list. The goal of the exercise is to ensure that they can manipulate strings.

This also helps distinguish this exercise from others; discussion [here](https://github.com/exercism/problem-specifications/pull/994#discussion_r148731929) was in favor of keeping the input list.

exercises/proverb/canonical-data.json
```diff
@@ -0,0 +1,94 @@
+{
+    "exercise": "proverb",
+    "version": "1.0.0",
+    "comments": [
+        "JSON doesn't allow for multi-line strings, so all expected outputs are presented ",
+        "here as arrays of strings. It's up to the test generator to join the ",
+        "lines together with line breaks. ",
+        "All cases expect the student to construct a proverb of the form ",
+        "'For want of an X, the Y was lost.', terminating with the stock phrase ",
+        "'And all for the want of a Z.'"
```
wolf99: Looks like `Z` is a special case for _one_ of the inputs - could be worth calling that out specifically in the description.md

exercises/proverb/canonical-data.json
rpottsoh: Thanks @coriolinus for clearing this up for me.  I have not worked this exercise yet so some of the what was getting past me.  Thanks for taking the time to produce the canonical data for this exercise.

exercises/proverb/canonical-data.json
coriolinus: No? I don't understand your comment; all inputs of non-zero length terminate with that stock phrase.

wolf99: The description.md doesn't specifically draw any explicit connection between the last line and the first input. This especially is confusing if you are aware of the proverb IRL, where the inputs are always the same and the last line is always "And all for the want of a nail".

Going solely by the description, it seems like Z should always be "nail". Yes, the data implicitly makes this a requirement in a single case - but surely the data should be informed by the requirement (i.e. the description) rather than it introducing a requirement by itself?

What I mean to say is that by the description it is obvious that the inputs affect X and Y, but it is not obvious that they affect Z; Z is really X[0] so to speak.

I should have added this note against the description.md rather than this line I realise 😅

coriolinus: I see now what you mean, though I'm not sure that I agree: the obvious terminating phrase, to me, would always be `f"And all for the want of a {input[0]}."`. Still, I don't mind adding a bit of text to the description to make that more clear.

Insti: The `description.md` specifies the problem so I think you can just remove these 3 lines (8-10) from the `config.json`.

Insti:
Thanks @coriolinus :heart:

PR #948, opened 2017-10-12
kindergarten-garden: grammatical improvements to description.md

rpottsoh:

PR #995, opened 2017-11-04
description.md: rm trailing spaces

petertseng:
One might assume that we can consider a CI check that there are no
trailing spaces, but we have to be careful because some embedded strings
that serve as example outputs are expected to have trailing spaces. The
currently only example is ocr-numbers. One might consider a CI check,
but simply exclude that file.

### DIFF
--- a/exercises/kindergarten-garden/description.md
+++ b/exercises/kindergarten-garden/description.md
@@ -7,8 +7,8 @@ actual dirt, and grow actual plants.
 
 They've chosen to grow grass, clover, radishes, and violets.
 
-To this end, the children have put little cups along the window sills, and 
-planted one type of plant in each cup, choosing randomly from the available 
+To this end, the children have put little cups along the window sills, and
+planted one type of plant in each cup, choosing randomly from the available
 types of seeds.
 
 ```text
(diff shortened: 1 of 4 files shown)


exercises/kindergarten-garden/description.md
```diff
@@ -7,8 +7,8 @@ actual dirt, and grow actual plants.
 
 They've chosen to grow grass, clover, radishes, and violets.
 
-To this end, the children have put little cups along the window sills, and 
-planted one type of plant in each cup, choosing randomly from the available
```
rpottsoh: Oops, that's my handy work from #948.  My bad!  Thanks @petertseng

exercises/kindergarten-garden/description.md
```diff
@@ -23,7 +23,7 @@ There are 12 children in the class:
 - Eve, Fred, Ginny, Harriet,
 - Ileana, Joseph, Kincaid, and Larry.
 
-Each child gets 4 cups, two on each row. Their teacher assigns cups to 
+Each child gets 4 cups, two on each row. Their teacher assigns cups to
```
rpottsoh: same as before

petertseng:
OK, the only one of these that I think is potentially controversial is the `rectangles` one. The spaces make it so that each line in the individual-rectangle diagrams is the same length as each line in the original diagram. I don't think this is necessary because it's for human consumption and does not in any way indicate what the output should look like (the output is a number!) so I still think I'm justified. So I am not really thinking I should wait on this one.

We did remove trailing whitespace in https://github.com/exercism/problem-specifications/pull/915/commits/33d9ada8b0d926a0577770981914024aa4956f84 though I also note that that commit did not remove the trailing whitespace on `rectangle` when it had a chance to do so! So, potentially deliberate! But I have stated my reason.

Issue #577, opened 2017-02-13
protein-translation: Implement canonical-data.json


kytrinyx:
We want to have a standard set of test inputs and outputs for each exercise to
make it easier to port them to new languages, as well as to help keep the
exercises in sync.

The Protein Translation exercise can be found in the [./exercises/protein-translation/][exercise-dir] directory.

The step-by-step instructions for how to compile a canonical-data.json file is
described in [this section of the contributing guide][contributing].

See http://exercism.io/contribute/canonical-data/protein-translation for the up-to-date list of
language tracks that have an implementation of the Protein Translation exercise.

The [tracking issue][tracking-issue] has more context.

[contributing]: https://github.com/exercism/docs/blob/master/you-can-help/improve-exercise-metadata.md#extracting-canonical-test-data
[tracking-issue]: https://github.com/exercism/x-common/issues/552
[exercise-dir]: https://github.com/exercism/x-common/tree/master/exercises/protein-translation

petertseng:
Here is a thing to consider for whoever takes this on:

As you will see in #268, we deprecated an exercise called nucleotide-codons in favour of protein-translation. Both these exercises deal with converting codons into protein names.

The exercises worked in slightly different ways, which I summarise but you can see the original analysis at https://github.com/exercism/xrust/issues/267#issuecomment-286613950.

* nucleotide-codons introduced in #93 introduced the concept of compressions.
* protein-translation introduced in #37 acts on multiple codons at once.

When creating the canonical data for this exercise, you should consider which of these two concepts to include.

ErikSchierboom:
Closed by #997.

PR #997, opened 2017-11-08
protein-translation: Add canonical data.

sjwarner-bp:
Add protein translation canonical data as per #577.

After going to implement this exercise in Java, and realising that there is not a `canonical-data.json` for the problem, I decided to do this. I have helped a friend complete this exercise on the Python track, and believe the Python tests to be very comprehensive, so I have used them as my main source of data.

Please let me know if this is too long for a `canonical-data.json`, as there are about 20 tests...

### DIFF
--- a/exercises/protein-translation/canonical-data.json
+++ b/exercises/protein-translation/canonical-data.json
@@ -11,127 +11,127 @@
       "cases": [
         {
           "description": "Methionine RNA sequence is identified",
-          "property": "translate",
+          "property": "translate_codon",
           "strand": "AUG",
           "expected": "Methionine"
         },
         {
           "description": "Phenylalanine RNA sequence 1 is identified",
-          "property": "translate",
+          "property": "translate_codon",
           "strand": "UUU",
           "expected": "Phenylalanine"
         },
         {
           "description": "Phenylalanine RNA sequence 2 is identified",
-          "property": "translate",
+          "property": "translate_codon",
           "strand": "UUC",
           "expected": "Phenylalanine"
         },
         {
           "description": "Leucine RNA sequence 1 is identified",
-          "property": "translate",
+          "property": "translate_codon",
           "strand": "UUA",
           "expected": "Leucine"
         },
         {
           "description": "Leucine RNA sequence 2 is identified",
-          "property": "translate",
+          "property": "translate_codon",
           "strand": "UUG",
           "expected": "Leucine"
         },
         {
           "description": "Serine RNA sequence 1 is identified",
-          "property": "translate",
+          "property": "translate_codon",
           "strand": "UCU",
           "expected": "Serine"
         },
         {
           "description": "Serine RNA sequence 2 is identified",
-          "property": "translate",
+          "property": "translate_codon",
           "strand": "UCC",
           "expected": "Serine"
         },
         {
           "description": "Serine RNA sequence 3 is identified",
-          "property": "translate",
+          "property": "translate_codon",
           "strand": "UCA",
           "expected": "Serine"
         },
         {
           "description": "Serine RNA sequence 4 is identified",
-          "property": "translate",
+          "property": "translate_codon",
           "strand": "UCG",
           "expected": "Serine"
         },
         {
           "description": "Tyrosine RNA sequence 1 is identified",
-          "property": "translate",
+          "property": "translate_codon",
           "strand": "UAU",
           "expected": "Tyrosine"
         },
         {
           "description": "Tyrosine RNA sequence 2 is identified",
-          "property": "translate",
+          "property": "translate_codon",
           "strand": "UAC",
           "expected": "Tyrosine"
         },
         {
           "description": "Cysteine RNA sequence 1 is identified",
-          "property": "translate",
+          "property": "translate_codon",
           "strand": "UGU",
           "expected": "Cysteine"
         },
         {
           "description": "Cysteine RNA sequence 2 is identified",
-          "property": "translate",
+          "property": "translate_codon",
           "strand": "UGC",
           "expected": "Cysteine"
         },
         {
           "description": "Tryptophan RNA sequence is identified",
-          "property": "translate",
+          "property": "translate_codon",
           "strand": "UGG",
           "expected": "Tryptophan"
         },
         {
           "description": "STOP codon RNA sequence is identified",
-          "property": "translate",
+          "property": "translate_codon",
           "strand": "UAA",
           "expected": "STOP"
         },
         {
           "description": "STOP codon RNA sequence is identified",
-          "property": "translate",
+          "property": "translate_codon",
           "strand": "UAG",
           "expected": "STOP"
         },
         {
           "description": "STOP codon RNA sequence is identified",
-          "property": "translate",
+          "property": "translate_codon",
           "strand": "UGA",
           "expected": "STOP"
         },
         {
           "description": "Translate RNA strand into correct protein list",
-          "property": "translate",
+          "property": "translate_rna",
           "strand": "AUGUUUUGG",
           "expected": ["Methionine","Phenylalanine","Tryptophan"]
         },
         {
           "description": "Translation stops if STOP codon present 1",
-          "property": "translate",
+          "property": "translate_rna",
           "strand": "AUGUUUUAA",
           "expected": ["Methionine","Phenylalanine"]
         },
         {
           "description": "Translation stops if codon present 2",
-          "property": "translate",
+          "property": "translate_rna",
           "strand": "UGGUGUUAUUAAUGGUUU",
           "expected": ["Tryptophan","Cysteine","Tyrosine"]
         },
         {
           "description": "Test invalid codons",
-          "property": "translate",
+          "property": "translate_codon",
           "strand": "CARROT",
           "expected": {
             "error": "Invalid nucleotide in strand"
--- a/exercises/protein-translation/canonical-data.json
+++ b/exercises/protein-translation/canonical-data.json
@@ -94,22 +94,22 @@
           "expected": ["Tryptophan"]
         },
         {
-          "description": "STOP codon RNA sequence",
+          "description": "STOP codon RNA sequence 1",
           "property": "translate",
           "strand": "UAA",
-          "expected": ["STOP"]
+          "expected": []
         },
         {
-          "description": "STOP codon RNA sequence",
+          "description": "STOP codon RNA sequence 2",
           "property": "translate",
           "strand": "UAG",
-          "expected": ["STOP"]
+          "expected": []
         },
         {
-          "description": "STOP codon RNA sequence",
+          "description": "STOP codon RNA sequence 3",
           "property": "translate",
           "strand": "UGA",
-          "expected": ["STOP"]
+          "expected": []
         },
         {
           "description": "Translate RNA strand into correct protein list",
@@ -121,13 +121,13 @@
           "description": "Translation stops if STOP codon at end of sequence",
           "property": "translate",
           "strand": "AUGUUUUAA",
-          "expected": ["Methionine","Phenylalanine", "STOP"]
+          "expected": ["Methionine","Phenylalanine"]
         },
         {
           "description": "Translation stops if codon in middle of sequence",
           "property": "translate",
           "strand": "UGGUGUUAUUAAUGGUUU",
-          "expected": ["Tryptophan","Cysteine","Tyrosine", "STOP"]
+          "expected": ["Tryptophan","Cysteine","Tyrosine"]
         },
         {
           "description": "Test invalid codons",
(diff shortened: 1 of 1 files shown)


exercises/protein-translation/canonical-data.json
```diff
@@ -0,0 +1,143 @@
...
+          "description": "STOP codon RNA sequence is identified",
+          "property": "translate",
+          "strand": "UGA",
+          "expected": "STOP"
+        },
+        {
+          "description": "Translate RNA strand into correct protein list",
+          "property": "translate",
+          "strand": "AUGUUUUGG",
+          "expected": ["Methionine","Phenylalanine","Tryptophan"]
```
cmccandless: The type returned by `translate` should be consistent; until this line, the output of `translate` has been a string, but now a list is expected. The python test cases use two different properties `of_codon` and `of_rna` for these; something similar would be appropriate.

exercises/protein-translation/canonical-data.json
```diff
@@ -0,0 +1,143 @@
+{
+  "exercise": "protein-translation",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "Translate input RNA sequences into proteins",
+      "comments": [
+        " Returns the name of the protein if given RNA is valid, "
```
ErikSchierboom: Maybe lose the extra space in the beginning of the string?

exercises/protein-translation/canonical-data.json
```diff
@@ -0,0 +1,143 @@
...
+      , " else throws an error. "
```
ErikSchierboom: Maybe lose the extra space in the beginning of the string?

exercises/protein-translation/canonical-data.json
```diff
@@ -0,0 +1,143 @@
...
+      ],
+      "cases": [
+        {
+          "description": "Methionine RNA sequence is identified",
+          "property": "translate_codon",
```
ErikSchierboom: Lately, we have tried to have move away from testing intermediate results in the canonical data, and only test end results. For this canonical data, I would suggest to do the same and only have a single `translate` property to test. The existing `translate_codon` property tests (like this one) would then use the `translate` property and have as its expected value an array with a single value.

exercises/protein-translation/canonical-data.json
```diff
@@ -0,0 +1,143 @@
...
+        },
+        {
+          "description": "STOP codon RNA sequence is identified",
+          "property": "translate_codon",
+          "strand": "UGA",
+          "expected": "STOP"
+        },
+        {
+          "description": "Translate RNA strand into correct protein list",
+          "property": "translate_rna",
```
ErikSchierboom: In keeping in line with my comment above on testing only one property, I would suggest renaming this to `translate`.

sjwarner-bp:
Oh dear - I've made changes but the Travis build has failed. Do I need to add the new property names (`translate_codon`,`translate_rna`) somewhere?

exercises/protein-translation/canonical-data.json
sjwarner-bp: I didn't think of this, but it makes much more sense. I'll work on that.

cmccandless:
@sjwarner-bp Regarding the failed Travis build, this from the [canonical-schema](https://github.com/exercism/problem-specifications/blob/e93924e30cc78384c692305314d47f6cdb86914b/canonical-schema.json#L114-L118):

```json
"property":
        { "description": "A letters-only, lowerCamelCase property name"
        , "type"       : "string"
        , "pattern"    : "^[a-z]+([A-Z][a-z]+)*[A-Z]?$"
        },
```

exercises/protein-translation/canonical-data.json
rpottsoh: This feels a lot like what has been discussed recently at #983.  "translate_codon" seems to be an intermediate test of what the description is really asking for.  `"The translate_codon"` may not belong.  @petertseng, @Insti thoughts?

cmccandless: @rpottsoh I believe @ErikSchierboom beat you to it. See his comments.

sjwarner-bp:
That makes sense - I'll keep this in mind for future.

exercises/protein-translation/canonical-data.json
```diff
@@ -0,0 +1,143 @@
...
+          "expected": ["Methionine","Phenylalanine","Tryptophan"]
+        },
+        {
+          "description": "Translation stops if STOP codon present 1",
+          "property": "translate_rna",
+          "strand": "AUGUUUUAA",
+          "expected": ["Methionine","Phenylalanine"]
+        },
+        {
+          "description": "Translation stops if codon present 2",
```
ErikSchierboom: "Translation stops if codon" -> "Translation stops if STOP codon present 2"

Maybe describe how this case differs from the first one?

exercises/protein-translation/canonical-data.json
```diff
@@ -0,0 +1,143 @@
...
+          "expected": ["Methionine"]
+        },
+        {
+          "description": "Phenylalanine RNA sequence 1 is identified",
+          "property": "translate",
+          "strand": "UUU",
+          "expected": ["Phenylalanine"]
+        },
+        {
+          "description": "Phenylalanine RNA sequence 2 is identified",
```
ErikSchierboom: I am wondering if there is another way to discern between the different Phenylalanine types? They are currently named "Phenylalanine RNA sequence 1" and "Phenylalanine RNA sequence 2", but maybe there is a more descriptive name? If there isn't, just ignore this :)

exercises/protein-translation/canonical-data.json
```diff
@@ -0,0 +1,143 @@
...
+  "cases": [
+    {
+      "description": "Translate input RNA sequences into proteins",
+      "comments": [
+        "Returns the name of the protein if given RNA is valid, "
+      , "else throws an error. "
+      ],
+      "cases": [
+        {
+          "description": "Methionine RNA sequence is identified",
```
ErikSchierboom: I think the " is identified" postfix can be removed, as it doesn't add much extra information I think. But maybe someone else disagrees?

exercises/protein-translation/canonical-data.json
```diff
@@ -118,13 +118,13 @@
           "expected": ["Methionine","Phenylalanine","Tryptophan"]
         },
         {
-          "description": "Translation stops if STOP codon present 1",
+          "description": "Translation stops if STOP codon present in short string",
           "property": "translate",
           "strand": "AUGUUUUAA",
           "expected": ["Methionine","Phenylalanine"]
```
cmccandless: `STOP` is included in the above test cases where it is the only codon; either the above cases should expected empty lists as a result or `STOP` should be included in the results here.

exercises/protein-translation/canonical-data.json
```diff
@@ -118,13 +118,13 @@
           "expected": ["Methionine","Phenylalanine","Tryptophan"]
         },
         {
-          "description": "Translation stops if STOP codon present 1",
+          "description": "Translation stops if STOP codon present in short string",
```
cmccandless: Perhaps `Translation stops if STOP codon at end of sequence` vs below `Translation stops if STOP codon in middle of sequence`?

exercises/protein-translation/canonical-data.json
sjwarner-bp: That makes sense to me - concise is good!

exercises/protein-translation/canonical-data.json
sjwarner-bp: From the problem description, I couldn't see a particularly good way of differentiating the two from each other - maybe if someone knows biology they could chime in, otherwise this might be as good as we can get it.

exercises/protein-translation/canonical-data.json
```diff
@@ -118,13 +118,13 @@
...
           "property": "translate",
           "strand": "AUGUUUUAA",
           "expected": ["Methionine","Phenylalanine"]
         },
         {
-          "description": "Translation stops if codon present 2",
+          "description": "Translation stops if codon present in longer string",
```
cmccandless: Per my above comment, perhaps a change in wording?

cmccandless:
Love how fast this one's going!

@sjwarner-bp please see my latest comments, I'm afraid they might've been covered up by the flurry of activity.

exercises/protein-translation/canonical-data.json
rpottsoh: or
`Translation stops when STOP codon encountered`

sjwarner-bp:
I think I got them - I've added 'STOP' to the end of each list where a 'STOP' codon is involved.

I also amended the description for the test of early termination - hope this was right! :slightly_smiling_face:

exercises/protein-translation/canonical-data.json
cmccandless: Still need to distinguish between this case and the next... perhaps `Translation stops when STOP codon encountered at end of sequence` vs `Translation stops when STOP codon encountered in middle of sequence`?

exercises/protein-translation/canonical-data.json
```diff
@@ -118,7 +118,7 @@
           "expected": ["Methionine","Phenylalanine","Tryptophan"]
         },
         {
-          "description": "Translation stops if STOP codon present in short string",
+          "description": "Translation stops if STOP codon at end of sequence",
           "property": "translate",
           "strand": "AUGUUUUAA",
           "expected": ["Methionine","Phenylalanine", "STOP"]
```
rpottsoh: Should "STOP" be included in "expected"?  The stop example in the README didn't include stop in the resulting protein.

sjwarner-bp: This was added because of @cmccandless point (above) regarding consistency between tests. If "STOP" is expected in the tests for "STOP" codons, it might make sense to include them in the list of expected proteins?

Equally, it might not make too much sense - the Python track does not include them in protein lists, and only identifies them if they are input as a single codon.

rpottsoh: Perhaps then:
```text
"strand":  "UAA",
"expected": ""
```

rpottsoh: or update the README to list the STOP protein in the stop example....

cmccandless: I think it would be better to expect an empty list in the case of `"expected": ["STOP"]`

sjwarner-bp: Ok, empty list coming up!

exercises/protein-translation/canonical-data.json
```diff
@@ -0,0 +1,143 @@
...
+          "expected": ["Methionine","Phenylalanine"]
+        },
+        {
+          "description": "Translation stops if codon in middle of sequence",
+          "property": "translate",
+          "strand": "UGGUGUUAUUAAUGGUUU",
+          "expected": ["Tryptophan","Cysteine","Tyrosine"]
+        },
+        {
+          "description": "Test invalid codons",
```
Insti: Don't test invalid codons.

exercises/protein-translation/canonical-data.json
```diff
@@ -0,0 +1,143 @@
...
+    {
+      "description": "Translate input RNA sequences into proteins",
+      "comments": [
+        "Returns the name of the protein if given RNA is valid, "
+      , "else throws an error. "
+      ],
+      "cases": [
+        {
+          "description": "Methionine RNA sequence",
+          "property": "translate",
```
Insti: What about `proteins` as the property?

exercises/protein-translation/canonical-data.json
```diff
@@ -0,0 +1,143 @@
...
+          "expected": []
+        },
+        {
+          "description": "STOP codon RNA sequence 2",
+          "property": "translate",
+          "strand": "UAG",
+          "expected": []
+        },
+        {
+          "description": "STOP codon RNA sequence 3",
```
Insti: I'd like to see tests for more of the simple STOP codon sequences.
`UAGUGG`
`UGGUAG`
`UGGUAGUGG`

exercises/protein-translation/canonical-data.json
```diff
@@ -0,0 +1,143 @@
...
+        {
+          "description": "Translation stops if STOP codon at end of sequence",
+          "property": "translate",
+          "strand": "AUGUUUUAA",
+          "expected": ["Methionine","Phenylalanine"]
+        },
+        {
+          "description": "Translation stops if codon in middle of sequence",
+          "property": "translate",
+          "strand": "UGGUGUUAUUAAUGGUUU",
```
Insti: Put some thought put into which sequences are easiest for humans to read and distinguish.

exercises/protein-translation/canonical-data.json
ErikSchierboom: Good point. Maybe remove this test?

sjwarner-bp: All done!

ErikSchierboom:
Merged. Thanks @sjwarner-bp! :tada:

PR #1000, opened 2017-11-10
all-your-base: Be explicit about error cases

petertseng:
all-your-base 2.0.0

This uses the error object defined in
https://github.com/exercism/problem-specifications/issues/401#issuecomment-274395645
and thenceforth added to the README's **Test Data Format**.

However, since `expected: null` is also used for cases where there is
always an error (and there is no decision being left to the tracks),
it becomes unreasonably hard to tell cases in these two categories
apart.

To make this easier, all cases in the latter category (always are
errors) can canonically be marked as such.

### DIFF
--- a/exercises/all-your-base/canonical-data.json
+++ b/exercises/all-your-base/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "all-your-base",
-  "version": "1.2.0",
+  "version": "2.0.0",
   "comments": [
     "This canonical data makes the following choices:",
     "1. Zero is always represented in outputs as [0] instead of [].",
@@ -9,11 +9,6 @@
     "4. An empty sequence of input digits is considered zero, rather than an error.",
     "",
     "Tracks that wish to make different decisions for these choices may translate appropriately.",
-    "",
-    "It's up to each track do decide:",
-    "How should invalid input be handled?",
-    "",
-    "All the undefined cases are marked as null.",
     "",
     "All your numeric-base are belong to [2..]. :)"
   ],
@@ -120,7 +115,7 @@
       "input_base": 1,
       "input_digits": [],
       "output_base": 10,
-      "expected": null
+      "expected": {"error": "input base must be >= 2"}
     },
     {
       "description": "first base is zero",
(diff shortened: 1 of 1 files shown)


exercises/all-your-base/canonical-data.json
```diff
@@ -124,71 +124,71 @@
...
+      "expected": {"error": "input base must be >= 2"}
     },
     {
       "description": "negative digit",
       "property": "rebase",
       "input_base": 2,
       "input_digits": [1, -1, 1, 0, 1, 0],
       "output_base": 10,
-      "expected": null
+      "expected": {"error": "all digits must satisfy 0 <= d < input base"}
```
Insti: pedantic_mode: Is `-1` really a digit?

petertseng: I see why one would ask this question, and it is true that in all bases allowed in this problem, digits can be considered to be unsigned values.

So thus, I present for academic curiosity https://en.wikipedia.org/wiki/Balanced_ternary, in which there is the possibility of negative digits.

petertseng:
Wait a second. I think I can now remove "4. How should invalid input be handled?" from the list of unanswered questions at the top of the file now, right?

petertseng:
So, like this then?

If not, Remind me to go back to 7983e19c0cdf2c35b954e6c7fb7f5542ed1bbd0c

Insti:
I'm becoming more convinced that invalid input handling should not be present in the canonical data. See also: https://github.com/exercism/problem-specifications/issues/902

petertseng:
I wonder if I find it funny that removing all these cases would move to version 1.2.0, whereas keeping them moves to version 2.0.0.

Regardless of that, I'll think about whether I want to remove them.

Insti:
That's more an indication of how screwed up our versioning system is.

petertseng:
What I will say today, and that I can reasonably say there is agreement on, is that **if** the invalid inputs should be tested, we agree that it is better to expect an `error` rather than `null`.

The merging of this PR is not to be taken as an acceptance or rejection of the statement "we should remove the invalid inputs from the canonical data"

Issue #473, opened 2016-12-20
Generating tests for "all-your-base" exercise


tommyschaefer:
Hi there!

I've been working on adding a test generator for the `all-your-base` exercise for [exercism/xruby](https://github.com/exercism/xruby) and I've run into a bit of a snag.

Everything works perfectly until I get to the last 13 (if I counted correctly) exercises that are left undefined. As far as I'm able to tell, and I certainly could be missing something, these tests aren't easy (possible?) to generate from individual tracks. Is that correct?

If so, I'm wondering if theres a way that we could add more information to the `canonical-data` so that different tracks could generate their tests based on the following (from the data documentation):

```
It's up to each track do decide:

1. What's the canonical representation of zero?
    - []?
    - [0]?

2. What representations of zero are allowed?
     - []?
     - [0]?
     - [0,0]?

3. Are leading zeroes allowed?

4. How should invalid input be handled?
```

Though, I'm really not to sure how that would work...

Have any other tracks successfully implemented a test generator for `all-your-base`? I looked through a couple of the other language repositories, but I wasn't able to find anything.

Thank you so much for your time!!

Insti:
It's useful to link to a file you are discussing: https://github.com/exercism/x-common/blob/master/exercises/all-your-base/canonical-data.json

Insti:
Aren't the last examples with `expected: null` error cases?
So you need to return/raise an appropriate error value.

dvberkel:
@stevejb71 has created a test generated and converted [`all-your-base`](https://github.com/exercism/xocaml/blob/master/tools/test-generator/templates/all-your-base/template.ml) with it for the [xocaml](https://github.com/exercism/xocaml) track.

He probably faces similar questions, so maybe he can weigh in on the matter?

tommyschaefer:
@Insti 
> Aren't the last examples with expected: null error cases?

Not necessarily, for example the [leading zeros test](https://github.com/exercism/x-common/blob/master/exercises/all-your-base/canonical-data.json#L88):
```json
{
  "description"  : "leading zeros",
  "input_base"   : 7,
  "input_digits" : [0, 6, 0],
  "output_base"  : 10,
  "expected"     : null
}
```

This is a case where tracks can either implement leading zeros as valid or invalid.

So in the [Ruby implementation](https://github.com/exercism/xruby/blob/master/exercises/all-your-base/all_your_base_test.rb#L159), we do allow leading zeros:
```ruby
def test_leading_zeros
    skip
[...]

PR #1001, opened 2017-11-10
all-your-base: Be decisive about zeroes

petertseng:
all-your-base 2.0.0

There are four affected cases.

https://github.com/exercism/problem-specifications/issues/473 results in
a discussion that some post-processing is necessary, based on the
decisions that a given track makes. One might even imagine that each
individual generator may have to reimplement base translation to arrive
at the proper answers for each case.

To ease the process of translation, instead we make some canonical
choices, explicitly show what the choices are, and offer that tracks may
make a different choice if desired.

A previous proposal did not receive support:
https://github.com/exercism/problem-specifications/pull/468

However, this proposal differs because it does change the comments at
the top of the file.

Closes #1002 by mutual exclusion

### DIFF
--- a/exercises/all-your-base/canonical-data.json
+++ b/exercises/all-your-base/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "all-your-base",
-  "version": "2.0.0",
+  "version": "1.2.0",
   "comments": [
     "This canonical data makes the following choices:",
     "1. Zero is always represented in outputs as [0] instead of [].",
@@ -9,6 +9,11 @@
     "4. An empty sequence of input digits is considered zero, rather than an error.",
     "",
     "Tracks that wish to make different decisions for these choices may translate appropriately.",
+    "",
+    "It's up to each track do decide:",
+    "How should invalid input be handled?",
+    "",
+    "All the undefined cases are marked as null.",
     "",
     "All your numeric-base are belong to [2..]. :)"
   ],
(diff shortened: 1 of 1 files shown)


petertseng:
The intent is that #1000 will be merged and zero or one of (#1001, #1002) will be merged.

I gain very little benefit from #1001.

exercises/all-your-base/canonical-data.json
```diff
@@ -1,23 +1,14 @@
 {
   "exercise": "all-your-base",
-  "version": "1.1.0",
+  "version": "2.0.0",
   "comments": [
-    "It's up to each track do decide:",
+    "This canonical data makes the following choices:",
+    "1. Zero is always represented in outputs as [0] instead of [].",
+    "2. In no other instances are leading zeroes present in any outputs.",
```
petertseng: this is not really a choice that I expect there to be controversy on, so I guess it is just here to be a foil to choice 1.

exercises/all-your-base/canonical-data.json
```diff
@@ -1,23 +1,14 @@
...
+    "3. Leading zeroes are accepted in inputs.",
```
petertseng: it should be relatively obvious that the opposite choice is "rejected", so a clause, "rather than being rejected" was not added.

petertseng:
I think I'll make this 1.2.0 since this commit doesn't introduce any new types that were not already present in this file. That way I can merge this then #1000 and end with version 2.0.0 rather than 3.0.0 or 2.1.0 or something. Not that there's any benefit to being minimal with version changes.

PR #1002, opened 2017-11-10
all-your-base: Show all the choices that tracks could make

petertseng:
all-your-base 2.0.0

There are four affected cases.

https://github.com/exercism/problem-specifications/issues/473 results in
a discussion that some post-processing is necessary, based on the
decisions that a given track makes. One might even imagine that each
individual generator may have to reimplement base translation to arrive
at the proper answers for each case.

To ease the process of translation, instead we show the possible choices
that a track might make and show the expected value under each possible
choice.

A previous proposal did not receive support:
https://github.com/exercism/problem-specifications/pull/468

However, this proposal differs because it offers multiple choices.

Closes #1001 by mutual exclusion


petertseng:
The intent is that #1000 will be merged and zero or one of (#1001, #1002) will be merged.

I gain almost no benefit from #1002.

exercises/all-your-base/canonical-data.json
```diff
@@ -92,31 +101,67 @@
...
       "description": "multiple zeros",
       "property": "rebase",
       "input_base": 10,
       "input_digits": [0, 0, 0],
       "output_base": 2,
-      "expected": null
+      "expected": {
+        "choices": {
+          "if_your_track_allows_leading_zeroes_in_input": {
+            "choices": {
```
petertseng: bug: For tracks that allow leading zeroes in **nonzero** inputs BUT expect their zeroes to nonetheless be `[]`, this should be an error, but currently isn't. Oh no, how could I make such a grievous error on such an important PR

petertseng:
In consideration of the reviewers' limited time, I really don't think you should expend your time reviewing this one.

Issue #996, opened 2017-11-08
JSON Schema proposal: All inputs should appear under the `input` key


petertseng:
In our schema, we know that `description`, `expected`, and `property` are fixed, but the inputs to each test case depend on the problem!
That means we can't know ahead of time what key(s) constitute the inputs; we have to examine the JSON in order to do so.

According to those test generators listed at https://github.com/exercism/discussions/issues/155 the current solutions to this problem take on one of two forms:

* Look for any key that isn't `description`, `expected`, `property`, or `comments`. That key (or those keys) must be the input, by process of elimination.
* Per-exercise configuration or code describing what keys hold the input for each exercise.

Well, what if we could do better?

Let's look at https://github.com/exercism/problem-specifications/blob/master/exercises/change/canonical-data.json where each case has two inputs: `coins` and `target`. Let's say that instead a case looks like:

```json
{
  "description": "single coin change",
  "property": "findFewestCoins",
  "input": {
    "coins": [1, 5, 10, 25, 100],
    "target": 25
  },
  "expected": [25]
}
```

As you can see, the key `input` contains an object whose keys and values represent the inputs.

And what about https://github.com/exercism/problem-specifications/blob/master/exercises/leap/canonical-data.json where each case has only one input? One might consider two possible ways to do this:

The first way is actually how the JSON is already: The key `input` contains a scalar, the one and only input.

```json
{
  "description": "year not divisible by 4: common year",
  "property": "leapYear",
  "input": 2015,
  "expected": false
}
```

The second way we might consider is that it will also be an object, for consistency. So then:

```json
{
  "description": "year not divisible by 4: common year",
  "property": "leapYear",
  "input": {
    "year": 2015
  },
  "expected": false
}
```

One could consider making that a one-liner: `"input": {"year": 2015}`) if vertical space savings are desired.
This has the advantage of consistency, of course: `input` always contains an object, but the file also becomes more verbose.

Were one of the two forms of this proposal to be implemented, would this benefit you and/or your track? Would it instead cause harm?

Alternative proposals to solve the problem of "we don't know the names of the keys that hold the input(s) ahead of time"?

petertseng:
EBWODP:

> And what about https://github.com/exercism/problem-specifications/blob/master/exercises//canonical-data.json 

[...]

PR #1006, opened 2017-11-14
meetup: correct example in description.md

wolf99:
The *third* Tuesday of January 2017 was the 17th and thus should either be specified as a "-teenth" or be a different Tuesday. I went with the latter.

This confused me no end when trying to grok what qualified a day to be "teenth" or not >.<

### DIFF
--- a/exercises/meetup/description.md
+++ b/exercises/meetup/description.md
@@ -6,7 +6,7 @@ will take a description of a meetup date, and return the actual meetup date.
 Examples of general descriptions are:
 
 - The first Monday of January 2017
-- The third Tuesday of January 2017
+- The second Tuesday of January 2017
 - The wednesteenth of January 2017
 - The last Thursday of January 2017
 


wolf99:
Hmm, checking the canonical data I think maybe I still haven't groked it!
I see the following examples (among others):

```json
    {
      "description": "second Tuesday of May 2013",
      "property": "meetup",
      "year": 2013,
      "month": 5,
      "week": "second",
      "dayofweek": "Tuesday",
      "dayofmonth": 14
    },
```

```json
    {
      "description": "second Wednesday of August 2013",
      "property": "meetup",
      "year": 2013,
      "month": 8,
      "week": "second",
      "dayofweek": "Wednesday",
      "dayofmonth": 14
    },
```

These specify days that have a date that is a teen, but that are not expected to return "teenth"? 
So what exactly is it that qualifies a day of the week as "teenth" or not?

mienaikage:
I think you may be looking at it backwards. The description of the date goes in e.g second/-teenth Tuesday of May 2013, and the actual date e.g. 2013-05-14 comes out.

Dates which are -teenth will always overlap with dates in either second or third.

ErikSchierboom:
> Dates which are -teenth will always overlap with dates in either second or third.

Well put. The input to this problem is the year/month/day (first, second, third, teenth, last) and the output is a date. Hence why there will be an overlap.

wolf99:
Light dawns. Thanks folks!

Insti:
The input should be a phrase and the output should be a date.
The exploded nature of the properties in the canonical data is not really helpful here.
Hopefully we can fix this.
See also: https://github.com/exercism/problem-specifications/issues/996

```
    {
      "description": "why this date is useful to test",
      "property": "meetup",
      "input": { "phrase": "second Wednesday of August 2013" },
      "expected": { "year": 2013, "month": 8, "dayofmonth": 14 }
    },
```

PR #1013, opened 2017-11-16
README.md: grammar corrections

### LINKED ISSUE #1010: documentation: README.md conventions corrections.
The [README.md test data format section](https://github.com/exercism/problem-specifications/blob/master/README.md#test-data-format-canonical-datajson) has some  issues that need fixing.

**Task**

Update the "There are also some convention about expected that you must follow:" section of [README.md](https://github.com/exercism/problem-specifications/blob/master/README.md#test-data-format-canonical-datajson)

Suggested changes:

>
> There are also some convention**s** ~~about expected~~ that you must follow:

>    All keys should follow the lowerCamelCase convention.
    **I**f the input is valid but there is no result for the input, the value at "expected" should be null.
    **I**f an error is expected (because the input is invalid, or any other reason), the value at "expected" should be an object containing exactly one property, "error", whose value is a string.
        The string should explain why the error would occur.
        A particular track's implementation of the exercise need not necessarily check that the error includes that exact string as the cause, depending on what is idiomatic in the language (it may not be idiomatic to check strings for error messages).

There may be other changes that need making within that section, these were just the ones that caused me to create the issue. 

(The bullet points should probably stay, they just disappeared when I copy/pasted the text.)

See also: #1009

rpottsoh:
Minor grammatical corrections.

Closes #1010

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,13 +30,13 @@ There are three metadata files:
 
 * `description.md` - the basic problem description
 * `metadata.yml` - additional information about the problem, such as where it came from
-* `canonical-data.json` (optional) - standardized test inputs and outputs that can be used to implement the problem
+* `canonical-data.json` - standardized test inputs and outputs that can be used to implement the problem
 
 ## Test Data Format (canonical-data.json)
 
 This data can be incorporated into test programs manually or extracted by a
 program.  The file format is described in `canonical-schema.json`, but it
-is easier to understand with a example:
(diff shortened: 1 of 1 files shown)


Insti:
Thanks. At first glance this looks fine. 👍
I can have a closer look tomorrow.

Insti:
Thanks @rpottsoh :heart:

PR #1014, opened 2017-11-19
pov: Remove extraneous heading

N-Parsons:
I noticed that there is an extra top-level heading in the `description.md` that I don't think should be there, so this PR removes it.

### DIFF
--- a/exercises/pov/description.md
+++ b/exercises/pov/description.md
@@ -1,7 +1,5 @@
 Reparent a graph on a selected node.
 
-# Tree Reparenting
-
 This exercise is all about re-orientating a graph to see things from a different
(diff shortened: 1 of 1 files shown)


exercises/pov/description.md
```diff
@@ -1,7 +1,5 @@
 Reparent a graph on a selected node.
 
-# Tree Reparenting
-
```
rpottsoh: I get the sense that at one point, perhaps, the exercise was intended to be called Tree Reparenting.  I don't think anything is lost by removing this heading.

Insti:
Thanks @N-Parsons :heart:

PR #1016, opened 2017-11-20
protein-translation: Remove repetition from canonical data

sjwarner-bp:
I noticed there was some redundancy in the test names - removed it :slightly_smiling_face:

### DIFF
--- a/exercises/protein-translation/canonical-data.json
+++ b/exercises/protein-translation/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "protein-translation",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "cases": [
     {
       "description": "Translate input RNA sequences into proteins",
--- a/exercises/protein-translation/canonical-data.json
+++ b/exercises/protein-translation/canonical-data.json
@@ -142,7 +142,7 @@
           "expected": ["Tryptophan"]
         },
         {
-          "description": "Translation stops if codon in middle of six-codon sequence",
+          "description": "Translation stops if STOP codon in middle of six-codon sequence",
           "property": "proteins",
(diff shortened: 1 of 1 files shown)


sjwarner-bp:
I wasn't sure if it was a big enough change to bother updating the version number or not - I'd be interested to hear thoughts.

exercises/protein-translation/canonical-data.json
```diff
@@ -136,13 +136,13 @@
...
         {
-          "description": "Translation stops if STOP codon in middle of sequence of three-codon sequence",
+          "description": "Translation stops if STOP codon in middle of three-codon sequence",
           "property": "proteins",
           "strand": "UGGUAGUGG",
           "expected": ["Tryptophan"]
         },
         {
-          "description": "Translation stops if codon in middle of sequence of six-codon sequence",
+          "description": "Translation stops if codon in middle of six-codon sequence",
```
ErikSchierboom: Shouldn't "Translation stops if codon" be "Translation stops if STOP codon"

sjwarner-bp: That is indeed the case - good spot! I'll fix that one too :slightly_smiling_face:

coriolinus:
@sjwarner-bp re: increment the version?

It's a matter of discretion, really. The [Semantic Version 2.0 spec](http://semver.org/) says that the PATCH field is for "backwards-compatible bug fixes"; this is certainly backwards-compatible, but is it a bug fix? 

My instinct is to say that as this change might change the output of exercise generators, you may as well update the patch field of the version. However, I don't feel strongly about this at all; I'm happy to leave it to your discretion.

sjwarner-bp:
On it :+1:

PR #1018, opened 2017-11-20
armstrong-numbers: new exercise

### LINKED ISSUE #1017: New exercise idea: Armstrong numbers
https://en.wikipedia.org/wiki/Narcissistic_number
I'd like to add this as an exercise.
The idea would be to implement a test for Armstrong-Numbers
Objections? Approvals?

pheanex:
How is this for a first draft on #1017 ?

Closes #1017.

### DIFF
--- a/exercises/armstrong-numbers/canonical-data.json
+++ b/exercises/armstrong-numbers/canonical-data.json
@@ -1,6 +1,7 @@
 {
   "exercise": "armstrong-numbers",
   "version": "1.0.0",
+  "description": "There are no 2 digit Armstrong numbers",
   "cases": [
     {
       "description": "0 is an Armstrong number",
--- a/exercises/armstrong-numbers/canonical-data.json
+++ b/exercises/armstrong-numbers/canonical-data.json
@@ -1,189 +1,86 @@
 {
   "exercise": "armstrong-number",
   "version": "1.0.0",
-  "comments": [
-    "Both input-parameters are always given in base-10 (decimal)."
-  ],
   "cases": [
     {
-      "description": "0 is an Armstrong number in base 10",
+      "description": "0 is an Armstrong number",
       "property": "armstrong",
       "input": {
-        "number": 0,
-        "base": 10
+        "number": 0
       },
       "expected": true
     },
     {
-      "description": "1 is an Armstrong number in base 10",
+      "description": "1 is an Armstrong number",
       "property": "armstrong",
       "input": {
-        "number": 1,
-        "base": 10
+        "number": 1
       },
       "expected": true
     },
     {
-      "description": "2 is an Armstrong number in base 10",
+      "description": "9 is an Armstrong number",
       "property": "armstrong",
       "input": {
-        "number": 2,
-        "base": 10
+        "number": 9
       },
       "expected": true
     },
     {
-      "description": "9 is an Armstrong number in base 10",
+      "description": "10 is not an Armstrong number",
       "property": "armstrong",
       "input": {
-        "number": 9,
-        "base": 10
-      },
-      "expected": true
-    },
-    {
-      "description": "10 is not an Armstrong number in base 10",
-      "property": "armstrong",
-      "input": {
-        "number": 10,
-        "base": 10
+        "number": 10
       },
       "expected": false
     },
     {
-      "description": "100 is not an Armstrong number in base 10",
+      "description": "100 is not an Armstrong number",
       "property": "armstrong",
       "input": {
-        "number": 100,
-        "base": 10
+        "number": 100
       },
       "expected": false
     },
     {
-      "description": "153 is an Armstrong number in base 10",
+      "description": "153 is an Armstrong number",
       "property": "armstrong",
       "input": {
-        "number": 153,
-        "base": 10
+        "number": 153
       },
       "expected": true
     },
     {
-      "description": "370 is an Armstrong number in base 10",
+      "description": "370 is an Armstrong number",
       "property": "armstrong",
       "input": {
-        "number": 370,
-        "base": 10
+        "number": 370
       },
       "expected": true
     },
     {
-      "description": "9474 is an Armstrong number in base 10",
+      "description": "9474 is an Armstrong number",
       "property": "armstrong",
       "input": {
-        "number": 9474,
-        "base": 10
+        "number": 9474
       },
       "expected": true
     },
     {
-      "description": "0 is an Armstrong number in base 8",
+      "description": "9926314 is not an Armstrong number",
       "property": "armstrong",
       "input": {
-        "number": 0,
-        "base": 8
-      },
-      "expected": true
-    },
-    {
-      "description": "1 is an Armstrong number in base 8",
-      "property": "armstrong",
-      "input": {
-        "number": 1,
-        "base": 8
-      },
-      "expected": true
-    },
-    {
-      "description": "2 is an Armstrong number in base 8",
-      "property": "armstrong",
-      "input": {
-        "number": 2,
-        "base": 8
-      },
-      "expected": true
-    },
-    {
-      "description": "7 is an Armstrong number in base 8",
-      "property": "armstrong",
-      "input": {
-        "number": 7,
-        "base": 8
-      },
-      "expected": true
-    },
-    {
-      "description": "10 is not an Armstrong number in base 8",
-      "property": "armstrong",
-      "input": {
-        "number": 10,
-        "base": 8
+        "number": 9926314
       },
       "expected": false
     },
     {
-      "description": "432 is an Armstrong number in base 8",
+      "description": "9926315 is an Armstrong number",
       "property": "armstrong",
       "input": {
-        "number": 432,
-        "base": 8
+        "number": 9926315
       },
       "expected": true
-    },
-    {
-      "description": "2 is an Armstrong number in base 3",
-      "property": "armstrong",
-      "input": {
-        "number": 2,
-        "base": 3
-      },
-      "expected": true
-    },
-    {
-      "description": "4 is not an Armstrong number in base 3",
-      "property": "armstrong",
-      "input": {
-        "number": 4,
-        "base": 3
-      },
-      "expected": false
-    },
-    {
-      "description": "17 is an Armstrong number in base 3",
-      "property": "armstrong",
-      "input": {
-        "number": 17,
-        "base": 3
-      },
-      "expected": true
-    },
-    {
-      "description": "10 is an Armstrong number in base 16",
-      "property": "armstrong",
-      "input": {
-        "number": 10,
-        "base": 16
-      },
-      "expected": true
-    },
-    {
-      "description": "10 is not an Armstrong number in base 3",
-      "property": "armstrong",
-      "input": {
-        "number": 10,
-        "base": 3
-      },
-      "expected": false
     }
   ]
 }
--- a/exercises/armstrong-numbers/canonical-data.json
+++ b/exercises/armstrong-numbers/canonical-data.json
@@ -28,7 +28,7 @@
       "expected": true
     },
     {
-      "description": "10 is not an Armstrong number",
+      "description": "There are no 2 digit Armstrong numbers",
       "property": "isArmstrongNumber",
       "input": {
         "number": 10
@@ -36,7 +36,7 @@
       "expected": false
     },
     {
-      "description": "100 is not an Armstrong number",
+      "description": "Three digit number that is not an Armstrong number",
       "property": "isArmstrongNumber",
       "input": {
         "number": 100
@@ -44,7 +44,7 @@
       "expected": false
     },
     {
-      "description": "153 is an Armstrong number",
+      "description": "Three digit number that is an Armstrong number",
       "property": "isArmstrongNumber",
       "input": {
         "number": 153
@@ -52,15 +52,7 @@
       "expected": true
     },
     {
-      "description": "370 is an Armstrong number",
-      "property": "isArmstrongNumber",
-      "input": {
-        "number": 370
-      },
-      "expected": true
-    },
-    {
-      "description": "9474 is an Armstrong number",
+      "description": "Four digit number that is an Armstrong number",
       "property": "isArmstrongNumber",
       "input": {
         "number": 9474
@@ -68,7 +60,15 @@
       "expected": true
     },
     {
-      "description": "9926314 is not an Armstrong number",
+      "description": "Four digit number that is not an Armstrong number",
+      "property": "isArmstrongNumber",
+      "input": {
+        "number": 9475
+      },
+      "expected": false
+    },
+    {
+      "description": "Seven digit number that is not an Armstrong number",
       "property": "isArmstrongNumber",
       "input": {
         "number": 9926314
@@ -76,7 +76,7 @@
       "expected": false
     },
     {
-      "description": "9926315 is an Armstrong number",
+      "description": "Seven digit number that is an Armstrong number",
       "property": "isArmstrongNumber",
       "input": {
         "number": 9926315
--- a/exercises/armstrong-numbers/canonical-data.json
+++ b/exercises/armstrong-numbers/canonical-data.json
@@ -1,10 +1,10 @@
 {
-  "exercise": "armstrong-number",
+  "exercise": "armstrong-numbers",
   "version": "1.0.0",
   "cases": [
     {
       "description": "0 is an Armstrong number",
-      "property": "armstrong",
+      "property": "isArmstrongNumber",
       "input": {
         "number": 0
       },
@@ -12,7 +12,7 @@
     },
     {
       "description": "1 is an Armstrong number",
-      "property": "armstrong",
+      "property": "isArmstrongNumber",
       "input": {
         "number": 1
       },
@@ -20,7 +20,7 @@
     },
     {
       "description": "9 is an Armstrong number",
-      "property": "armstrong",
+      "property": "isArmstrongNumber",
       "input": {
         "number": 9
       },
@@ -28,7 +28,7 @@
     },
     {
       "description": "10 is not an Armstrong number",
-      "property": "armstrong",
+      "property": "isArmstrongNumber",
       "input": {
         "number": 10
       },
@@ -36,7 +36,7 @@
     },
     {
       "description": "100 is not an Armstrong number",
-      "property": "armstrong",
+      "property": "isArmstrongNumber",
       "input": {
         "number": 100
       },
@@ -44,7 +44,7 @@
     },
     {
       "description": "153 is an Armstrong number",
-      "property": "armstrong",
+      "property": "isArmstrongNumber",
       "input": {
         "number": 153
       },
@@ -52,7 +52,7 @@
     },
     {
       "description": "370 is an Armstrong number",
-      "property": "armstrong",
+      "property": "isArmstrongNumber",
       "input": {
         "number": 370
       },
@@ -60,7 +60,7 @@
     },
     {
       "description": "9474 is an Armstrong number",
-      "property": "armstrong",
+      "property": "isArmstrongNumber",
       "input": {
         "number": 9474
       },
@@ -68,7 +68,7 @@
     },
     {
       "description": "9926314 is not an Armstrong number",
-      "property": "armstrong",
+      "property": "isArmstrongNumber",
       "input": {
         "number": 9926314
       },
@@ -76,7 +76,7 @@
     },
     {
       "description": "9926315 is an Armstrong number",
-      "property": "armstrong",
+      "property": "isArmstrongNumber",
       "input": {
         "number": 9926315
       },
--- a/exercises/armstrong-numbers/canonical-data.json
+++ b/exercises/armstrong-numbers/canonical-data.json
@@ -177,15 +177,6 @@
       "expected": true
     },
     {
-      "description": "10 is not an Armstrong number in base 8",
-      "property": "armstrong",
-      "input": {
-        "number": 10,
-        "base": 8
-      },
-      "expected": false
-    },
-    {
       "description": "10 is an Armstrong number in base 16",
       "property": "armstrong",
       "input": {
--- a/exercises/armstrong-numbers/canonical-data.json
+++ b/exercises/armstrong-numbers/canonical-data.json
@@ -1,7 +1,6 @@
 {
   "exercise": "armstrong-numbers",
   "version": "1.0.0",
-  "description": "There are no 2 digit Armstrong numbers",
   "cases": [
     {
       "description": "0 is an Armstrong number",
--- a/exercises/armstrong-numbers/canonical-data.json
+++ b/exercises/armstrong-numbers/canonical-data.json
@@ -1,0 +1,207 @@
+{
+  "exercise": "armstrong-number",
+  "version": "1.0.0",
+  "comments": [
+    "Both input-parameters are always given in base-10(decimal)."
+  ],
+  "cases": [
+    {
+      "description": "0 is an Armstrong number in base 10",
+      "property": "armstrong",
+      "input": {
+        "number": 0,
+        "base": 10
+      },
+      "expected": true
+    },
+    {
+      "description": "1 is an Armstrong number in base 10",
+      "property": "armstrong",
+      "input": {
+        "number": 1,
+        "base": 10
+      },
+      "expected": true
+    },
+    {
+      "description": "2 is an Armstrong number in base 10",
+      "property": "armstrong",
+      "input": {
+        "number": 2,
+        "base": 10
+      },
+      "expected": true
+    },
+    {
+      "description": "9 is an Armstrong number in base 10",
+      "property": "armstrong",
+      "input": {
+        "number": 9,
+        "base": 10
+      },
+      "expected": true
+    },
+    {
+      "description": "10 is not an Armstrong number in base 10",
+      "property": "armstrong",
+      "input": {
+        "number": 10,
+        "base": 10
+      },
+      "expected": false
+    },
+    {
+      "description": "100 is not an Armstrong number in base 10",
+      "property": "armstrong",
+      "input": {
+        "number": 100,
+        "base": 10
+      },
+      "expected": false
+    },
+    {
+      "description": "111 is not an Armstrong number in base 10",
+      "property": "armstrong",
+      "input": {
+        "number": 111,
+        "base": 10
+      },
+      "expected": false
+    },
+    {
+      "description": "153 is an Armstrong number in base 10",
+      "property": "armstrong",
+      "input": {
+        "number": 153,
+        "base": 10
+      },
+      "expected": true
+    },
+    {
+      "description": "370 is an Armstrong number in base 10",
+      "property": "armstrong",
+      "input": {
+        "number": 370,
+        "base": 10
+      },
+      "expected": true
+    },
+    {
+      "description": "9474 is an Armstrong number in base 10",
+      "property": "armstrong",
+      "input": {
+        "number": 9474,
+        "base": 10
+      },
+      "expected": true
+    },
+    {
+      "description": "0 is an Armstrong number in base 8",
+      "property": "armstrong",
+      "input": {
+        "number": 0,
+        "base": 8
+      },
+      "expected": true
+    },
+    {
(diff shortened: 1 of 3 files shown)


exercises/armstrong-numbers/canonical-data.json
```diff
@@ -0,0 +1,207 @@
...
+      "description": "17 is an Armstrong number in base 3",
+      "property": "armstrong",
+      "input": {
+        "number": 17,
+        "base": 3
+      },
+      "expected": true
+    },
+    {
+      "description": "10 is not an Armstrong number in base 8",
```
coriolinus: Isn't this a duplicate of [L135](https://github.com/exercism/problem-specifications/pull/1018/files#diff-e2fa98853c36612fa9ec636bd3055158R135)?

exercises/armstrong-numbers/description.md
```diff
@@ -0,0 +1,10 @@
+An [Armstrong number](https://en.wikipedia.org/wiki/Narcissistic_number) is a number that is the sum of its own digits each raised to the power of the number of digits.
+It also depends on the base of the number being used (for example base = 10 for the decimal system or base = 2 for the binary system).
+
+For example:<sup></sup>
```
coriolinus: While HTML is valid Markdown, we're also aiming to keep these files readable without rendering. As such, let's get rid of empty tags.

exercises/armstrong-numbers/description.md
```diff
@@ -0,0 +1,10 @@
...
+153<sub>10</sub> is &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;an armstrong number in base 10, because: 153<sub>10</sub> = 1<sup>3</sup> + 5<sup>3</sup> + 3<sup>3</sup> = 1 + 125 + 27 = 154<sub>10</sub>
```
coriolinus: Why six non-breaking spaces in a row?

exercises/armstrong-numbers/metadata.yml
```diff
@@ -0,0 +1,4 @@
+---
+blurb: "Given two decimal parameters (number and base) return True if the number 'number' is an armstrong number with respect to the numeral system with base 'base'."
```
coriolinus: Minor nitpick: instead of "return True if", possibly phrase this as "determine whether".

coriolinus:
Closes #1017.

exercises/armstrong-numbers/description.md
ErikSchierboom: Although I agree that using `<sup></sup>` makes for the best display of exponentiation, I don't really like using HTML in markdown. I've looked at the all-your-base exercise, and [it's description](https://github.com/exercism/problem-specifications/blob/master/exercises/all-your-base/description.md) uses the caret sign (^), which I think is still quite readable.

ErikSchierboom: It also doesn't render [well](https://github.com/pheanex/problem-specifications/blob/091476825e82ed4a7f5255f8c96d3464e56ec71d/exercises/armstrong-numbers/description.md).

exercises/armstrong-numbers/canonical-data.json
```diff
@@ -0,0 +1,207 @@
...
+      "description": "100 is not an Armstrong number in base 10",
+      "property": "armstrong",
+      "input": {
+        "number": 100,
+        "base": 10
+      },
+      "expected": false
+    },
+    {
+      "description": "111 is not an Armstrong number in base 10",
```
Insti: (Feedback applies to all tests.)

The description here doesn't tell me anything already covered by the input and output values.
What is this test testing?
Why is 111 useful number to test?
If we had a solution that passed the previous n tests, what code would need to change that would make this test pass?

Insti:
Is this problem about Armstrong numbers or base calculations?

Given one could express the Armstrong number portion of this exercise using only "standard" base 10. What does the additional base parameter add to the problem?

exercises/armstrong-numbers/metadata.yml
Insti: "Determine if a number is an Armstrong number" 

would be enough.

The blurb doesn't need to completely specify what you have to do. It should be a teaser to encourage the student to look at the detail of the problem.

Insti:
> How is this for a first draft on #1017 ?

This is a great first draft. 👍  

Thanks. :heart:

coriolinus:
@Insti I requested that @pheanex write the tests across several bases in #1017, when the idea was proposed. The reason is that whether a given integer is an Armstrong number varies according to the base in use. That is, `153` is an Armstrong number in base 10, but not necessarily in base 16. Therefore, the base used is a relevant property to vary across in the tests.

I'd expect most students not to bother with base conversions on their own for most cases, instead using the relevant string formatting options of their chosen language to generate a string representation, and then iterating over the digits. If you like, we can restrict the options to octal, decimal, and hexadecimal in order to minimize the chance that students need to manually implement any kind of base conversion.

coriolinus:
@Insti Another relevant property to vary over is the number of digits in the number; that affects the way the property is computed. Therefore, looking at the tests, I'd expect to see for each of some number of bases, for each of some number of digits in that base, a few numbers to check which are Armstrong numbers, and a few which are not. This is what I observe in these tests, so I'm satisfied. 

The precise details of which bases are tested, which digit lengths are tested, and how many and which numbers are tested could be debated, but I don't have any objection to the examples already chosen.

Insti:
> 153 is an Armstrong number in base 10, but not necessarily in base 16. Therefore, the base used is a relevant property to vary across in the tests.

`5 + 5 = 10` in base 10 but not in base 16, but we don't ask for base as an argument to every exercise that uses addition.

I think my point is that the base stuff adds unnecessary complexity to the problem.

coriolinus:
`5+5 == 10` is always true, regardless of the number system; you can write it as `0o5 + 0o5 == 0o12`, or `0x5+0x5 == A`, but the math doesn't change. `18`, `0o22`, and `0x12` are all even, regardless of base. Regardless of what base a number is represented in, it will always be even; we can detect evenness by use of the modulo operator.

Armstrong numbers are different: whether or not a number is an Armstrong number is intrinsically linked to the base in which it is represented. That's precisely why Armstrong numbers are interesting; most numerical properties hold true regardless of representation, and this one doesn't. To test only a single base would ignore the property which makes Armstrong numbers interesting.

Or, to put it otherwise: as the README states, Armstrong numbers come from recreational math. We could create an exercise which tested them in a single base, but it would be pure symbolic manipulation, uninteresting in itself. What makes it interesting is what makes it recreational: the fact that the representation changes the property. 

We're not even asking students to parse numbers of varying base; all the inputs are given as standard integers. We just ask them to compute a property which varies according to base. Testing only base 10 would both be less interesting for the student, and incomplete within the scope of the problem.

exercises/armstrong-numbers/canonical-data.json
pheanex: Removed, Thank you

exercises/armstrong-numbers/description.md
pheanex: I removed the empty tag. Do you want no html at all in the description-Markdown then?

exercises/armstrong-numbers/description.md
pheanex: The idea was to keep the indentation for the two example the same, so that differences are more obvious. => Removed

exercises/armstrong-numbers/metadata.yml
pheanex: Ok, I wasn't really sure what the "blurb" was good for ... now i know :smile:

exercises/armstrong-numbers/description.md
pheanex: To be consistent here, I'd also remove the "sub" additional to the "sup" then.
This would "cost" quite a bit of readability for the rendered version because we lose the hint in what base a number is represented. If we decide to drop all html, then I'd also vote for dropping the "base" completely and go with @Insti argument at the bottom. Do we have a consensus here?

exercises/armstrong-numbers/canonical-data.json
pheanex: I just wanted to have another false 3-digit 10-base test additionally to 100.
I guess you are right and this does not catch anything special additionally.
=> Removed

pheanex:
@Insti @coriolinus I'm fine with both your versions.
How do we proceed? I don't want to bikeshed this :smiley:

exercises/armstrong-numbers/canonical-data.json
```diff
@@ -0,0 +1,189 @@
+{
+  "exercise": "armstrong-number",
+  "version": "1.0.0",
+  "comments": [
+    "Both input-parameters are always given in base-10(decimal)."
```
ErikSchierboom: I think "base-10(decimal)" should be "base-10 (decimal)"

pheanex: Missing space added

exercises/armstrong-numbers/canonical-data.json
```diff
@@ -0,0 +1,189 @@
...
+      "description": "153 is an Armstrong number in base 10",
+      "property": "armstrong",
+      "input": {
+        "number": 153,
+        "base": 10
+      },
+      "expected": true
+    },
+    {
+      "description": "370 is an Armstrong number in base 10",
```
Insti: I had no specific objection to 111, but rather a general objection to ALL the tests.

The description here doesn't tell me anything already covered by the input and output values.
What is this test testing?
Why is  this number useful to test?
If we had a solution that passed the previous n tests, what code would need to change that would make this test pass?

What is it about each of these numbers that makes them interesting/important to test?

exercises/armstrong-numbers/description.md
```diff
@@ -0,0 +1,10 @@
+An [Armstrong number](https://en.wikipedia.org/wiki/Narcissistic_number) is a number that is the sum of its own digits each raised to the power of the number of digits.
+It also depends on the base of the number being used (for example base = 10 for the decimal system or base = 2 for the binary system).
+
+For example:
+153<sub>10</sub> is an armstrong number in base 10, because: 153<sub>10</sub> = 1<sup>3</sup> + 5<sup>3</sup> + 3<sup>3</sup> = 1 + 125 + 27 = 154<sub>10</sub>
```
Insti: Use a capital `A` for all usage of `Armstrong` in English.

exercises/armstrong-numbers/description.md
```diff
@@ -0,0 +1,10 @@
...
+154<sub>10</sub> is *not* an armstrong number in base 10, because: 154<sub>10</sub> &ne; 1<sup>3</sup> + 5<sup>3</sup> + 4<sup>3</sup> = 1 + 125 + 64 = 190<sub>10</sub>
+17<sub>10</sub> is an armstrong number in base 3, because: 17<sub>10</sub> = 122<sub>3</sub> = 1<sup>3</sup> + 2<sup>3</sup> + 2<sup>3</sup> = 1 + 8 + 8 = 17<sub>10</sub>
+18<sub>10</sub> is *not* an armstrong number in base 3, because: 18<sub>10</sub> = 200<sub>3</sub> &ne; 2<sup>3</sup> + 0<sup>3</sup> + 0<sup>3</sup> = 8 + 0 + 0 = 8<sub>10</sub>
```
Insti: Make sure all these lines are separated in the markdown rendered view.
They currently confusingly run into each other.

exercises/armstrong-numbers/canonical-data.json
Insti: Why is 100 useful to test?

Insti:
> as the README states, Armstrong numbers come from recreational math.

The README doesn't actually state this.

Including more of the information from https://github.com/exercism/problem-specifications/pull/1018#issuecomment-345985070 in the README would help.

>  We just ask them to compute a property which varies according to base. Testing only base 10 would both be less interesting for the student, and incomplete within the scope of the problem.

It is still necessary to do base conversion to get the digits to test though?

What do you think the core of this exercise is?

exercises/armstrong-numbers/canonical-data.json
pheanex: I guess its easier if we turn this around:
What numbers do you suggest we test/are interesting to test?

Insti:
> What makes it interesting is what makes it recreational: the fact that the representation changes the property.

I guess I just don't find it interesting. You're still just taking `n` digits and raising them to the power of `n` and summing them. How you get those `n` digits and what you want them to equal is arbitrary. 

You could just as easily say the digits are "122" and you want them to equal 17.
No mention of bases required.

I suspect there is an interesting problem in here somewhere, but I don't think we've found it yet.

coriolinus:
> The README doesn't actually state this.

I apologize, you're right; that info is in the wikipedia page which describes Armstrong numbers.

> You could just as easily say the digits are "122" and you want them to equal 17.
No mention of bases required.

I think this is the key issue. You're absolutely right that we _could_ just say "sum the digits 122 and compare them to 17", and we'd have an exercise which might introduce students to iterating through characters of a string and string to int conversion. However, without context, that exercise is pointless and boring. In contrast, if the exercise talks about Armstrong numbers and links me to a wiki page, and I know that 17<sub>3</sub> is `122`, then I'm perfectly happy solving it; that context is enough to interest me. Perhaps I'm unusually susceptible to mathy nerd-sniping? I don't think so, but I suppose few do. 

I do see the point that we don't want to send students down a base-conversion rabbit hole, particularly if we expect this exercise to appear fairly early in the tracks. To that end, I propose that we limit the canonical data to using bases 8, 10, and 16. Most languages should have some equivalent to `sprintf`, `str.format`, or equivalent, which can handle the base conversion to these targets automatically; we might add a `hints.md` pointing students in that direction. Would that satisfy your need to minimize the base conversion portion?

coriolinus:
To phrase it more concisely, the point of this exercise could be to teach intermediate use of the chosen language's string formatting system, using Armstrong numbers so that it doesn't come across as too contrived.

Insti:
> the point of this exercise could be to teach intermediate use of the chosen language's string formatting system, using Armstrong numbers so that it doesn't come across as too contrived.

In this case the description would need to be significantly rewritten to emphasise this.

BUT string formatting seems to be a rather implementation specific thing.
There are ways to solve this problem that don't involve using string formatting, and languages that do not include such string formatting tricks. 
Exercism problems try to be language agnostic and not force a particular implementation.

But I think we've found the source of our differences.
As I understand it you would like the exercise to be about base conversion via string formatting and I would prefer it to be more focused on simply verifying Armstrong numbers.

coriolinus:
@pheanex As the person submitting this exercise, it's only fair that you get to pick the direction to take it. If in your mind this is more interesting with the bases, you'll need to remove the base-3 tests and update the description to suggest using string formatting to the student. If you think it's more interesting as a pure Armstrong number verification exercise, you'll need to remove the non-base-10 tests and refactor to remove mention of the bases. I'm ok with either variant.

coriolinus:
If it's decided to use the version with the bases, it might be desirable to put a phrase in the description similar to the following:

> Challenge: implement this exercise without using `sprintf`, `str.format`, or your language's equivalent.

This suggests to the student that string formatting may be a simple way to implement the cases, without doing so explicitly. Simultaneously, it reminds them that there are other methods available for doing base conversion. This should serve the goals of maximizing student flexibility while hinting at a possible path for those who need help.

pheanex:
okay, let's take the simple version without base-conversions then

exercises/armstrong-numbers/description.md
```diff
@@ -0,0 +1,9 @@
+An [Armstrong number](https://en.wikipedia.org/wiki/Narcissistic_number) is a number that is the sum of its own digits each raised to the power of the number of digits.
+
+For example:
+9 is an Armstrong number, because 9 = 9^1 = 9
+10 is *not* an Armstrong number, because 10 != 1^2 + 0^2 = 2
+153 is an Armstrong number, because: 153 = 1^3 + 5^3 + 3^3 = 1 + 125 + 27 = 154
```
coriolinus: `153 = ... = 154` should be corrected.

pheanex: obvious mistake. Thank you ;-)

exercises/armstrong-numbers/canonical-data.json
```diff
@@ -0,0 +1,86 @@
+{
+  "exercise": "armstrong-number",
```
Insti: missing an 's' here.

exercises/armstrong-numbers/canonical-data.json
```diff
@@ -0,0 +1,86 @@
...
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "0 is an Armstrong number",
+      "property": "armstrong",
```
Insti: For boolean options the preferred property would be: "isArmstrongNumber"

exercises/armstrong-numbers/canonical-data.json
```diff
@@ -0,0 +1,86 @@
...
+    {
+      "description": "9 is an Armstrong number",
+      "property": "armstrong",
+      "input": {
+        "number": 9
+      },
+      "expected": true
+    },
+    {
+      "description": "10 is not an Armstrong number",
```
Insti: These test descriptions are still not telling me anything about the tests.
Why is 10 a useful number to test?

exercises/armstrong-numbers/canonical-data.json
```diff
@@ -0,0 +1,86 @@
...
+      "property": "armstrong",
+      "input": {
+        "number": 10
+      },
+      "expected": false
+    },
```
Insti: Are there no 2 digit Armstrong numbers?

exercises/armstrong-numbers/description.md
```diff
@@ -0,0 +1,9 @@
+An [Armstrong number](https://en.wikipedia.org/wiki/Narcissistic_number) is a number that is the sum of its own digits each raised to the power of the number of digits.
+
+For example:
+9 is an Armstrong number, because 9 = 9^1 = 9
+10 is *not* an Armstrong number, because 10 != 1^2 + 0^2 = 2
+153 is an Armstrong number, because: 153 = 1^3 + 5^3 + 3^3 = 1 + 125 + 27 = 153
+154 is *not* an Armstrong number, because: 154 != 1^3 + 5^3 + 4^3 = 1 + 125 + 64 = 190
+
+Implement a function, that returns True if the number is an Armstrong number and False otherwise.
```
Insti: There was a suggestion at some point for a better wording for this line.

exercises/armstrong-numbers/canonical-data.json
pheanex: Fixed

exercises/armstrong-numbers/canonical-data.json
pheanex: Fixed

exercises/armstrong-numbers/canonical-data.json
pheanex: I think it's a useful test since it is the only test for 2 digit numbers and since there seems to be no armstrong 2-digit armstrong numbers it is a representative of its class.

exercises/armstrong-numbers/canonical-data.json
pheanex: According to Wikipedia https://en.wikipedia.org/wiki/Narcissistic_number and https://oeis.org/A005188 there is none

exercises/armstrong-numbers/description.md
pheanex: If it is that important to you, I would ask you to be more specific here.

pheanex:
Aren't we driving this a little far with the reviews here?
I am glad for all the reviews, but some seem a little too pedantic for my taste for an exercise of this magnitude ... at least compared to all the other already existing exercises.

exercises/armstrong-numbers/description.md
Insti: I think I was thinking of this: https://github.com/exercism/problem-specifications/pull/1018#discussion_r152147164

But can see now that didn't refer specifically to this text in the description.

"Write some code to determine whether a number is an Armstrong number."

A lot of different programming languages that all have their own syntaxes and conventions, so being as generic as possible in the common description which is used by all of them is helpful. Language tracks can add their own specific details if required.

See also: https://github.com/exercism/problem-specifications/issues/321

Insti:
> Aren't we driving this a little far with the reviews here?

We are nearly there. :smile:

> I am glad for all the reviews, but some seem a little too pedantic for my taste for an exercise of this magnitude ... at least compared to all the other already existing exercises.

It is important that new exercises meet the current standards for formatting and clarity. 
Many of the existing were added before these standards were determined so perhaps do not achieve the levels expected of new exercises.

There are already a lot of exercises, why should we add a new ones?
If something is worth doing it's worth doing well.

exercises/armstrong-numbers/description.md
pheanex: Fixed

exercises/armstrong-numbers/description.md
```diff
@@ -0,0 +1,9 @@
+An [Armstrong number](https://en.wikipedia.org/wiki/Narcissistic_number) is a number that is the sum of its own digits each raised to the power of the number of digits.
+
+For example:
+9 is an Armstrong number, because 9 = 9^1 = 9
+10 is *not* an Armstrong number, because 10 != 1^2 + 0^2 = 2
+153 is an Armstrong number, because: 153 = 1^3 + 5^3 + 3^3 = 1 + 125 + 27 = 153
+154 is *not* an Armstrong number, because: 154 != 1^3 + 5^3 + 4^3 = 1 + 125 + 64 = 190
```
Insti: These line all still need some way to separate them in the rendered output, they currently run into each other in one paragraph.

see: https://github.com/pheanex/problem-specifications/blob/ff8b492d3b1863a7456c66f60fd7f646335745a9/exercises/armstrong-numbers/description.md

pheanex:
Is there anything still preventing this PR getting merged?

Insti:
This exercise is mostly fine. Thanks for all your work on it. :heart:

I'm not merging it because I still don't think the test case descriptions are good, and would like to see more logical and instructive test cases chosen.

But I'm not the only one here. There are many others with commit access who have different opinions. Maybe they will review and merge this.

It may also be that no-one has had a chance to review it yet, we're all volunteers here and do this in our spare time.

exercises/armstrong-numbers/canonical-data.json
rpottsoh: Then the description should include this information.
```text
"description": "There are no 2 digit Armstrong numbers",
```

exercises/armstrong-numbers/canonical-data.json
```diff
@@ -0,0 +1,86 @@
...
+    {
+      "description": "100 is not an Armstrong number",
+      "property": "isArmstrongNumber",
+      "input": {
+        "number": 100
+      },
+      "expected": false
+    },
+    {
+      "description": "153 is an Armstrong number",
```
rpottsoh:
```text
"description": "Three digit number that is an Armstrong number",
```

exercises/armstrong-numbers/canonical-data.json
```diff
@@ -0,0 +1,86 @@
...
+    {
+      "description": "10 is not an Armstrong number",
+      "property": "isArmstrongNumber",
+      "input": {
+        "number": 10
+      },
+      "expected": false
+    },
+    {
+      "description": "100 is not an Armstrong number",
```
rpottsoh:
```text
"description"; "Three digit number that is not an Armstrong number",
```

exercises/armstrong-numbers/canonical-data.json
```diff
@@ -0,0 +1,86 @@
...
+    {
+      "description": "153 is an Armstrong number",
+      "property": "isArmstrongNumber",
+      "input": {
+        "number": 153
+      },
+      "expected": true
+    },
+    {
+      "description": "370 is an Armstrong number",
```
rpottsoh: This test is redundant.  Previous test already tests for three digit numbers, unless there is some other relevance regarding 370.

exercises/armstrong-numbers/canonical-data.json
```diff
@@ -0,0 +1,86 @@
...
+      "property": "isArmstrongNumber",
+      "input": {
+        "number": 370
+      },
+      "expected": true
+    },
+    {
+      "description": "9474 is an Armstrong number",
```
rpottsoh:
```text
"description": "Four digit Armstrong number",
```

rpottsoh: Perhaps a test of a non-Armstrong four digit number would be prudent here as well.

exercises/armstrong-numbers/canonical-data.json
```diff
@@ -0,0 +1,86 @@
...
+      "property": "isArmstrongNumber",
+      "input": {
+        "number": 9474
+      },
+      "expected": true
+    },
+    {
+      "description": "9926314 is not an Armstrong number",
```
rpottsoh:
```text
"description": "Seven digit number that is not an Armstrong number",
```

exercises/armstrong-numbers/canonical-data.json
```diff
@@ -0,0 +1,86 @@
...
+      "property": "isArmstrongNumber",
+      "input": {
+        "number": 9926314
+      },
+      "expected": false
+    },
+    {
+      "description": "9926315 is an Armstrong number",
```
rpottsoh:
```text
"description": "Seven digit number that is an Armstrong number",
```

exercises/armstrong-numbers/canonical-data.json
coriolinus: I'd expect a test with that description to iterate over all 2 digit numbers and ensure none of them are Armstrong.

Perhaps `"description": "Two digit number which is not an Armstrong number"`?

rpottsoh:
@insti said:
> I'm not merging it because I still don't think the test case descriptions are good

I agree.  I have taken a quick crack at the case descriptions...

exercises/armstrong-numbers/canonical-data.json
pheanex: Is there a way i can technically specify it in this json format??

exercises/armstrong-numbers/canonical-data.json
```diff
@@ -0,0 +1,87 @@
+{
+  "exercise": "armstrong-numbers",
+  "version": "1.0.0",
+  "description": "There are no 2 digit Armstrong numbers",
```
rpottsoh: I don't think this line is needed.  It is also causing Travis CI to fail as it violates the [canonical-schema.json](https://github.com/exercism/problem-specifications/blob/master/canonical-schema.json)

exercises/armstrong-numbers/canonical-data.json
```diff
@@ -0,0 +1,86 @@
...
+    {
+      "description": "1 is an Armstrong number",
+      "property": "isArmstrongNumber",
+      "input": {
+        "number": 1
+      },
+      "expected": true
+    },
+    {
+      "description": "9 is an Armstrong number",
```
rpottsoh: Appears as though the first three tests could be consolidated into a single test since all single digit numbers are Armstrong numbers.

```json
"description": "Single digit numbers are Armstrong numbers",
```
My wording may need some work.  I don't want to imply that all single digit number need to be checked.

exercises/armstrong-numbers/canonical-data.json
Insti: This should be the description for a test of a 2 digit number.

rpottsoh: Wouldn't it still be `1.0.0`?

Insti: The canonical-data `version` key is correct.

It was an out of date "commit" I was looking at.
Confusing word choice by me there sorry.

pheanex:
@Insti is there anything concrete (maybe on the test-cases?) i can do to get your approval as well?

rpottsoh:
I am going to go ahead and Squash & Merge since everyone is holding their peace...  If something is discovered later, a new issue/PR can be opened.

Thanks @pheanex for all your time and energy and helping to improve Exercism!

Insti:
Sorry @pheanex, I've been busy the last few days.

This looks fine. Thanks for all your work on it :heart:

Issue #987, opened 2017-10-30
json schema validation: enforce lowerCamelCase keys


Insti:
There is a json schema definition for the `canoncial-data.json` file.
https://github.com/exercism/problem-specifications/blob/master/canonical-schema.json

However this still allows for arbitrary keys containing spaces. 
These keys are often used for specifying test input arguments.

https://github.com/exercism/problem-specifications/blob/637cca615dbbb52d61d0a7a74d71c09360bdebf4/exercises/food-chain/canonical-data.json#L21

**Task**

Update the https://github.com/exercism/problem-specifications/blob/master/canonical-schema.json to require all keys to be lowerCamelCase

Insti:
Is this a good idea?

I'm personally fine with keys that contain spaces, but am aware that for some languages  being able to assume that json keys will not contain spaces will make parsing easier.

NobbZ:
I'd not say that parsing is easier or harder, but usage.

In some languages keys are transformed in a way that we can use the keys name as an accessor of a struct/object/record like thing (eg. `case.startVerse`). If though they contain spaces they need to be often treated differently by using array like accessors (eg `case["start verse"]`).

I'm not affected by that in erlang anyway. But I have to match on the keys using atoms, and usually atoms in erlang are alphanumerics + underscore, beginning with lowercase alpha. I can though construct atoms that contain all kind of characters (in modern erlang even unicode characters) as long as they do not exceed 256 (or was it 255) bytes in their UTF8 representation. But I have to quote them then using single-quote character: (`startVerse` vs. `'start verse'`). Even though quoted atoms are possible, a rule of thumb is to avoid them, since the quoting often leads to problems in IDEs and when talking with people from the outside world. They often consider them plain strings which is wrong.

So even here I'd prefer a way that avoids quoting the key.

ErikSchierboom:
> So even here I'd prefer a way that avoids quoting the key.

Agreed. I'm also in favor of enforcing `lowerCamelCase` keys.

rbasso:
Nice change, @Insti. We didn't enforced that in the schema in #602 at the time because my json-schema-fu [is|was] too weak.

ErikSchierboom:
@Insti Are you still interested in working on this issue?

ErikSchierboom:
Looking at the current [canonical-data schema](https://github.com/exercism/problem-specifications/blob/main/canonical-data.schema.json#L140), it looks like this has been implemented.

petertseng:
[...]

PR #1020, opened 2017-11-21
series: Improve documentation

### LINKED ISSUE #584: series: Implement canonical-data.json
exercises in sync.

The Series exercise can be found in the [./exercises/series/][exercise-dir] directory.

See http://exercism.io/contribute/canonical-data/series for the up-to-date list of
language tracks that have an implementation of the Series exercise.

[exercise-dir]: https://github.com/exercism/x-common/tree/master/exercises/series

N-Parsons:
Resolves #584 by implementing canonical data for `series`.

I've noticed that lots of implementations expect the output to be a list of lists of integers rather than a list of strings:
- [F#](https://github.com/exercism/fsharp/blob/master/exercises/series/SeriesTest.fs)
- [Java](https://github.com/exercism/java/blob/master/exercises/series/src/test/java/SeriesTest.java)
- [Haskell](https://github.com/exercism/haskell/blob/master/exercises/series/test/Tests.hs)

However, I would argue that a list of strings is more reasonable, particularly given the current description, which mentions substrings. Some tracks have implemented it this way:
- [Go](https://github.com/exercism/go/blob/master/exercises/series/series_test.go)
- [Ruby](https://github.com/exercism/ruby/blob/master/exercises/series/series_test.rb)
- Python, subject to exercism/python#1117 (as a result of exercism/python#1109)

As a result, this implementation of the canonical data expects the output to be a list of strings.

### DIFF
--- a/exercises/series/canonical-data.json
+++ b/exercises/series/canonical-data.json
@@ -1,0 +1,105 @@
+{
+  "exercise": "series",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "slices of one from one",
+      "property": "slices",
+      "input": {
+        "series": "1",
+        "sliceLength": 1
+      },
+      "expected": ["1"]
+    },
+    {
+      "description": "slices of one from two",
+      "property": "slices",
+      "input": {
+        "series": "12",
+        "sliceLength": 1
+      },
+      "expected": ["1", "2"]
+    },
+    {
+      "description": "slices of two",
+      "property": "slices",
+      "input": {
+        "series": "35",
+        "sliceLength": 2
+      },
+      "expected": ["35"]
+    },
+    {
+      "description": "slices of two overlap",
+      "property": "slices",
+      "input": {
+        "series": "9142",
+        "sliceLength": 2
+      },
+      "expected": ["91", "14", "42"]
+    },
+    {
+      "description": "slices can include duplicates",
+      "property": "slices",
+      "input": {
+        "series": "777777",
+        "sliceLength": 3
+      },
+      "expected": ["777", "777", "777", "777"]
+    },
+    {
+      "description": "slices of a long series",
+      "property": "slices",
+      "input": {
+        "series": "918493904243",
+        "sliceLength": 5
+      },
+      "expected": ["91849", "18493", "84939", "49390",
+                   "93904", "39042", "90424", "04243"]
+    },
+    {
+      "description": "slice length is too large",
+      "property": "slices",
+      "input": {
+        "series": "12345",
+        "sliceLength": 6
+      },
+      "expected": {
+        "error": "slice length cannot be greater than series length"
+      }
+    },
+    {
+      "description": "slice length cannot be zero",
+      "property": "slices",
+      "input": {
+        "series": "12345",
+        "sliceLength": 0
+      },
+      "expected": {
+        "error": "slice length cannot be zero"
+      }
+    },
+    {
+      "description": "slice length cannot be negative",
+      "property": "slices",
+      "input": {
+        "series": "123",
+        "sliceLength": -1
+      },
+      "expected": {
+        "error": "slice length cannot be negative"
+      }
+    },
+    {
+      "description": "empty series is invalid",
+      "property": "slices",
+      "input": {
+        "series": "",
+        "sliceLength": 1
+      },
+      "expected": {
+        "error": "series cannot be empty"
+      }
+    }
+  ]
+}
--- a/exercises/series/description.md
+++ b/exercises/series/description.md
@@ -1,16 +1,16 @@
 Given a string of digits, output all the contiguous substrings of length `n` in
-that string.
+that string in the order that they appear.
 
 For example, the string "49142" has the following 3-digit series:
 
-- 491
-- 914
-- 142
+- "491"
+- "914"
+- "142"
 
 And the following 4-digit series:
 
-- 4914
-- 9142
+- "4914"
+- "9142"
 
 And if you ask for a 6-digit series from a 5-digit string, you deserve
 whatever you get.


exercises/series/canonical-data.json
```diff
@@ -0,0 +1,73 @@
+{
+    "exercise": "series",
+    "version": "1.0.0",
+    "cases": [
+        {
+            "description": "slices of one",
+            "property": "slices",
+            "series": "01234",
+            "slice_length": 1,
```
ErikSchierboom: We are in the process of enforcing that all keys should be `lowerCamelCase` (see #987). Could you rename this property to `sliceLength` (and also in the other cases)?

exercises/series/canonical-data.json
```diff
@@ -0,0 +1,73 @@
+{
+    "exercise": "series",
+    "version": "1.0.0",
+    "cases": [
+        {
+            "description": "slices of one",
+            "property": "slices",
+            "series": "01234",
```
ErikSchierboom: We are in the processing of ensuring that all input properties are gathered in a single `input` key (see #996). Could you put the `series` and `slice_length` properties in an `input` field (and also in the other cases)?

exercises/series/canonical-data.json
```diff
@@ -0,0 +1,73 @@
...
+        },
+        {
+            "description": "slices can include duplicates",
+            "property": "slices",
+            "series": "123123",
+            "slice_length": 3,
+            "expected": ["123", "231", "312", "123"]
+        },
+        {
+            "description": "overly long slice",
```
ErikSchierboom: Maybe rename to "too long slice"?

exercises/series/canonical-data.json
```diff
@@ -0,0 +1,73 @@
...
+            "property": "slices",
+            "series": "01234",
+            "slice_length": 6,
+            "expected": {
+                "error": "slice length cannot be greater than series length"
+            }
+        },
+        {
+            "description": "overly short slice",
```
ErikSchierboom: Maybe rename to "slice length cannot be zero" or something like that?

N-Parsons:
Thanks for the feedback @ErikSchierboom! I've made the suggested changes.

I've also changed the indentation level to 2-spaces instead of 4 for consistency with the rest of the repo.

ErikSchierboom:
Looking much better already! I think the next step is to think about the various test cases. Ideally, each subsequent test case should introduce one new concept, that the previous tests didn't cover. And also in ascending difficulty.

As an example, consider the following two test cases:

```json
{
  "description": "slices of three",
  "property": "slices",
  "input": {
    "series": "01234",
    "sliceLength": 3
  },
  "expected": ["012", "123", "234"]
},
...
{
  "description": "order preserved with descending numbers",
  "property": "slices",
  "input": {
    "series": "43210",
    "sliceLength": 3
  },
  "expected": ["432", "321", "210"]
},
```

Does the second test case actually introduce a new concept that the first test doesn't cover? I don't think it does in this case (the preservation of the ordering is also tested in the first test case). So perhaps we can try to come up with a set of test cases that gradually introduce more complex concepts not covered previously?

Insti:
Looks good.

Should there be a test for a series made up of a single repeating digit? `7777777`
Most of the tested series are length 5, what about different lengths?
Maybe start with some really easy cases ("1",1), ("12",1)?

Insti:
https://github.com/exercism/problem-specifications/blob/master/exercises/series/description.md

ErikSchierboom:
One way to come up with test cases is to look at the description. E.g. the description notes:

"the digits need not be numerically consecutive."

So maybe one test could be with consecutive numbers, whereas a later test could test with non-consecutive numbers.

AtelesPaniscus:
> One way to come up with test cases is to look at the description.

Hello.  It was me who kicked off this review of this exercise.  Why ?  Because I looked at the description and saw that it said "substrings" and was miffed because it did not occur to me that whoever wrote the test cases for Python must have read a version that read "lists of integers".

I read the description and it said **nothing** about handling duplicates (not tested), **nothing** about the order in which results were to be returned and **nothing** about returning a list.

Ergo: the test cases were expecting a **set containing results in no particular order with no duplicates**.

Perhaps the choice of return type is, or should be, track specific.  I would suggest that all 'generated' READMEs add a track specific note of "what to return unless the generic read me specifies otherwise".

AtelesPaniscus:
> (the preservation of the ordering is also tested in the first test case)

It is but on its own does not prove the solution preserves ordering in the general case ... a pass for one test may simply be a coincidence.

I suspect somewhere in the TDD bible it is written that each test case should test only one requirement.  Amen.  That is not the same as each requirement should have only one test case.

What one test case does is test your assumptions about the solution's implementation.  I would suggest that a good test suite makes no such assumptions.

Even two test cases don't prove ordering, they simply reduce the chance of a buggy solution passing all tests.

Consider again the 'set'.  Contents may be ordered (I don't assume so).  An ordered set converted back to a list might pass one of the two tests case but not both.

It is possible that converting a list (without duplicates) to an unordered set and back again preserves order (because nothing happened to alter the initial order) .  This would pass all the original tests.

For an unordered set.  In some languages, a duplicate is discarded, in others it replaces the original.  That, I expect, would make a difference to the order of results.

The general principle for all systems (road networks for example) including software systems and also test suites for software is that optimisation tends to eliminate redundancy.  The corollary of this is that is makes the system 'brittle'.  Road networks experience gridlock, test suites pass buggy code.

Redundancy is necessary for robust and, ultimately, secure systems.

AtelesPaniscus:
| I've noticed that lots of implementations expect the output to be a list of lists of integers rather than a list of strings:

I suspect this exercise was imported from somewhere and something in the description was lost in the process.  After all, there must be a reason the input consists of only digits.

N-Parsons:
Sorry for the delay in getting back to this - I should have some time to look at this again at the weekend.

rpottsoh:
This PR is changing two files.  I think the `description.md` changes should be moved to a separate PR or the title of this PR should be re-worded.  It doesn't matter to me which occurs.

ErikSchierboom:
@N-Parsons Small bump. Are you still working on this?

N-Parsons:
Sorry for the delay, @ErikSchierboom - I've been quite ill recently. I'll take another look at this in the next few days.

ErikSchierboom:
@N-Parsons Sorry to hear that! I hope you have recovered fully. Good luck continuing the work on this!

exercises/series/canonical-data.json
```diff
@@ -0,0 +1,134 @@
...
+      "input": {
+        "series": "",
+        "sliceLength": 1
+      },
+      "expected": {
+        "error": "string cannot be empty"
+      }
+    },
+    {
+      "description": "empty string is invalid even if slice 0",
```
ErikSchierboom: This is basically a combination of the "empty string is invalid" and "slice length cannot be zero" test cases, so I think this can be removed.

exercises/series/canonical-data.json
```diff
@@ -0,0 +1,134 @@
...
+    },
+    {
+      "description": "empty string is invalid",
+      "property": "slices",
+      "input": {
+        "series": "",
+        "sliceLength": 1
+      },
+      "expected": {
+        "error": "string cannot be empty"
```
ErikSchierboom: Maybe instead of "string cannot be empty" we could use "series cannot be empty'?

exercises/series/canonical-data.json
```diff
@@ -0,0 +1,134 @@
...
+      "description": "slices can include duplicates",
+      "property": "slices",
+      "input": {
+        "series": "123123",
+        "sliceLength": 3
+      },
+      "expected": ["123", "231", "312", "123"]
+    },
+    {
+      "description": "strings can have repeating digits",
```
ErikSchierboom: Does this test case add anything to the "slices can include duplicates" test case?

rpottsoh:
In the original README the input is described as a string of digits and the output appears to be in integer form.  [0540d50](https://github.com/exercism/problem-specifications/pull/1020/commits/0540d5000889026d9a8f809bc2bbf525679377d4) saw fit to redefine the output as strings.  What was the reasoning behind this?  I see from the initial comment that some tracks seem to have implemented the exercise as described and other tracks didn't.  I understand the comment regarding the word `substring` appearing in the README.  Another way that the README could have been changed would have been to replace the word `output` with the word `find`.  Now I am lead to think the expected output should be in integer form since the sample output is in integer form.

I don't like the word `digit` in the README.  Strings contain characters, not digits.  To me, digits are numbers. https://www.eskimo.com/~scs/cclass/progintro/sx6.html

I am not saying that the change to the README should be reversed.  I have not completed the exercise myself nor have I implemented it.  I think requiring integer output over strings is more difficult (interesting maybe), but only in a trivial way.  If we are talking about only strings, why limit the input to only digits?  Since the exercise is only interested in digits, why does the input have to be a string?    Enough rambling and babbling from me for now.

I'll try to take some time over the weekend and read through the JSON.

Thanks for the vote of confidence @ErikSchierboom :+1:

exercises/series/canonical-data.json
```diff
@@ -0,0 +1,134 @@
...
+    },
+    {
+      "description": "slices of two with nonsequential numbers",
+      "property": "slices",
+      "input": {
+        "series": "35",
+        "sliceLength": 2
+      },
+      "expected": ["35"]
+    },
```
rpottsoh: Since we are dealing with characters and not numbers / digits how is "slices of two" any different than "slices of two with nonsequential numbers"?

exercises/series/canonical-data.json
rpottsoh: These two cases are essentially the same in my opinion.

ErikSchierboom:
> If we are talking about only strings, why limit the input to only digits? Since the exercise is only interested in digits, why does the input have to be a string? Enough rambling and babbling from me for now.

This captures my thoughts nicely. The current exercise seems to be leaning on two ideas:

1. Capturing slices of characters from a string.
1. Capturing a slice of digits from a number.

Although I think the second case is slightly harder (and thus more interesting), I would actually be fine with just going with the first option: capturing a slice of characters from a string. In that case, the test cases can probably be significantly reduced and should also include not just digits but also alphabetic characters (and even other characters like spaces, punctuation, etc.).

petertseng:
The only choices to really make are:

* Negative slice length? Probably error. Rather than `[]`.
* Slice length too large? `[]` or error?
* Slice length 0? `[[]]`? `[[]] * (series.length + 1)`? error?

substitute `[[]]` with `[""]` if using strings, of course.

I don't particularly care which choice is made in any of the above cases

petertseng:
EBWODP:

> I only am comfortable with the following arrangement of commits:

Feel free to delay rearranging commits until the very end, of course.

> I don't particularly care which choice is made in any of the above cases

because all are justifiable

ErikSchierboom:
@N-Parsons What do you think about the above comments made by @petertseng?

petertseng:
I have decided to justify each alternative I present.

The only choices to really make are:

* Negative slice length?
  * Probably should be an error.
  * Probably unjustifiable alternative: `[]`, but I'd rather not.
* Slice length too large?
  * Justifiable alternative: `[]`. There are zero ways to make a slice of length six from five elements, for example.
  * Justifiable alternative: `error`. If we define the acceptable range of inputs to be bounded above at the slice length.
* Slice length 0?
  * Justifiable alternative: `[[]] * (series.length + 1)`. Consider three elements. There are:
    * One way to make a slice of length three.
    * Two ways to make a slice of length two.
    * Three ways to make a slice of length one.
    * Therefore, to fit the pattern, there should be four slices of length zero in three elements.
  * Justifiable alternative: `error`. If we define the acceptable range of inputs to be bounded below at 1.
  * Unjustifiable alternative: `[]`. This implies there are zero ways to make a slice of length zero, which is trivially demonstrable to be false: `[]` is a slice of length zero, so there must be at least one way to do it.
  * Unjustifiable alternative: `[[]]`. Just as `slices("555", 2)` results in two instances of `[5, 5]`, so too should `slices("555", 0)` result in four instances of `[]`, not just one.

substitute `[[]]` with `[""]` if using strings, of course.

I don't particularly care which justifiable alternative is chosen in any of the above decisions.

ErikSchierboom:
To me, all three scenarios presented above should result in an error.

coriolinus:
I disagree: exceptions should be exceptional, whether they're implemented
as throws or Result monads or whatever. If there exists a justifiable
non-error alternative, it is to be preferred over an error. GIGO applies as
always, but it's the caller's responsibility to watch for that.

In the Rust track, this exercise is implemented such that the function
cannot return an error: unsigned inputs mean that there can be no negative
slice length, and in the other cases the slice-based return value is used.
I believe that this is fundamentally more elegant than designing it to
return an error. If we specify here that the function should return errors
in these scenarios, that means that the Rust track's implementation of this
exercise is not in compliance, and should be changed to a fundamentally
worse design in order to conform with the canonical data.

I believe that the case testing negative slice length should return an
error, as there is no other justifiable return for that case, and languages
which support unsigned ints can simply discard the test case. The other two
cases should return the slices as described by @petertseng.

ErikSchierboom:
Can we perhaps come to a decision here? I'm fine with any option.

rpottsoh:
> I'm fine with any option.

As am I.  @N-Parsons what do you think?  Have you been able to follow all of the discussion that has been happening since the beginning of February?

ErikSchierboom:
@rpottsoh @petertseng @coriolinus There has not been a response for 24 days. I prefer not to have long standing PR's open.

coriolinus:
My stance remains:

> I believe that the case testing negative slice length should return an
error, as there is no other justifiable return for that case, and languages
which support unsigned ints can simply discard the test case. The other two
cases should return the slices as described by @petertseng.

ErikSchierboom:
@coriolinus Okay, clear. After considering those changes, I'm fine with them. However, there being no response for a while, begs the question if we want to wait for someone to work on this PR, or should someone create a new PR?

coriolinus:
I'm fine leaving this open and waiting for someone with interest to come in and implement the requested changes.

petertseng:
I am not able to offer my time to make any changes personally.

In general, from my point of view, I have no qualms about leaving any PR open indefinitely as long as:

1.  our total number of open PRs does not exceed one page (as defined by GitHub pagination)
2. any PR that I don't need to spend further time on has Changes Requested, because that status is visible from https://github.com/exercism/problem-specifications/pulls.
3. the branch is on a fork instead of on Exercism's copy of the repo

If some means of clearly indicating that an interested party may work on this PR would make this more attractive to those interested parties, I support applying those means. For example, that might be to either: Label, comment, or close this PR. I don't feel a need to express a preference between these options at this time.

exercises/series/canonical-data.json
rpottsoh: > This is basically a combination of the "empty string is invalid" and "slice length cannot be zero" test cases, so I think this can be removed.

Done.

exercises/series/canonical-data.json
rpottsoh: > Maybe instead of "string cannot be empty" we could use "series cannot be empty'?

Done.

exercises/series/canonical-data.json
rpottsoh: > Does this test case add anything to the "slices can include duplicates" test case?

It seemed to be duplicating efforts.  I combined `"slices can include duplicates"` and `"strings can have repeating digits"`.

rpottsoh:
I had an hour this evening so I have donated here.  I am not sure that how I have committed my changes is kosher.  If there is another / better way I should have gone about contributing to this PR / issue please let me know.

exercises/series/description.md
```diff
@@ -1,16 +1,16 @@
...
 
 And the following 4-digit series:
 
-- 4914
-- 9142
+- "4914"
+- "9142"
 
 And if you ask for a 6-digit series from a 5-digit string, you deserve
 whatever you get.
```
rpottsoh: I interpret this last statement to mean that each track can decide how to deal with potentially erroneous input.  I know that has been debated to an extent in this PR.  If the consent were to treat erroneous a certain way then this statement should probably be reworded to reflect that decision.

rpottsoh:
At present there are three different error messages related to `"sliceLength"`.  Would it be preferred to have one error message that states what the criteria is for a good `"sliceLength"` or to leave things as they are, pointing out what is wrong with `"sliceLength"` each time there is an error?  I don't really have a strong preference either way.

cmccandless:
Personally, I prefer specific feedback as it makes it easier to immediately see what the current issue is. 

If I am playing a game, and I violate a rule, it is not useful for another player who has observed the violation to simply hand me the rule book and tell me "you can't do that" (ignoring the fact this is in itself is a rule in some games). It is much better to cite the violated rule and reference the rules (or, to bring the example back home, the README) for further clarification.

rpottsoh:
> Personally, I prefer specific feedback as it makes it easier to immediately see what the current issue is.

Thanks for the feedback @cmccandless.

exercises/series/canonical-data.json
ErikSchierboom: I don't think this test case adds anything over the previous one.

ErikSchierboom:
I think this is looking great, with just one single nit. Thanks for doing this @rpottsoh!

ErikSchierboom:
@petertseng Could you perhaps also review this PR, as you have also given it some thought previously.

petertseng:
Please do not wait for my review, other than I can vouch for the fact that the following code verifies the proposed JSON file:

```ruby
require 'json'

verify(json['cases'], property: 'slices') { |i, _|
  i['series'].each_char.each_cons(i['sliceLength']).map(&:join).tap { |x|
    raise 'Nope, empty is bad' if x.empty?
  }
}
```

I have no additional commentary to add at this time.

ErikSchierboom:
Then I find this ready to be merged. Thanks a lot @N-Parsons for starting with this and @rpottsoh for finishing it!

PR #1021, opened 2017-11-23
book-store: Add test case requiring two moves to solve correctly

coriolinus:
In the Rust track, @shybyte [implemented a very simple solver](http://exercism.io/submissions/9ede17337e074a108c69d767caa3ea79) for this problem which looked like it shouldn't work, but passed all existing test cases, and then requested my input.

I came up with a case which correctly failed their implementation. This PR adds that case to the canonical data.

### DIFF
--- a/exercises/book-store/canonical-data.json
+++ b/exercises/book-store/canonical-data.json
@@ -1,13 +1,13 @@
 {
   "exercise": "book-store",
-  "version": "1.0.1",
+  "version": "1.1.0",
   "cases": [
     {
       "description": "Return the total basket price after applying the best discount.",
       "comments": [
         "Calculate lowest price for a shopping basket containing books only from ",
         "a single series.  There is no discount advantage for having more than ",
-        "one copy of any single book in a grouping."      
+        "one copy of any single book in a grouping."
       ],
       "cases": [
         {
@@ -100,6 +100,13 @@
          "basket": [1,1,2,2,3,3,4,4,5,5,1,2],
          "targetgrouping": [[1,2,3,4,5],[1,2,3,4,5],[1,2]],
          "expected": 75.20
+        },
+        {
+         "property": "total",
+         "description": "Two moves are required to find minimal price",
+         "basket": [1,1,2,2,3,3,4,5,1,1,2,2,3,3,4,5],
+         "targetgrouping": [[1,2,3,4],[1,2,3,5],[1,2,3,4],[1,2,3,5]],
+         "expected": 102.4
         }
      ]
     }
--- a/exercises/book-store/canonical-data.json
+++ b/exercises/book-store/canonical-data.json
@@ -103,7 +103,7 @@
         },
         {
          "property": "total",
-         "description": "Two moves are required to find minimal price",
+         "description": "Four groups of four are cheaper than two groups each of five and three",
          "basket": [1,1,2,2,3,3,4,5,1,1,2,2,3,3,4,5],
          "targetgrouping": [[1,2,3,4],[1,2,3,5],[1,2,3,4],[1,2,3,5]],
          "expected": 102.4


exercises/book-store/canonical-data.json
```diff
@@ -100,6 +100,13 @@
          "basket": [1,1,2,2,3,3,4,4,5,5,1,2],
          "targetgrouping": [[1,2,3,4,5],[1,2,3,4,5],[1,2]],
          "expected": 75.20
+        },
+        {
+         "property": "total",
+         "description": "Two moves are required to find minimal price",
```
Insti: This description does not make sense to me as it seems to be assuming some knowledge of what "moves" are which seems like a very implementation specific thing.
Is there another way to explain this? Is this the only test with four target groupings? Is that somehow relevant?

Insti:
👍 for adding test cases not detected by real life solutions.

exercises/book-store/canonical-data.json
coriolinus: This is in fact the only case with four target groupings, and that is relevant. The issue is that sticking to the naming convention of the rest of the file, you end up with a description like the following:

- "Four groups of four are cheaper than one group of three and two groups of four and one group of five"
- "Four groups of four are cheaper than one group of three and two groups of four and one group of five and also cheaper than two groups of five and two groups of three"
- "Four groups of four are the cheapest organization in this case"

All of these options seem excessively verbose, and none of them express that what this case is really intended to test is that the solver runs the "find a cheaper grouping option" function more than once, if it is designed that way. 

I'm not sure what's best in this case: stick to the "two moves" verbiage, or move to one of the longer ones. Do you have a preference?

Insti: Do not stick with the naming convention of the rest of the file. These are all terribly described tests.

Insti: > none of them express that what this case is really intended to test is that the solver runs the "find a cheaper grouping option" function more than once, if it is designed that way.

That's getting closer to a good description. What is it about the input that means that the "find cheaper  grouping option" needs to be run more than once?

coriolinus: It's easy to get the most compact grouping using a very simple greedy strategy, and most of the time, the optimal grouping will be near the most compact grouping. I imagine most students implement the greedy search for the minimal grouping right away. Sometimes, but not all the time, you can get a more optimal set of groups by moving an item from one group to another. This is what I refer to as a move when I say "two moves are required to find the minimal price". 

This test's input is simply that of the "Two groups of four is cheaper than group of five plus group of three" case, twice. The obvious greedy grouper will produce two groups of five and two groups of three, given that input. Therefore, to find the optimal set of groups, you need to move an item from a group of five into a group of three at least twice. Implementations which attempt to move an item into another group only once will pass every other test, as none of them require more than one move to get from the greedy set of groups to the optimal set of groups, but will not pass this test because a single move leaves you with one group of three, two groups of four, and one group of five.

coriolinus: @Insti I think this has been addressed, then? I still don't have a better concise description for this case than the one which is already in the PR, and if your objection to it has been resolved to your satisfaction, I'd merge this.

Insti: No it has not.  

I don't have time right now but I will try to explain further.

Please bump this later so I get notified if I forget.

Insti: If you could compress [your comment above](https://github.com/exercism/problem-specifications/pull/1021/files/a0e5ea6dcd0dca4b622cbca1a482eeaa07f6b667#r152842175)  down into one line then you'd have a great description.

What about something like:

"groups of four can be cheaper when there are a lot of books"
or 
"four sets of four are cheaper than two sets of five and three"

I like the second one here, as it gives the student a hint about what might be wrong with their solution.
(although there are probably also many other ways of getting it wrong..)

coriolinus: > If you could compress your comment above down into one line then you'd have a great description

That's the crux of the problem: that's the shortest complete description I know how to make, and it's two paragraphs. I'm verbose by default, so the other option is to go with an incomplete description and hope for the best.

"four groups of four are cheaper than two groups each of five and three" works for me. It's incomplete, but it's in line with the style of the rest of the file. It doesn't describe the actual expected failure case (two sets of four, one set of five, one set of three), but the student can figure that out on their own.

Insti: "four groups of four are cheaper than sets that contain groups of five"?

Issue #1022, opened 2017-11-23
word-count: Test for handling of multiple whitespace


jonmcalder:
There is a suggestion in [exercism/r #85](https://github.com/exercism/r/issues/85) that this exercise should include a test for multiple whitespace (i.e. check that words are trimmed of whitespace and that repeated whitespace isn't treated as another "word").

This could be done either by adding a simple new test case, or via adapting the inputs for an existing test case e.g. 
```
"description": "ignore punctuation",
      "property": "countwords",
      "input": " car : carpet as java :  javascript  !!&@$%^&",       # note extra spaces here
      "expected": {
        "car": 1,
        "carpet": 1,
        "as": 1,
        "java": 1,
        "javascript": 1
      }
```

Was this type of test case omitted deliberately in order to simplify the exercise? If not, this seems to me like it could be a worthwhile addition to the canonical test data.

Insti:
[Description](https://github.com/exercism/problem-specifications/blob/master/exercises/word-count/description.md)

Adding a new test case which contains a specific example of this is a good idea.
The input text should be the simplest possible example that will trigger this bug.
The example provided in the linked issue looks fine.

jonmcalder:
Cool - will do so a little later.

ErikSchierboom:
This has been closed by #1023.

PR #1023, opened 2017-11-25
word-count: Add test for multiple whitespace

jonmcalder:
See #1022 for reference

### DIFF
--- a/exercises/word-count/canonical-data.json
+++ b/exercises/word-count/canonical-data.json
@@ -113,7 +113,7 @@
       }
     },
     {
-      "description": "with multiple whitespace",
+      "description": "multiple spaces not detected as a word",
       "property": "countwords",
       "input": " multiple   whitespaces",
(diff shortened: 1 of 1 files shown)


exercises/word-count/canonical-data.json
```diff
@@ -111,6 +111,15 @@
         "large": 2,
         "and": 1
       }
+    },
+    {
+      "description": "with multiple whitespace",
```
Insti: A more helpful description would be nice.
What is it about multiple whitespace that makes it interesting to test?

"space surrounded by spaces not detected as a word"

jonmcalder: Thanks - a valid point. I fell into the trap of just imitating the overly brief descriptions from some of the prior test cases...

jonmcalder:
Thanks @ErikSchierboom!

PR #1025, opened 2017-12-02
Bob: Add fifth rule

### LINKED ISSUE #943: bob: Precedence of the four rules
The https://github.com/exercism/problem-specifications/blob/master/exercises/bob/description.md lists the rules in the order:

* Question -> sure
* Yell -> whoa, chill out
* Silent -> fine, be that way
* Otherwise -> whatever

Is it ever possible that a given input will match two (or more) of these rules? Let's see.

Since the last rule is an "otherwise" rule, it is necessarily exclusive of all others.

The silence rule is exclusive of yelling (assuming that we can only consider a text to be yelling if it has some non-whitespace character) and of questions (since questions necessarily contain the question mark).

But you could have a yelling question, matching the first two rules. For a yelling question, which rule takes precedence? Looking at the description only, I wouldn't know.

The tests have something to say about it though. https://github.com/exercism/problem-specifications/blob/master/exercises/bob/canonical-data.json#L54-L57 says that yelling takes precedence.

What shall we do about the fact that the description does not say which rule takes precedence?

1. Do not define precedence in the description.
2. Define yell > question in description.
3. Define question > yell in description, and change the test in canonical-data.json.

My choice: I do not care.

And, shall we keep the test of the yelling question?

My choice: I do not care.

rpottsoh:
closes #943 

This first commit adds a description for an emphatic question.  The second commit will update the `forceful question` case with the new response I am proposing in the description.

### DIFF
--- a/exercises/bob/description.md
+++ b/exercises/bob/description.md
@@ -4,7 +4,7 @@ Bob answers 'Sure.' if you ask him a question.
 
 He answers 'Whoa, chill out!' if you yell at him.
 
-He answers 'Calm down, I know what I'm doing!' if you question him emphaticlly.
+He answers 'Calm down, I know what I'm doing!' if you yell a question at him.
 
 He says 'Fine. Be that way!' if you address him without actually saying
 anything.
--- a/exercises/bob/description.md
+++ b/exercises/bob/description.md
@@ -4,6 +4,8 @@ Bob answers 'Sure.' if you ask him a question.
 
 He answers 'Whoa, chill out!' if you yell at him.
 
+He answers 'Calm down, I know what I'm doing!' if you question him emphaticlly.
+
 He says 'Fine. Be that way!' if you address him without actually saying
 anything.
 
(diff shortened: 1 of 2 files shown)


exercises/bob/description.md
```diff
@@ -4,6 +4,8 @@ Bob answers 'Sure.' if you ask him a question.
 
 He answers 'Whoa, chill out!' if you yell at him.
 
+He answers 'Calm down, I know what I'm doing!' if you question him emphaticlly.
```
Insti: spelling issue on this line.

Also a bit worried about complex language use here, since there are many non-english speakers.

Insti:
I like adding a new specific non-ambiguous response for this case. 👍 

Remember to bump the version number.

exercises/bob/description.md
rpottsoh: Yup, I see the spelling issue now.  I chose `emphatically` because it helped to keep the line short too.  

...`if you yell your question at him.`  

should suffice.

rpottsoh:
> bump the version number.

1.1.0 you think?

PR #1030, opened 2017-12-10
readme.md: Update example to reflect "input key as an object" policy

### LINKED ISSUE #1007: documentation: Update README.md to reflect the "input key as an object" policy.
In https://github.com/exercism/problem-specifications/issues/996 it was decided that the canonical data should have an explicit input key whose value is an object containing the input(s) to the code under test.

For example:

```json
{
  "description": "year not divisible by 4: common year",
  "property": "leapYear",
  "input": {
    "year": 2015
  },
  "expected": false
}
```

**Task**

Update the example canonical-data.json in the [Readme.md](https://github.com/exercism/problem-specifications/blob/master/README.md) to reflect the "input key as an object" policy.

**Questions**
Is this documented anywhere else?
Should there also be json-schema changes?

rpottsoh:
In #996 it was decided that the canonical data should have an explicit input key whose value is an object containing the input(s) to the code under test.

closes #1007

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,8 +71,10 @@ is easier to understand with an example:
       }
     , { "description": "Bar'ing a name returns its parts combined"
       , "property"   : "bar"
-      , "firstName"  : "Alan"
-      , "lastName"   : "Smithee"
+      , "input"      : {
+          "firstName"  : "Alan",
+          "lastName"   : "Smithee"
+        }  
       , "expected"   : "ASlmainthee"
       }
     , { "comments":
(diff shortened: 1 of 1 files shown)


Stargator:
Not related, but I assume there will be efforts to update the current canonical data to reflect this change as well? I imagine most people will follow a "monkey see, monkey do" approach.

Additionally, the Dart track has a tool to generate a test suite based on the canonical-data.json of a track (if it exists) and will currently need to be updated to handle cases of no canonical-data, canonical-data with old schema, and canonical-data with new schema).

rpottsoh:
I am going to go ahead and merge this since several PRs have been opened to update canonical data based on #996 which is what this PR is illustrating by updating the example canonical data in README.

PR #1039, opened 2017-12-16
binary: Update json for new input policy and fix typo

Stargator:
#996 Updated the schema for canonical data. So I am updating the exercise's json to reflect the change as well as fixed one small typo.

### DIFF
--- a/exercises/binary/canonical-data.json
+++ b/exercises/binary/canonical-data.json
@@ -1,95 +1,125 @@
 {
   "exercise": "binary",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "cases": [
     {
       "description": "binary 0 is decimal 0",
       "property": "decimal",
-      "binary": "0",
+      "input": {
+        "binary": "0"
+      },
       "expected": 0
     },
     {
       "description": "binary 1 is decimal 1",
       "property": "decimal",
-      "binary": "1",
+      "input": {
+        "binary": "1"
+      },
       "expected": 1
     },
     {
       "description": "binary 10 is decimal 2",
       "property": "decimal",
-      "binary": "10",
+      "input": {
+        "binary": "10"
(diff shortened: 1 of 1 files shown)


petertseng:
But the https://github.com/exercism/docs/blob/master/contributing/pull-request-guidelines.md say that the title should not include the word "and", therefore there should be two commits, not one:

1. update JSON for new input policy
2. fix typo

it's the rules I'm very angry about this I can't believe someone would do this blah blah blah blah... OK I admit it I don't actually care, this exercise is deprecated anyway. But I was obligated to point out the guidelines. And it means that I won't approve it since I do agree with the thought that each commit should contain only one change. But since the exercise is deprecated I'm not going to complain if someone else chooses to approve it. No need to wait for me.

Stargator:
@petertseng I like you, don;t ever change :)

petertseng:
(I wonder if someone will complain "but the typo fix should make it 1.1.1. If someone does, there's also the possibility of reordering the commits so that the typo fix makes it 1.0.1 and then the input change makes it 1.1.0 so that we remain at 1.1.0 instead of 1.1.1. If there is an interested party who wishes that to happen, I trust that party to say so.

petertseng:
And then squash and merge completely defeated the purpose of splitting the commits. Well, if that was going to happen anyway, I might as well not have asked. I feel insulted. But there is nothing to be done about it now (I'm not going to ask master to be rebased)

petertseng:
To express it in the form of my hopes and dreams: It is my hope and dream that if I ask for a PR to be split into multiple commits then those commits should not be squashed when being merged. If this was my fault for not being clear, then first I apologise and second I offer that I can be explicit in saying "I approve but only if these commits ARE squashed" or "I approve but only if these commits ARE NOT squashed", etc.

PR #1041, opened 2017-12-16
problem-specifications: Update ajv-cli dependency to latest

Stargator:
As I have been updating the canonical-data.json for a few exercises, I realized the only dependency in package.json was outdated. So this PR bumps the version for ajv-cli to 2.1.0.

I had both scripts within package.json and I have not seen any issues since the upgrade.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,13 +1,13 @@
 {
   "name": "problem-specifications",
-  "version": "1.0.0",
+  "version": "1.1.0",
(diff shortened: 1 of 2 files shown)


petertseng:
> It was my experience in the past that version 1 only had the oneOf error which was distinctly unhelpful.

Confirmed in #1076 as well.

ErikSchierboom:
Thanks @Stargator! 🎉

PR #1042, opened 2017-12-16
problem-specifications: Replace use of node/npm with yarn

Stargator:
I have noticed that in the README we instruct people to use node/npm. Yet we use Yarn to install our dependencies.

So I figured it would make sense to drop any mention of node/npm and instead use yarn. Yarn provides everything we need and we wouldn't be asking people to install two tools for one job.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,4 +6,4 @@ cache: yarn
 
 script:
   - bin/check_required_files_present
-  - npm test
+  - yarn test
--- a/README.md
+++ b/README.md
@@ -170,25 +170,25 @@ PATCH changes would never break well-designed test generators, because the test 
 ## Automated Tests
 
 `canonical-data.json` for each exercise is checked for compliance against the [canonical-schema.json](canonical-schema.json).
-In order to run these tests, you will need to have `node` and `npm` installed on your system.
-Install them from [here](https://nodejs.org/en/). (`npm` comes bundled with most installations of `node`).
+In order to run these tests, you will need to have `yarn` installed on your system.
+Install them from [here](https://yarnpkg.com/en/docs/install).
 
 Install the required packages:
 
 ```shell
-npm install
+yarn install
 ```
 
 Run for all exercises:
 
 ```shell
-npm test
+yarn test
 ```
 
 Run for single exercise:
 
 ```shell
-npm run test-one exercises/<exercise>/canonical-data.json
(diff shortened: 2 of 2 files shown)


ErikSchierboom:
Well, maybe the correct question is: should we be using `yarn` instead of `npm`? There was a significant performance difference when yarn was released, but with NPM 5 and the introduction of lock files, the difference is not that big anymore. I would actually be in favor of just using NPM, as it means one less dependency to install.

tleen:
`npm` ships with Node, `yarn` does not. Usually the instructions are with NPM or both. It is rare to see a project say "yarn" and not say "npm/yarn" isn't it? (This is my anecdotal observation.)

ErikSchierboom:
@petertseng @rpottsoh What do you think about reverting back to using plain NPM instead of yarn?

petertseng:
I don't think about it.

ErikSchierboom:
@Stargator What is your opinion on this matter?

rpottsoh:
I'm fine either way, thanks for asking.

masters3d:
npm should suffice here. there are no packages being imported, any speed improvements of yarn are not effective here.

masters3d:
i guess there is one package but there are only a handful of dependencies. im sticking to my point.

Stargator:
@ErikSchierboom Sorry for the delay.

From my own projects at work, I notice a warning from bower when building the project to switch from using bower to using yarn and webpack instead.

We aren't doing anything frontend, but the [bower team's recommendations](https://bower.io/blog/) amount to:
1. Yarn uses checksums to verify the integrity of every installed package (like npm@5)
1. Yarn uses lockfile to exactly reproduce installed packages each time (like npm@5)
1. Yarn supports most features npm supports, and is able to force flattening of dependencies

I am not sure if npm 5 forces flattening of dependencies.

That's where this got started. But overall, I think either use npm or yarn. Having both seems odd and confusing.

ErikSchierboom:
I agree, and I have no real strong opinion on the matter.

tleen:
An update to my previous comment. Shopify [recommends Yarn over NPM](https://shopify.github.io/slate/commands/) for their templating tool.

ErikSchierboom:
Well, let's go with yarn then for the moment. We can always change this back to NPM should we deem it necessary.

Stargator:
Cool

PR #1044, opened 2017-12-19
two-fer: fix case inconsistency

sjwarner-bp:
I implemented this exercise for Clojure and noticed the case was inconsistent. I hope this fix is ok, and the odd case was not intentional - let me know :slightly_smiling_face:


ErikSchierboom:
Thanks @sjwarner-bp!

PR #1045, opened 2017-12-20
alphametics: Update test description

nikamirrr:
The test contains 199 addends, not 41. Sorry

### DIFF
--- a/exercises/alphametics/canonical-data.json
+++ b/exercises/alphametics/canonical-data.json
@@ -98,7 +98,7 @@
           }
         },
         {
-          "description": "puzzle with ten letters and 41 addends",
+          "description": "puzzle with ten letters and 199 addends",
           "property": "solve",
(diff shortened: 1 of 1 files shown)


petertseng:
Yup.

Please prefix commit message and PR description with "alphametics:"

I would like to take this without changing version since this is something that should have happened with 1.1.0

However I guess there is a possibility that someone could have used 1.1.0 to generate. I could check all the repos to see whether anyone in fact did so, but lazy.

So, I think we have to have version 1.1.1, unless there is more agreement that staying 1.1.0 is OK. I don't care either way but if I don't hear from anyone else I'm going to ask 1.1.1.

rpottsoh:
I think to be PC the version should be bumped.

nikamirrr:
Like this?
 "version": "1.1.0" ->  "version": "1.1.1",

ErikSchierboom:
@nikamirrr Yes, precisely!

ErikSchierboom:
Thanks @nikamirrr!

nikamirrr:
Absolutely! Thank you, I am not a pro developer, and this is the opportunity for me to get more familiar with github and such.

PR #1047, opened 2017-12-21
transpose: add test case for mixed line length

### LINKED ISSUE #1046: Transpose: add test-case
The current version of the transpose-test cases (1.0.0) seem to miss out on a case:
There are two test-cases which cover the correct right-stripping of white-space
("first line longer than second line" and "many lines"), yet they do not cover the case, where you have to right-trim whitespace with multiple lines at the end (Idk if i describe it correctly).
I noticed it, when I checked some implementations in the Python track which failed my following suggested test:

```json
{
  "description": "descending line length",
  "property": "transpose",
  "input": [
    "The longest line.",
    "A longer line.",
    "A long line."
  ],
  "expected": [
    "TAA",
    "h  ",
    "ell",
    " oo",
    "lnn",
    "ogg",
    "ne ",
    "grl",
    "e i",
    "sln",
    "tie",
    " n.",
    "le",
    "i.",
    "n",
    "e",
    "."
  ]
}
```

pheanex:
closes #1046


rpottsoh:
@pheanex #996 states that all inputs should now appear with an "input" object.  I am not sure if that should be handled as a separate commit of this PR, or held off and taken care of in a new PR.  In either event `version` should be bumped to "1.1.0".  It also appears that a key name will need to be invented as there is presently none.

If the decision is made to hold off on changes per #996 for another PR, then I think this PR is OK.

Also, I agree with @ErikSchierboom's [comment](https://github.com/exercism/problem-specifications/issues/1046#issuecomment-353328175) in #1046.

pheanex:
I thought I'll just add the new schema, when I am already at it.

rpottsoh:
This looks ok to me.  Some may baulk at the fact that you have performed more than one change in a single commit.  That change to the test case should be in one commit and changed for the input policy in one that follows the first.  The PR would likely be merged without squashing in order to maintain the traceability of the changes.  Are you able to make the necessary changes to this PR so that first commit is for the change to the test case followed by another commit for the policy update?

pheanex:
Like so?

rpottsoh:
Excellent,  thank you.

PR #1050, opened 2017-12-27
meetup: Update json for new input policy

ErikSchierboom:
It also adds an "expected" property, that was previously missing but now makes even more sense as the input properties are contained in a separate property.


mienaikage:
Was the removal of `dayofmonth` intentional?

ErikSchierboom:
@mienaikage Yes it was, as that was only used to determine the `expected` value, but we know use a date string for that (which matches existing expected values).

PR #1054, opened 2017-12-30
luhn: Apply new "input" policy

rpottsoh:
Per #996

### DIFF
--- a/exercises/luhn/canonical-data.json
+++ b/exercises/luhn/canonical-data.json
@@ -1,83 +1,109 @@
 {
   "exercise": "luhn",
-  "version": "1.0.1",
+  "version": "1.1.0",
   "cases": [
     {
       "description": "single digit strings can not be valid",
       "property": "valid",
-      "input": "1",
+      "input": {
+        "ccNumber": "1"
+      },
       "expected": false
     },
     {
       "description": "a single zero is invalid",
       "property": "valid",
-      "input": "0",
+      "input": {
+        "ccNumber": "0"
+      },
       "expected": false
     },
     {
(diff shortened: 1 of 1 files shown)


exercises/luhn/canonical-data.json
```diff
@@ -1,83 +1,109 @@
...
   "exercise": "luhn",
-  "version": "1.0.1",
+  "version": "1.1.0",
   "cases": [
     {
       "description": "single digit strings can not be valid",
       "property": "valid",
-      "input": "1",
+      "input": {
+        "ccNumber": "1"
```
ErikSchierboom: I assume `ccNumber` stands for `creditCardNumber`. If so, why not write it out and use `creditCardNumber` instead?

rpottsoh: No reason not to other than the description.md also says it could be used for Canadian Social Insurance Numbers.  I didn't want the field to be overly long and I couldn't come up with a good, meaningful, name that was generic enough for this first commit. 

Yes, `ccNumber` stands for `creditCardNumber`.

ErikSchierboom: Aha, I see. Maybe we should then revert to something plain like "input"?

rpottsoh: "input" inside of "input"? 👎 Technically, I guess, still satisfies #996.  How about `"aValue"`?

ErikSchierboom: I don't really like the "a" prefix. How about just "value"?

rpottsoh: Sold!  Thanks @ErikSchierboom.

PR #1055, opened 2017-12-30
isbn-verifier: add empty isbn case

### LINKED ISSUE #1052: isbn-verifier: Test Suite Omits Important Edge Case
Reported by @p0st0culus in https://github.com/exercism/exercism.io/issues/3702

Confirmed missing from canonical data.

----

The test suite for the ISBN Verifier problem in the Swift track should test for an empty string ISBN.  This is a an important edge case that @Damlar noted was missing from my solutions which is also omitted from the test suite.

It's something that, in hindsight, I should have known to test for, but I didn't.  That's the whole point of a well thought out test suite though:  Code against assertions of what a valid solution should produce.  I tend to implicitly put trust in the test suites, so that's my lesson learned.  I'll take an active role in the test suites going forward.

I recommend modifying IsbnVerifierTests.swift as follows.  Other content from the test suite omitted with ellipses for clarity:

````swift
class IsbnVerifierTests: XCTestCase {

    //...

    func testEmptyIsbn() {
        XCTAssertFalse(IsbnVerifier.isValid(""))
    }

    //...

    static var allTests: [(String, (IsbnVerifierTests) -> () throws -> Void)] {
        return [
            //...
            ("testEmptyIsbn", testEmptyIsbn),
            //...
        ]
    }
}
````

Thanks.

rpottsoh:
I have inserted the new test case at the end of  the test data.  I am not sure where it might fit better.
[description](https://github.com/exercism/problem-specifications/blob/master/exercises/isbn-verifier/description.md)
closes #1052

### DIFF
--- a/exercises/isbn-verifier/canonical-data.json
+++ b/exercises/isbn-verifier/canonical-data.json
@@ -1,6 +1,6 @@
 {
     "exercise": "isbn-verifier",
-    "version": "2.1.0",
+    "version": "2.2.0",
     "comments": [
         "An expected value of true indicates a valid ISBN-10, ",
         "whereas false means the ISBN-10 is invalid."
@@ -107,6 +107,14 @@
             "property": "isValid",
             "input": {
               "isbn": "3-598-21515-X"
+            },
+            "expected": false
+        },
+        {
+            "description": "empty isbn",
+            "property": "isValid",
+            "input": {
+              "isbn": ""
             },
             "expected": false
         }


rpottsoh:
@p0st0culus please give this PR a look.  Let me know if you think the new test case should be moved to a different location, and where.

p0st0culus:
@rpottsoh, I’m not sure where I should be looking...

rpottsoh:
@p0st0culus try [this](https://github.com/rpottsoh/exercism-problem-specifications/blob/4288ee0400c5f37c318f0ceb8ee2a88ef883bdb7/exercises/isbn-verifier/canonical-data.json).  The new test case appears at the end of the file.

p0st0culus:
Looks great. I don’t think the order is much important, so long as it’s in there. Thanks. :)

PR #1072, opened 2018-01-05
 travis: Choose schema based on USE_OLD_SCHEMA file

petertseng:
If it is present, use the old schema, else use the new schema.

This plan is proposed in:
#998

Its merits are that it allows us to:

* Track progress by simply counting number of USE_OLD_SCHEMA file.
* Verify, as each PR is made, that each JSON file intending to use new
  schema does in fact validate.
  * Otherwise we have to manually run a schema check at periodic
    intervals.
* Keep compatibility with old schema, to support moving over one file at
  a time instead of all at once.

Note that old-schema.json is the same canonical-schema.json as is
currently on master.

The intention is that when all exercises use the new schema, **the last
commit in this PR can/should be reverted** and then all exercises will in fact use
the new schema, which will replace the old completely.

### DIFF
--- a/canonical-schema.json
+++ b/canonical-schema.json
@@ -3,11 +3,9 @@
    [ " This is a JSON Schema for 'canonical-data.json' files.     "
    , "                                                            "
    , " It enforces just a general structure for all exercises,    "
-   , " without specifying how the test data should be organized   "
-   , " for each type of test. We do this to keep generality and   "
-   , " allow support for tests the do not fit well in the         "
-   , " 'function (input) == output' structure, like property      "
-   , " tests.                                                     "
+   , " optimising for the ability to represent example-based tests"
+   , " of the form 'function (input) == output'. Future expansions"
+   , " may allow for other types of tests such as property-based. "
    , "                                                            "
    , " The only thing enforced regarding how test data should be  "
    , " structured is the error encoding, because it was agreed    "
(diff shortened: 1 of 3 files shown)


petertseng:
As we can see in #1073, we enforce the condition: "If USE_OLD_SCHEMA is not present, the file is a correct file under the new schema".

Note that since the new schema is a superset of the old, I have a difficult time enforcing the **converse** condition: "If the file is a correct file under the new schema, USE_OLD_SCHEMA is not present".

This means someone may convert a JSON file to the new schema and then *forget to remove USE_OLD_SCHEMA*, and we have no way of detecting it.

If it is very important, we can take the strategy: "If USE_OLD_SCHEMA is present, **also** enforce that the file is **invalid** under the **new** schema". That will enforce that we do not forget to remove the USE_OLD_SCHEMA file. The cost is simply extra complexity in .travis.yml

petertseng:
> If it is very important, we can take the strategy: "If USE_OLD_SCHEMA is present, **also** enforce that the file is **invalid** under the **new** schema". That will enforce that we do not forget to remove the USE_OLD_SCHEMA file. The cost is simply extra complexity in .travis.yml


OK, done in #1074

petertseng:
Uh guess it behooves me to rebase this now that alphametics is on the thing

canonical-schema.json
```diff
@@ -85,13 +83,15 @@
       "labeledTest":
           { "description": "A single test with a description"
           , "type"       : "object"
-          , "required"   : ["description", "property"]
+          , "required"   : ["description", "property", "input", "expected"]
           , "properties" :
                 { "description": { "$ref": "#/definitions/description" }
                 , "comments"   : { "$ref": "#/definitions/comments"    }
                 , "property"   : { "$ref": "#/definitions/property"    }
+                , "input"      : { "$ref": "#/definitions/input"    }
```
petertseng: how did I cause this to get misaligned?

petertseng: Because I was lazy in #998.

petertseng:
Because #1074 exists, I refuse to maintain this PR any further. I do not prefer the ability to forget to remove a USE_OLD_SCHEMA file.

PR #1077, opened 2018-01-07
[reverse-string] Remove title from metadata

budmc29:


petertseng:
Weird. Guess many people (including me) missed it...

I would rather remove the `title` line completely to signal that the default title should be used. It is more consistent with all other exercises if the `title` line is only present if a non-default title is being used.

exercises/reverse-string/metadata.yml
```diff
@@ -1,5 +1,5 @@
 ---
 blurb: "Reverse a string"
-title: "reverse-string"
+title: "Reverse String"
 source: "Introductory challenge to reverse an input string"
-source_url: "https://medium.freecodecamp.org/how-to-reverse-a-string-in-javascript-in-3-different-ways-75e4763c68cb"
\ No newline at end of file
+source_url: "https://medium.freecodecamp.org/how-to-reverse-a-string-in-javascript-in-3-different-ways-75e4763c68cb"
```
petertseng: unrelated whitespace change should go in its own commit in this PR, or in a different PR

budmc29: Sorry about that, I'm not bothered by the newline missing so I'll leave it as it is.

budmc29:
Thank you.

PR #1080, opened 2018-01-08
gigasecond: Apply new "input" policy

ErikSchierboom:
See #996


ErikSchierboom:
> "birthdate", should it be "birthDate" per #987?

Well, it depends. I've checked some resources, and found this nice article on quora: https://www.quora.com/Is-it-birthdate-or-birth-date It posits that the word is "birthdate" and not "birth date".

rpottsoh:
My mind wondered about that too after I posted my comment; birthdate is an actual word, even though spell check balks at the spelling.  Birthdate is the first person spelling.

PR #1084, opened 2018-01-08
two-bucket: Apply new "input" policy

ErikSchierboom:
See #996


ErikSchierboom:
@rpottsoh WelI spotted. I would prefer to do that in a separate PR.

PR #1089, opened 2018-01-08
book-store: convert to fixed point output

### LINKED ISSUE #964: book-store: should we refrain from using floating point?
When I search for "why not use floating point for money" in my preferred search engine, I get many results, such as https://stackoverflow.com/questions/3730019/why-not-use-double-or-float-to-represent-currency (by the way, searching for "why use floating point for money" also gives reasons in opposition rather than reasons in support).

Should we take their advice, and instead use integer number of cents?

It's a little tricky because it is sure possible that calculating the discounts may involve floating point math. But not in this case; here the discount values always result in whole numbers of cents. Thus one can conceive of a way to do this without any floating point math.

It also may be a bit jarring to students, since the units will not be the same as those in the description.

I do not maintain any track that has this exercise, so consider this as a curious question of an unknowing outsider. It does not particularly matter to me which choice we take.

There are previous discussions about floating point in space-age (https://github.com/exercism/problem-specifications/pull/880) and triangle (https://github.com/exercism/problem-specifications/issues/798) but my providing these links is not to be construed as an assertion or denial that those past discussions are relevant to this one.

rpottsoh:
Convert "expected" output to fixed point (cents) to remove floating point values.
closes #964

### DIFF
--- a/exercises/book-store/canonical-data.json
+++ b/exercises/book-store/canonical-data.json
@@ -9,7 +9,8 @@
         "a single series.  There is no discount advantage for having more than ",
         "one copy of any single book in a grouping.",
         "implementors should use proper fixed-point or currency data types of the ",
-        "corresponding language and not float."
+        "corresponding language and not float.",
+        "All 'expected' amounts are in cents."
       ],
       "cases": [
         {
--- a/exercises/book-store/canonical-data.json
+++ b/exercises/book-store/canonical-data.json
@@ -1,13 +1,15 @@
 {
   "exercise": "book-store",
-  "version": "1.2.0",
+  "version": "1.3.0",
   "cases": [
     {
       "description": "Return the total basket price after applying the best discount.",
       "comments": [
         "Calculate lowest price for a shopping basket containing books only from ",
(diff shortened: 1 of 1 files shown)


exercises/book-store/canonical-data.json
```diff
@@ -1,13 +1,15 @@
...
+  "version": "1.3.0",
   "cases": [
     {
       "description": "Return the total basket price after applying the best discount.",
       "comments": [
         "Calculate lowest price for a shopping basket containing books only from ",
         "a single series.  There is no discount advantage for having more than ",
-        "one copy of any single book in a grouping."
+        "one copy of any single book in a grouping.",
+        "implementors should use proper fixed-point or currency data types of the ",
```
petertseng: maybe it would be helpful to be explicit in saying that the values in the JSON file are numbers of cents?

rpottsoh:
Added a quick blurb about "expected" currency denomination.

rpottsoh:
Travis is complaining about the quotes I included in the comment...  I'll take care of that.

PR #1090, opened 2018-01-09
phone-number: add missing edge cases

cmccandless:
The existing descriptions for invalid area codes and exchange codes both cite a range, but only test a single value. This PR tests for all values outside the valid range.

### DIFF
--- a/exercises/phone-number/canonical-data.json
+++ b/exercises/phone-number/canonical-data.json
@@ -93,7 +93,7 @@
           "property": "clean",
           "phrase": "(223) 156-7890",
           "expected": null
-        },
+        }
       ]
     }
   ]
--- a/exercises/phone-number/canonical-data.json
+++ b/exercises/phone-number/canonical-data.json
@@ -71,39 +71,29 @@
           "expected": null
         },
         {
-          "description": "invalid if area code does not start with 2-9",
-          "cases": [
-            {
-              "description": "invalid if area code starts with 0",
-              "property": "clean",
-              "phrase": "(023) 456-7890",
-              "expected": null
-            },
-            {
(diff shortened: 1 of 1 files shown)


rpottsoh:
Thanks for bumping that.  It is so easy to miss.

exercises/phone-number/canonical-data.json
```diff
@@ -72,15 +72,37 @@
...
+              "expected": null
+            },
+            {
+              "description": "invalid if area code starts with 1",
+              "property": "clean",
+              "phrase": "(123) 456-7890",
+              "expected": null
+            }
+          ]
         },
```
leenipper: Agree with the new edge test cases.  But, would prefer it without another level of "cases": [ ].  I suggest the two invalid cases be at same level as the other clean test cases.  Otherwise, the change will break some generators unnecessarily.

exercises/phone-number/canonical-data.json
```diff
@@ -72,15 +72,37 @@
...
+              "phrase": "(223) 056-7890",
+              "expected": null
+            },
+            {
+              "description": "invalid if exchange code starts with 1",
+              "property": "clean",
+              "phrase": "(223) 156-7890",
+              "expected": null
+            }
+          ]
```
leenipper: Ditto here. Prefer two new invalid cases, but without another nested level of cases.

exercises/phone-number/canonical-data.json
cmccandless: Fair enough; done.

leenipper: Thanks!

ErikSchierboom:
Merged. Thanks!

PR #1093, opened 2018-01-12
two-bucket: remove broken link

sjwarner-bp:
Looks like the link in two-bucket READMEs to http://lindsaylevine.com/ doesn't seem to work anymore.

I've had a quick look and http://lindsay.engineer is a website that does seem to work, though I am not 100% sure if it is the same person or not! I've removed the link for now and replaced with a full name, though if other people know that the other website is actually correct then I'm happy to update it to that too :slightly_smiling_face:

### DIFF
--- a/exercises/two-bucket/description.md
+++ b/exercises/two-bucket/description.md
@@ -25,4 +25,4 @@ To conclude, the only valid moves are:
 - emptying one bucket and doing nothing to the other
 - filling one bucket and doing nothing to the other
 
-Written with <3 at [Fullstack Academy](http://www.fullstackacademy.com/) by [Lindsay](http://lindsaylevine.com).
+Written with <3 at [Fullstack Academy](http://www.fullstackacademy.com/) by Lindsay Levine.


sjwarner-bp:
Agreed :slightly_smiling_face:

ErikSchierboom:
Did anybody manage to find another link?

sjwarner-bp:
Only the one linked in the original description 🙂 Still not 100%, but I haven't found anything better!

petertseng:
I cannot at this time reliably tell the difference between "I needed to change my URL for various reasons and could not set a 302 on my old one but it'd be great if all my old URLs could be changed to my new one" versus "I needed to change my URL because I want to dissociate with my old identity and it is very important that the old URL not be changed to the new one". It may be the case that the most courteous thing to do is to ask the author of https://github.com/exercism/problem-specifications/pull/40 what to do with it.

ErikSchierboom:
@lindslev You are the original author of #40, perhaps you can help us with this issue. The old link for the two-bucket problem doesn't work anymore, is there perhaps another link?

ErikSchierboom:
@petertseng In my comment above, I've tried to reach out to the original link's author, but there was no responsive unfortunately. Therefore I think we should merge this PR to at least not have a broken link. We can always add back a link later.

sjwarner-bp:
Thanks for doing that @ErikSchierboom - I agree with you, if we get more info later it isn't hard to add the appropriate link back in :sli

petertseng:
👍  of course.

ErikSchierboom:
Merged. Thanks @sjwarner-bp!

PR #1095, opened 2018-01-12
complex-numbers: exponential of complex number

cmccandless:
For some reason, there was no test for exp(a + bi) where a + bi has both a real part and an imaginary part. This PR adds such a test with exact values that are simple to define (requiring no rounded values).

Proof:
```
e^(ln(2) + PI * i) = e^ln(2) * e^(PI * i)
e^ln(2) = 2
e^(PI * i) = cos(PI) + isin(PI) = -1 + 0i = -1
e^(ln(2) + PI * i) = 2 * -1 = -2
```

### DIFF
--- a/exercises/complex-numbers/canonical-data.json
+++ b/exercises/complex-numbers/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "complex-numbers",
-  "version": "1.2.0",
+  "version": "1.3.0",
   "comments": [
     " The canonical data assumes mathematically correct real   ",
     " numbers. The testsuites should consider rounding errors  ",
--- a/exercises/complex-numbers/canonical-data.json
+++ b/exercises/complex-numbers/canonical-data.json
@@ -327,7 +327,9 @@
         {
           "description": "Exponential of a number with real and imaginary part",
           "property": "exp",
-          "input": ["ln(2)", "pi"],
+          "input": {
+            "z": ["ln(2)", "pi"]
+          },
           "expected": [2, 0]
         }
       ]
--- a/exercises/complex-numbers/canonical-data.json
+++ b/exercises/complex-numbers/canonical-data.json
@@ -323,6 +323,12 @@
             "z": [1, 0]
           },
           "expected": ["e", 0]
+        },
+        {
+          "description": "Exponential of a number with real and imaginary part",
+          "property": "exp",
+          "input": ["ln(2)", "pi"],
+          "expected": [2, 0]
         }
       ]
     }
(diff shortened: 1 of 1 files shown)


wolf99:
Good catch @cmccandless

ErikSchierboom:
@cmccandless This PR currently has a merge conflict. Could you perhaps look at it?

Insti:
https://github.com/exercism/problem-specifications/blob/master/exercises/complex-numbers/description.md

Version should probably be `1.3.0` (once the new test is applied to the new json) since this PR is adding a test.

rpottsoh:
I think something didn't go as intended in [e73ac93](https://github.com/exercism/problem-specifications/pull/1095/commits/e73ac93feaab5fb3fe00cd220a62aa9970022474).  I am showing 40 modified and 4 added files.

rpottsoh:
Conflict resolved.  The rev number still needs to be bumped to 1.3.0.

cmccandless:
@rpottsoh Thanks for resolving that. I was lazy and tried to resolve the merge in-browser. I will bump the version to 1.3.0.

exercises/complex-numbers/canonical-data.json
```diff
@@ -323,6 +323,12 @@
             "z": [1, 0]
           },
           "expected": ["e", 0]
+        },
+        {
+          "description": "Exponential of a number with real and imaginary part",
+          "property": "exp",
+          "input": ["ln(2)", "pi"],
```
Insti: This seems to be missing the `{ z: }` bits.

cmccandless: These changes were originally made before the adaptation to the new schema; I will correct it.

exercises/complex-numbers/canonical-data.json
```diff
@@ -323,6 +323,14 @@
...
           },
           "expected": ["e", 0]
+        },
+        {
+          "description": "Exponential of a number with real and imaginary part",
+          "property": "exp",
+          "input": {
+            "z": ["ln(2)", "pi"]
+          },
+          "expected": [2, 0]
```
petertseng: According to my calculations, and also according to your calculations in https://github.com/exercism/problem-specifications/pull/1095#issue-288173324, the answer is `-2`. But given that I see here `[2, 0]` that would be lead me to believe the answer is 2.

* Should this be `[-2, 0]` instead?
* Did we both  mess up in our calculations?

cmccandless: > Should this be [2, 0] instead?

@petertseng Yes; very good catch. I must've missed the hyphen when creating the test case. I will make the change.

PR #1097, opened 2018-01-14
list-ops: Normalize format of functions

ErikSchierboom:
There were two formats being used to define a function:

1. A unary function where the parameter is not defined by assumed to be named `value`. Example: `value + 1`
2. A binary function where the parameters are defined by `(x, y)`. Example: `(x, y) -> x + y`

### DIFF
--- a/exercises/list-ops/canonical-data.json
+++ b/exercises/list-ops/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "list-ops",
-  "version": "2.0.0",
+  "version": "2.1.0",
   "comments": [
     "Though there are no specifications here for dealing with large lists,",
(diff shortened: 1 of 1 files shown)


exercises/list-ops/canonical-data.json
```diff
@@ -57,14 +57,14 @@
           "description": "empty list",
           "property": "filter",
           "list": [],
-          "function": "value modulo 2 == 1",
+          "function": "(x) -> x modulo 2 == 1",
```
rpottsoh: @ErikSchierboom would "list" and "function" be considered part of "input" in a future PR?

ErikSchierboom: Yes, they would!

PR #1116, opened 2018-01-15
isbn-verifier: Provide metadata title to fix generated name

kytrinyx:
When regenerating the ISBN Verifier exercise READMEs, the title is generated as _Isbn Verifier_. In order to provide the correct acronym for the generator, this adds the optional `title:` property to the metadata file.


exercises/isbn-verifier/metadata.yml
```diff
@@ -1,4 +1,5 @@
 ---
+title: "ISBN Verifier"
 blurb: "Check if a given string is a valid ISBN-10 number."
 source: "Converting a string into a number and some basic processing utilizing a relatable real world example."
-source_url: "https://en.wikipedia.org/wiki/International_Standard_Book_Number#ISBN-10_check_digit_calculation"
\ No newline at end of file
+source_url: "https://en.wikipedia.org/wiki/International_Standard_Book_Number#ISBN-10_check_digit_calculation"
```
petertseng: I'm obliged to request that the whitespace change goes in its own commit. It is of no concern to me whether the request is accepted or not, but please understand that I must request it.

kytrinyx: Sure, no problem. The whitespace difference is what I got from saving the file in my editor. I'll go ahead and submit a general fix for all the metadata files to ensure that the last line includes a newline, since I believe that is the default behavior of text editors.

PR #1117, opened 2018-01-15
Normalize whitespace in metadata yaml files

kytrinyx:
While tweaking one of the exercise I realized that a small handful of Yaml files didn't contain a newline at the end, which means that if we ever change it, it would have a noisy and unnecessary whitespace change.


ErikSchierboom:
Thanks! 🎉

PR #1119, opened 2018-01-15
Anagram: Improve Test Coverage

### LINKED ISSUE #1067: Anagram: Test simplification and/or removal
I think that the suite of tests in `Anagram` can be improved.

1. Some of the tests are already covered by other tests and can be removed.
2. Some tests are testing for too many things and can be simplified.

I've opened a PR in [exercism/java](https://github.com/exercism/java/pull/1164), please do take a look and see if what I'm doing makes sense :)

Zhiyuan-Amos:
Fixes #1067

### DIFF
--- a/exercises/anagram/canonical-data.json
+++ b/exercises/anagram/canonical-data.json
@@ -83,9 +83,9 @@
       "property": "anagrams",
       "input": {
         "subject": "corn",
-        "candidates": ["corn", "dark", "Corn", "rank", "CORN", "cron", "park"]
+        "candidates": ["corn", "Corn", "CORN"]
       },
-      "expected": ["cron"]
+      "expected": []
     },
     {
       "description": "does not detect non-anagrams with identical checksum",
--- a/exercises/anagram/canonical-data.json
+++ b/exercises/anagram/canonical-data.json
@@ -79,15 +79,6 @@
       "expected": ["gallery", "regally", "largely"]
     },
     {
-      "description": "does not detect identical words",
-      "property": "anagrams",
-      "input": {
-        "subject": "corn",
-        "candidates": ["corn", "Corn", "CORN"]
-      },
-      "expected": []
-    },
-    {
       "description": "does not detect non-anagrams with identical checksum",
       "property": "anagrams",
       "input": {
--- a/exercises/anagram/canonical-data.json
+++ b/exercises/anagram/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "anagram",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "comments": [
     "The string argument cases possible matches are passed in as",
     "individual arguments rather than arrays. Languages can include",
--- a/exercises/anagram/canonical-data.json
+++ b/exercises/anagram/canonical-data.json
@@ -18,15 +18,6 @@
       "expected": []
     },
     {
-      "description": "detects simple anagram",
-      "property": "anagrams",
-      "input": {
-        "subject": "ant",
-        "candidates": ["tan", "stand", "at"]
-      },
-      "expected": ["tan"]
-    },
-    {
       "description": "detects two anagrams",
       "property": "anagrams",
       "input": {
--- a/exercises/anagram/canonical-data.json
+++ b/exercises/anagram/canonical-data.json
@@ -115,15 +115,6 @@
       "expected": ["Carthorse"]
     },
     {
-      "description": "does not detect a word as its own anagram",
-      "property": "anagrams",
-      "input": {
-        "subject": "banana",
-        "candidates": ["Banana"]
-      },
-      "expected": []
-    },
-    {
       "description": "does not detect a anagram if the original word is repeated",
       "property": "anagrams",
       "input": {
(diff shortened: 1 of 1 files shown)


Zhiyuan-Amos:
@rpottsoh & @coriolinus tagging both of you here since both of you reviewed my issue :)

exercises/anagram/canonical-data.json
```diff
@@ -131,6 +131,15 @@
         "candidates": ["Banana"]
       },
       "expected": []
+    },
+    {
+      "description": "detects mixed-case anagrams",
```
rpottsoh: This appears to be a repeat of `"detects anagrams case-insensitively"`

Zhiyuan-Amos: Oh yes you're right! That way, should I make the `"detects anagrams case-insensitively"` harder?

E.g.:
```
 {
   "description": "detects anagrams case-insensitively",
   "property": "anagrams",
   "input": {
     "subject": "OrChEsTrA",
     "candidates": ["cashregister", "CaRtHoRsE", "radishes"]
   },
   "expected": ["CaRtHoRsE"]
 }
```

rpottsoh: Why is that harder?  Still seems redundant.

Zhiyuan-Amos: It's harder in the sense that there's more mixed-case letters. For e.g.:

```
 {
   "description": "detects simple anagram",
   "property": "anagrams",
   "input": {
     "subject": "ant",
     "candidates": ["tan", "stand", "at"]
    },
    "expected": ["tan"]
 },

 {
   "description": "detects anagram",
   "property": "anagrams",
   "input": {
     "subject": "listen",
     "candidates": ["enlists", "google", "inlets", "banana"]
   },
   "expected": ["inlets"]
 },
```

Both tests that I mentioned above are actually testing for the same thing (lower-cased anagram matches). The length differentiates between the difficulty.

Or do you think it's unnecessary, and I should remove one of the test case above as well? :P

Zhiyuan-Amos:
@rpottsoh your call on whether we should remove test cases? As I mentioned [here](https://github.com/exercism/problem-specifications/issues/1067#issuecomment-356461870), it seems to me that having multiple test cases performing the same thing, yet having different descriptions, can be confusing for students doing these exercises.

A possible middle-ground: Duplicate tests have the same description. E.g.
```
{
  "description": "anagrams must use all letters exactly once (1)",
  "property": "anagrams",
  "input": {
    "subject": "tapper",
    "candidates": ["patter"]
  },
  "expected": []
}

{
  "description": "anagrams must use all letters exactly once (2)",
  "property": "anagrams",
  "input": {
    "subject": "galea",
    "candidates": ["eagle"]
  },
  "expected": []
}
```

This makes it clear to the students that these tests are testing for the same functionality.

exercises/anagram/canonical-data.json
rpottsoh: The more I think about this the less I know what to do.  I think at the very least `"detects mixed-case anagrams"` could be moved up to be with the other `orchestra` test cases.

@Insti, @petertseng or @ErikSchierboom do you have any thoughts on this?

Maybe nothing should be done, and just merge this PR as is?

petertseng:
I quote from https://github.com/exercism/docs/blob/master/language-tracks/exercises/anatomy/test-suites.md that a good individual test "introduces a single new requirement". For those who accept that documentation and its definitions, those tests that introduce zero new requirements are not good according to that documentation (which I'll refer to as Good in the rest of this comment, to avoid ambiguity). The file also talks about Good test suites, and it's probably reasonable to assume that a Good test suite is one that is comprised entirely of Good tests. It's also probably reasonable to assume that it is desirable for the canonical-data.json to be a Good test suite.

exercises/anagram/canonical-data.json
Zhiyuan-Amos: @rpottsoh following what's been mentioned [here](https://github.com/exercism/problem-specifications/pull/1119#issuecomment-357855859) I should: 
1. Have a new commit to remove the test with description `"detects simple anagram"`, 
1. Remove this test.

ErikSchierboom:
> [...] those tests that introduce zero new requirements are not good according to that documentation

I agree with this. If there is a test that does not introduce a new concept and just works if the previous test also works, I think that test should be removed.

Zhiyuan-Amos:
I've updated this PR and updated the commit messages to include the link: https://github.com/exercism/docs/blob/master/language-tracks/exercises/anatomy/test-suites.md

Concerning the version, I only updated it as the last commit.

rpottsoh:
@Zhiyuan-Amos thanks for making these changes.  I was hung up on the orchestra cases, and never considered the cases you ended up removing.  I think you have done a good job here and have improved this exercise, thanks!
@coriolinus, @petertseng, @ErikSchierboom, thanks guys for the help and weighing in on this. :+1: 

To Squash and merge or just merge?  I'm leaning towards just a merge commit.

ErikSchierboom:
I'm also leaning towards a merge commit, seeing how the commit messages are very descriptive.

ErikSchierboom:
Thanks @Zhiyuan-Amos!

Zhiyuan-Amos:
@ErikSchierboom @rpottsoh glad to be of help! ^^

PR #1123, opened 2018-01-15
pig-latin: Apply new "input" policy

ErikSchierboom:
See #996


ErikSchierboom:
> would "phrase" or "string" be better than "word" since the last case has more than one word as input?

Agreed. I've changed the name to "phrase".

Issue #761, opened 2017-04-18
Identify computer science algorithm exercises


kytrinyx:
We currently have at least one, maybe several exercises that asks the learner to implement a specific algorithm.

We discussed whether this type of exercise is a good fit for Exercism, in https://github.com/exercism/discussions/issues/124 and concluded that:

> Instead of asking "Write a program that implements erastosthenes sieve." we should ask "Write a program which returns a list of prime numbers, there are a couple of efficient algorithms: erastostenes sive, wheel sieve, etc", perhaps even with links.

We need to identify which exercises target an algorithm rather than targeting _what the algorithm does_.

The one that started the discussion in `sieve`, but I'm not sure if that's the only one.

Once we've identified them, we need to open a new issue to implement the new exercise and deprecate the old one.

petertseng:
`binary-search` asks for a specific algorithm.

ErikSchierboom:
The `diffie-hellman` exercise is all about a specific algorithm.

tejasbubane:
I guess not related to CS, but `luhn` is a specific algorithm. Although I am not aware of any similar ones like the `sieve` example above. Maybe rename `luhn` to `credit-card-validator` to remove the specificity?

Insti:
Are there other algorithms that will pass the tests for `diffie-hellman` and `luhn`?
With `sieve` and `binary-search` there are.

Is that the difference between a "good" and a "bad" specfic algorithm question.

ErikSchierboom:
@Insti I think that's it!

petertseng:
Oh. In that case, it's time to suggest `parallel-letter-frequency`... because I imagine most tracks' parallel-letter-frequency test suites don't specifically enforce that the implementation is indeed parallel, so a serial solution (whether it is intentionally or unintentionally serial) could pass the tests, right?

(Or *should* the tests enforce parallelism, and how? Like https://github.com/exercism/xhaskell/pull/542 ?)

Insti:
[`hangman`](https://github.com/exercism/x-common/blob/master/exercises/hangman/description.md)?

Insti:
[binary](https://github.com/exercism/x-common/blob/master/exercises/binary/description.md)? 
(although it is deprecated.)

emilianbold:
The core of the matter for me is not about implementing a specific algorithm. It's that the algorithm does not help me improve for that particular language and that sometimes I spend more time on the algorithm than on the language.

[...]

PR #1128, opened 2018-01-15
added simple-min-heap problem

masters3d:


masters3d:
Will require to be more general

https://github.com/exercism/problem-specifications/issues/761

PR #1130, opened 2018-01-16
 react: Move inputs (cells, operations) to input object

petertseng:
react 1.1.0

As proposed and accepted in #996

```ruby
indent = false
operations = false

ARGF.each_line { |l|
  if l.include?('version')
    ver = l.split(?")[3]
    ver_components = ver.split(?.).map(&:to_i)
    ver_components[1] += 1
    ver_components[2] = 0
    l[ver] = ver_components.join(?.)
  end

  first_non_space = l.index(/\S/)
  if l.include?('"cells"')
    puts ' ' * first_non_space + '"input": {'
    indent = true
  end
  operations ||= l.include?('"operations"')
  l.delete!(?,) if operations && l.strip == '],'
  puts "#{'  ' if indent}#{l}"
  if operations && l.strip == ?]
    puts ' ' * first_non_space + '},'
    operations = false
    indent = false
  end
}
```

---

also note the additional commit that adds `"expected": {}`, since it's required but we expect nothing other than what was already expected in `operations`.

### DIFF
--- a/exercises/react/canonical-data.json
+++ b/exercises/react/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "react",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "comments": [
     "Note that, due to the nature of this exercise,",
     "the tests are specified using their cells and a series of operations to perform on the cells.",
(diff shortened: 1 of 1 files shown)


exercises/react/canonical-data.json
```diff
@@ -46,313 +46,340 @@
...
+        "cells": [
+          {
+            "name": "input",
+            "type": "input",
+            "initial_value": 10
+          }
+        ],
+        "operations": [
+          {
+            "type": "expect_cell_value",
```
petertseng: In its current state, it is not possible to **consistently** place `expect_cell_value` in `expected`, so I do not.

exercises/react/canonical-data.json
```diff
@@ -46,313 +46,340 @@
...
+          {
+            "name": "output",
+            "type": "compute",
+            "inputs": ["times_two", "times_thirty"],
+            "compute_function": "inputs[0] + inputs[1]"
+          }
+        ],
+        "operations": [
+          {
+            "type": "expect_cell_value",
```
petertseng: Note that this `expect_cell_value` comes **before** a `set_value`, therefore it is not possible to **consistently** place `expect_cell_value` in `expected`. Ordering is important. To place `expect_cell_value` in `expected`, we would need to redefine how `expect_cell_value` works so that it may refer to a point in time before/after certain other `operations`.

PR #1133, opened 2018-01-17
rna-transcription: correct DNA to RNA map

### LINKED ISSUE #1078: RNA Transcription: source gives slightly different problem definition
As per [exercism.io/#3705](https://github.com/exercism/exercism.io/issues/3705):

> Looking through the problem statement found on Exercism vs the problem statement found on ROSALIND (which is linked to as the source in the README.md), the problem on Exercism is incorrect.
>
> The transcription would actually be as follows, according to http://rosalind.info/problems/rna/:
>
> * `G` -> `G`
> * `C` -> `C`
> * `T` -> `U`
> * `A` -> `A`
>
> The complements listed on Exercism are as follows:
>
> * `G` -> `C`
> * `C` -> `G`
> * `T` -> `A`
> * `A` -> `U`
>
> My guess is the source might be http://rosalind.info/problems/revc/ instead, which is a related problem which closely matches but still isn't quite the same as the premise here.

I'm not too informed on biology, so I don't know if this exercise has been implemented incorrectly, or if it is just that the source isn't quite right (I expect the second option).

rpottsoh:
Fixes #1078

### DIFF
--- a/exercises/rna-transcription/canonical-data.json
+++ b/exercises/rna-transcription/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "rna-transcription",
-  "version": "1.2.0",
+  "version": "1.3.0",
   "cases": [
     {
       "description": "RNA complement of cytosine is guanine",
@@ -8,7 +8,7 @@
       "input": {
         "dna": "C"
       },
-      "expected": "G"
+      "expected": "C"
     },
     {
       "description": "RNA complement of guanine is cytosine",
@@ -16,7 +16,7 @@
       "input": {
         "dna": "G"
       },
-      "expected": "C"
+      "expected": "G"
     },
     {
       "description": "RNA complement of thymine is adenine",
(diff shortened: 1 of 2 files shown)


rpottsoh:
Next commit will update the test data.

exercises/rna-transcription/canonical-data.json
```diff
@@ -1,46 +1,46 @@
 {
   "exercise": "rna-transcription",
-  "version": "1.2.0",
+  "version": "1.3.0",
   "cases": [
     {
       "description": "RNA complement of cytosine is guanine",
```
ErikSchierboom: Do the descriptions also need to be updated?

rpottsoh: I don't think so at this point.  Give #1078 a read.  I am considering closing this PR in favor of making minor changes to metadata.yml and description.md instead (in another PR).  I am currently researching another source to site in metadata.yml.

PR #1134, opened 2018-01-18
rna-transcription: provide more accurate source info

### LINKED ISSUE #1078: RNA Transcription: source gives slightly different problem definition

rpottsoh:
closes #1078 and is meant to replace #1133.

This exercise is currently designed around the Complement Transcription idea, but the source_url (in metadata.yml) was pointing to a site that describes the T -> U transcription method.

### DIFF
--- a/exercises/rna-transcription/metadata.yml
+++ b/exercises/rna-transcription/metadata.yml
@@ -1,5 +1,5 @@
 ---
 title: "RNA Transcription"
-blurb: "Given a DNA strand, return its RNA complement (per RNA transcription)."
-source: "Rosalind"
-source_url: "http://rosalind.info/problems/rna"
+blurb: "Given a DNA strand, return its RNA Complement Transcription."
+source: "Hyperphysics"
+source_url: "http://hyperphysics.phy-astr.gsu.edu/hbase/Organic/transcription.html"


rpottsoh:
Originally I thought I would also tweak the README too, but for now I think I will leave it alone.  If anyone feels it would benefit the exercise to adjust its wording I'm willing to make changes.

exercises/rna-transcription/metadata.yml
```diff
@@ -1,5 +1,5 @@
 ---
 title: "RNA Transcription"
-blurb: "Given a DNA strand, return its RNA complement (per RNA transcription)."
+blurb: "Given a DNA strand, return its RNA Complement Transcription."
 source: "Rosalind"
```
rpottsoh: Ack!  I need to change `source:`.

PR #1139, opened 2018-01-20
roman-numerals: add test case for input value 49 to test double-normalization

### LINKED ISSUE #1138: roman-numerals: Suggest test case addition
I suggest adding a test case where more than one normalization has to occur on a given number.

For example, 49 would have two normalizations. Starting with "XXXXVIIII", "XXXX" should be "XL" and "VIIII" should be "IX", resulting in a final string of "XLIX".

This will catch solutions that do after-the-fact normalizations but only do one pass.

I can submit a pull request if nobody objects.

https://github.com/exercism/problem-specifications/blob/master/exercises/roman-numerals/canonical-data.json

Zarimax:
closes #1138

adds a test case for input value 49, which contains two substrings that need to be normalized, ("XXXX" should be "XL" and "VIIII" should be "IX".

### DIFF
--- a/exercises/roman-numerals/canonical-data.json
+++ b/exercises/roman-numerals/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "roman-numerals",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "cases": [
     {
       "description": "1 is a single I",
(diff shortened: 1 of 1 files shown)


rpottsoh:
@Zarimax, we'll allow the PR to sit for several days, this should give adequate time for anyone that may want to weigh in to do so, before we merge it.

rpottsoh:
@Zarimax thanks for taking the time and making the effort to improve this exercise!  I am going to go ahead and merge this PR.

PR #1081, opened 2018-01-08
word-count: Apply new "input" policy

ErikSchierboom:
See #996

PR #1143, opened 2018-01-22
word-count: Use camel-case for property name

ErikSchierboom:
As countwords actually consists of two words, its correct name is `countWords`.


ErikSchierboom:
> It amazes me that the trailing newline was removed in #1081 (comment) and then added back.

I did this edit using GitHub's editor, which apparently does not leave newlines untouched. Sorry about that.

PR #1144, opened 2018-01-23
space-age: Update description.md

fedavies:
Clarification of the English in this problem specification.

### DIFF
--- a/exercises/space-age/description.md
+++ b/exercises/space-age/description.md
@@ -1,4 +1,4 @@
- Given an age in seconds, calculate that age in terms of years of the different planets. Here is a list of the length of one 'year' on each planet:
+Given a person's age in seconds, calculate their age in terms of years of the different planets.
 
    - Earth: orbital period 365.25 Earth days, or 31557600 seconds
    - Mercury: orbital period 0.2408467 Earth years
--- a/exercises/space-age/description.md
+++ b/exercises/space-age/description.md
@@ -1,4 +1,4 @@
-Given an age in seconds, calculate how old someone would be on:
+ Given an age in seconds, calculate that age in terms of years of the different planets. Here is a list of the length of one 'year' on each planet:
 
    - Earth: orbital period 365.25 Earth days, or 31557600 seconds
    - Mercury: orbital period 0.2408467 Earth years


exercises/space-age/description.md
```diff
@@ -1,4 +1,4 @@
-Given an age in seconds, calculate how old someone would be on:
+ Given an age in seconds, calculate that age in terms of years of the different planets. Here is a list of the length of one 'year' on each planet:
```
rpottsoh: Just a nit pick, but it appears like there is a leading space that isn't needed.  Also, `"that"` sounds strange to me.  How about `a persons`?  It isn't clear what "that" is referring to.  Is it meant to be the age of the planet, or in the context of the original description, a persons age on a given planet?

wolf99: Both sound a little odd to me!
Maybe a more concise version would be
`Given a person's age in seconds, calculate their age in terms of years of the different planets. ...`

fedavies: I agree that what you have written is clearer

rpottsoh:
@fedavies just a friendly nudge.  Do you need assistance in making the suggested changes?

fedavies:
I might need a hand, thanks. I think that I have done it now but I'm not sure - I'm new to all this!

exercises/space-age/description.md
```diff
@@ -1,4 +1,4 @@
-Given an age in seconds, calculate how old someone would be on:
+Given a person's age in seconds, calculate their age in terms of years of the different planets.
```
Insti: > a person's 

That it's a person is largely irrelevant.

> years of the different planets

This does not make grammatical sense to me.


Disclaimer: I don't think there is anything wrong with the original wording.

wolf99: > That it's a person is largely irrelevant.

True, but supplying an object for the noun makes it easier for students to understand by adding a little more context. Of course _we_ know what it means as-is, but people that don't already know may struggle - as evidenced by the fact that the issue was raised.

>>years of the different planets
>
>This does not make grammatical sense to me.

Maybe this should be "years _on_ the different planets"? Or maybe even "according to the length of a year on a different planet"

kotp:
Thank you @fedavies even if this is not merged in, it shows good effort and we appreciate it.

PR #1146, opened 2018-01-23
clock: apply "input" policy

rpottsoh:
per #996

### DIFF
--- a/exercises/clock/canonical-data.json
+++ b/exercises/clock/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "clock",
-  "version": "2.1.0",
+  "version": "2.2.0",
   "comments": [
     "Most languages require constructing a clock with initial values,",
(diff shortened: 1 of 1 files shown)


exercises/clock/canonical-data.json
```diff
@@ -291,195 +361,225 @@
...
           "property": "equal",
-          "clock1": {
-            "hour": 15,
-            "minute": 37
-          },
-          "clock2": {
-            "hour": 15,
-            "minute": 37
+          "input": {
+            "clock1": {
```
rpottsoh: For these equality cases I was not sure if each clock should have its own "input" object or if the policy only pertains to the test case directly.  I will add input objects for each of the two clocks if the consensus is to do so.

ErikSchierboom: I think the way you've set it up currently is fine.

PR #1147, opened 2018-01-24
clock: remove mention of negative minutes in comment

rpottsoh:
This exercise does not directly test adding or subtracting negative minutes.  I thought it prudent to revise the comment at the top of the testdata to reflect this recent change to the exercise.

### DIFF
--- a/exercises/clock/canonical-data.json
+++ b/exercises/clock/canonical-data.json
@@ -1,9 +1,9 @@
 {
   "exercise": "clock",
-  "version": "2.2.0",
+  "version": "2.2.1",
   "comments": [
     "Most languages require constructing a clock with initial values,",
-    "adding a positive or negative number of minutes, and testing equality",
+    "adding or subtracting some number of minutes, and testing equality",
     "in some language-native way.  Some languages require separate add and",
     "subtract functions.  Negative and out of range values are generally",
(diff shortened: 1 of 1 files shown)


exercises/clock/canonical-data.json
```diff
@@ -1,9 +1,9 @@
 {
   "exercise": "clock",
-  "version": "2.2.0",
+  "version": "2.2.1",
   "comments": [
     "Most languages require constructing a clock with initial values,",
-    "adding a positive or negative number of minutes, and testing equality",
+    "adding or subtracting some number of minutes, and testing equality",
     "in some language-native way.  Some languages require separate add and",
```
rpottsoh: > "Some languages require separate add and subtract functions."

I am not sure that this sentence is really necessary.  Opinions?

ErikSchierboom: I agree that that sentence is not really necessary.

petertseng: in the previous state, it contrasts with the previous sentence's approach of having only `add` but having add support both +ve and -ve. In the current state it doesn't add more information since the previous sentence already has both add and subtract

because of this I think it makes more sense to remove it

rpottsoh: Thanks guys, I'll work on removing it.

PR #1152, opened 2018-01-26
palindrome-products: apply "input" policy

rpottsoh:
per #996

### DIFF
--- a/exercises/palindrome-products/canonical-data.json
+++ b/exercises/palindrome-products/canonical-data.json
@@ -1,11 +1,13 @@
 {
     "exercise": "palindrome-products",
-    "version": "1.0.0",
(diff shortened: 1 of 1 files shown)


rpottsoh:
Yes, I admit I jumped the gun a little on this one.  I agree, in hindsight, that renaming of the properties should have been a separate commit / PR.  I did so many of these types of PRs over the past few days I think my mind went in the direction of testdata's that already had "input" with no variables, and I had to create variables and in that moment equated creating variables or renaming variables as the same thing. :frowning_face: .

Unless there is a strong need for two commits here, or just one that leaves the variable names alone, I will plan on merging this as is.  I'll wait for now.

PR #1163, opened 2018-01-29
robot-simulator: apply "input" policy

rpottsoh:
per #996

PR #1166, opened 2018-01-31
robot-simulator: move "instructions": into "input":

rpottsoh:
This is a follow up to PR #1163.  I had missed moving `"instructions":` into `"input":`.


rpottsoh:
Feel free to merge if you're OK with this.  I'll get it later otherwise...

PR #1168, opened 2018-02-04
crypto-square: apply lowerCamelCase policy

rpottsoh:
per #987

### DIFF
--- a/exercises/crypto-square/canonical-data.json
+++ b/exercises/crypto-square/canonical-data.json
@@ -1,60 +1,60 @@
 {
   "exercise": "crypto-square",
-  "version": "3.2.0",
+  "version": "3.2.1",
   "cases": [
     {
       "description": "empty plaintext results in an empty ciphertext",
-      "property": "ciphertext",
+      "property": "cipherText",
       "input": {
-        "plaintext": ""
+        "plainText": ""
       },
       "expected": ""
     },
     {
       "description": "Lowercase",
-      "property": "ciphertext",
+      "property": "cipherText",
       "input": {
-        "plaintext": "A"
+        "plainText": "A"
       },
       "expected": "a"
     },
     {
       "description": "Remove spaces",
-      "property": "ciphertext",
+      "property": "cipherText",
       "input": {
-        "plaintext": "  b "
+        "plainText": "  b "
       },
       "expected": "b"
     },
     {
       "description": "Remove punctuation",
-      "property": "ciphertext",
+      "property": "cipherText",
       "input": {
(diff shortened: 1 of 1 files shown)


exercises/crypto-square/canonical-data.json
```diff
@@ -1,60 +1,60 @@
 {
   "exercise": "crypto-square",
-  "version": "3.2.0",
+  "version": "3.2.1",
   "cases": [
     {
-      "description": "empty plaintext results in an empty ciphertext",
-      "property": "ciphertext",
+      "description": "empty plainText results in an empty cipherText",
```
ErikSchierboom: I assume you deliberately chose to use the property names, but I think it would maybe be better to use `plain text` and `cipher text` here.

rpottsoh: fixed descriptions to not use property names in text.

rpottsoh:
Not at all @petertseng.  Thanks for feedback.

rpottsoh:
@petertseng you were only referring to their use in the descriptions right?

exercises/crypto-square/canonical-data.json
```diff
@@ -1,60 +1,60 @@
 {
   "exercise": "crypto-square",
-  "version": "3.2.0",
+  "version": "3.2.1",
   "cases": [
     {
       "description": "empty plaintext results in an empty ciphertext",
-      "property": "ciphertext",
+      "property": "cipherText",
```
petertseng: I think ciphertext is a better property than cipherText, so I would revert this change.

exercises/crypto-square/canonical-data.json
```diff
@@ -1,60 +1,60 @@
...
       "input": {
-        "plaintext": ""
+        "plainText": ""
```
petertseng: I think plaintext is a better input key than plainText, so I would revert this change.

exercises/crypto-square/canonical-data.json
```diff
@@ -1,60 +1,60 @@
 {
   "exercise": "crypto-square",
-  "version": "3.2.0",
+  "version": "3.2.1",
   "cases": [
     {
       "description": "empty plaintext results in an empty ciphertext",
```
petertseng: I think this is good. I think "plaintext" and "ciphertext" are better in descriptions than "plain text" and "cipher text"

rpottsoh:
Closing this PR.  Citing https://en.wikipedia.org/wiki/Plaintext.  For this exercise "plaintext" and "ciphertext" are treated as single words.

PR #1170, opened 2018-02-04
two-fer: Remove assumption about track ordering

kytrinyx:
The exercise was originally made to have the same format as the old hello-world exercise:
provide a default value if none is passed, and use the passed argument when present.

This let us simplify hello-world so that it is a simple check to ensure that everything is
wired together correctly.

That said, there are languages where implementing this is non-trivial, so while we did make the
exercise to replace the old hello-world, it doesn't necessarily follow that it would always
directly follow a hello-world exercise.

I tagged you, @sjwarner-bp since you're listed as having touched that file recently ✨

### DIFF
--- a/exercises/two-fer/metadata.yml
+++ b/exercises/two-fer/metadata.yml
@@ -1,4 +1,3 @@
 ---
 blurb: 'Create a sentence of the form "One for X, one for me."'
-source: "This is an exercise to introduce users to basic programming constructs, just after Hello World."
 source_url: "https://en.wikipedia.org/wiki/Two-fer"


rpottsoh:
Why is it nontrivial for some tracks?  I am not opposed to the change, just curious.

kytrinyx:
> Why is it nontrivial for some tracks?

In some languages it's quite difficult to provide a parameter conditionally (e.g. sometimes provide the parameter and sometimes leave it off and use a default).

petertseng:
I agree because https://github.com/exercism/problem-specifications/issues/949#issuecomment-336322163

PR #1177, opened 2018-02-10
all-your-base: improve "input base is one" test

Cohen-Carlisle:
the previous version was not properly testing that input base was validated
passing the "empty list" test would also pass this test

fixed in elixir here: exercism/elixir#389


ErikSchierboom:
Thanks!

PR #1182, opened 2018-02-12
grep: Apply new "input" policy

ErikSchierboom:
See #996.

### DIFF
--- a/exercises/grep/canonical-data.json
+++ b/exercises/grep/canonical-data.json
@@ -60,9 +60,9 @@
           "description": "One file, one match, no flags",
           "property": "grep",
           "input": {
-              "pattern": "Agamemnon",
-              "flags": [],
-              "files": ["iliad.txt"]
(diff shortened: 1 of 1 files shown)


exercises/grep/canonical-data.json
```diff
@@ -59,69 +59,83 @@
         {
           "description": "One file, one match, no flags",
           "property": "grep",
-          "pattern": "Agamemnon",
-          "flags": [],
-          "files": ["iliad.txt"],
+          "input": {
+              "pattern": "Agamemnon",
```
petertseng: I notice that this line is indented four spaces past the previous line, as opposed to other indentations which are two spaces from the previous line. Is there a reason why this should be so?

ErikSchierboom: No reason, I will fix this. Good catch.

ErikSchierboom:
@petertseng I've fixed the indentation.

PR #1183, opened 2018-02-12
zipper: Apply new "input" policy

ErikSchierboom:
See #996.


ErikSchierboom:
@petertseng You are correct. I've re-added the newline.

PR #1195, opened 2018-02-22
go-counting: Updating documentation

### LINKED ISSUE #560: go-counting: Implement canonical-data.json
exercises in sync.

The Go Counting exercise can be found in the [./exercises/go-counting/][exercise-dir] directory.

See http://exercism.io/contribute/canonical-data/go-counting for the up-to-date list of
language tracks that have an implementation of the Go Counting exercise.

[exercise-dir]: https://github.com/exercism/x-common/tree/master/exercises/go-counting

jssander:
Closes #560 

Here is a canonical data file for the go-counting exercise. I have a couple questions:

Why are some parts highlighted in red? Also should I specify that the expected territory should be a set and not an array?

### DIFF
--- a/exercises/go-counting/canonical-data.json
+++ b/exercises/go-counting/canonical-data.json
@@ -1,0 +1,200 @@
+{
+  "exercise": "go-counting",
+  "version": "1.0.0",
+  "comments": [
+    "Territory consists of [x, y] coordinate pairs."
+  ],
+  "cases": [
+    {
+      "description": "Black corner territory on 5x5 board",
+      "property": "territory",
+      "input": {
+        "board": [
+          "  B  ",
+          " B B ",
+          "B W B",
+          " W W ",
+          "  W  "
+        ],
+        "x": 0,
+        "y": 1
+      },
+      "expected": {
+        "owner": "BLACK",
+        "territory": [[0, 0], [0, 1], [1, 0]]
+      }
+    },
+    {
+      "description": "White center territory on 5x5 board",
+      "property": "territory",
+      "input": {
+        "board": [
+          "  B  ",
+          " B B ",
+          "B W B",
+          " W W ",
+          "  W  "
+        ],
+        "x": 2,
+        "y": 3
+      },
+      "expected": {
+        "owner": "WHITE",
+        "territory": [[2, 3]]
+      }
+    },
+    {
+      "description": "Open corner territory on 5x5 board",
+      "property": "territory",
+      "input": {
+        "board": [
+          "  B  ",
+          " B B ",
+          "B W B",
+          " W W ",
+          "  W  "
+        ],
+        "x": 1,
+        "y": 4
+      },
+      "expected": {
+        "owner": "NONE",
+        "territory": [[0, 3], [0, 4], [1, 4]]
+      }
+    },
+    {
+      "description": "A stone and not a territory on 5x5 board",
+      "property": "territory",
+      "input": {
+        "board": [
+          "  B  ",
+          " B B ",
+          "B W B",
+          " W W ",
+          "  W  "
+        ],
+        "x": 1,
+        "y": 1
+      },
+      "expected": {
+        "owner": "NONE",
+        "territory": []
+      }
+    },
+    {
+      "description": "Invalid because X is too low for 5x5 board",
+      "property": "territory",
+      "input": {
+        "board": [
+          "  B  ",
+          " B B ",
+          "B W B",
+          " W W ",
+          "  W  "
+        ],
+        "x": -1,
+        "y": 1
+      },
+      "expected": {
+        "error": "Invalid coordinate"
+      }
+    },
+    {
+      "description": "Invalid because X is too high for 5x5 board",
+      "property": "territory",
+      "input": {
+        "board": [
+          "  B  ",
+          " B B ",
+          "B W B",
+          " W W ",
+          "  W  "
+        ],
+        "x": 5,
+        "y": 1
+      },
+      "expected": {
+        "error": "Invalid coordinate"
+      }
+    },
+    {
+      "description": "Invalid because Y is too low for 5x5 board",
+      "property": "territory",
+      "input": {
+        "board": [
+          "  B  ",
+          " B B ",
+          "B W B",
+          " W W ",
+          "  W  "
+        ],
+        "x": 1,
+        "y": -1
+      },
+      "expected": {
+        "error": "Invalid coordinate"
+      }
+    },
+    {
+      "description": "Invalid because Y is too high for 5x5 board",
+      "property": "territory",
+      "input": {
+        "board": [
+          "  B  ",
+          " B B ",
+          "B W B",
+          " W W ",
+          "  W  "
+        ],
+        "x": 1,
+        "y": 5
+      },
+      "expected": {
+        "error": "Invalid coordinate"
+      }
+    },
+    {
+      "description": "One territory is the whole board",
+      "property": "territories",
+      "input": {
+        "board": [
+          " "
+        ]
+      },
+      "expected": {
+        "territoryBlack": [],
+        "territoryWhite": [],
+        "territoryNone": [[0, 0]]
+      }
+    },
+    {
+      "description": "Two territory rectangular board",
+      "property": "territories",
+      "input": {
+        "board": [
+          " BW ",
+          " BW "
+        ]
+      },
+      "expected": {
+        "territoryBlack": [[0, 0], [0, 1]],
+        "territoryWhite": [[3, 0], [3, 1]],
+        "territoryNone": []
+      }
+    },
+    {
+      "description": "Two region rectangular board",
+      "property": "territories",
+      "input": {
+        "board": [
+          " B "
+        ]
+      },
+      "expected": {
+        "territoryBlack": [[0, 0], [2, 0]],
+        "territoryWhite": [],
+        "territoryNone": []
+      }
+    }
+  ]
+}
--- a/exercises/go-counting/description.md
+++ b/exercises/go-counting/description.md
@@ -9,6 +9,8 @@ Write a function that determines the territory of each player. You may
 assume that any stones that have been stranded in enemy territory have
 already been taken off the board.
 
+Write a function that determines the territory which includes a specified coordinate.
+
 Multiple empty intersections may be encircled at once and for encircling
 only horizontal and vertical neighbours count. In the following diagram
 the stones which matter are marked "O" and the stones that don't are


exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+		    " B B ",
+		    "B W B",
+		    " W W ",
+		    "  W  "
+		],
+        "x": 0,
+        "y": 1
+      },
+      "expected": {
+        "stone": BLACK
```
cmccandless: A couple notes on syntax:

- Values must be one of: string (with quotes), number, array (`[...]`), object (`{ ... }`)
- Values must be followed by a comma unless it is the last value in the current array/object

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+        "territory": [(0, 0), (0, 1), (1, 0)]
```
cmccandless: Tuples are not supported by JSON.

cmccandless:
@jssander 
> Also should I specify that the expected territory should be a set and not an array?

The use of sets is an implementation detail, so arrays are better for the canonical-data.

exercises/go-counting/canonical-data.json
rpottsoh: `BLACK` is probably hi-lighted because it should probably be in `"`'s and should have a trailing `,` on the line

exercises/go-counting/canonical-data.json
rpottsoh: I think this wants to be:
```JSON
"territory": [[0,0], [0,1], [1,0]]
```
Are you trying to portray an array of tuples?  I am not sure if it is really possible to convey a tuple in JSON.

petertseng:
Please edit the PR description and amend the commit message to use one of the words in https://help.github.com/articles/closing-issues-using-keywords/ and `https://github.com/exercism/problem-specifications/issues/560` (https://github.com/exercism/problem-specifications/issues/560) so that it will automatically close that issue when this PR is merged. We wouldn't want us to forget to do that.

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+		    " B B ",
+		    "B W B",
+		    " W W ",
+		    "  W  "
+		],
+        "x": 2,
+        "y": 3
+      },
+      "expected": {
+        "stone": WHITE
```
rpottsoh:
```JSON
"stone": "WHITE",
```

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+        "territory": [(2, 3)]
```
rpottsoh:
```JSON
"territory": [[2, 3]]
```

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+		    " B B ",
+		    "B W B",
+		    " W W ",
+		    "  W  "
+		],
+        "x": 1,
+        "y": 4
+      },
+      "expected": {
+        "stone": NONE
```
rpottsoh:
```JSON
"stone": "NONE",
```

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+        "territory": [(0, 3), (0, 4), (1, 4)]
```
rpottsoh:
```JSON
"territory": [[0, 3], [0, 4], [1,4]]
```

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+		    " B B ",
+		    "B W B",
+		    " W W ",
+		    "  W  "
+		],
+        "x": 1,
+        "y": 1
+      },
+      "expected": {
+        "stone": NONE
```
rpottsoh:
```JSON
"stone": "NONE",

File: exercises/go-counting/canonical-data.json [hunk 8]
```diff
@@ -0,0 +1,172 @@
...
+        "territory": []
```
rpottsoh:
```JSON
"territory": {}
```

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+		    " B B ",
+		    "B W B",
+		    " W W ",
+		    "  W  "
+		],
+        "x": -1,
+        "y": 1
+      },
+      "expected": {
+        "stone": NONE
```
rpottsoh:
```JSON
"stone": "NONE",
```

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+		    " B B ",
+		    "B W B",
+		    " W W ",
+		    "  W  "
+		],
+        "x": 1,
+        "y": 5
+      },
+      "expected": {
+        "stone": NONE
```
rpottsoh:
```JSON
"stone": "NONE",
```

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+    {
+      "description": "Test when one territory is the whole board",
+      "property": "territories",
+      "input": {
+        "board": [
+		    " "
+		],
+      },
+      "expected": {
+        "TerritoryBlack": []
```
rpottsoh: should have a `,` at EOL.

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+        "TerritoryWhite": []
```
rpottsoh: should have a `,` at EOL.

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+	"TerritoryNone": [(0, 0)]
```
rpottsoh:
```json
"territoryNone": [[0, 0]]
```
Note use of lowerCamelCase.  Be careful of the indentation too.

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+        "board": [
+		    " BW ",
+		    " BW "
+		],
+      },
+      "expected": {
+        "TerritoryBlack": [(0, 0), (0, 1)]
+        "TerritoryWhite": [(3, 0), (3, 1)]
+	"TerritoryNone": []
+      }
```
rpottsoh:
```json
"expected": {
  "territoryBlack": [[0, 0], [0, 1]],
  "territoryWhite": [[3, 0], [3, 1]],
  "territoryNone": []
}
```

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+		    "B B   B B",
+		    " W BBB W ",
+		    "   B B   "
+		],
+        "x": 0,
+        "y": 8
+      },
+      "expected": {
+        "stone": NONE
+        "territory": [(2, 7), (2, 8), (1, 8), (0, 8), (0, 7)]
```
rpottsoh:
```json
"stone": "NONE",
"territory": [[2, 7], [2, 8], [1, 8], [0, 8], [0, 7]]

**Comment by rpottsoh** [comment]
problem [description](https://github.com/exercism/problem-specifications/blob/master/exercises/go-counting/description.md).

**Comment by jssander** [comment]
Thanks for all the comments, I'll make all the changes. Sorry for all the problems.

**Comment by rpottsoh** [comment]
@jssander this looks like a good start, thanks for taking on the challenge and deciding to help out.

I have created a [gist](https://jsonschemalint.com/#/version/draft-04/markup/json?gist=423ab28d50dac3edd253ad18b36a7b3c) of your canonical data file.  The left-hand pane is the JSON schema used by Exercism to confirm that canonical data files are formatted properly.  In the right-hand pane is your submitted JSON.  Feel free to paste your JSON into this window as you go to check it against the schema.  The same check is also made when you commit.

**Comment by rpottsoh** [comment]
No problems. 


Sent from my android device.

File: exercises/go-counting/canonical-data.json (hunk 2 above)
Comment (petertseng) [inline exercises/go-counting/canonical-data.json]: Here is a suggestion I have. What if these were to be `{"x": whatever, "y": whatever}` instead?

For queen-attack that was suggested in https://github.com/exercism/problem-specifications/issues/721 and implemented in https://github.com/exercism/problem-specifications/pull/935
For robot-simulator that was suggested in https://github.com/exercism/problem-specifications/issues/722 and implemented in https://github.com/exercism/problem-specifications/pull/936

Does the same reasoning apply here? Why or why not?

**Comment by petertseng** [comment]
I plan to verify with code similar to the following.

```ruby
require 'json'

cases = by_property(json['cases'], %w(territory territories))

class UnionFind
  def initialize(things)
    @parent = things.map { |x| [x, x] }.to_h
    @rank = things.map { |x| [x, 0] }.to_h
    @frozen = false
  end

  def keys
    @parent.keys
  end

  def union(x, y)
    raise FrozenError if @frozen
    xp = find(x)
    yp = find(y)

    return if xp == yp

    if @rank.fetch(xp) < @rank.fetch(yp)
      @parent[xp] = yp
    elsif @rank.fetch(xp) > @rank.fetch(yp)
      @parent[yp] = xp
    else
      @parent[yp] = xp
      @rank[xp] += 1
    end
  end

  def find(x)
    @parent[x] = find(@parent.fetch(x)) if @parent.fetch(x) != x
    @parent.fetch(x)
  end

  def freeze
    @frozen = true
    self
  end
end

module EachGridPair refine Array do
  def each_grid_pair
    each_with_index { |row, y|
      row.each_char.with_index { |c, x|
        [[y + 1, x], [y, x + 1]].each { |y2, x2|
          next unless (c2 = self[y2]&.[](x2))
          yield c, [y, x], c2, [y2, x2]
        }
      }
    }
  end
end end

using EachGridPair

def territories(board)
  uf = UnionFind.new(board.each_with_index.flat_map { |row, y| row.size.times.map { |x| [y, x] } })
  board.each_grid_pair { |c1, place1, c2, place2|
    uf.union(place1, place2) if c1 == ' ' && c2 == ' '
  }
  uf.freeze
  regions = uf.keys.group_by { |k| uf.find(k) }
  colours = regions.keys.map { |k| [k, [false, false]] }.to_h
  board.each_grid_pair { |c1, place1, c2, place2|
    next if (c1 == ' ') == (c2 == ' ')
    place = c1 == ' ' ? place1 : place2
    colour = [c1, c2].sort.last == ?B ? 0 : 1
    colours.fetch(uf.find(place))[colour] = true
  }
  colours.transform_values! { |black, white|
    (black && !white ? 'BLACK' : white && !black ? 'WHITE' : nil).freeze
  }
  regions.map { |parent, v|
    y, x = parent
    [parent, {
      # TODO: call this colour instead?
      'stone' => colours.fetch(parent),
      'territory' => v,
    }] if board[y][x] == ' '
  }.compact.to_h.freeze
end

verify(cases['territory'], property: 'territory') { |i, _|
  territories(i['board']).values.find { |t|
    t['territory'].include?([i[?y], i[?x]])
  }
}

verify(cases['territories'], property: 'territories') { |i, _|
  # TODO: Type subject to discussion of multiple territories of a given colour.
  territories(i['board']).values
}
```

If for some reason I should become unavailable, feel free to verify in my stead. I would highly suggest that verification be performed before merging.

jssander:
I think I made all the changes mentioned, but it still doesn't seem to be proper JSON.

exercises/go-counting/canonical-data.json
```diff
@@ -125,9 +125,9 @@
 		],
       },
       "expected": {
-        "TerritoryBlack": []
-        "TerritoryWhite": []
-	"TerritoryNone": [(0, 0)]
+        "territoryBlack": [],
+        "territoryWhite": [],
+	"territoryNone": [[0, 0]]
```
rpottsoh: Watch the indents on the `"territoryNone":`'s.  they should be the same as the two above.

jssander: Is this why it is not valid JSON? Or is there some other problem?

rpottsoh:
It appears as though you have a mix of tabs and spaces to for indenting.  It would be easier to read through the file if the tabs were replaced with a sufficient number of spaces.  Tabs mean different things in different editors, it isn't always the same number of spaces.

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+        "territory": []
+      }
+    },
+    {
+      "description": "Test when one territory is the whole board",
+      "property": "territories",
+      "input": {
+        "board": [
+		    " "
+		],
```
rpottsoh: `,` not necessary.
Try www.jsonlint.com.  It will help point out where the syntax is wrong.

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+      }
+    },
+    {
+      "description": "Test for two territory rectangular board",
+      "property": "territories",
+      "input": {
+        "board": [
+		    " BW ",
+		    " BW "
+		],
```
rpottsoh: `,` no needed.

jssander:
Thanks for all the help and patience. It passes the checks now.

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+                    "B W B",
+                    " W W ",
+                    "  W  "
+                 ],
+        "x": 0,
+        "y": 1
+      },
+      "expected": {
+        "stone": "BLACK",
+        "territory": [[0, 0], [0, 1], [1, 0]]
```
petertseng: Here is a suggestion I have. What if these were to be {"x": whatever, "y": whatever} instead?

For queen-attack that was suggested in #721 and implemented in #935
For robot-simulator that was suggested in #722 and implemented in #936

Does the same reasoning apply here? Why or why not?

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+                    " B B ",
+                    "B W B",
+                    " W W ",
+                    "  W  "
+                 ],
+        "x": 0,
+        "y": 1
+      },
+      "expected": {
+        "stone": "BLACK",
```
petertseng: I don't think `stone` makes sense, because by definition there is no stone in a territory. This is meant to represent the owner of the territory, so perhaps `player`, `colour`, or `owner`?

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
+{
+  "exercise": "go-counting",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "Test for black territory on 5x5 board",
+      "property": "territory",
+      "input": {
+        "board": [
```
petertseng: it would probably look better if this line's indentation lined up with the above line. Same with all `board` line.

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+                    "  B  ",
```
petertseng: it would probably look better if this line's indentation was indented exactly two spaces beyond the start of the `board` line. Same with all other lines making up a board.

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+    {
+      "description": "Test when one territory is the whole board",
+      "property": "territories",
+      "input": {
+        "board": [
+                    " "
+                ]
+      },
+      "expected": {
+        "territoryBlack": [],
```
petertseng: I notice this says `territory`, singular. What if a board has more than one black territory? How should that be handled?

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
+{
+  "exercise": "go-counting",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "Test for black territory on 5x5 board",
```
petertseng: I suggest to remove the words `test for` or `test to` from the descriptions because it is already the case that everything in this file is a test; the descriptio ndoes not need to reaffirm it. Like in https://github.com/exercism/problem-specifications/pull/771.

petertseng: Let's have descriptions say **why** a test case is present, rather than **what** it is testing. For example, some new functionality that this case tests that previous cases did not test. For example, "territory surrounded entirely by stones" versus "territory bordering the edge of the board", etc.

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+        "x": 1,
+        "y": 4
+      },
+      "expected": {
+        "stone": "NONE",
+        "territory": [[0, 3], [0, 4], [1, 4]]
+      }
+    },
+    {
+      "description": "Test for none territory on 5x5 board",
```
petertseng: "none territory" doesn't seem to be sufficiently distinctive description from the above". This is testing that a stone is not considered part of a territory; the description should say so

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+        "x": 1,
+        "y": 1
+      },
+      "expected": {
+        "stone": "NONE",
+        "territory": []
+      }
+    },
+    {
+      "description": "Test for valid coordinate on 5x5 board",
```
petertseng: the coordinate is not valid, even though the description claims it is

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+                    "B W B",
+                    " W W ",
+                    "  W  "
+                ],
+        "x": -1,
+        "y": 1
+      },
+      "expected": {
+        "stone": "NONE",
+        "territory": []
```
petertseng: I would think invalid coordinates should cause an `"error": "some error message"` instead of just having a none `stone` and empty `territory` array to signal that. It makes it more distinguishable from valid cases. Any particular arguments for keeping `stone` and `territory` in invalid cases?

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+      }
+    },
+    {
+      "description": "Test for valid coordinate #2 on 5x5 board",
```
petertseng: > the coordinate is not valid, even though the description claims it is

same comment here

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+                    "B W B",
+                    " W W ",
+                    "  W  "
+                ],
+        "x": 1,
+        "y": 5
+      },
+      "expected": {
+        "stone": "NONE",
+        "territory": []
```
petertseng: > I would think invalid coordinates should cause an `"error": "some error message"` instead of just having a none `stone` and empty `territory` array to signal that. It makes it more distinguishable from valid cases. Any particular arguments for keeping `stone` and `territory` in invalid cases?

same comment here

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+                    "B W B",
+                    " W W ",
+                    "  W  "
+                ],
+        "x": 1,
+        "y": 1
+      },
+      "expected": {
+        "stone": "NONE",
+        "territory": []
```
petertseng: I'm wondering whether noting a stone as `"stone": "NONE"` and `"territory": []`  is sufficiently distinctive from a territory with no owner. I was considering whether this should be considered an `error` instead. But I do see that `"territory": []` is distinguishable because anything that's not a stone will have at least one element in here.

So, would anyone like to argue either for or against having this be an `error`?

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+      "description": "Test when one territory is the whole board",
+      "property": "territories",
+      "input": {
+        "board": [
+                    " "
+                ]
+      },
+      "expected": {
+        "territoryBlack": [],
+        "territoryWhite": [],
```
petertseng: > I notice this says `territory`, singular. What if a board has more than one black territory? How should that be handled?

Same, but for white

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+        "territoryNone": [[0, 0]]
```
petertseng: > I notice this says `territory`, singular. What if a board has more than one black territory? How should that be handled?

Same, but for no owner

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+      "input": {
+        "board": [
+                    " BW ",
+                    " BW "
+                ]
+      },
+      "expected": {
+        "territoryBlack": [[0, 0], [0, 1]],
+        "territoryWhite": [[3, 0], [3, 1]],
+        "territoryNone": []
```
petertseng: > I notice this says `territory`, singular. What if a board has more than one black territory? How should that be handled?

Same, but for no owner

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+      "property": "territories",
+      "input": {
+        "board": [
+                    " BW ",
+                    " BW "
+                ]
+      },
+      "expected": {
+        "territoryBlack": [[0, 0], [0, 1]],
+        "territoryWhite": [[3, 0], [3, 1]],
```
petertseng: > I notice this says `territory`, singular. What if a board has more than one black territory? How should that be handled?

Same, but for white

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,172 @@
...
+      "description": "Test for two territory rectangular board",
+      "property": "territories",
+      "input": {
+        "board": [
+                    " BW ",
+                    " BW "
+                ]
+      },
+      "expected": {
+        "territoryBlack": [[0, 0], [0, 1]],
```
petertseng: > I notice this says `territory`, singular. What if a board has more than one black territory? How should that be handled?

Same

exercises/go-counting/canonical-data.json
jssander: Well I use object notation for the input where there is just one x coordinate and one y coordinate. However, there can be many x and ys in the expected output, so array notation (which was also suggested in #721 and #722 seems cleaner, shorter, and simpler.

exercises/go-counting/canonical-data.json
jssander: Well it would still all be considered black's territory, but I'll handle it however you want me to. I can change it to territoriesBlack if you like.

exercises/go-counting/canonical-data.json
jssander: I understand what you're saying and it makes sense. I did it this way though because I think it is how it is currently implemented in the Python and C# exercises.

exercises/go-counting/canonical-data.json
petertseng: > it would still all be considered black's territory

true story - then the question is whether they should all go in the same array. For the following board:

```json
"board": [
  " B "
]
```

Should the answer for black be a single array `[[0, 0], [0, 2]]`, or should it be two, one for each contiguous region?

```json
[
 [
    [0, 0]
 ],
 [
    [0, 2]
 ]
]
```

I think I had misunderstood the concept of territory. I thought that each contiguous region is a single territory. This understanding stems from the fact that `territory` designates a single intersection on the board and only examines the single region that contains that intersection. But it could be instead that each player has a single territory that may comprise multiple regions.

If I didn't misunderstand, it would want to be `territories` for each player (and each contiguous region would be in its own array), OR it would just be an array whose elements are the same type as the `expected` value of each `territory`.

If someone would confirm to me that I misunderstood, then it would make sense to keep it as `territory`.

petertseng: EBWODP: Previous comment did not sufficiently make clear the distinction between `territory`/`territories` the properties being tested versus the abstract concept in the Go game. I hope this edit makes it clearer:

I think I had misunderstood the concept of territory. I thought that each contiguous region is a single territory. This understanding stems from the fact that `territory` (the property being tested earlier in this file) designates a single intersection on the board and only examines the single region that contains that intersection. But it could be instead that each player has a single territory that may comprise multiple regions.

If I didn't misunderstand, then the keys in the expected object of `territories` would each want to be named `territories` for each player (and each contiguous region would be in its own array), OR the `expected` object of a `territories` case would just be an array whose elements are the same type as the `expected` value of a `territory` case.

If someone would confirm to me that I misunderstood, then for the expected answer to `territories`, it would make sense to keep the key names as `territory` for each player .

exercises/go-counting/canonical-data.json
petertseng: Then I think it should be changed here.

petertseng: Given that the argument to make it more distinguishable makes sense, then I think it should be changed here.

jssander: If it is changed here what will happen to the current implementations of the exercise?

exercises/go-counting/canonical-data.json
petertseng: If so, then let's define in a `"comments": ["The coordinates are ........."]` at the top level of the file which coordinate is which, otherwise it is not clear.

(Feel free to look at https://github.com/exercism/problem-specifications/blob/master/exercises/anagram/canonical-data.json#L4-L9 for example of syntax for the comment)

For me, my opinion will be that the order should be `[y, x]` because that matches the order in which one would index into the board (y selects which string in the array, x selects which character in the string)

exercises/go-counting/canonical-data.json
petertseng: They can be updated, if the maintainers of the respective tracks deem a change suitable for the track in question.

petertseng:
I looked at https://github.com/exercism/problem-specifications/blob/master/exercises/go-counting/description.md which says "Write a function that determines the territory of each player." . Given that, I think `territories` is the only functionality being called for. If we must keep `territory` tests, I think the description should have a change.

jssander:
I've made some, but perhaps not all the changes. I added a line to the description for the territory function. I haven't put in the comment saying which coordinate goes first (x or y). Would you like me to do that and change it so it goes [y, x] instead of [x, y]?

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,170 @@
+{
+  "exercise": "go-counting",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "Black corner territory on 5x5 board",
+      "property": "territory",
+      "input": {
+      "board": [
```
petertseng: I realise that I gave bad advice. Since this is a k/v pair under `input`, I would indent the `board` line and all the strings making up the board two spaces in

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,170 @@
...
+        "  B  ",
+        " B B ",
+        "B W B",
+        " W W ",
+        "  W  "
+                 ],
```
petertseng: I would align this with the same column that `"board"` is; same with all other closing bracket lines

```
"input": {
  "board": [
    "stuff"
  ],
  "x": 0,
}

**Comment by petertseng** [comment]
> I haven't put in the comment saying which coordinate goes first (x or y). Would you like me to do that 

Whichever of `[y, x]` or `[x, y]` you choose, I think it will be necessary to put a comment in, yes. So that it is clear to readers of this file.

> and change it so it goes [y, x] instead of [x, y]?

I think last time I argued that `[y, x]` matches how a computer would index into the board, right? On the other hand, I think I also recall from education that `[x, y]` more closely matches the conventional way humans choose to designate a coordinate in 2d space? It is not a perfect analogy though since I think in such a conventional coordinate system `y` is increasing as you go up, but in this file `y` is increasing as you go down.

Unless someone comes forward with more arguments as to whether it is desirable to perfer one or the other, I don't object to either. Maybe someone wishes to retrieve some arguments from https://github.com/exercism/problem-specifications/issues/313 .

**Comment by jssander** [comment]
I left it as [x, y], but put in a comment about it. I also changed the indentation.

File: exercises/go-counting/canonical-data.json [hunk 38]
```diff
@@ -0,0 +1,173 @@
...
+        "x": 1,
+        "y": 1
+      },
+      "expected": {
+        "owner": "NONE",
+        "territory": []
+      }
+    },
+    {
+      "description": "Invalid coordinate on 5x5 board",
```
petertseng: Would it be better to explain in what way the coordinate is invalid? Here, it is because X is too low.

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,173 @@
...
+        ],
+        "x": -1,
+        "y": 1
+      },
+      "expected": {
+        "error": "Invalid coordinate"
+      }
+    },
+    {
+      "description": "Invalid coordinate #2 on 5x5 board",
```
petertseng: Would it be better to explain in what way the coordinate is invalid, so that it is clearer than "#2"? Here, it is because Y is too high.

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,173 @@
...
+      "expected": {
+        "error": "Invalid coordinate"
+      }
+    },
+    {
+      "description": "One territory is the whole board",
+      "property": "territories",
+      "input": {
+        "board": [
+                    " "
```
petertseng: let's re-indent to match the other `board`s

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,173 @@
...
+        ]
+      },
+      "expected": {
+        "territoryBlack": [[0, 0], [0, 1]],
+        "territoryWhite": [[3, 0], [3, 1]],
+        "territoryNone": []
+      }
+    },
+    {
+      "description": "Open corner territory on 9x9 board",
```
petertseng: I wasn't sure what this case tests that (for `territory`) that any previous cases don't. I know the board is larger, but I don't think that adds any extra considerations to the algorithm. May I ask if any insight can be gleaned from any of the four implementing tracks (http://exercism.io/contribute/canonical-data/go-counting C# F# Haskell Python) why they have this test, what it does for them?

If there is nothing new being tested, I would actually think it should be removed.

jssander:
Ok, I remove the 9x9 board test, fixed an indentation, made the two invalid coordinate test case descriptions clearer, and added a test case for the " B " board. I haven't done anything about the "territory" term problem yet.

exercises/go-counting/canonical-data.json
```diff
@@ -147,26 +147,17 @@
...
-        ],
-        "x": 0,
-        "y": 8
+          " B "
+        ]
       },
       "expected": {
-        "owner": "NONE",
-        "territory": [[2, 7], [2, 8], [1, 8], [0, 8], [0, 7]]
+        "territoryBlack": [[0, 0], [0, 2]],
```
petertseng: inconsistent: This uses `[y, x]` coordinate instead of `[x, y]`, since a y coordinate of 2 does not exist (the x coordinate should be 2).

petertseng:
EBWODP: 

> Use an array whose element type is the same as the element type of `territory`

Should instead be: Use an array whose element type is the same as the expected answer to `territory`

petertseng:
I prepared code to verify any of the three possible expected types, so just replace the last four lines of the verify script and specify the output type we use.

The current type in the submitted JSON is `by_owner_combined`. (As I pointed out in my earlier review, the currently submitted JSON does not verify, so I can't Approve yet)

```ruby
verify(cases['territories'], property: 'territories') { |i, _|
  default = {
    'territoryBlack' => [],
    'territoryWhite' => [],
    'territoryNone' => [],
  }

  case which_output_type_should_we_use = :by_owner_combined
  when :array_of_territory
    territories(i['board'])
  when :by_owner
    default.merge(territories(i['board']).group_by { |t| t['owner'] }.map { |k, vs|
      ['territory' + k.capitalize, vs.map { |t| t['territory'] }.sort.freeze]
    }.to_h)
  when :by_owner_combined
    default.merge(territories(i['board']).group_by { |t| t['owner'] }.map { |k, vs|
      ['territory' + k.capitalize, vs.flat_map { |t| t['territory'] }.sort.freeze]
    }.to_h)
  end
}
```

petertseng:
EBWODP: Clarification:

> so just replace the last four lines of the verify script and specify the output type we use.

This is not a **command** that some reader of this text **should** verify anything. This is to say: For anyone who **wants** to verify (willingly choosing to do), to do that, you can follow those instructions.

jssander:
Ok I fixed the x/y coordinate pair order for one of the tests and added tests for x and y being too high or too low.

rpottsoh:
Sorry for bringing it up so late.  The change to description.md should probably be in its own commit and / or the title of the PR should be updated as it is no longer just implementing canonical test data for this exercise.

exercises/go-counting/canonical-data.json
```diff
@@ -100,7 +100,43 @@
...
+        ],
+        "x": 1,
+        "y": -1
+      },
+      "expected": {
+        "error": "Invalid coordinate"
+      }
+    },
+    {
+      "description": "Invalid because Y is too high for 5x5 board",
```
rpottsoh: This looks good, thanks for making this change.

jssander:
I updated the title of the PR. Do you still want me to put the changes to description.md in a different commit? Can I keep it in the same PR?

rpottsoh:
I think it is ok as you have things now.  I wouldn't worry about it unless any one else has a different opinion.   Thanks for updating the PR title.

petertseng:
This file now verifies. This means that my Request for Changes is hereby dismissed.

Reminder: In a little over 3 days, it will have been one week since I submitted https://github.com/exercism/problem-specifications/pull/1195#pullrequestreview-99863372 which asks for opinions on which expected type to use for `territories`. Reviewers have until that time to express an opinion. If no opinion is expressed, I will make the choice of "arbitrary i.e. whatever the author of the JSON file chooses" which means I will Approve this PR in its current state.

petertseng:
I hope you will not consider it presumptuous that I dismiss https://github.com/exercism/problem-specifications/pull/1195#pullrequestreview-98720462. It should be easy to see why I am doing so: The review mentions invalid JSON syntax, and the syntax is valid today.

petertseng:
Great job, thank you.

jssander:
You're welcome. Thank you for merging.

exercises/go-counting/canonical-data.json
```diff
@@ -0,0 +1,200 @@
...
+        "x": 1,
+        "y": 4
+      },
+      "expected": {
+        "owner": "NONE",
+        "territory": [[0, 3], [0, 4], [1, 4]]
+      }
+    },
+    {
+      "description": "A stone and not a territory on 5x5 board",
```
cmccandless: @jssander @petertseng I know this PR has already been merged, but should this be "A stone is not a territory..."?

petertseng: If you want. I admit I don't see a reason to prefer one or the other, but that could just be because of English not being my first language. I'm happy to defer to others on this matter.

PR #1196, opened 2018-02-23
react: Fold expect_callback_values into set_value

petertseng:
react 2.0.0

expect_callback_values is often not precise enough about which
`set_value` calls should result in which callbacks being called or not
called. "callbacks can be added and removed" is the worst offender,
expecting a `[32, 42]` from callback2 only after both `set_value` have
fired.

That means for all we know, maybe an implementation:

* magically manages to predict the future and calls the callback twice
  on the first `set_value`, with the correct value, and then zero times
  on the second `set_value`.
* calls the callback zero times on the first `set_value`, but twice
  on the second `set_value`.

To remedy this, we observe that the only time callbacks might be called
is if there was a `set_value`. So, let's explicitly associate our
callback expectations along with each `set_value`.

### DIFF
--- a/exercises/react/canonical-data.json
+++ b/exercises/react/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "react",
-  "version": "1.3.0",
+  "version": "2.0.0",
   "comments": [
     "Note that, due to the nature of this exercise,",
     "the tests are specified using their cells and a series of operations to perform on the cells.",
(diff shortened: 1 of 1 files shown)


petertseng:
The necessary diff to the verify script to verify this should remain the most recent commit of https://github.com/petertseng/exercism-problem-specifications/commits/verify-react . As of the current moment, that would be https://github.com/petertseng/exercism-problem-specifications/commit/6a8b82d98a1de8cb5c1dd11a1d051347e0d607a6, but obviously I can't correct that stays correct forever, since I may need to rebase.

petertseng:
aside: This makes it abundantly clear that there is something that has never been tested: Ensuring that a `set_value` that changes two output cells should in fact cause callbacks on both to be triggered. Separate PR for that.

petertseng:
If #1197 is merged before #1196, the version number progression will be 1.2.0 -> 1.3.0 -> 2.0.0.
If the opposite happens, and #1196 is merged before #1197, the version number progression will be 1.2.0 -> 2.0.0 -> 2.1.0.

It does not matter to me which path is taken. In the absence of other feedback, I will arrange so that the final version is 2.0.0, but if there is a large disparity in time of Approval (in the GitHub sense) of the two PRs, I will of course take the appropriate path dictated by those Approvals.

ErikSchierboom:
@petertseng Great improvement. Is this ready to be merged?

petertseng:
For those wondering if any test suites will actually need to change: Mostly not; the two ways of expressing callback expectations are isomorphic. However, my hope is that this way makes the intent clearer to implementors of this exercise.

However, I did sneak in one functional change, and that is to "callbacks can be added and removed", as the PR description states. For those looking to go from 1.3.0 to 2.0.0, look at what changes in that function (explicit expectations after each `set_value`, instead of only after the second).

It was not the wisest to combine the two operations into one commit, so I apologise for that.

PR #1197, opened 2018-02-23
react: Test callbacks on multiple cells

petertseng:
react 1.3.0

Consider an implementation that stops looking for cells to fire
callbacks on after it has found any single cell that has callbacks. That
would fail this test.

### DIFF
--- a/exercises/react/canonical-data.json
+++ b/exercises/react/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "react",
-  "version": "1.2.0",
+  "version": "1.3.0",
   "comments": [
     "Note that, due to the nature of this exercise,",
     "the tests are specified using their cells and a series of operations to perform on the cells.",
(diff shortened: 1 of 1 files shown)


petertseng:
If this is merged before #1196, the version number progression will be 1.2.0 -> 1.3.0 -> 2.0.0. If this is merged after #1197, the version number progression will be 1.2.0 -> 2.0.0 -> 2.1.0.

It does not matter to me which path is taken. In the absence of other feedback, I will arrange so that the final version is 2.0.0, but if there is a large disparity in time of Approval (in the GitHub sense) of the two PRs, I will of course take the appropriate path dictated by those Approvals.

ErikSchierboom:
@petertseng Is this ready to be merged?

petertseng:
> Is this ready to be merged?

Yes.

petertseng:
Thank you for reviewing!

PR #1207, opened 2018-03-10
yacht: create problem specification for scoring a single hand of yacht.

zeimusu:
Create the three files that give a short description of the problem: a metadata file, a description file and a canonical test data file.

An implementation for the go track to come.

### DIFF
--- a/exercises/yacht/canonical-data.json
+++ b/exercises/yacht/canonical-data.json
@@ -203,6 +203,15 @@
             "expected":0
         },
         {
+            "description":"Minimum is 1, maximum is 5, but not a little straight",
+            "property":"score",
+            "input":{
+                "dice":[1, 1, 3, 4, 5],
+                "category":"little straight"
+            },
+            "expected":0
+        },
+        {
             "description":"Big Straight",
             "property":"score",
             "input":{
--- a/exercises/yacht/canonical-data.json
+++ b/exercises/yacht/canonical-data.json
@@ -1,261 +1,260 @@
-{
-    "exercise" : "yacht",
-    "version" : "1.0.0",
-    "comments" : [
+{  
+    "exercise":"yacht",
+    "version":"1.0.0",
+    "comments":[  
         "The dice are represented always as a list of exactly five integers",
-        "with values beteen 1 and 6 inclusive. The category is an integer.",
-        "the numbers 0 to 5 represent the categories 'ones' to 'sixes'",
-        "Then 6 is full house,",
-        "     7 is four of a kind",
-        "     8 is a little straight 1-5",
-        "     9 is b big straight 2-6",
-        "    10 is Choise, sometimes called Chance",
-        "    11 is Yacht, or five of a kind"
-    ],
-    "cases": [
-	{
-	
-        "description": "Yacht",
-        "property":    "score",
-        "input":       {
-		    "dice":    [5, 5, 5, 5, 5],
-		    "category":11
+        "with values beteen 1 and 6 inclusive. The category is an string.",
+        "the categories are 'ones' to 'sixes' then",
+        "Then 'full house',",
+        "     'four of a kind'",
+        "     'little straight' 1-5",
+        "     'big straight' 2-6",
+        "     'choice', sometimes called Chance",
+        "     'yacht', or five of a kind"
+   ],
+    "cases":[  
+        {  
+            "description":"Yacht",
+            "property":"score",
+            "input":{  
+                "dice":[5, 5, 5, 5, 5],
+                "category":"yacht"
+            },
+            "expected":50
         },
-		"expected":    50
-	},
-	{
-		"description": "Not Yacht",
-        "property":    "score",
-        "input":       {
-		    "dice":    [1, 3, 3, 2, 5],
-		    "category":11
+        {  
+            "description":"Not Yacht",
+            "property":"score",
+            "input":{  
+                "dice":[1, 3, 3, 2, 5],
+                "category":"yacht"
+            },
+            "expected":0
         },
-		"expected":    0
-	},
-	{
-		"description": "Ones",
-        "property":    "score",
-        "input":       {
-		    "dice":    [1, 1, 1, 3, 5],
-		    "category":0
+        {  
+            "description":"Ones",
+            "property":"score",
+            "input":{  
+                "dice":[1, 1, 1, 3, 5],
+                "category":"ones"
+            },
+            "expected":3
         },
-		"expected":    3
-	},
-	{
-		"description": "Ones, out of order",
-        "property":    "score",
-        "input":       {
-		    "dice":    [3, 1, 1, 5, 1],
-		    "category":0
+        {  
+            "description":"Ones, out of order",
+            "property":"score",
+            "input":{  
+                "dice":[3, 1, 1, 5, 1],
+                "category":"ones"
+            },
+            "expected":3
         },
-		"expected":    3
-	},
-	{
-		"description": "No ones",
-        "property":    "score",
-        "input":       {
-		    "dice":    [4, 3, 6, 5, 5],
-		    "category":0
+        {  
+            "description":"No ones",
+            "property":"score",
+            "input":{  
+                "dice":[4, 3, 6, 5, 5],
+                "category":"ones"
+            },
+            "expected":0
         },
-		"expected":    0
-	},
-	{
-		"description": "Fours",
-        "property":    "score",
-        "input":       {
-		    "dice":    [1, 4, 1, 4, 1],
-		    "category":3
+        {  
+            "description":"Fours",
+            "property":"score",
+            "input":{  
+                "dice":[1, 4, 1, 4, 1],
+                "category":"fours"
+            },
+            "expected":8
         },
-		"expected":    8
-	},
-	{
-		"description": "Yacht counted as threes",
-        "property":    "score",
-        "input":       {
-		    "dice":    [3, 3, 3, 3, 3],
-		    "category":2
+        {  
+            "description":"Yacht counted as threes",
+            "property":"score",
+            "input":{  
+                "dice":[3, 3, 3, 3, 3],
+                "category":"threes"
+            },
+            "expected":15
         },
-		"expected":    15
-	},
-	{
-		"description": "Yacht of 3s counted as fives",
-        "property":    "score",
-        "input":       {
-		    "dice":    [3, 3, 3, 3, 3],
-		    "category":4
+        {  
+            "description":"Yacht of 3s counted as fives",
+            "property":"score",
+            "input":{  
+                "dice":[3, 3, 3, 3, 3],
+                "category":"fives"
+            },
+            "expected":0
         },
-		"expected":    0
-	},
-	{
-		"description": "Full house",
-        "property":    "score",
-        "input":       {
-		    "dice":    [2, 2, 4, 4, 4],
-		    "category":6
+        {  
+            "description":"Full house",
+            "property":"score",
+            "input":{  
+                "dice":[2, 2, 4, 4, 4],
+                "category":"full house"
+            },
+            "expected":16
         },
-		"expected":    16
-	},
-	{
-		"description": "Full house unsorted",
-        "property":    "score",
-        "input":       {
-		    "dice":    [5, 3, 3, 5, 3],
-		    "category":6
+        {  
+            "description":"Full house unsorted",
+            "property":"score",
+            "input":{  
+                "dice":[5, 3, 3, 5, 3],
+                "category":"full house"
+            },
+            "expected":19
         },
-		"expected":    19
-	},
-	{
-		"description": "Not a full house",
-        "property":    "score",
-        "input":       {
-		    "dice":    [2, 2, 4, 4, 5],
-		    "category":6
+        {  
+            "description":"Not a full house",
+            "property":"score",
+            "input":{  
+                "dice":[2, 2, 4, 4, 5],
+                "category":"full house"
+            },
+            "expected":0
         },
-		"expected":    0
-	},
-	{
-		"description": "Yacht is not a full house",
-        "property":    "score",
-        "input":       {
-		    "dice":    [2, 2, 2, 2, 2],
-		    "category":6
+        {  
+            "description":"Yacht is not a full house",
+            "property":"score",
+            "input":{  
+                "dice":[2, 2, 2, 2, 2],
+                "category":"full house"
+            },
+            "expected":0
         },
-		"expected":    0
-	},
-	{
-		"description": "Four of a Kind",
-        "property":    "score",
-        "input":       {
-		    "dice":    [1, 2, 2, 2, 2],
-		    "category":7
+        {  
+            "description":"Four of a Kind",
+            "property":"score",
+            "input":{  
+                "dice":[1, 2, 2, 2, 2],
+                "category":"four of a kind"
+            },
+            "expected":8
         },
-		"expected":    8
-	},
-	{
-		"description": "Four of a Kind unsorted",
-        "property":    "score",
-        "input":       {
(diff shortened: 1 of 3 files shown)


exercises/yacht/canonical-data.json
```diff
@@ -0,0 +1,261 @@
...
+    "exercise" : "yacht",
+    "version" : "1.0.0",
+    "comments" : [
+        "The dice are represented always as a list of exactly five integers",
+        "with values beteen 1 and 6 inclusive. The category is an integer.",
+        "the numbers 0 to 5 represent the categories 'ones' to 'sixes'",
+        "Then 6 is full house,",
+        "     7 is four of a kind",
+        "     8 is a little straight 1-5",
+        "     9 is b big straight 2-6",
```
petertseng: I don't think `b big straight` is correct here. I think it should be `a bit straight`, unless the `b` has a special significance I missed.

exercises/yacht/canonical-data.json
```diff
@@ -0,0 +1,261 @@
...
+        "    10 is Choise, sometimes called Chance",
```
petertseng: I don't think Choise is a correct spelling. I think it should be Choice, unless there is a game-specific spelling I am not aware of (in that case, simply make me aware of it)

exercises/yacht/canonical-data.json
```diff
@@ -0,0 +1,261 @@
+{
+    "exercise" : "yacht",
+    "version" : "1.0.0",
+    "comments" : [
+        "The dice are represented always as a list of exactly five integers",
+        "with values beteen 1 and 6 inclusive. The category is an integer.",
```
petertseng: > The category is an integer

I found that this makes the JSON file unnecessarily hard to read. A human reader of this file will have to constantly refer to the top of this file to see what category a given number refers to. A machine reader will have the mapping built in, but a human reader must first read this file to build the mapping.

If this JSON file instead used strings such as `"Ones"` or `"LittleStraight"` for categories, the advantage is that it would be more readable for humans. Any potential disadvantages we should be aware of?

exercises/yacht/canonical-data.json
```diff
@@ -0,0 +1,261 @@
...
+    "cases": [
+	{
+	
+        "description": "Yacht",
+        "property":    "score",
+        "input":       {
+		    "dice":    [5, 5, 5, 5, 5],
+		    "category":11
+        },
+		"expected":    50
```
petertseng: I have a hard time reading this `expected` line since it is not indented to the same column as other keys on its level (`description`, `property`, `input`). I would request that `expected`'s indentation be changed to be the same column, for the purpose of readability.

I have a similar hard time reading further cases below in this file where `description` and `expected` are indented to some column but `property` and `input` are indented to some other column, and I make the same request (change them so that they indent to the same column).

exercises/yacht/canonical-data.json
```diff
@@ -0,0 +1,261 @@
...
+        "    10 is Choise, sometimes called Chance",
+        "    11 is Yacht, or five of a kind"
+    ],
+    "cases": [
+	{
+	
+        "description": "Yacht",
+        "property":    "score",
+        "input":       {
+		    "dice":    [5, 5, 5, 5, 5],
```
petertseng: I have a hard time reading this since the column distance between levels is inconsistent. The distance between the opening brace on line 16 and `"description"` on line 18 is only one space, much smaller than the distance between `"input"` and `"dice"`. I would request that the distance between levels be made consistent, for the purpose of readability.

Same request applies to all other cases.

exercises/yacht/canonical-data.json
petertseng: EBWODP:

> a bit straight

I was wrong as well.

It should be `a big straight`

zeimusu:
Changes requested by Peter Tseng made.

exercises/yacht/canonical-data.json
```diff
@@ -0,0 +1,260 @@
...
+                "category":"big straight"
+            },
+            "expected":30
+        },
+        {  
+            "description":"Big Straight as little straight",
+            "property":"score",
+            "input":{  
+                "dice":[6, 5, 4, 3, 2],
+                "category":"big straight"
```
petertseng: little straight (given the test description is "Big Straight as little straight")

exercises/yacht/canonical-data.json
```diff
@@ -0,0 +1,260 @@
+{  
+    "exercise":"yacht",
+    "version":"1.0.0",
+    "comments":[  
+        "The dice are represented always as a list of exactly five integers",
+        "with values beteen 1 and 6 inclusive. The category is an string.",
```
petertseng: beteen -> between

exercises/yacht/canonical-data.json
```diff
@@ -0,0 +1,260 @@
...
+        "the categories are 'ones' to 'sixes' then",
```
petertseng: the word "then" is repeated on the end of this line and the beginning of the next. One of the two should be removed.

exercises/yacht/canonical-data.json
```diff
@@ -0,0 +1,260 @@
+{
```
petertseng: it would be preferable if all trailing spaces at the end of a line would be removed, since they don't change the meaning of the file and just take up space

zeimusu:
Changes requested by Peter Tseng done:
Spelling and doubled words removed. 
Coverage of the twos and sixes.
Correction of big straight to little straight.
Whitespace removed at end of lines.

rpottsoh:
@zeimusu this is looking very good.  I am wondering if the testdata really needs to have sorted and unsorted tests.  It seems to me that the unsorted tests should be enough.  This is only an opinion, I am not saying that the sorted cases should be removed.  I am just looking for yours and others opinions on the idea before jumping to making any more changes.

exercises/yacht/description.md
```diff
@@ -0,0 +1,4 @@
+Score a single throw of dice in *Yacht*.
+
+See the description of [Yacht](https://en.wikipedia.org/wiki/Yacht_(dice_game))  on Wikipedia for the scoring system.
```
rpottsoh: Minor nit pick.  There is an extra space between the url and the word `on` or perhaps reword:
```json
A description of the Yacht scoring system, used by this exercise, can be found at [Wikipedia](https://en.wikipedia.org/wiki/Yacht_(dice_game)).
```
Alternativly you could incorporate the scoring system into the description.md file.  This file becomes a part of the README that each student receives when they fetch this exercise.  It would be more convenient if the student didn't have to follow a trail of bread crumbs to get to the documentation.  I think this would also serve to protect the exercise from any changes that might crop up on the Wikipedia page that would not be immediately detected at Exercism.

zeimusu: The sorted/unsorted tests were added on the basis of "make early tests easy to pass". It is perhaps unnecessary to have sorted/unsorted tests throughout and some can be trimmed.  It is a good idea to copy the scoring system into the description.md file, and I'll do that.

zeimusu:
I've completely rewritten the description, It now contains a full description of the scoring mechanics and an extended description of the task. I've removed some of the sorted/unsorted tests also some of the fairly redundant tests of the choice category. I've left in the sorted ones/unsorted ones near the start and there are two tests for full house (for the cases xxxyy and xxyyy)  

I've also got an implementation of this exercise for the go track ready, when this is merged.

exercises/yacht/description.md
```diff
@@ -1,4 +1,34 @@
...
+    Fives           5 × number of fives     5 1 5 2 5 scores 15
+    Sixes           6 × number of sixes     2 3 4 5 6 scores 6
+    Full House      Total of the dice       3 3 3 5 5 scores 25
+    Four of a Kind  Total of the four dice  4 4 4 4 6 scores 16
+    Little Straight 30 points               1 2 3 4 5 scores 30 
+    Big Straight    30 points               2 3 4 5 6 scores 30
+    Choice          Sum of the dice         2 3 3 4 6 scores 18
+    Yacht           50 points               4 4 4 4 4 scores 50
+
+If the dice to not satisfy the requirements of a category, the score is zero.
```
rpottsoh: `to` should be `do`.

exercises/yacht/description.md
```diff
@@ -1,4 +1,34 @@
...
+If, for example, *Four Of A Kind* is entered in the *Yacht* category, zero
+points are scored. A *Yacht* scores zero if entered in the *Full House* category.
+
+## Task 
+Given a list of values of the dice and a category you should return the
+score of the dice in that category. If the dice don't satify the requirements
+of the category you should return 0. You can assume that you will receive the
+values of five dice, and the value of each will be between one and six. You
+should not assume that the dice are ordered.
```
rpottsoh: Given a list of values for five dice and a category, your solution should return
the score of the dice for that category. If the dice do not satisfy the requirements
of the category your solution should return 0. You can assume that five values
will always be presented, and the value of each will be between one and six
inclusively. You should not assume that the dice are ordered.

exercises/yacht/canonical-data.json
```diff
@@ -113,7 +113,7 @@
             "expected":16
         },
         {
-            "description":"Full house unsorted",
+            "description":"Full house three small, two big, unsorted",
```
rpottsoh: I don't think  you need to include `unsorted` in the description.

exercises/yacht/canonical-data.json
```diff
@@ -139,15 +139,6 @@
...
-            "description":"Four of a Kind",
-            "property":"score",
-            "input":{
-                "dice":[1, 2, 2, 2, 2],
-                "category":"four of a kind"
-            },
-            "expected":8
-        },
         {
             "description":"Four of a Kind unsorted",
```
rpottsoh: I don't think you need to include the word `unsorted`.

exercises/yacht/canonical-data.json
```diff
@@ -175,15 +166,6 @@
...
-            "description":"Little Straight",
-            "property":"score",
-            "input":{
-                "dice":[1, 2, 3, 4, 5],
-                "category":"little straight"
-            },
-            "expected":30
-        },
         {
             "description":"Little Straight unsorted",
```
rpottsoh: drop the word `unsorted`.

exercises/yacht/canonical-data.json
```diff
@@ -211,15 +193,6 @@
...
-            "description":"Big Straight",
-            "property":"score",
-            "input":{
-                "dice":[2, 3, 4, 5, 6],
-                "category":"big straight"
-            },
-            "expected":30
-        },
         {
             "description":"Big Straight unsorted",
```
rpottsoh: don't need `unsorted`.

exercises/yacht/canonical-data.json
```diff
@@ -0,0 +1,233 @@
...
+            "description":"Yacht can be scored as Four of a Kind",
+            "property":"score",
+            "input":{
+                "dice":[3, 3, 3, 3, 3],
+                "category":"four of a kind"
+            },
+            "expected":12
+        },
+        {
+            "description":"Not Four of a Kind",
```
rpottsoh: Should follow the same naming scheme you used on `Two pair is not a full house`
```
"description": "Full House is not Four of a Kind",
```

exercises/yacht/canonical-data.json
```diff
@@ -0,0 +1,233 @@
...
+            "description":"Little Straight as Big Straight",
+            "property":"score",
+            "input":{
+                "dice":[1, 2, 3, 4, 5],
+                "category":"big straight"
+            },
+            "expected":0
+        },
+        {
+            "description":"Not a little straight",
```
rpottsoh:
```json
"description": "Maybe Choice but not a Little Straight",
```

exercises/yacht/canonical-data.json
```diff
@@ -0,0 +1,233 @@
...
+            "property":"score",
+            "input":{
+                "dice":[1, 2, 3, 4, 6],
+                "category":"little straight"
+            },
+            "expected":0
+        },
+        {
+            "description":"Big Straight unsorted",
```
rpottsoh: drop `unsorted`.

rpottsoh:
@zeimusu thanks for your perseverance.  New exercise ideas get scrutinized pretty thoroughly and they kind of have to, these files are the basis for almost all the exercises offered by all the language tracks at Exercism.  I think _Yacht_ will end up making a great addition and as @petertseng has already mentioned, it will be a good counterpart to `Poker`.  Hopefully a few others will review what you have accomplished so far and will offer up some further thoughts.  In my opinion I think this exercise is close to merging.

zeimusu:
The spelling corrections have been done.
The word "unsorted" has been removed.
The phrasing of the description of the tests that return zero has been made more consistent.
I've not used your suggestion "Maybe Choice but not a Little Straight" (since everything is a "choice") but used the phrasing "No pairs but not a little straight" This captures better what is distinctive about this test.

exercises/yacht/description.md
```diff
@@ -0,0 +1,34 @@
...
+## Scores in Yacht
+
+    Category    Score                   Example
+    Ones            1 × number of ones      1 1 1 4 5 scores 3
+    Twos            2 × number of twos      2 2 3 4 5 scores 4
+    Threes          3 × number of threes    3 3 3 3 3 scores 15
+    Fours           4 × number of fours     1 2 3 3 5 scores 0
+    Fives           5 × number of fives     5 1 5 2 5 scores 15
+    Sixes           6 × number of sixes     2 3 4 5 6 scores 6
+    Full House      Total of the dice       3 3 3 5 5 scores 25
```
petertseng: Should this be 19? Or is the scoring method to be changed?

petertseng:
I notice the "No pairs but not a little straight" is interesting since it looks for an particular mistake that a student might make when implementing. So, I will also suggest the opposite mistake: `1 1 3 4 5`

zeimusu:
I've fixed the silly error, (my thinking that 3+3+3 = 15)
On Peter's suggestion, I've added a test for 1 1 2 3 4, as "Four in order but not a little straight" I think that address a potential student error (Yahtzee has a "short straight" which is 4 dice in order)

exercises/yacht/canonical-data.json
```diff
@@ -0,0 +1,242 @@
...
+                "dice":[1, 2, 3, 4, 5],
+                "category":"big straight"
+            },
+            "expected":0
+        },
+        {
+            "description":"Four in order but not a little straight",
+            "property":"score",
+            "input":{
+                "dice":[1, 1, 2, 3, 4],
```
petertseng: 1, 1, 2, 3, 4 is an interesting case (so I think it should be kept?), but I had a different case in mind: 1, 1, 2, 3, 5. An implementation may just ask the question: is the max five and the min 1? but that implementation would be mistaken here.

zeimusu: I see what you're getting at.  I've added that as well.

rpottsoh:
Congratulations @zeimusu, thanks for your time and efforts.  Squashing and merging your PR.

petertseng:
The commit message of https://github.com/exercism/problem-specifications/commit/ea4689df394281182eb2615a0b2c327ea59a59bf is quite long. I would prefer future squashes take care to only include relevant content in the commit messages.

rpottsoh:
@petertseng understood...  Commit message can be modified...  A no no here I think though.  If not I will modify the commit message.

petertseng:
I don't think this is important enough to rebase master. I do not recommend modifying the commit message after the fact, since then everyone who has acquired the old commit will have to be instructed to discard it and acquire the new commit. I recommend modifying the commit message before merging, in the future.

masahx:
don't know where I can write this, but you don't have test for FIVES category, when you have 5-s. There is when 5s are absent only, so I don't have to check for fives.

PR #1221, opened 2018-04-09
isbn-verifier: adding test case for 0 < length < 10

### LINKED ISSUE #1216: isbn-verifier: lacks a test case for 0 < length < 10
https://github.com/exercism/problem-specifications/blob/master/exercises/isbn-verifier/canonical-data.json has a case for 0 length, but no case for a nonzero length that is less than 10.

Without such a case, incorrect implementations such as those in https://github.com/petertseng/exercism-problem-specifications/blob/verify/exercises/isbn-verifier/verify.rb#L59-L74 will pass the tests, even though they should not.

Task:

* Craft an input(s) with 9 digits that those incorrect implementations would incorrectly declare to be a valid ISBN. This may be one input that covers both implementations, or one for each.
* Add those input(s) to the JSON file.
* Increase second version component of the JSON file. If current version is `x . y . z`, make the new version `x . y+1 . 0`
* Use a word in https://help.github.com/articles/closing-issues-using-keywords/ in your commit message to indicate that you have closed this issue.

mikestratton:
fixes #1216

### DIFF
--- a/exercises/isbn-verifier/canonical-data.json
+++ b/exercises/isbn-verifier/canonical-data.json
@@ -1,6 +1,6 @@
 {
     "exercise": "isbn-verifier",
-    "version": "2.3.0",
+    "version": "2.4.0",
     "comments": [
         "An expected value of true indicates a valid ISBN-10, ",
         "whereas false means the ISBN-10 is invalid."
@@ -115,6 +115,14 @@
             "property": "isValid",
             "input": {
               "isbn": ""
+            },
+            "expected": false
+        },
+        {
+            "description": "input is 9 characters",
+            "property": "isValid",
+            "input": {
+                "isbn": "134456729"
             },
             "expected": false
         }


mikestratton:
209/11 = 19

petertseng:
My mistake. Let me see why the implementation isn't rejecting the case.

petertseng:
(I had written `apppend` which was obviously incorrect, therefore my implementation rejected but only because the method name didn't exist)

petertseng:
EBWODP: Unnecessarily overloading of the word "accept". I'll change this to: I'd like to **merge** this.

For a second case that the other implementation will incorrectly accept, I'm happy to simply open another issue and decrease the scope of the existing issue.

petertseng:
When merging, I plan to add a few details from #1216 into the commit message. It is better to have them in the commit message than make someone have to open GitHub to see what #1216 is.

petertseng:
We have two approvals, and the linked issue has been open long enough that if there were any objections, people have had enough time to raise them in the issue. So I think we should merge it, without needing to wait longer.

I have a path forward for the remaining issue I want to tackle (just make a new issue about the other implementation), so I'm just going with that plan.

Thanks!

exercises/isbn-verifier/canonical-data.json
```diff
@@ -117,6 +117,14 @@
               "isbn": ""
             },
             "expected": false
+        },
+        {
+            "description": "input is 9 characters",
+            "property": "isValid",
+            "input": {
+                "isbn": "134456729"
```
petertseng: indentation interesting, but since we are not enforcing it automatically, my only argument is "it should be the same so that it is consistent, which will maybe make it easier to read"

PR #966, opened 2017-10-21
simple-cipher: Add canonical data

aos:
Closes #586 

Hello. I've been working on this one. There are a few places where I wasn't sure what to do, specifically L27, and others where you have to do multiple encode into decodes. Or when adding a new cipher key. Let me know what you think!

PR #1224, opened 2018-04-11
diffie-hellman: Add canonical data

petertseng:
Resources:

* https://github.com/exercism/problem-specifications/blob/master/exercises/diffie-hellman/description.md
* https://en.wikipedia.org/wiki/Diffie%E2%80%93Hellman_key_exchange
* http://cryptopals.com/sets/5

Implementing tracks (10):

* https://github.com/exercism/csharp/blob/master/exercises/diffie-hellman/DiffieHellmanTest.cs
* https://github.com/exercism/ecmascript/blob/master/exercises/diffie-hellman/diffie-hellman.spec.js
* https://github.com/exercism/elixir/blob/master/exercises/diffie-hellman/diffie_hellman_test.exs
* https://github.com/exercism/fsharp/blob/master/exercises/diffie-hellman/DiffieHellmanTest.fs
* https://github.com/exercism/go/blob/master/exercises/diffie-hellman/diffie_hellman_test.go
* https://github.com/exercism/java/blob/master/exercises/diffie-hellman/src/test/java/DiffieHellmanTest.java
* https://github.com/exercism/javascript/blob/master/exercises/diffie-hellman/diffie-hellman.spec.js
* https://github.com/exercism/python/blob/master/exercises/diffie-hellman/diffie_hellman_test.py
* https://github.com/exercism/rust/blob/master/exercises/diffie-hellman/tests/diffie-hellman.rs
* https://github.com/exercism/typescript/blob/master/exercises/diffie-hellman/diffie-hellman.test.ts

In the below discussion:

* "The ECMAScript family" is used to refer to the three tracks:
  ECMAScript, JavaScript, TypeScript.
* MITM means man-in-the-middle
* DH means Diffie-Hellman

All implementing tracks have the same tests, those shown here, with the
following differences:

**Inconsequential changes** based on different values selected:

* Rust used different values for the secret key calculation.
* All three members of the ECMAScript family used only fixed values for
  private keys in the key exchange, rather than random values.

Fundamentally, these changes make no difference, as the same operations
are being tested either way. It is, however, instructive to show the
entire process of a key exchange, so the test is kept.

All three members of the ECMAScript family **require implementations to
perform extra validation**:

* values of `p` are sensible (> 2, prime)
* (though they do not test `g`)
* private keys provided to the public key calculation are < `p`.

These tests are **not** included, for the following reasons:

* The README states that `p` is to be provided by the test suite. Thus,
  validating `p` is not inherently in scope for this exercise. This is
  despite the fact that validating `p` is important when actually doing
  DH! What happens when parameters are not validated:
  http://cryptopals.com/sets/5/challenges/35
* Generating a private key is part of the exercise, and an earlier test
  has already ensured that private keys are in range. Because private
  keys are kept private and not received from others, a MITM cannot
  cause either party to suddenly use a bogus value for a private key.
  In contrast, a MITM might attack DH by tricking the parties into using
  bogus public keys:
  http://cryptopals.com/sets/5/challenges/34
* Allows the core operations to declare Garbage In Garbage Out and not
  concern themselves with validation, whereas the parameters should be
  validated at the points where they may be received from outside.

On **testing large numbers**:

* C#, F#, Java, and Python test that secret calculation works.
* Elixir and Go test that all operations (private key generation, public
  key calculation, secret calculation) work.
* Rust and the ECMAScript family do not test large numbers at all.

These are not included because past discussion indicates large numbers
hold no additional value, but may impose implementation hardships in
some languages:
https://github.com/exercism/problem-specifications/pull/944#issuecomment-336730635

Closes https://github.com/exercism/problem-specifications/issues/557


exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,67 @@
...
+    "  this is a reasonable real-world use for them",
+    "  and it may be instructive to have an exercise on their use.",
+    "  Consult tracks with this exercise (such as the Go track) for possible inputs to use."
+  ],
+  "cases": [
+    {
+      "description": "private key is in range 1 .. p",
+      "property": "privateKeyIsInRange",
+      "input": {},
+      "expected": {}
```
petertseng: I'm not here to tell individual tracks how they test this, so I have no value for `expected` to put here. Though I guess I can throw out the following alternative proposals:

* `"expected": {}` (what I currently have)
* `"expected": "1 < privateKey < p` (expressing it as a string)
* `"expected": {"greaterThan": 1, "lessThan": "p"} (expressing it in a more structured way)
* Your suggestion here.

exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,67 @@
...
+    "  Although the calculations fundamentally do not require large numbers,",
+    "  this is a reasonable real-world use for them",
+    "  and it may be instructive to have an exercise on their use.",
+    "  Consult tracks with this exercise (such as the Go track) for possible inputs to use."
+  ],
+  "cases": [
+    {
+      "description": "private key is in range 1 .. p",
+      "property": "privateKeyIsInRange",
+      "input": {},
```
petertseng: I mean, I know `p` is a necessary input to generating a private key, but I figure languages can determine for themselves what `p` they wish to test with. If you really care, I guess I can say:

`"input": {"p": "figure out a value(s) for p that makes sense based on how your track will test this"}`

exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,67 @@
...
+      "expected": {}
+    },
+    {
+      "description": "private key is random",
+      "property": "privateKeyIsRandom",
+      "input": {},
+      "expected": {}
```
petertseng: I'm definitely not here to tell individual tracks how they test this.

There is no precedent. Of the exercises that test randomness:

* `robot-name` also has no JSON file yet.
* `simple-cipher`'s https://github.com/exercism/problem-specifications/pull/966 received no suggestions.
* `diffie-hellman`'s https://github.com/exercism/problem-specifications/pull/944#discussion_r147567932 received no suggestions.

Therefore, I decided what it should look like.

exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,67 @@
...
+    {
+      "description": "private key is in range 1 .. p",
+      "property": "privateKeyIsInRange",
+      "input": {},
+      "expected": {}
+    },
+    {
+      "description": "private key is random",
+      "property": "privateKeyIsRandom",
+      "input": {},
```
petertseng: same comment about `input`. `p` is a necessary input to generating a private key, but I want to let tracks determine what value(s) they should test with.

exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,67 @@
...
+        "p": 23,
+        "g": 5,
+        "alicePrivateKey": "privateKey(p)",
+        "bobPrivateKey": "privateKey(p)",
+        "alicePublicKey": "publicKey(p, g, alicePrivateKey)",
+        "bobPublicKey": "publicKey(p, g, bobPrivateKey)",
+        "secretA": "secret(p, bobPublicKey, alicePrivateKey)",
+        "secretB": "secret(p, alicePublicKey, bobPrivateKey)"
+      },
+      "expected": "secretA == secretB"
```
petertseng: A departure from the usual "your calculated value should equal the pre-calculated value known to the test". This is "these two values that you calculate should be equal". 

The only other instance of this I remember is https://github.com/exercism/problem-specifications/blob/master/exercises/zipper/canonical-data.json; see the sameResultFromOperations test and possibly the related discussion around https://github.com/exercism/problem-specifications/pull/934#issuecomment-335202217 and https://github.com/exercism/problem-specifications/pull/934#issuecomment-335952991

exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,67 @@
...
+      },
+      "expected": 2
+    },
+    {
+      "description": "key exchange",
+      "property": "keyExchange",
+      "input": {
+        "p": 23,
+        "g": 5,
+        "alicePrivateKey": "privateKey(p)",
```
petertseng: it doesn't seem worth it to come up with a mini-DSL that expresses these in a more structured way.

petertseng:
Nothing to verify automatically here, really, but sure you can if you want...

```ruby
require 'json'

cases = by_property(json['cases'], %w(privateKeyIsInRange privateKeyIsRandom publicKey secret keyExchange))

verify(cases['publicKey'], property: 'publicKey') { |i, _|
  i[?g] ** i['privateKey'] % i[?p]
}

verify(cases['secret'], property: 'secret') { |i, _|
  i['theirPublicKey'] ** i['myPrivateKey'] % i[?p]
}
```

exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,67 @@
...
+        "privateKey": 6
+      },
+      "expected": 8
+    },
+    {
+      "description": "can calculate secret using other party's public key",
+      "property": "secret",
+      "input": {
+        "p": 23,
+        "theirPublicKey": 19,
```
petertseng: it will not be possible to convince me that I should just say `publicKey` and `privateKey` here because I want to make it clear that the public key and private key belong to different people. However, you could convince me of something like `alicePublicKey` and `bobPrivateKey`.

I considered `receiverPublicKey` and `senderPrivateKey` but I thought both parties may play both roles, so I discarded that

exercises/diffie-hellman/canonical-data.json
```diff
@@ -0,0 +1,67 @@
...
+    "* Large numbers:",
+    "  Although the calculations fundamentally do not require large numbers,",
+    "  this is a reasonable real-world use for them",
+    "  and it may be instructive to have an exercise on their use.",
+    "  Consult tracks with this exercise (such as the Go track) for possible inputs to use."
+  ],
+  "cases": [
+    {
+      "description": "private key is in range 1 .. p",
+      "property": "privateKeyIsInRange",
```
cmccandless: Typically `property` is used as a function name... perhaps this should simply be `"privateKey"`?

exercises/diffie-hellman/canonical-data.json
cmccandless: > - `"expected": {"greaterThan": 1, "lessThan": "p"}`

Could we indicate that this test calls `property` multiple times, verifying all returned values are in range?

petertseng: > > - `"expected": {"greaterThan": 1, "lessThan": "p"}`
>
> Could we indicate that this test calls `property` multiple times, verifying all returned values are in range?

While I don't want to force tracks to implement their tests in any particular way, I can be convinced that it is universally wise (across all tracks) to generate multiple private keys, since even an incorrect implementation may, by chance, generate a single correct private key and therefore not be found to be incorrect by a test that generated only once.

Conclusion: I'm happy to use any means suggested to indicate this, since there is no precedent.

exercises/diffie-hellman/canonical-data.json
petertseng: > Typically `property` is used as a function name... perhaps this should simply be `"privateKey"`?

In the following discussion, I will use the https://github.com/exercism/problem-specifications/blob/master/README.md 's definition of `property`, which to define the type of test. Further, I acknowledge that the README agrees with what you say: Typically, it is the function name.

Here, I think it made sense to use the broader definition of simply the type of test. Specifically, I say that means this: If two tests have the same `property`, then necessarily their inputs and outputs should be interpreted in the same way.

Let's brainstorm some suggestions here. Should it be the case that *both* `privateKeyIsRandom` and `privateKeyIsInRange` be changed to just `privateKey`?

If so, I think we will have to rely heavily on `expected` to express the difference between the two. This is, of course, very possible. I'm happy to go with anything suggested, since there is no precedent.

coriolinus:
>  If two tests have the same property, then necessarily their inputs and
outputs should be interpreted in the same way.

If the above statement is ever false, it will break the Rust track's
exercise generator.

exercises/diffie-hellman/canonical-data.json
ErikSchierboom: I don't think there is any good way to model an expected random outcome.

exercises/diffie-hellman/canonical-data.json
ErikSchierboom: Agreed.

exercises/diffie-hellman/canonical-data.json
ErikSchierboom: Well, it is of course due to the randomness of the input that you cannot use fixed output values, so I think this will do fine.

exercises/diffie-hellman/canonical-data.json
ErikSchierboom: I would be in favor of being as explicit as possible, meaning I prefer `"expected": {"greaterThan": 1, "lessThan": "p"}`.

exercises/diffie-hellman/canonical-data.json
ErikSchierboom: > Let's brainstorm some suggestions here. Should it be the case that both privateKeyIsRandom and privateKeyIsInRange be changed to just privateKey?

This would be my preference too (same as @cmccandless).

> If so, I think we will have to rely heavily on expected to express the difference between the two. This is, of course, very possible. I'm happy to go with anything suggested, since there is no precedent.

I think using the expected value to discern between what is being tested for the private key is the best option here.

exercises/diffie-hellman/canonical-data.json
ErikSchierboom: The next test case uses `alicePublicKey`, so I think it would make sense to have a common terminology.

exercises/diffie-hellman/canonical-data.json
cmccandless: As noted above, there is currently no precedent for creating canonical data to test randomness. As such, this exercise will become that precedent, so we will want to think carefully about the best way to do this that is also reusable in other instances (i.e. `robot-name` and company)

coriolinus:
I'm not sure that there exists a reasonable way to do this given the
canonical data format in use. It feels like we're going to encounter
https://xkcd.com/221/ rather quickly.

exercises/diffie-hellman/canonical-data.json
petertseng: > The next test case uses alicePublicKey, so I think it would make sense to have a common terminology.

I really wanted to emphasise that only I should know my private key. Therefore, I chose `myPrivateKey`. To me, this was not emphasised enough if I simply use `alicePrivateKey`. However, maybe that can be made obvious in an alternate way.

For example, I can rename the keys to `alicePrivateKey` and `bobPublicKey` and have the description be "Alice can calculate a secret using Bob's public key".

Okay. If I use `alicePrivateKey` and `bobPublicKey`, I will insist that I change the description in that way (or any equivalent way to convey the message that only Alice has the power to calculate the secret in this way).

I prefer `myPrivateKey` at a strength of 1/10 so you must tell me if you prefer `alicePrivateKey` and `bobPublicKey` with any strength > 1/10. The scale is subjective, so anything that the person speaking feels comfortable calling > 1/10 is acceptable.

exercises/diffie-hellman/canonical-data.json
petertseng: I'll *start* with `"expected": {"random": true}`, but I still get the feeling it could be better. I will write a comment later saying why.

exercises/diffie-hellman/canonical-data.json
petertseng: > If so, I think we will have to rely heavily on expected to express the difference between the two. This is, of course, very possible. I'm happy to go with anything suggested, since there is no precedent.

I am having second thoughts about using `expected` for this. I will have to write a comment later explaining why I have second thoughts.

petertseng:
Notice the current state of the tests for two properties of private keys: that they are random, and that they are within range `(1, p)`

```json
    {
      "description": "private key is in range 1 .. p",
      "property": "privateKeyIsInRange",
      "input": {},
      "expected": {
        "greaterThan": 1,
        "lessThan": "p"
      }
    },
    {
      "description": "private key is random",
      "property": "privateKeyIsRandom",
      "input": {},
      "expected": {
        "random": true
      }
    },
```

Let's pretend that, instead, both `property` were `privateKey`. This will require a test generator to make a non-trivial decision about what test code to emit based on the presence or absence of keys in the `expected` object. Define a non-trivial decision as anything other than "insert this value here". For exmaple, a conditional decision is non-trivial.

Compare to a test generator that reads in https://github.com/exercism/problem-specifications/blob/master/exercises/leap/canonical-data.json : It makes no non-trivial decisions; it simply inserts the `year` and `expected` in the correct place, and all is well.

It's never before been explicitly written that test generators should not have to make non-trivial decisions based on `input` or `expected`. I *believe* it is true today that no test generator needs to make a non-trivial decision based on `input` or `expected`; the only decision is based on `property`. If this is to be changed, I need some help to change my way of thinking so that I can understand how to think about `input` and `expected`. Please help me to do that.

petertseng:
EBWODP: Call to action is unclear.

> Please help me to do that.

If you wish for me to use `privateKey` as the `property` for both of these tests, please help me in the above-described way.

petertseng:
> I *believe* it is true today that no test generator needs to make a non-trivial decision based on input or expected

I don't need to *believe*, I can demonstrate based on all the code that's been written in https://github.com/petertseng/exercism-problem-specifications.

I know from grepping on `accept:` that I can enumerate every acceptance check (and therefore, what a possible test generator might need to do) that isn't exact equality. Here they are:

These obviously have no decisions:

* https://github.com/exercism/problem-specifications/blob/master/exercises/circular-buffer/canonical-data.json uses operations, thus has no `expected`, so trivially no decisions are being made.
* https://github.com/exercism/problem-specifications/blob/master/exercises/react/canonical-data.json uses operations, thus has no `expected`, so trivially no decisions are being made.
* https://github.com/exercism/problem-specifications/blob/master/exercises/pov/canonical-data.json allows for differences in child ordering, no decisions.
* https://github.com/exercism/problem-specifications/blob/master/exercises/space-age/canonical-data.json allows values within a delta, no decisions.

Maybe you could argue these make non-trivial decisions:

* https://github.com/exercism/problem-specifications/blob/master/exercises/allergies/canonical-data.json: a test generator for `allergicTo` needs to only check the allergens that are present in the `expected` array. This is just iteration over an array; no conditional decision is being made.
* https://github.com/exercism/problem-specifications/blob/master/exercises/dominoes/canonical-data.json maps false -> "did code under test indicate no chain is possible?" and true -> "did code under test actually give a correct chain?"
* https://github.com/exercism/problem-specifications/blob/master/exercises/robot-simulator/canonical-data.json needs to test only for the key/value pairs actually present in expected. So, this is an iteration over the k/v pairs.

robot-simulator makes decisions based on presence/absence of keys in its `expected` object, the most similar to what might happen in these two `privateKey` tests.

So my only real argument is "But no test generator has ever had to make a decision between two *completely different* tests. Randomness is *completely* different from a bounds check!"

Okay, that means I can only prefer `privateKeyIsInRange` and `privateKeyIsRandom` with strength 3/10. If you want me to change it, you must report a preference for `privateKey` at any higher preference strength.

ErikSchierboom:
> Okay, that means I can only prefer privateKeyIsInRange and privateKeyIsRandom with strength 3/10. If you want me to change it, you must report a preference for privateKey at any higher preference strength.

I'm quite impressed by the thoroughness of your research. Having given it a second thought, using the `property` field to describe what is being tested actually makes a lot of sense.

ErikSchierboom:
@petertseng Is there anything you want to work on this PR, or is it ready to be merged?

petertseng:
I have no changes planned.

ErikSchierboom:
Thanks a lot @petertseng!

PR #1227, opened 2018-04-18
pov: add test for paths not involving root

paiv:
I've seen solutions that always hit root node, even when the shortest path should not. So here is a test for that.


paiv:
@petertseng thank you for your review, here is updated description.

paiv:
@petertseng you are right, having a minimum test is the way; here is updated test case.

petertseng:
Thanks for taking the time to improve our tests!

Issue #1208, opened 2018-03-12
All READMEs for exercises where whitespace is significant should have a placeholder for the whitespace


shingtaklam1324:
For example, the README for crypto-square has this line

> ```
> imtgdvs fearwer mayoogo anouuio ntnnlvt wttddes aohghn  sseoau 
> ```

Which doesn't automatically indicate that there is a whitespace after the `sseoau` when rendered or in many text editors. To make it clearer, it should be like this

> `_` is used for space (ASCII Code 32)
> ```
> imtgdvs_fearwer_mayoogo_anouuio_ntnnlvt_wttddes_aohghn__sseoau_
> ```

cmccandless:
[description](https://github.com/exercism/problem-specifications/blob/master/exercises/crypto-square/description.md)

Personally, I would prefer wrapping text with leading/trailing whitespace in quotes.

> "imtgdvs fearwer mayoogo anouuio ntnnlvt wttddes aohghn  sseoau "

This is common in many shells.

coriolinus:
I strongly prefer the quotes solution over the underscore variant.

petertseng:
Since this issue designates "All READMES for exercises where whitespace is significant", we need to find them.

```
$ git grep ' $' exercises/*/description.md
exercises/crypto-square/description.md:imtgdvs fearwer mayoogo anouuio ntnnlvt wttddes aohghn  sseoau
exercises/ocr-numbers/description.md: _
exercises/ocr-numbers/description.md:|_
exercises/ocr-numbers/description.md:
exercises/ocr-numbers/description.md:    _  _
exercises/ocr-numbers/description.md:
exercises/ocr-numbers/description.md:    _  _
exercises/ocr-numbers/description.md:|_||_ |_
exercises/ocr-numbers/description.md:
exercises/ocr-numbers/description.md: _  _  _
exercises/ocr-numbers/description.md:
exercises/yacht/description.md:    Little Straight 30 points               1 2 3 4 5 scores 30
exercises/yacht/description.md:## Task
```

One by one:

* crypto-square: This issue must be applied. This is important because this describes the expected output.
* ocr-numbers: This describes the input, so failure to notice the trailing whitespaces shouldn't affect the problem. The inputs are going to be in the tests. The lines in question are already in triple backticks. Note that replacement with `_` is going to be undesirable because `_` is a possible character in the input. I recommend that no action need be taken.
* yacht: I recommend removing trailing whitespaces from the description.

[...]

PR #1230, opened 2018-04-22
crypto-square: add quotes to indicate relevant whitespace

rpottsoh:
per #1208.  Quotes (`"`) added to indicate relevant white space in description.

### DIFF
--- a/exercises/crypto-square/description.md
+++ b/exercises/crypto-square/description.md
@@ -10,11 +10,15 @@ regarded as forming a rectangle when printed with intervening newlines.
 
 For example, the sentence
 
-> If man was meant to stay on the ground, god would have given us roots.
+```text
+"If man was meant to stay on the ground, god would have given us roots."
+```
 
 is normalized to:
 
-> ifmanwasmeanttostayonthegroundgodwouldhavegivenusroots
+```text
+"ifmanwasmeanttostayonthegroundgodwouldhavegivenusroots"
+```
 
 The plaintext should be organized in to a rectangle.  The size of the
 rectangle (`r x c`) should be decided by the length of the message,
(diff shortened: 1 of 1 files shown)


exercises/crypto-square/description.md
```diff
@@ -49,19 +53,19 @@ with `c` chunks of `r` length, separated by spaces. For phrases that are
...
-wttddes
-aohghn
-sseoau
+"imtgdvs"
+"fearwer"
+"mayoogo"
+"anouuio"
+"ntnnlvt"
+"wttddes"
+"aohghn"
```
Insti: Should the trailing space be still present here?

exercises/crypto-square/description.md
```diff
@@ -49,19 +53,19 @@ with `c` chunks of `r` length, separated by spaces. For phrases that are
...
+"sseoau"
```
Insti: and here

exercises/crypto-square/description.md
rpottsoh: @Insti long time no see!  Welcome back.
> Should the trailing space be still present here?

I wasn't sure.  There was no white space there to begin with.  #1208 makes no mention of adding white space.  So I thought it best for the first commit to not add any white space.  I thought it best to leave it open for discussion.  I have no issues with adding white space if the consensus is it is needed.

Insti: (Thanks, I'm not sure if I'm back back but I'm here now :smile: )

>  There was no white space there to begin with.

I think that this file has been a casualty of of some of the trailing whitespace clearing attempts which has removed them. 

Either the appropriate trailing whitespace should be added or the quotes should not be included in this section. The trailing whitspace here is a pretty important part of the exercise.

PR #1232, opened 2018-04-29
Added "Four of a kind is not a full house" test case.

Kopfstein:
A full house is three of one number and two of another. Added test case to exclude four of a kind as full house.


NobbZ:
How could this ever been considered a full house?

Kopfstein:
Humans would certainly not consider this as full house. Still, I managed to implement a solution that was happy to count this as full house (checked only if there are two unique numbers). My implementation was certainly not smart, still I would expect the test suite to catch the error.

Kopfstein:
Thanks for sharing the link to the best practice for commit messages!

petertseng:
Thanks for this contribution!

petertseng:
Ah, I said 72 but I think it's barely been 48. The transition between months confused me too much. I am sorry. Given the number of approvals it is probably fine anyway.

PR #1239, opened 2018-05-10
bracket-push: remove ambiguity in description

### LINKED ISSUE #1178: bracket-push: description is ambiguous
[A reasonable question](https://github.com/exercism/go/issues/1072) popped up on the Go track. 

Our API for the problem contains a true/false bool if they match and an error value. Looking at the [problem description](https://github.com/exercism/problem-specifications/blob/master/exercises/bracket-push/description.md): 

> Given a string containing brackets [], braces {} and parentheses (), verify that all the pairs are matched and nested correctly.

I could see the **Given a string** being interpreted in two ways. 

- The string *will* have brackets [], braces {} and(/or?) parentheses ()
- The string *may* have brackets [], braces {} and(/or?) parentheses ()

If it is the former an error condition is an indicator of an incorrect string, if the latter an error condition in probably unnecessary.

[The test cases for the problem](https://github.com/exercism/problem-specifications/blob/master/exercises/bracket-push/canonical-data.json#L14-L19) suggest the latter. If it is the latter perhaps we should tighten up the description a bit? If it is the former we should adjust the test cases.

Thoughts?

rpottsoh:
fixes #1178
Rewording now indicates that the input string `may` have instead of `will` have.

### DIFF
--- a/exercises/bracket-push/description.md
+++ b/exercises/bracket-push/description.md
@@ -1,2 +1,3 @@
-Given a string containing brackets `[]`, braces `{}` and parentheses `()`,
-verify that all the pairs are matched and nested correctly.
+Given a string containing brackets `[]`, braces `{}`, parentheses `()`,
+or any combination thereof, verify that any and all pairs are matched
+and nested correctly.


rpottsoh:
> because empty string is a test case

@petertseng I have made a small change to the wording.  My hope is that it addresses your concern re: empty strings.

PR #1240, opened 2018-05-11
list-ops: clarify purpose of concat function

### LINKED ISSUE #1231: Clarify purpose of list-ops concat function
I think it's supposed to take a list of lists, and concatenate all the lists together.
But I've seen some implementations that think it's supposed to flatten the lists.
Both implementations pass, though they are radically different.

Ref: https://github.com/exercism/problem-specifications/pull/521

Cohen-Carlisle:
closes #1231

### DIFF
--- a/exercises/list-ops/canonical-data.json
+++ b/exercises/list-ops/canonical-data.json
@@ -40,7 +40,7 @@
       ]
     },
     {
-      "description": "given a list of lists, concatenate each list to the previous list",
+      "description": "concatenate a list of lists",
       "cases": [
         {
           "description": "empty list",
--- a/exercises/list-ops/canonical-data.json
+++ b/exercises/list-ops/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "list-ops",
-  "version": "2.2.0",
+  "version": "2.3.0",
   "comments": [
     "Though there are no specifications here for dealing with large lists,",
     "implementers may add tests for handling large lists to ensure that the",
--- a/exercises/list-ops/canonical-data.json
+++ b/exercises/list-ops/canonical-data.json
@@ -57,6 +57,14 @@
             "lists": [[1, 2], [3], [], [4, 5, 6]]
           },
           "expected": [1, 2, 3, 4, 5, 6]
+        },
+        {
+          "description": "list of nested lists",
+          "property": "concat",
+          "input": {
+            "lists": [[[1, 2]], [[3]], [[]], [[4, 5, 6]]]
+          },
+          "expected": [[1, 2], [3], [], [4, 5, 6]]
         }
       ]
     },
--- a/exercises/list-ops/canonical-data.json
+++ b/exercises/list-ops/canonical-data.json
@@ -40,7 +40,7 @@
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


exercises/list-ops/canonical-data.json
```diff
@@ -40,7 +40,7 @@
       ]
     },
     {
-      "description": "concat lists and lists of list into a new list",
+      "description": "given a list of lists, concatenate each list to the previous list",
```
yurrriq: **nitpick**: "concatenate ... to" feels awkward to me, as does "the previous list."

Perhaps something more like the following would be better.
> concatenate a list of lists

exercises/list-ops/canonical-data.json
```diff
@@ -57,6 +57,14 @@
...
           },
           "expected": [1, 2, 3, 4, 5, 6]
+        },
+        {
+          "description": "list of nested lists",
+          "property": "concat",
+          "input": {
+            "lists": [[[1, 2]], [[3]], [[]], [[4, 5, 6]]]
+          },
+          "expected": [[1, 2], [3], [], [4, 5, 6]]
```
yurrriq: :+1: I think this clearly distinguishes `concat` from `flatten`.

exercises/list-ops/canonical-data.json
```diff
@@ -57,6 +57,14 @@
             "lists": [[1, 2], [3], [], [4, 5, 6]]
           },
           "expected": [1, 2, 3, 4, 5, 6]
+        },
+        {
+          "description": "list of nested lists",
+          "property": "concat",
+          "input": {
+            "lists": [[[1, 2]], [[3]], [[]], [[4, 5, 6]]]
```
petertseng: Initial reaction: It would have been more instructive to display some necessary merging, perhaps for example:

```
[[[1], [2]], [[3]], [[]], [[4, 5, 6]]]
```

(with the same expected result)

Upon review, I don't think these two differ meaningfully in what implementations they will accept/reject, so I should not request such a change.

Cohen-Carlisle: Wouldn't the output of this concat be `[[1], [2], [3], [], [4, 5, 6]]` and not `[[1, 2], [3], [], [4, 5, 6]]`? Or perhaps I'm misunderstanding something. :thinking:
I do like this suggestion, if I understand it correctly, because the number of elements from input to output change, whereas in the current test it does not.

petertseng: You got it, I was wrong. Output of `[[[1], [2]], [[3]], [[]], [[4, 5, 6]]]` is your suggested `[[1], [2], [3], [], [4, 5, 6]]` instead of my bogus suggestion.

Cohen-Carlisle: I like this idea and it should also be compatible with any languages that require all elements of a list be of the same type. Pushing now!

petertseng:
> Since the submitter has the ability to merge, I am assuming that it is not my responsibility to merge.

Uh I have been informed that that is incorrect; submitter is in the Alumni group so actually cannot merge.

Okay, in that case, I would probably merge somewhere around 48 hours from now, in case I missed anything.

Issue #1225, opened 2018-04-13
Schema optimises for testing `f(inputs) = output`. What of other types of tests?


petertseng:
Our JSON schema right now most ergonomically expresses tests of the form:

**The code under test, given this input, produces this exact output.**

This is expressed in https://github.com/exercism/problem-specifications/blob/master/canonical-schema.json#L6-L8, which today reads:

```
   , " optimising for the ability to represent example-based tests"
   , " of the form 'function (input) == output'. Future expansions"
   , " may allow for other types of tests such as property-based. "
```

What other types might we care about:

* The code under test, given X input, produces the same result as when it is given Y input
  * https://github.com/exercism/problem-specifications/pull/1224#discussion_r180638468: diffie-hellman key exchange should result in two parties calculating the same secret
  * https://github.com/exercism/problem-specifications/pull/934#issuecomment-335202217 and https://github.com/exercism/problem-specifications/pull/934#issuecomment-335952991: You can arrive at equal `zipper`s using two different series of operations.
* Create a thing then perform some number of operations on it, each of which should produce some expected result
  * Circular-buffer https://github.com/exercism/problem-specifications/pull/488
  * React https://github.com/exercism/problem-specifications/pull/358
  * possibly bank-account, pending JSON: https://github.com/exercism/problem-specifications/issues/554
* Property-based tests
  * diamond for sure
  * possibly "answer is within this range" such as diffie-hellman's https://github.com/exercism/problem-specifications/pull/1224#discussion_r180637557
  * possibly "answer matches this regular expression" such as simple-cipher's https://github.com/exercism/problem-specifications/pull/966#discussion_r150900147
  * possibly "answer is random" such as diffie-hellman's https://github.com/exercism/problem-specifications/pull/1224#discussion_r180638016, simple-cipher's , and robot-name

The number of uses for each of these alternate test types is not large enough that I feel it is worth any possible proposal at this time. Therefore I'm closing this issue immediately and taking no action on it.

If you want to make it easy to make a proposal when the time is right for it, it is in your interest to continue reporting any such use cases here.

petertseng:
forgot to link to simple-cipher's https://github.com/exercism/problem-specifications/pull/966#discussion_r150899857 for random.

petertseng:
Our JSON schema right now most ergonomically expresses tests of the form:

**The code under test, given this input, produces this exact output.**

[...]

PR #1241, opened 2018-05-11
simple-cipher: Add canonical-data.json

### LINKED ISSUE #586: simple-cipher: Implement canonical-data.json
exercises in sync.

The Simple Cipher exercise can be found in the [./exercises/simple-cipher/][exercise-dir] directory.

See http://exercism.io/contribute/canonical-data/simple-cipher for the up-to-date list of
language tracks that have an implementation of the Simple Cipher exercise.

[exercise-dir]: https://github.com/exercism/x-common/tree/master/exercises/simple-cipher

gustavosobral:
Can anyone clarify the CI building output for me? I see that is failing but I didn't understand why. 

As I understood from the [canonical-data.json specification](https://github.com/exercism/problem-specifications#test-data-format-canonical-datajson), I should be able to nest `cases ` keywords.

For this problem in specific, there is a test case where we don't assert, but we use a match test to check a property against a regex, so I used a `match` key for it.

Closes #586

### DIFF
--- a/exercises/simple-cipher/canonical-data.json
+++ b/exercises/simple-cipher/canonical-data.json
@@ -1,6 +1,11 @@
 {
   "exercise": "simple-cipher",
   "version": "1.0.0",
+  "comments":
+    ["Some of the strings used in this file are symbolic and doesn't represent their literal value. They are:",
+      "cipher.key - Represents the Cipher key",
+      "cipher.encode - Represents the output of the Cipher encode method",
+      "new - Represents the Cipher initialization"],
   "cases": [
     {
       "description": "Random key cipher",
--- a/exercises/simple-cipher/canonical-data.json
+++ b/exercises/simple-cipher/canonical-data.json
@@ -1,0 +1,141 @@
+{
+  "exercise": "simple-cipher",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "Random key cipher",
+      "cases": [
+        {
+          "description": "Key is made only of lowercase letters",
+          "property": "key",
+          "matches": "^[a-z]+$"
+        },
+        {
+          "description": "It can encode",
+          "property": "encode",
+          "input": {
+            "plaintext": "aaaaaaaaaa"
+          },
+          "expected": "cipher.key"
+        },
+        {
+          "description": "It can decode",
+          "property": "decode",
+          "input": {
+            "ciphertext": "cipher.key"
+          },
+          "expected": "aaaaaaaaaa"
+        },
+        {
+          "description": "It is reversible. I.e., if you apply decode in a encoded result, you must see the same plaintext encode parameter as a result of the decode method",
+          "property": "decode",
+          "input": {
+            "plaintext": "abcdefghij",
+            "ciphertext": "cipher.encode"
+          },
+          "expected": "abcdefghij"
+        }
+      ]
+    },
+    {
+      "description": "Incorrect key cipher",
+      "cases": [
+        {
+          "description": "It throws an error with an all uppercase key",
+          "property": "new",
+          "input": {
+            "key": "ABCDEF"
+          },
+          "expected": { "error": "Bad key" }
+        },
+        {
+          "description": "It throws an error with a numeric key",
+          "property": "new",
+          "input": {
+            "key": "12345"
+          },
+          "expected": { "error": "Bad key" }
+        },
+        {
+          "description": "It throws an error with empty key",
+          "property": "new",
+          "input": {
+            "key": ""
+          },
+          "expected": { "error": "Bad key" }
+        }
+      ]
+    },
+    {
+      "description": "Substitution cipher",
+      "cases": [
+        {
+          "description": "It keeps submitted key",
+          "property": "key",
+          "input": {
+            "key": "abcdefghij"
+          },
+          "expected": "abcdefghij"
+        },
+        {
+          "description": "It can encode",
+          "property": "encode",
+          "input": { 
+            "plaintext": "aaaaaaaaaa"
+          },
+          "expected": "abcdefghij"
+        },
+        {
+          "description": "It can decode",
+          "property": "decode",
+          "input": { 
+            "ciphertext": "abcdefghij"
+          },
+          "expected": "aaaaaaaaaa"
+        },
+        {
+          "description": "It is reversible. I.e., if you apply decode in a encoded result, you must see the same plaintext encode parameter as a result of the decode method",
+          "property": "decode",
+          "input": {
+            "plaintext": "abcdefghij",
+            "ciphertext": "cipher.encode"
+          },
+          "expected": "abcdefghij"
+        },
+        {
+          "description": "It can double shift encode",
+          "input": {
+            "key": "iamapandabear",
+            "plaintext": "iamapandabear"
+          },
+          "expected": "qayaeaagaciai"
+        },
+        {
+          "description": "It can wrap on encode", 
+          "property": "encode",
+          "input": {
+            "plaintext": "zzzzzzzzzz"
+          },
+          "expected": "zabcdefghi"
+        },
+        {
+          "description": "It can wrap on decode",
+          "property": "decode",
+          "input": {
+            "ciphertext": "zabcdefghi"
+          },
+          "expected": "zzzzzzzzzz"
+        },
+        {
+          "description": "It can handle messages longer than the key",
+          "property": "encode",
+          "input": {
+            "key": "abc",
+            "plaintext": "iamapandabear"
+          },
+          "expected": "iboaqcnecbfcr"
+        }
+      ]
+    }
+  ]
+}
--- a/exercises/simple-cipher/canonical-data.json
+++ b/exercises/simple-cipher/canonical-data.json
@@ -2,7 +2,7 @@
   "exercise": "simple-cipher",
   "version": "1.0.0",
   "comments":
-    ["Some of the strings used in this file are symbolic and doesn't represent their literal value. They are:",
+    ["Some of the strings used in this file are symbolic and do not represent their literal value. They are:",
       "cipher.key - Represents the Cipher key",
       "cipher.encode - Represents the output of the Cipher encode method",
       "new - Represents the Cipher initialization"],
--- a/exercises/simple-cipher/canonical-data.json
+++ b/exercises/simple-cipher/canonical-data.json
@@ -6,15 +6,7 @@
       "description": "Random key cipher",
       "cases": [
         {
-          "description": "Key is made only of lowercase letters",
-          "property": "key",
-          "input": {},
-          "expected": {
-            "match": "^[a-z]+$"
-          }
-        },
-        {
-          "description": "It can encode",
+          "description": "Can encode",
           "property": "encode",
           "input": {
             "plaintext": "aaaaaaaaaa"
@@ -22,7 +14,7 @@
           "expected": "cipher.key"
         },
         {
-          "description": "It can decode",
+          "description": "Can decode",
           "property": "decode",
           "input": {
             "ciphertext": "cipher.key"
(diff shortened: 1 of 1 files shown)


gustavosobral:
Okay, I figured out: The building output was not so clear, I was missing some required properties and the output was saying too much.

exercises/simple-cipher/canonical-data.json
```diff
@@ -0,0 +1,145 @@
...
+            "match": "^[a-z]+$"
+          }
+        },
+        {
+          "description": "It can encode",
+          "property": "encode",
+          "input": {
+            "plaintext": "aaaaaaaaaa"
+          },
+          "expected": "cipher.key"
```
petertseng: this is interesting because we are to understand that this is not the literal string `"cipher.key"` but instead the key of the cipher, whatever that may be.

This therefore falls under one of the categories of https://github.com/exercism/problem-specifications/issues/1225 . 

This suggests some possible courses of action:

1. Make no change to this PR. It's up to readers of this JSON file to understand that certain fields are symbolic instead of literal strings:
    * The `expected` of every `encode`
    * The `input.ciphertext` of every `decode`
2. Make no change to this PR. It's up to readers of this JSON file to understand that all strings beginning with the seven characters `cipher.` are symbolic instead of literal strings.
3. Make a change to this PR that uses a different type (perhaps some JSON object) for such symbolic inputs and/or outputs.
4. Add your suggestion here

I will not yet express an preference ordering between these options

exercises/simple-cipher/canonical-data.json
```diff
@@ -0,0 +1,145 @@
...
+        {
+          "description": "It can decode",
+          "property": "decode",
+          "input": {
+            "ciphertext": "cipher.key"
+          },
+          "expected": "aaaaaaaaaa"
+        },
+        {
+          "description": "It is reversible. I.e., if you apply decode in a encoded result, you must see the same plaintext encode parameter as a result of the decode method",
```
petertseng: This test is useful to show the student what is possible. However, it is not possible for the test to reject any implementation that the previous two tests have not already rejected, since it just composes the previous two tests together.

Thus, the question is: Is the advantage of showing the student the possibilities worth the disadvantage of the test that does not help catch any more mistakes?

I personally thought it was not worth keeping this test. Let me know your thoughts on it.

exercises/simple-cipher/canonical-data.json
```diff
@@ -0,0 +1,145 @@
...
+          },
+          "expected": "abcdefghij"
+        }
+      ]
+    },
+    {
+      "description": "Incorrect key cipher",
+      "cases": [
+        {
+          "description": "It throws an error with an all uppercase key",
```
petertseng: Since not all languages throw errors, I am considering asking for more language-agnostic wording. For example, "it rejects an all-uppercase key" or "it treats an all-uppercase key as an error" or something.

exercises/simple-cipher/canonical-data.json
```diff
@@ -0,0 +1,145 @@
...
+          "property": "new",
```
petertseng: This is interesting since `new` might imply too much an object-oriented interface, which would raise questions as to how this file is to be understood by languages less amenable to object-oriented interfaces.

I don't have a concrete suggestion yet. Let's see if there are any.

exercises/simple-cipher/canonical-data.json
```diff
@@ -0,0 +1,145 @@
...
+        {
+          "description": "It throws an error with a numeric key",
+          "property": "new",
+          "input": {
+            "key": "12345"
+          },
+          "expected": { "error": "Bad key" }
+        },
+        {
+          "description": "It throws an error with empty key",
```
petertseng: I understand that we can immediately know, upon providing an empty key, that we will never be able to encipher or decipher any texts, so we can immediately reject the empty key.

~~What about cases where the key is shorter than the text? How are they handled?~~

For anyone asking that question, the answer is in the currently last test case, entitled "It can handle messages longer than the key"

exercises/simple-cipher/canonical-data.json
```diff
@@ -0,0 +1,145 @@
...
+        {
+          "description": "It can decode",
+          "property": "decode",
+          "input": { 
+            "ciphertext": "abcdefghij"
+          },
+          "expected": "aaaaaaaaaa"
+        },
+        {
+          "description": "It is reversible. I.e., if you apply decode in a encoded result, you must see the same plaintext encode parameter as a result of the decode method",
```
petertseng: same q as above re: whether showing the possibilities is worth the test which catches no extra mistakes

exercises/simple-cipher/canonical-data.json
cmccandless: > 3. Make a change to this PR that uses a different type (perhaps some JSON object) for such symbolic inputs and/or outputs.
> 4. Add your suggestion here

As a sort of combination of the above, what about the following:
```json
"expected": {
  "match": "^{cipher.key}$"
}
```

Where `{symbol}` is to be interpolated with local variables before matching is applied.

On the surface, this approach seems to simply add complication to the current syntax. However, if this is to act as a precedent for future situations, the interpolation approach allows multiple variable values to be used. 

Examples of such use:
- `"^{name.first} {name.last}$"`
- `"^{a + b / c}$"`

However, if my fellow maintainers think this is too complicated, my vote would be option 1.

gustavosobral:
Thank you so much for your extensive feedback @petertseng! I will get back to it as soon as I can

exercises/simple-cipher/canonical-data.json
gustavosobral: The first test guarantee that if we pass `aaaaaaaaaa` as argument to the `encode`, the result should be the same as the key. And the second one guarantee that if we pass `key` as argument to the `decode` we expect a `aaaaaaaaaa` as result. This is the first test that uses a different `plaintext`.

I believe that is good to explicitly verify that the algorithm is reversible, since is a characteristic of it. But you're right, if the encode and the decode are already verified with valid `"reversibles"` outputs, the nesting of the encode and decode must be reversible.

I personally like this reversibles test cases. But I see your point

exercises/simple-cipher/canonical-data.json
gustavosobral: Good point! I think it will be all wrong if I had used `it throws an exception`, which is very specific. But throw a program error is quite universal for me, it's language agnostic and operating systems oriented.

For sure there is room for improvement, I liked your second suggestion more.

What do you think about my points?

exercises/simple-cipher/canonical-data.json
gustavosobral: Good point, this is a tricky one.

I think the problem is that we're not testing the constructor method itself, but verifying that assigning a invalid key to `key` must cause a error, what is done in the constructor here. 

One option for me would be set the property to `key` or maybe something like `key=` will be also valid for me. I don't have any better/concrete suggestion too.

exercises/simple-cipher/canonical-data.json
gustavosobral: Yes, you're right

exercises/simple-cipher/canonical-data.json
gustavosobral: Please look at my reply in the other review.

exercises/simple-cipher/canonical-data.json
gustavosobral: This one is very specific to the canonical-data specification/syntax. I believe you guys know way more than me about where the we are and where we should go. I'm completely open to suggestions.

gustavosobral:
> How would we be able to determine the ciphertext if we don't have both a plaintext and a key?

If we pass `aaaaaaaaaa` as parameter to the encode method, it should return the same value as the key (which can be randomly generated).
If we pass the `key` as argument to the decode method, it should return `aaaaaaaaaa`.

So you don't necessarily need both values for testing.

exercises/simple-cipher/canonical-data.json
```diff
@@ -0,0 +1,145 @@
+{
+  "exercise": "simple-cipher",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "Random key cipher",
```
ErikSchierboom: Reading the [exercise's description](https://github.com/gustavosobral/problem-specifications/blob/master/exercises/simple-cipher/description.md), the random key part is actually the third and last step in the description. To me, this implies that the test cases involving the random key should also be moved to the bottom of the test cases.

exercises/simple-cipher/canonical-data.json
ErikSchierboom: We could also use the `encode` property and name the case something like `Encode using uppercase key`.

exercises/simple-cipher/canonical-data.json
```diff
@@ -0,0 +1,145 @@
...
+          },
+          "expected": { "error": "Bad key" }
+        }
+      ]
+    },
+    {
+      "description": "Substitution cipher",
+      "cases": [
+        {
+          "description": "It keeps submitted key",
```
ErikSchierboom: Is this a useful test? The current format subtly implies state (keeping track of the key). Why not modify the input for the other test cases to have the key explicitly passed as an input value? This is how virtually canonical data cases are usually setup: all input is specified within the test case itself. As an example, the next case would then look like this:

```json
{
    "description": "It can encode",
    "property": "encode",
    "input": { 
        "plaintext": "aaaaaaaaaa",
        "key": "abcdefghij"
    },
    "expected": "abcdefghij"
},
```

exercises/simple-cipher/canonical-data.json
gustavosobral: The third step is making it fault tolerant. The first step is the cipher without key assignment, that's the random key. But yeah, I will organize the order.

exercises/simple-cipher/canonical-data.json
gustavosobral: Very good point. Thank you

gustavosobral:
> We could also use the encode property and name the case something like Encode using uppercase key.

Is not only the `encode` method that should fail, `decode` too. I don't know why, but I could not comment this directly on your review.

> A general remark: I'm not a huge fan of starting every test case description with "it". To me, that does not add anything of value to the description.

I totally agree.

Thanks for the feedback @ErikSchierboom

gustavosobral:
Thank you guys for the extensive feedback! Please check my comments and my changes @petertseng @cmccandless and @ErikSchierboom

exercises/simple-cipher/canonical-data.json
ErikSchierboom: @gustavosobral Is is correct that you have not yet applied the re-ordering?

ErikSchierboom:
Great set of improvements! I've just added a small question.

exercises/simple-cipher/canonical-data.json
gustavosobral: The first step is the Cipher without key, so that's the `Random key` set of cases. The second is step is the cypher accepting keys, this is the `Substitution` set of cases. The third step is making it fault tolerant, that's the `Incorrect key` cases. I believe they are in order now, no?

ErikSchierboom: Aha, I think I did not explain it correctly. What I meant was not to re-order the individual test cases with the "Random key cipher" case, but the "Random key cipher" case itself. At the moment, the order is:

1. "Random key cipher"
1. "Substitution cipher"
1. "Incorrect key cipher"

I would argue that the first case should actually be the last, as it is a special case of the "Substitution cipher" case. Furthermore, the README also lists it as the third case.

gustavosobral: Oh, I see what you mean. So if I understood correctly, your suggestion is to change the order to:

1. Substitution cipher
2. Random key cipher
3. Incorrect key cipher

Right? I think we all agree that the last step is making it fault tolerant against incorrect keys?

I see your point and believe that we can really see the Random key as a expansion of the Substitution part, but the thing is that in the exercise [description.md](https://github.com/exercism/problem-specifications/blob/master/exercises/simple-cipher/description.md) it explicitly say that the step 2 is the random key cipher:

> ### Step 2
> Shift ciphers are no fun though when your kid sister figures it out. Try amending the code to allow us to specify a key and use that for the shift distance. This is called a substitution cipher.

So, the exercise description sees the functionality of accepting a key parameter as a expansion of having the random generated case. Which I also believe is correct. Did you see the point here?

Would you still suggest the re-order in this case?

Thank you so much for the feedback and sorry for the late reply.

ErikSchierboom: Ah, I then misread the description. It's fine to leave it as-is then.

ErikSchierboom:
@petertseng @rpottsoh Would you mind reviewing this PR in the current state?

rpottsoh:
Overall I think this is great!  Initially I was confused by the use of symbolic and literal strings.  I don't think I have seen this used before in the testdata.  It caught me be surprise.  I didn't review all the comments above before reviewing the JSON.  I just dove in.  I think it could be worthwhile to include a comment of sorts at the beginning of the JSON to indicate that some of the strings are symbolic and perhaps to indicate which of those are symbolic.

Once I finally reviewed the comments and the README the JSON made perfect sense.

Thanks @gustavosobral for working on this. 👍

gustavosobral:
Thank you for the feedback @rpottsoh. It makes total sense to add comments in the file. 

I just did it, thanks.

exercises/simple-cipher/canonical-data.json
```diff
@@ -0,0 +1,147 @@
+{
+  "exercise": "simple-cipher",
+  "version": "1.0.0",
+  "comments":
+    ["Some of the strings used in this file are symbolic and doesn't represent their literal value. They are:",
```
rpottsoh: `do not` instead of doesn't.

ErikSchierboom:
@petertseng Are you interested in reviewing this PR? If not, not worries of course.

ErikSchierboom:
With three approvers, I think we are now ready to merge this. Thanks a lot @gustavosobral!

PR #1243, opened 2018-05-13
forth: add test for multiple word redefinition

Disasm:
All solutions I have seen don't support "recursive" word definitions and/or word "versioning". New test case will check both situations.

### DIFF
--- a/exercises/forth/canonical-data.json
+++ b/exercises/forth/canonical-data.json
@@ -378,7 +378,7 @@
           "expected": [5, 6]
         },
         {
-          "description": "can define word that use word with the same name",
+          "description": "can define word that uses word with the same name",
           "property": "evaluate",
           "input": {
             "instructions": [
--- a/exercises/forth/canonical-data.json
+++ b/exercises/forth/canonical-data.json
@@ -365,16 +365,29 @@
           "expected": [12]
         },
         {
-          "description": "can override words multiple times",
+          "description": "can use different words with the same name",
           "property": "evaluate",
           "input": {
             "instructions": [
-              ": add_x 1 + ; 5 add_x",
-              ": add_x add_x 1 + ; 5 add_x",
-              ": add_x add_x 1 + ; 5 add_x"
+              ": foo 5 ;",
+              ": bar foo ;",
+              ": foo 6 ;",
+              "bar foo"
             ]
           },
-          "expected": [6, 7, 8]
+          "expected": [5, 6]
+        },
+        {
+          "description": "can define word that use word with the same name",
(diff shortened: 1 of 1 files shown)


petertseng:
As a person who doesn't know Forth, it was necessary for me to make an inquiry, in the form of the following Forth code:

```forth
: foo 5 ;
: bar foo ;
: foo 6 ;
bar
```

Mercifully, when I tried to run that it resulted in stack 5. If it had resulted in stack 6, true confusion was sure to result.

ErikSchierboom:
> OK, so how about merging 7*24 hours from PR submission if nobody else has anything to say, potentially to shorten to 3*24 if at least one other person Approves (in the GitHub sense of the word)

I agree with this.

Disasm:
@petertseng, your test case is much better for "versioning" part. I propose splitting my test case into two: one for "recursive" definitions and one for word "versioning".

exercises/forth/canonical-data.json
```diff
@@ -365,16 +365,29 @@
...
+              ": bar foo ;",
+              ": foo 6 ;",
+              "bar foo"
             ]
           },
-          "expected": [6, 7, 8]
+          "expected": [5, 6]
+        },
+        {
+          "description": "can define word that use word with the same name",
```
petertseng: for grammar, I think it would be ... "that *uses* word" ...

Disasm: My bad. Thanks.

petertseng:
Oh, my implementation actually can handle it. But one could easily imagine an implementation that could not.

petertseng:
Let's pretend that in the future there is a person who looks at this PR and thinks the cases added are too hard, then asks whether they are in scope.

https://github.com/exercism/problem-specifications/blob/master/exercises/forth/description.md is relatively silent on this affair, only telling us that we support defining new words, with no additional detail on that.

I currently have no particular opinion on whether the description should be updated to specify this more precisely, because I'd need to see a proposal on what the update should be.

I do think it is important to test that user-defined words can refer to other user-defined words. I'm not adamant about keeping the behaviours defined in these particular tests, but I think the fact that the behaviour surprised me is a reason to keep it. I'm all for encouraging exploration of the surprising behaviour, and the realisation that "Oh, this behaviour surprised me" hopefully sparks curiosity.

Disasm:
I think that description update is not necessary because the first line in the description implicitly contains almost everything you need to know. And yes, any surprising behaviour that is not in line with programmer's understanding of Forth should encourage to experiment with other implementations and explore Forth a bit more.

PR #1245, opened 2018-05-25
isbn-verifier: add a test case that would be accepted by an implementation where an invalid character is simply dropped

### LINKED ISSUE #1218: isbn-verifier: Lacks a test case that would be accepted by an implementation where an invalid character is simply dropped
https://github.com/exercism/problem-specifications/issues/1212 has added a test case where an invalid character is treated as zero (3598P1581X), resulting in:

```
  3 * 10
+ 5 * 9
+ 9 * 8
+ 8 * 7
+ 0 * 6
+ 1 * 5
+ 5 * 4
+ 8 * 3
+ 1 * 2
+ 10 * 1
= 264
```

An incorrect implementation treating P as zero will incorrectly accept this case.

Next step: Please craft a case that would be incorrectly accepted by an implementation that silently drops invalid characters (instead of rejecting the invalid ISBN). This will look something like: `????P?????` resulting in:

```
  ? * 9
+ ? * 8
+ ? * 7
+ ? * 6
+ ? * 5
+ ? * 4
+ ? * 3
+ ? * 2
+ ? * 1
= ? (some number divisible by 11)
```

The incorrect implementation https://github.com/petertseng/exercism-problem-specifications/blob/verify/exercises/isbn-verifier/verify.rb#L96-L102 should incorrectly accept this new case (a correct implementation would reject).

Tasks:

* Add the case to https://github.com/exercism/problem-specifications/blob/master/exercises/isbn-verifier/canonical-data.json
* Increase second version component of the JSON file. If current version is `x . y . z`, make the new version `x . y+1 . 0`
* Use a word in https://help.github.com/articles/closing-issues-using-keywords/ in your commit message to indicate that you have closed this issue.

cmccandless:
Closes #1218

### DIFF
--- a/exercises/isbn-verifier/canonical-data.json
+++ b/exercises/isbn-verifier/canonical-data.json
@@ -1,6 +1,6 @@
 {
     "exercise": "isbn-verifier",
-    "version": "2.4.0",
+    "version": "2.5.0",
     "comments": [
         "An expected value of true indicates a valid ISBN-10, ",
         "whereas false means the ISBN-10 is invalid."
--- a/exercises/isbn-verifier/canonical-data.json
+++ b/exercises/isbn-verifier/canonical-data.json
@@ -125,6 +125,14 @@
                 "isbn": "134456729"
             },
             "expected": false
+        },
+        {
+            "description": "invalid characters are not ignored",
+            "property": "isValid",
+            "input": {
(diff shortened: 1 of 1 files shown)


rpottsoh:
The math checks out.  This should be squashed when it is  merged...  And I see that the version has been bumped while I types this.  :+1:

PR #1246, opened 2018-05-25
luhn: add a test case

idealhack:
https://github.com/exercism/problem-specifications/blob/master/exercises/luhn/description.md

### DIFF
--- a/exercises/luhn/canonical-data.json
+++ b/exercises/luhn/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "luhn",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "cases": [
     {
       "description": "single digit strings can not be valid",
@@ -105,6 +105,14 @@
         "value": "091"
       },
       "expected": true
+    },
+    {
+      "description": "strings with non-digits is invalid",
+      "property": "valid",
+      "input": {
+        "value": ":9"
+      },
+      "expected": false
     }
   ]
 }


exercises/luhn/canonical-data.json
```diff
@@ -105,6 +105,14 @@
         "value": "091"
       },
       "expected": true
+    },
+    {
+      "description": "strings with non-digits is invalid",
```
rpottsoh: How is this new case different from `"valid strings with a non-digit included become invalid"`?

idealhack:
```go
for i, char := range str {
    n := int(char - '0')
    if (len(str)-i)%2 == 0 {
        n *= 2
        if n >= 10 {
            n -= 9
        }
    }
    sum += n
}
```

The code above will pass current tests, but not this one.

So we have to check non-digits by adding:

```go
if n > 9 || n < 0 {
    return false
}
```

idealhack:
@rpottsoh I've increased the minor version number, please take a look

exercises/luhn/canonical-data.json
rpottsoh: Thanks for the explanation.  I can see now how `"valid strings with a non-digit included become invalid"` passes without the new test case.

Other reviewers may ask that your new test case be moved to an earlier location in the JSON file.  I'll leave that open to others to address.

@idealhack thanks for your time in improving this exercise! :+1:

rpottsoh:
`version` looks good.  I am going to let this PR sit for now, to allow time for others to weigh in the change you are proposing.

idealhack:
@rpottsoh You are welcome. Exercism is great because of your work.

Issue #1223, opened 2018-04-10
isbn-verifier: lacks a test case for 0 < length < 10 that would be accepted by an implementation that pads left with 0s


petertseng:
In https://github.com/exercism/problem-specifications/pull/1221 we added a case `134456729 ` with 9 digits that would be incorrectly accepted by an implementation that pads **right** with zeroes:

```
  1 * 10
+ 3 * 9
+ 4 * 8
+ 4 * 7
+ 5 * 6
+ 6 * 5
+ 7 * 4
+ 2 * 3
+ 9 * 2
= 209
```

209 is divisible by 11, so such an implementation would incorrectly accept `134456729`.

As a follow-up, please craft a case that would be incorrectly accepted by an implementation that pads **left** with zeroes. This would be some input with digits `abcdefghi`, where

```
  a * 9
+ b * 8
+ c * 7
+ d * 6
+ e * 5
+ f * 4
+ g * 3
+ h * 2
+ i * 1
= ? (some number divisible by 11)
```

Without such a case, incorrect implementations such as that in https://github.com/petertseng/exercism-problem-specifications/blob/verify/exercises/isbn-verifier/verify.rb#L70-L77 will pass the tests, even though it should not.

Task:

* Craft an input with 9 digits of the form described above.
* Add that input to the JSON file.
* Increase second version component of the JSON file. If current version is `x . y . z`, make the new version `x . y+1 . 0`
* Use a word in https://help.github.com/articles/closing-issues-using-keywords/ in your commit message to indicate that you have closed this issue.

petertseng:
As for how this new case interacts with the existing case added in #1221, you can handle this one of two ways:

* crafting an input that only the pad-left implementation would incorrectly accept, then add this input *in addition to* the existing case for pad-right implementations
* crafting an input that both pad-left and pad-right implementations incorrectly accept, then replace the existing case with the new input

PR #1251, opened 2018-06-17
Add an ISBN Verifier test case for short ISBN

jgomo3:
If your code verify the isbn length to be only less than or equal to 10, it will accept some ISBN with less than 10 digits which still satisfy other rules, being "00" a case which could be considered ISBN by some "almost correct" solutions missing this consideration.

I.e: http://exercism.io/submissions/f2fac1e79bf441e08afecd196ec0a442

### DIFF
--- a/exercises/isbn-verifier/canonical-data.json
+++ b/exercises/isbn-verifier/canonical-data.json
@@ -87,6 +87,14 @@
             "expected": false
         },
         {
+            "description": "too short isbn",
+            "property": "isValid",
+            "input": {
+              "isbn": "00"
+            },
+            "expected": false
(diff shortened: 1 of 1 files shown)


petertseng:
Thank you. Please mark that this closes https://github.com/exercism/problem-specifications/issues/1223 using a word in https://help.github.com/articles/closing-issues-using-keywords/ .

petertseng:
(unless someone would like to convince me that #1223 should be done with a case that has nonzero digits)

rpottsoh:
I like it.  I agree that this PR should put #1223 to bed.

PR #1253, opened 2018-06-20
simple-cipher: specify substring length for symbolic values

### LINKED ISSUE #1250: simple-cipher: Random key length and symbolic values in canonical data
### From the canonical data
https://github.com/exercism/problem-specifications/blob/267f1eae9bbe5938d015f64b2e22809c4abd7a77/exercises/simple-cipher/canonical-data.json#L21-L28

### The problem
While trying to implement the canonical data for this exercise in the Python track, I ran into the following:
```Bash
self = <simple_cipher_test.RandKeyCipherTest testMethod=test_can_decode>

    def test_can_decode(self):
        cipher = Cipher()
>       self.assertEqual(cipher.decode(cipher.key), 'aaaaaaaaaa')
E       AssertionError: 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa[45 chars]aaaa' != 'aaaaaaaaaa'
E       - aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
E       + aaaaaaaaaa
```

To summarize, a face-value translation of the canonical data results in failure, because in the canonical data no substring length is specified for `cipher.key`. Thus, when then full string is used, the length of the actual value is much greater than the length of the expected value.

cmccandless:
Closes #1250 

To summarize, a face-value translation of the canonical data results in failure, because in the canonical data no substring length is specified for cipher.key. Thus, when then full string is used, the length of the actual value is much greater than the length of the expected value.

### DIFF
--- a/exercises/simple-cipher/canonical-data.json
+++ b/exercises/simple-cipher/canonical-data.json
@@ -17,13 +17,13 @@
           "input": {
             "plaintext": "aaaaaaaaaa"
           },
-          "expected": "cipher.key.substring(0, 10)"
+          "expected": "cipher.key.substring(0, plaintext.length)"
         },
         {
           "description": "Can decode",
           "property": "decode",
           "input": {
-            "ciphertext": "cipher.key.substring(0, 10)"
+            "ciphertext": "cipher.key.substring(0, expected.length)"
           },
           "expected": "aaaaaaaaaa"
         },
--- a/exercises/simple-cipher/canonical-data.json
+++ b/exercises/simple-cipher/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "simple-cipher",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "comments":
     ["Some of the strings used in this file are symbolic and do not represent their literal value. They are:",
       "cipher.key - Represents the Cipher key",
--- a/exercises/simple-cipher/canonical-data.json
+++ b/exercises/simple-cipher/canonical-data.json
@@ -5,8 +5,8 @@
     ["Some of the strings used in this file are symbolic and do not represent their literal value. They are:",
       "cipher.key - Represents the Cipher key",
       "cipher.encode - Represents the output of the Cipher encode method",
-      "new - Represents the Cipher initialization"
-      "string.substring(start, length) - Represents a substring of 'string' that begins at index 'start' and is 'length' characters long],
+      "new - Represents the Cipher initialization",
+      "string.substring(start, length) - Represents a substring of 'string' that begins at index 'start' and is 'length' characters long"],
   "cases": [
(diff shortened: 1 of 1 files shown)


exercises/simple-cipher/canonical-data.json
```diff
@@ -16,13 +17,13 @@
           "input": {
             "plaintext": "aaaaaaaaaa"
           },
-          "expected": "cipher.key"
+          "expected": "cipher.key.substring(0, 10)"
```
rpottsoh: @cmccandless I think this all looks fine and agrees with the discussion had in #1250.  I have another idea that I want to float...  How about:
```JSON
"expected": "cipher.key.substring(0,plaintext.length)"
```

exercises/simple-cipher/canonical-data.json
```diff
@@ -16,13 +17,13 @@
...
         },
         {
           "description": "Can decode",
           "property": "decode",
           "input": {
-            "ciphertext": "cipher.key"
+            "ciphertext": "cipher.key.substring(0, 10)"
```
rpottsoh: Similar to above:
```JSON
"ciphertext": "cipher.key.substring(0, expected.length)"
```

rpottsoh:
Please squash before merging.

exercises/simple-cipher/canonical-data.json
ErikSchierboom: Ooh, I like that even better (albeit with a space before `plaintext`)!

exercises/simple-cipher/canonical-data.json
ErikSchierboom: I like that!

exercises/simple-cipher/canonical-data.json
```diff
@@ -16,13 +17,13 @@
...
           },
-          "expected": "cipher.key"
+          "expected": "cipher.key.substring(0, plaintext.length)"
         },
         {
           "description": "Can decode",
           "property": "decode",
           "input": {
-            "ciphertext": "cipher.key"
+            "ciphertext": "cipher.key.substring(0, expected.length)"
```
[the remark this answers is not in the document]
rpottsoh: I agree, but I think we all realize that we are really just trying to communicate that length of "another" string should be used here.  I saw this as a way to avoid the need to make two changes if someone later determined that fewer or more characters were needed.  To me, this is a kin to short hand notation and not necessarily code.

ErikSchierboom:
Thanks for doing this PR @cmccandless!

PR #1255, opened 2018-06-25
isbn-verifier: Crafted input to catch more incorrect algorithms

### LINKED ISSUE #1199: isbn-verifier: Tune too-long inputs to catch more incorrect algorithms
This might be more detail than is needed, but it's something I noticed when completing the exercise.

The approached the problem by iterating over the first ten digits of the input and applying the algorithm in the description; any trailing digits were ignored. This passed all the test cases in the example. However, the test cases for ISBNs that were too long failed because the digits that they *did* happen to test were not a valid ISBN, and it was pretty easy to craft a string that was too long (and therefore not an ISBN) but passed the validator.

So, the proposal:

For exercises validating ISBN length ("too long isbn" and "too long isbn no dashes"), the first ten digits of the input should be a valid ISBN in order to catch solutions that ignore trailing digits.

As an extra twist, to catch folks approaching the problem by iterating in reverse, the test input could be eleven digits, the first ten and last ten of which are valid ISBNs (e.g: `35982150886`).

neganp:
This commit adds an input that will be incorrectly validated by
algorithms that don't explicitly check the length of their inputs, as
discussed in #1199 and #993.

For example, if an algorithm only checks the first (or last) ten
digits of the input, ignoring leftovers, it will pass the current test
suite even though it doesn't correctly implement the spec.

closes #1199

### DIFF
--- a/exercises/isbn-verifier/canonical-data.json
+++ b/exercises/isbn-verifier/canonical-data.json
@@ -103,14 +103,6 @@
             "expected": false
         },
         {
-            "description": "too long isbn",
-            "property": "isValid",
-            "input": {
-              "isbn": "3-598-21507-XX"
-            },
-            "expected": false
-        },
-        {
             "description": "check digit of X should not be used for 0",
             "property": "isValid",
             "input": {
--- a/exercises/isbn-verifier/canonical-data.json
+++ b/exercises/isbn-verifier/canonical-data.json
@@ -1,6 +1,6 @@
 {
     "exercise": "isbn-verifier",
-    "version": "2.6.0",
+    "version": "2.7.0",
     "comments": [
         "An expected value of true indicates a valid ISBN-10, ",
         "whereas false means the ISBN-10 is invalid."
@@ -139,6 +139,14 @@
             "property": "isValid",
             "input": {
                 "isbn": "3132P34035"
+            },
+            "expected": false
+        },
+        {
+            "description": "input is too long but contains a valid isbn",
+            "property": "isValid",
+            "input": {
+                "isbn": "98245726788"
             },
             "expected": false
         }


neganp:
Big thanks to @petertseng for helping me with this submission. :+1:

petertseng:
(to avoid any mistakes, the old case I'm referring to that I suggest to remove is `3-598-21507-XX`)

neganp:
I think it's worth keeping both cases. Having both "too long" and "too long with valid ISBN" cases highlights that a solution can have incorrect features even if it incidentally passes the cases designed to test those features.

petertseng:
> highlights that a solution can have incorrect features even if it incidentally passes the cases designed to test those features.

This is an interesting point! Let's talk about it (to make sure I understand the reasoning).

So I think that means that a student who initially writes a solution that checks all 11 digits (corresponding to https://github.com/petertseng/exercism-problem-specifications/blob/verify/exercises/isbn-verifier/verify.rb#L46-L52 ) would correctly reject the `3-598-21507-XX` case, and think "good, I am indeed rejecting ISBNs that are too long". Then when they get to `98245726788` the student sees "ah, in fact I am not rejecting all ISBNs that are too long", and this is a good learning moment.

Okay, I can see this being a good thing. I wish that this could be a part of the experience for everyone as well. Currently a student who writes a solution that checks the left 10 digits (corresponding to https://github.com/petertseng/exercism-problem-specifications/blob/verify/exercises/isbn-verifier/verify.rb#L54-L61) would already incorrectly accept the `3-598-21507-XX` case, and therefore they don't get this extra moment of learning. Is it OK that the experience is different for students who write such a solution?

(The answer might be yes, I haven't thought about it. I am just trying to figure out why and explain my approach to deciding whether to keep a case, and seeing if I should change my approach)

neganp:
> So I think that means that a student who initially writes a solution that checks all 11 digits (corresponding to https://github.com/petertseng/exercism-problem-specifications/blob/verify/exercises/isbn-verifier/verify.rb#L46-L52 ) would correctly reject the 3-598-21507-XX case, and think "good, I am indeed rejecting ISBNs that are too long". Then when they get to 98245726788 the student sees "ah, in fact I am not rejecting all ISBNs that are too long", and this is a good learning moment.

Yeah, this is exactly what I meant! :+1:

I think it's okay for learners to have different experiences as they move through the test suite (indeed, some will have learned this lesson before and be wise to this sort of ploy), but it's worth making sure we're getting the best possible outcome for as many folks as possible.

The cases we're considering are solutions that:

- validate length and check exactly 10 digits
- don't validate length and:
  - validate only the first 10 digits of input
  - validate only the last 10 digits of input
  - validate all digits of input with a generalized checksum

If the learner's solution validates length, I don't think we can catch them out with the ploy of a "treacherous test" that gives the right result for the wrong reason. If they're in the other category, we want a progression of tests where everybody gets a false positive that they'll catch eventually.

By keeping both exercises in, we get the "right outcome for the wrong reasons" in two out of three of the above cases (incorrect solutions that check 11 digits or the last 10 digits of input will reject 3-598-21507-XX for the wrong reasons). The goal would be to find an input that also gives the "right outcome for the wrong reason" for solutions that only check the first ten digits. I think that might just end up with a test case that's a random string of numbers that any solution would reject.

I have to sit down and game this out in more detail tonight... :confounded:

neganp:
Upon review, I think we can take out the "too long isbn" test(3-598-21507-XX). If all my checks are correct, the previous "too long isbn and no dashes" test (3598215078X) will spuriously pass for each of the three solutions types we discussed above. The new test case will catch the oversight later, and everybody gets the lesson.

I've updated the PR removing the old test case.

rpottsoh:
I don't really have the time right now to do a thorough review.  I trust @petertseng and @ErikSchierboom judgment here.  Peter and I have worked together on issues pertaining to this particular exercise before.

PR #1262, opened 2018-07-16
affine-cipher: New Exercise

### LINKED ISSUE #1260: Exercise Idea: Affine Cipher
After doing an atbash exercise I researched the affine cipher and found it to be much more engrossing.
I believe that it could make an excellent addition to the exercism family of problems!

Here you can see the files I have prepared for the bash track:

[README.md.txt](https://github.com/exercism/problem-specifications/files/2194562/README.md.txt)

[affine_cipher_test.sh.txt](https://github.com/exercism/problem-specifications/files/2194564/affine_cipher_test.sh.txt)

[example.sh.txt](https://github.com/exercism/problem-specifications/files/2194565/example.sh.txt)

In the above test I check for pipe functionality.  The pipe functionality is pretty bash specific so I'm not expecting it for other languages.

The example.sh(.txt) above can be invoked as:
    `bash example.sh encode 3 7 "string/or/file"` or 
    `echo "string/or/file" | bash example.sh encode/decode 15 4`

The files for bash are most polished but here are two (unpolished) ruby example solutions to an affine cipher:

This one, named first_affine uses true affine encryption and decryption with the functions
 E(x) = (ax + b) % m and D(y) = a^-1(y - b) % m where a^-1 is the modular multiplicative inverse of a
[first_affine.rb.txt](https://github.com/exercism/problem-specifications/files/2194576/first_affine.rb.txt)
 
This one, simply called affine, encrypts the alphabet from the key and saves resources (and eyes) by encoding and decoding with a simple call of tr with the encrypted alphabet.  I'm developing this implementation into an object with a lot of possibilities
[affine.rb.txt](https://github.com/exercism/problem-specifications/files/2194583/affine.rb.txt)

I haven't made a test for ruby yet. I have tested this class in irb.  The advantage for learners using
OO languages is in building the real decryption function first -  requiring a method to find the modal multiplicative inverse of a - and then developing the simpler implementation. Then, in the tr implementation adding functionality such as changing the key of the object and regenerating the cipherbet without making a new object.

Well, I hope this hasn't been too long winded.  Everyone, please let me know what you think.
I would really like to get this available for any tracks with interest as it really gets participants to dig deeper into ciphers.

Cheers
Guy Gastineau

guygastineau:
I think I got everything correct for this PR.  Let me know if you want to see a bash example.sh that can pass all the tests.

closes #1260

### DIFF
--- a/exercises/affine-cipher/canonical-data.json
+++ b/exercises/affine-cipher/canonical-data.json
@@ -1,0 +1,163 @@
+{
+  "exercise": "affine-cipher",
+  "version": "1.0.0",
+  "comments": [
+    "The test are divided into two groups: ",
+    "* Encoding from English to affine cipher",
+    "* Decoding from affine cipher to all-lowercase-mashed-together English"
+  ],
+  "cases": [
+    {
+      "description": "encode",
+      "comments": [ "Test encoding from English to ciphertext with keys" ],
+      "cases": [
+        {
+          "description": "encode yes",
+          "property": "encode",
+          "input": {
+            "phrase": "yes",
+            "key_a": 5,
+            "key_b": 7
+          },
+          "expected": "xbt"
+        },
+        {
+          "description": "encode no",
+          "property": "encode",
+          "input": {
+            "phrase": "no",
+            "key_a": 15,
+            "key_b": 18
+          },
+          "expected": "fu"
+        },
+        {
+          "description": "encode OMG",
+          "property": "encode",
+          "input": {
+            "phrase": "OMG",
+            "key_a": 21,
+            "key_b": 3
+          },
+          "expected": "lvz"
+        },
+        {
+          "description": "encode O M G",
+          "property": "encode",
+          "input": {
+            "phrase": "O M G",
+            "key_a": 25,
+            "key_b": 47
+          },
+          "expected": "hjp"
+        },
+        {
+          "description": "encode mindblowingly",
+          "property": "encode",
+          "input": {
+            "phrase": "mindblowingly",
+            "key_a": 11,
+            "key_b": 15
+          },
+          "expected": "rzcwa gnxzc dgt"
+        },
+        {
+          "description": "encode numbers",
+          "property": "encode",
+          "input": {
+            "phrase": "Testing,1 2 3, testing.",
+            "key_a": 3,
+            "key_b": 4
+          },
+          "expected": "jqgjc rw123 jqgjc rw"
+        },
+        {
+          "description": "encode deep thought",
+          "property": "encode",
+          "input": {
+            "phrase": "Truth is fiction.",
+            "key_a": 5,
+            "key_b": 17
+          },
+          "expected": "iynia fdqfb ifje"
+        },
+        {
+          "description": "encode all the letters",
+          "property": "encode",
+          "input": {
+            "phrase": "The quick brown fox jumps over the lazy dog.",
+            "key_a": 17,
+            "key_b": 33
+          },
+          "expected": "swxtj npvyk lruol iejdc blaxk swxmh qzglf"
+        },
+        {
+          "description": "encode with a not coprime to m",
+          "property": "encode",
+          "input": {
+            "phrase": "This is a test.",
+            "key_a": 6,
+            "key_b": 17
+         },
+          "expected": "Error: a and m must be coprime."
+        }
+      ]
+    },
+    {
+      "description": "decode",
+      "comments": [ "Test decoding from ciphertext to English with keys" ]
+      "cases": [ 
+        {
+          "description": "decode exercism",
+          "property": "decode",
+          "input": {
+            "phrase": "tytgn fjr",
+            "key_a": 3,
+            "key_b": 7
+          },
+          "expected": "exercism"
+        },
+        {
+          "description": "decode a sentence",
+          "property": "decode",
+          "input": {
+            "phrase": "qdwju nqcro muwhn odqun oppmd aunwd o",
+            "key_a": 19,
+            "key_b": 16
+          },
+          "expected": "anobstacleisoftenasteppingstone"
+        },
+        {
+          "description": "decode numbers",
+          "property": "decode",
+          "input": {
+            "phrase": "odpoz ub123 odpoz ub",
+            "key_a": 25,
+            "key_b": 7
+          },
+          "expected": "testing123testing"
+        },
+        {
+          "description": "decode all the letters",
+          "property": "decode",
+          "input": {
+            "phrase": "swxtj npvyk lruol iejdc blaxk swxmh qzglf",
+            "key_a": 17,
+            "key_b": 33
+          },
+          "expected": "thequickbrownfoxjumpsoverthelazydog"
+        },
+        {
+          "description": "decode with a not coprime to m",
+          "property": "decode",
+          "input": {
+            "phrase": "Test",
+            "key_a": 13,
+            "key_b": 5
+          },
+          "expected": "Error: a and m must be coprime."
+        }
+      ]
+    }
+  ]
+}
--- a/exercises/affine-cipher/canonical-data.json
+++ b/exercises/affine-cipher/canonical-data.json
@@ -174,6 +174,30 @@
           "expected": "thequickbrownfoxjumpsoverthelazydog"
         },
         {
+          "description": "decode with no spaces in input",
+          "property": "decode",
+          "input": {
+            "phrase": "swxtjnpvyklruoliejdcblaxkswxmhqzglf",
+            "key": {
+              "a": 17,
+              "b": 33
+            }
+          },
+          "expected": "thequickbrownfoxjumpsoverthelazydog"
+        },
+        {
+          "description": "decode with too many spaces",
+          "property": "decode",
+          "input": {
+            "phrase": "vszzm    cly   yd cg    qdp",
+            "key": {
+              "a": 15,
+              "b": 16
+            }
+          },
+          "expected": "jollygreengiant"
+        },
+        {
           "description": "decode with a not coprime to m",
           "property": "decode",
           "input": {
--- a/exercises/affine-cipher/canonical-data.json
+++ b/exercises/affine-cipher/canonical-data.json
@@ -16,8 +16,10 @@
           "property": "encode",
           "input": {
             "phrase": "yes",
-            "key_a": 5,
-            "key_b": 7
+            "key": {
+              "a": 5,
+              "b": 7
+            }
           },
           "expected": "xbt"
         },
@@ -26,8 +28,10 @@
           "property": "encode",
           "input": {
             "phrase": "no",
-            "key_a": 15,
-            "key_b": 18
+            "key": {
+              "a": 15,
+              "b": 18
+            }
           },
           "expected": "fu"
         },
@@ -36,8 +40,10 @@
           "property": "encode",
           "input": {
             "phrase": "OMG",
-            "key_a": 21,
-            "key_b": 3
+            "key": {
+              "a": 21,
+              "b": 3
+            }
           },
           "expected": "lvz"
         },
@@ -46,8 +52,10 @@
           "property": "encode",
           "input": {
             "phrase": "O M G",
-            "key_a": 25,
-            "key_b": 47
+            "key": {
+              "a": 25,
+              "b": 47
+            }
           },
           "expected": "hjp"
         },
@@ -56,8 +64,10 @@
           "property": "encode",
           "input": {
             "phrase": "mindblowingly",
-            "key_a": 11,
-            "key_b": 15
+            "key": {
+              "a": 11,
+              "b": 15
+            }
           },
           "expected": "rzcwa gnxzc dgt"
         },
@@ -66,8 +76,10 @@
           "property": "encode",
           "input": {
             "phrase": "Testing,1 2 3, testing.",
-            "key_a": 3,
-            "key_b": 4
+            "key": {
+              "a": 3,
+              "b": 4
+            }
           },
           "expected": "jqgjc rw123 jqgjc rw"
         },
@@ -76,8 +88,10 @@
           "property": "encode",
           "input": {
             "phrase": "Truth is fiction.",
-            "key_a": 5,
-            "key_b": 17
+            "key": {
+              "a": 5,
+              "b": 17
+            }
           },
           "expected": "iynia fdqfb ifje"
         },
@@ -86,8 +100,10 @@
           "property": "encode",
           "input": {
             "phrase": "The quick brown fox jumps over the lazy dog.",
-            "key_a": 17,
-            "key_b": 33
+            "key": {
+              "a": 17,
+              "b": 33
+            }
           },
           "expected": "swxtj npvyk lruol iejdc blaxk swxmh qzglf"
         },
@@ -96,8 +112,10 @@
           "property": "encode",
           "input": {
             "phrase": "This is a test.",
-            "key_a": 6,
-            "key_b": 17
+            "key": {
+              "a": 6,
+              "b": 17
+            }
          },
           "expected": "Error: a and m must be coprime."
         }
@@ -112,8 +130,10 @@
           "property": "decode",
           "input": {
             "phrase": "tytgn fjr",
-            "key_a": 3,
-            "key_b": 7
+            "key": {
+              "a": 3,
+              "b": 7
+            }
           },
           "expected": "exercism"
         },
@@ -122,8 +142,10 @@
           "property": "decode",
           "input": {
             "phrase": "qdwju nqcro muwhn odqun oppmd aunwd o",
-            "key_a": 19,
-            "key_b": 16
+            "key": {
+              "a": 19,
+              "b": 16
+            }
           },
           "expected": "anobstacleisoftenasteppingstone"
         },
@@ -132,8 +154,10 @@
           "property": "decode",
           "input": {
             "phrase": "odpoz ub123 odpoz ub",
-            "key_a": 25,
-            "key_b": 7
+            "key": {
+              "a": 25,
+              "b": 7
+            }
           },
           "expected": "testing123testing"
         },
@@ -142,8 +166,10 @@
           "property": "decode",
           "input": {
             "phrase": "swxtj npvyk lruol iejdc blaxk swxmh qzglf",
-            "key_a": 17,
-            "key_b": 33
+            "key": {
+              "a": 17,
+              "b": 33
+            }
           },
           "expected": "thequickbrownfoxjumpsoverthelazydog"
         },
@@ -152,8 +178,10 @@
           "property": "decode",
           "input": {
             "phrase": "Test",
-            "key_a": 13,
-            "key_b": 5
+            "key": {
+              "a": 13,
+              "b": 5
+            }
           },
           "expected": "Error: a and m must be coprime."
         }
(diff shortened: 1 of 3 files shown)


guygastineau:
Travis CI said there was an unexpected string in the .json
I think it was the ':' after cases in line 109.

`      "cases": [` 

I don't understand why it isn't right.  I don't have much experience with .json, but
I modeled it off of other ones in the repo.  The syntax appeared identical to me.
It shows up red in the PR changed files.

I believe that is the only problem with the PR.  If anyone with more experience could help me fix this, 
I would appreciate it very much. :heart_eyes:

exercises/affine-cipher/description.md
```diff
@@ -0,0 +1,53 @@
...
+  -  where y is the numeric value of an encrypted letter, ie. y = E(x)
+  -  it is important to note that a^-1 is the modal multiplicative inverse
+  -  the modal multiplicative inverse of a only exists is a and m are
+  -  coprime.
+ 
+To find the MMI of a:
+
+  **1 = an % m** # where n is the modal multiplicative inverse of a mod m
+ 
+Because automatice decryption, fails if a is not coprime to m your
```
rpottsoh: `automatice`?  Do you mean `automatic`?

rpottsoh: I don't think the comma is needed after `decryption` as well.

exercises/affine-cipher/description.md
```diff
@@ -0,0 +1,53 @@
...
+program should return status 1 and "Error: a and m must be coprime."
+if they are not.  Otherwise it should encode or decode with the
+provided key.
+ 
+The Caesar (shift) cipher is a simple affine cipher where a is 1 and
+b as the magnitude results in a static displacement of the letters.
```
rpottsoh: Consider writing a and b as `a` and `b`.
```
`a` and `b`
```

exercises/affine-cipher/metadata.yml
```diff
@@ -0,0 +1,4 @@
+---
+blurb: "Create an implementation of the Affine cipher, an ancient encryption algorithm for the Middle East."
```
rpottsoh: Should it be `from` instead of `for`?

exercises/affine-cipher/description.md
guygastineau: yes 'automatic' and that comma is a typo

exercises/affine-cipher/metadata.yml
guygastineau: yes it should be for.

exercises/affine-cipher/description.md
guygastineau: you mean in the same places but with the backticks?

guygastineau:
I added backticks to function vars in the description.  Fixed for in the metadata.  Will commit now.

rpottsoh:
Correct

Sent from my android device.

guygastineau:
####One thing that has me feeling uncertain:
- in the json I named both parts of the key separately as:
  - `key_a` and `key_b`
- I don't know if there is a more elegant solution
######Here it is for reference
   ```yml
        "description": "encode",
+      "comments": [ "Test encoding from English to ciphertext with keys" ],
+      "cases": [
+        {
+          "description": "encode yes",
+          "property": "encode",
+          "input": {
+            "phrase": "yes",
+            "key_a": 5,
+            "key_b": 7
+          },
+          "expected": "xbt"
+        },```

ErikSchierboom:
@guygastineau Maybe you could define the key as an object:

```json
"input": {
  "phrase": "yes",
  "key": {
    "a": 5,
    "b": 7
   }
},
````

**Comment by guygastineau** [comment]
@ErikSchierboom thought about doing that.  Will this cause difficulty in some language tracks? 
I am still not completely sure how the .json gets turned into test files yet.

**Comment by rpottsoh** [comment]
This will not cause difficulty.  It is up the track maintainers to process the JSON either by hand or in some cases write test generators to process the JSON into a test suite for that particular language.  The JSON serves as a guide for the maintainers to follow.


Sent from my android device.

**Comment by guygastineau** [comment]
Great. I'll update the .json with @ErikSchierboom suggestion today.
I think that is exactly what I was hoping we could do for the keys in that file!

**Comment by ErikSchierboom** [comment]
> I am still not completely sure how the .json gets turned into test files yet.

Each track has their own, custom generator logic to convert the JSON canonical data to the actual test file.

**Comment by guygastineau** [comment]
Just to clarify.  do I or do I not need to put a comma after the closing bracket for the key object when I change this?

**Comment by cmccandless** [comment]
> Just to clarify. do I or do I not need to put a comma after the closing bracket for the key object when I change this?

No comma after key object. As a rule, the last property in an object or array is not followed by a comma in proper JSON.

**Comment by guygastineau** [comment]
@cmccandless Thank you for the clarification.
- Passed Travis CI with the last commit.
- I think this is ready now.
- Are we just waiting for a merge pending further review?

File: exercises/affine-cipher/canonical-data.json [hunk 4]
```diff
@@ -0,0 +1,191 @@
...
+          "description": "encode with a not coprime to m",
+          "property": "encode",
+          "input": {
+            "phrase": "This is a test.",
+            "key": {
+              "a": 6,
+              "b": 17
+            }
+         },
+          "expected": "Error: a and m must be coprime."
```
coriolinus: If the keys are `a` and `m`, why are they listed as `a` and `b` in the canonical data?

exercises/affine-cipher/canonical-data.json
```diff
@@ -0,0 +1,191 @@
...
+        }
+      ]
+    },
+    {
+      "description": "decode",
+      "comments": [ "Test decoding from ciphertext to English with keys" ],
+      "cases": [
```
coriolinus: There should be a test decoding encoded text longer than five characters which does not have spaces. 

I'd expect such a test to require that the decoder successfully decode the input even without the spaces, but that's arguable; one could reasonably say that it should throw an error instead. However, right now, we're not defining any behavior in that case by our tests.

exercises/affine-cipher/description.md
```diff
@@ -0,0 +1,54 @@
...
+a mathematical function and then converted to the letter relating to
+its new numeric value. Although all monoalphabetic ciphers are weak,
+the affine cypher is much stronger than the atbash cipher,
+because it has many more keys.
+ 
+the encryption function is:
+ 
+  **E(x) = (ax + b) % m**
+  -  where `x` is the letter's index from 0 - length of alphabet - 1
+  -  `m` is the length of the alphabet - roman alphabet = 26
```
coriolinus: Presumably `m` is not the difference between the length of the alphabet and the roman alphabet. Consider rephrasing:

> `m` is the length of the alphabet. For the roman alphabet, `m == 26`.

exercises/affine-cipher/description.md
```diff
@@ -0,0 +1,54 @@
...
+  **D(y) = a^-1(y - b) % m**
+  -  where `y` is the numeric value of an encrypted letter, ie. y = E(x)
+  -  it is important to note that a^-1 is the modal multiplicative inverse
+  -  the modal multiplicative inverse of a only exists if `a` and `m` are
+  -  coprime.
+ 
+To find the MMI of `a`:
+
+  **1 = an % m**
+  -  where `n` is the modal multiplicative inverse of `a` mod `m`
```
coriolinus: This appears to be a recursive definition. In pseudocode:

```python
def mmi(a, m):
    return a * mmi(a % m) % m
```

Is this meant to be a recursive computation? If not, then I'd like to see some expansion of how to compute the MMI of a value.

exercises/affine-cipher/canonical-data.json
guygastineau: `m` is the length of the alphabet. `a` and `b` make the key for encryption and decryption.
Since the description of the problem has it working with English, `m` will always be 26.

guygastineau: Also, `a` and `m` must be coprime but `b` can be any integer value.

exercises/affine-cipher/description.md
guygastineau: I agree this would be a good change.  That line kept looking funny to me.

exercises/affine-cipher/description.md
guygastineau: Checking the MMI in ruby is as simple as
```ruby
  raise ArgumentError, "Error: a and m must be coprime!" if a.gcd(26) != 1
I could change the `%` in the description to `mod`.  It is literally just the math for finding MMI.

Comment (guygastineau) [reply exercises/affine-cipher/description.md]: - I could change the `%` in the description to `mod`.  It is literally just the math for finding MMI.

Comment (guygastineau) [reply exercises/affine-cipher/description.md]: I agree it is nice to be clear in the README.md's but I also want learners to search the internet to learn more about the math that is going on.  When I built this exercise I didn't know what a `modal multiplicative inverse` was.  I think that learning through self driven research helps us learn how to solve problems better.

File: exercises/affine-cipher/canonical-data.json (hunk 5 above)
Comment (guygastineau) [reply exercises/affine-cipher/canonical-data.json]: This type of test is one that I think each track team should add if they want.
- I have a number of tests that I did not include here because they test to see if input can be piped into the command, and if the program can accept multiline files as an argument.
  - those tests are bash specific, so while I will submit to the bash track they would be inappropriate here.
- Additionally, returning an error if the spaces aren't right in the input would be a nightmare in Bash.
  - It might have a quick (although not very readable) solution with `sed` or `awk`, but the guidelines for the bash track stipulate that we shouldn't use those external programs.
#### If some languages want to add that I think they should, but for some languages it will make the program unnecessarily messy and hard to read.
- most solutions are going to strip out all non-alphanumeric characters for encoding and decoding
- on my host machine I made another script that uses my bash Affine.sh to test the possible keys on any given string or file and spit out the decryption with the keys tagged.
  - That way if I forget the key to something I encrypted I can find it again.
  - Also, if someone has opened an encrypted file and inadvertently added or deleted a character, it is much more pragmatic for the program to decode the text with the error instead of throwing an error and forcing the user to look through encoded text to fix the issue before decoding.

Comment (guygastineau) [reply exercises/affine-cipher/canonical-data.json]: I think a test to show that encoded text without spaces can be decoded successfully makes more sense than throwing an error on non-spaced or incorrectly spaced input.  
Most sensible solutions should already do this though.

File: exercises/affine-cipher/description.md (hunk 7 above)
Comment (guygastineau) [reply exercises/affine-cipher/description.md]: In ruby the MMI can be found with something like this:
```ruby
  inverse = 0
  (1..25).each {|i| inverse = i if (@a * i) % 26 == 1}

Comment (guygastineau) [reply exercises/affine-cipher/description.md]: #### To be clear:
- I think that it is beneficial for most learners to implement a solution which includes a full decryption and finds the `MMI`
- For a fast, resource light implementation is better to revise the iteration so that:
  - the alphabet is encrypted with the function `E(x) = (ax + b) mod m`
  - all encoding and decoding simply translates between the order of the alphabet and its encrypted counterpart.

**Comment by rpottsoh** [comment]
@coriolinus said:
>  I'd like to see some expansion of how to compute the MMI of a value.

@guygastineau the exercises that are available to the track maintainers in this repo should remain language agnostic.  Explaining a concept with code may not be a good idea as the nature of the code you've written may not translate correctly to other languages way of doing things.  It may be necessary to describe concepts mathematically or in long form to more precisely communicate meaning.  Rely more on the maintainers' and strudents' ability to understand the instructions and to not have to infer meaning from code, that they may not be able to understand.

**Comment by rpottsoh** [comment]
I agree with the [comment](https://github.com/exercism/problem-specifications/pull/1262#discussion_r203045384) @coriolinus made.  Individual tracks may decide which cases they want to include or exclude.  I think the definition of the exercise, here, should be as complete as possible.  It is not uncommon for these exercise to be revised later, once individual tracks implement an exercise and discover weaknesses in (or ways to improve) the cases.

**Comment by guygastineau** [comment]
#### Regarding how to compute MMI.

I don't wish to be stubborn, but the equation:
- `1 = an mod m` seems so clear to me.
- example:
  - `3 mod 5 = 3`
  - `3 * 2 mod 5 = 6 mod 5 = 1`
    - `2` is the MMI of `3 mod 5`
- another example:
  - `9 mod 26 = 9`
  - `9 * 3 mod 26 = 27 mod 26 = 1`
  - `3` is the MMI of `9 mod 26`
- a more complicated example:
  - `15 mod 26 = 15`
  - `15 * 7 mod 26 = 105 mod 26 = 1`

##### Would you all like an explanation similar to this one in the description?

File: exercises/affine-cipher/description.md [hunk 8]
```diff
@@ -0,0 +1,54 @@
...
+  -  `m` is the length of the alphabet. For the roman alphabet `m == 26`.
+  -  and `a` and `b` make the key
+ 
+the decryption function is:
+ 
+  `D(y) = a^-1(y - b) mod m`
+  -  where `y` is the numeric value of an encrypted letter, ie. y = E(x)
+  -  it is important to note that a^-1 is the modal multiplicative inverse
+  -  the modal multiplicative inverse of a only exists if `a` and `m` are
+  -  coprime.
```
rpottsoh: I don't think you want a line break between are and coprime.  Coprime appears on the next line as its own bullet point.

guygastineau: Definitely an accident. I'll get to it when I get back home.  BTW, you thoughts on my previous questions would be very welcome.

exercises/affine-cipher/description.md
```diff
@@ -0,0 +1,54 @@
...
+ 
+To find the MMI of `a`:
+
+  `1 = an mod m`
+  -  where `n` is the modal multiplicative inverse of `a` mod `m`
```
rpottsoh: I have been trying to Google `modal multiplicative inverse` and only finding multiplicative inverse, which is simply a reciprocal.  I think it might be worth while to further simplify what is being asked for by trying to use simpler language, perhaps including a link to the definition of MMI that you used?  For example, Googling `MMI` was not helpful for me.

rpottsoh: I am confused by the language, to me it reads like you are using the word you are trying to define in the definition of the word you are trying to define....

rpottsoh:
@guygastineau your [comment](https://github.com/exercism/problem-specifications/pull/1262#issuecomment-405668497) containing the examples is very helpful.  If no one else has any objections I think there would be no harm in an example like this being added.

rpottsoh:
> So does everyone support cases that decode text without spaces or throw errors for decoding text without spaces?

I am not sure why it matters about the spaces.  Any input spaces seem to be ignored and output spaces are inserted after every 5th character.  Maybe I am missing something.

exercises/affine-cipher/description.md
guygastineau: Okay. I will add sources for it.
In short it is literally the modular reciprocal. That's why it is also called `multiplicative inverse of x mod b` or whatever vars you want.  When I get home I have the link in my browser bookmarks.

guygastineau:
@rpottsoh
You are right.  Spaces get ignored in both inputs and are added every five spaces in encoded output. No spaces should be present in decided output.

@coriolinus said there should be tests for irregular spaces in input to be decided. My question was clarifying what those tests should expect.

@rpottsoh Are you saying you don't think those tests are necessary?

guygastineau:
[Here](https://en.wikipedia.org/wiki/Modular_multiplicative_inverse) is the wiki link which I will add to the description.md
It is also referenced in the wiki page link in the source section of the description.md

exercises/affine-cipher/description.md
```diff
@@ -0,0 +1,54 @@
...
+  `E(x) = (ax + b) mod m`
+  -  where `x` is the letter's index from 0 - length of alphabet - 1
+  -  `m` is the length of the alphabet. For the roman alphabet `m == 26`.
+  -  and `a` and `b` make the key
+ 
+the decryption function is:
+ 
+  `D(y) = a^-1(y - b) mod m`
+  -  where `y` is the numeric value of an encrypted letter, ie. y = E(x)
+  -  it is important to note that a^-1 is the modal multiplicative inverse
```
rpottsoh: please rename `modal multiplicative inverse` to `modular multiplicative inverse`.  This makes more sense to me, and was the source of my confusion.  This change should apply to all instances.

exercises/affine-cipher/canonical-data.json
coriolinus: Makes sense; that's my error.

exercises/affine-cipher/canonical-data.json
coriolinus: Agree that it makes more sense to expect text without spaces to decode correctly than to expect an error in that case. Agree that most sensible solutions should already do that.

However, behavior which is untested is undefined, and undefined behavior is the devil. I can't speak for why the Atbash cipher doesn't include a test case for that, but it doesn't matter: we're working on this test suite, not that one. We can do better.

exercises/affine-cipher/description.md
coriolinus: I strongly disagree about the desirability of forcing learners to search the internet for information required to solve the problem. When I was going through the exercism exercises for the first time, I was frustrated to the point of turning in bogus unimplemented solutions with certain of the exercises which required knowledge of that sort. I was a motivated learner, but I wanted to learn about programming in a new language, not about the rules by which nucleotides bind to each other.

Using `%` for `mod` is fine; that's not the issue. The issue is that I don't know what the modal multiplicative inverse is, and the definition provided here is brief to the point of incomprehensibility.  This exercise should include a precise and useful definition of what exactly the MMI is, because students aren't coming to the exercise to learn about the MMI, they're coming here to learn the language of their choice.

coriolinus:
> @guygastineau:
> `1 = an mod m` seems so clear to me.

That is lucky, but not everyone is so lucky. Your worked examples made it much clearer, it would be very helpful to include examples of that sort in the README.

Perhaps it is a math failing, but I find it easier to work with equations where the literal constant is on the right: `ax^2 + by + c = 0`. In this case, that would be `an mod m = 1`.

exercises/affine-cipher/description.md
guygastineau: Yes, I realized I misspelled it `modal` instead of `modular`.  I will be home soon and can commit more fixes.

guygastineau:
@coriolinus I will add those examples in the next commit.  

Regarding the literal constant on the left side: 
- I was just following convention I noticed during my research.  I've no problem changing which side it is on if everyone agrees with it.

exercises/affine-cipher/canonical-data.json
guygastineau: Okay, thank you for the response.  I didn't see this earlier.

guygastineau:
Let me know, everybody, if you think those two test cases added to the bottom of the decode section in the .json (right above the last one - check if `a` and `m` aren't coprime) test incorrect spaces in the decoding input thoroughly enough.

guygastineau:
Thank you @coriolinus I really appreciate your input.
I know I seem stubborn and feisty sometimes, but it is all in pursuit of fruitful discourse.
Your contributions have been very valuable.

rpottsoh:
👍  @guygastineau!  Thanks for putting this all together.
I am going to wait a day at least before merging.  If someone else feels like it is OK to merge sooner that is fine.  This PR should be squashed when it is merged.

guygastineau:
Great. Do I need to rebase this or can it be squashed by whoever merges it?

rpottsoh:
Whomever merges can take care of it.  It wouldn't be good to squash it now as it is easier to review I think with all the commits still available.  A rebase is not needed.


Sent from my android device.

rpottsoh:
Merged.  Congrats @guygastineau on adding your first exercise to Exercism!  Bravo!

guygastineau:
Thank you.

PR #1265, opened 2018-07-19
complex-numbers: reword explanation of `e^(a + bi)`

### LINKED ISSUE #1094: complex-numbers: description for exponential function is unclear
The `description.md` of the complex-numbers exercise is rather unclear on exactly what mathematical operation is meant to implemented by the algorithm mentioned as "exponent of a complex number". The full text of that item reading:

>Exponent of a complex number can be expressed as exp(a + i * b) = exp(a) * exp(i * b), and the last term is given by Euler's formula exp(i * b) = cos(b) + i * sin(b).

Is this supposed to be `ln(complex-number)` or `e^(complex-number)` or even something else?

The C track has recently seen an [issue](https://github.com/exercism/c/issues/256) for this exact question, where I (a maintainer) had no real idea from reading the description, how to answer the OP without more mathematical knowledge than would strictly be required for a programming exercise.

Luckily in this case for the C track, the exercise was implemented very recently and the implementer was still around and able to help out, other cases might not be so lucky.

Essentially the `description.md` could do with rewriting in a somewhat less terse manner.

cmccandless:
Closes #1094

### DIFF
--- a/exercises/complex-numbers/description.md
+++ b/exercises/complex-numbers/description.md
@@ -17,10 +17,7 @@ The reciprocal of a non-zero complex number is
 Dividing a complex number `a + i * b` by another `c + i * d` gives:
 `(a + i * b) / (c + i * d) = (a * c + b * d)/(c^2 + d^2) + (b * c - a * d)/(c^2 + d^2) * i`.
 
-Exponent of a complex number can be expressed as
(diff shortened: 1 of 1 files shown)


cmccandless:
@rpottsoh yes, it's good to go. Merging...

PR #1267, opened 2018-07-19
bob: limit scope of incoming messages

### LINKED ISSUE #1257: bob: consider sentence ending with question mark followed by other punctuation marks
It looks like the current test suite for the Bob exercise is missing a case when a sentence is ending with a question mark followed by other punctuation marks, e.g.:
- What the heck?!
- Does it make any sense to you?..

cmccandless:
Resolves #1257


cmccandless:
As this already has two approvals and this is a minor change that has been discussed in detail in #1257, I will go ahead and merge.

PR #1270, opened 2018-07-21
Remove error-related comment from protein-translation

tejasbubane:
Since all tests assume the inputs to be valid.


tejasbubane:
@rpottsoh Changed the version.

PR #1272, opened 2018-07-23
Saddle-points: Add test for multiple saddle-points in a row

DocFogel:
There is a test for multiple saddle points in a column, but there is no test for multiple points in a row.

These changes rename the existing test to indicate it tests for saddle points in a column and adds a test for multiple points in a row.

### DIFF
--- a/exercises/saddle-points/canonical-data.json
+++ b/exercises/saddle-points/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "saddle-points",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "comments": [
     "Matrix rows and columns are 0-indexed."
   ],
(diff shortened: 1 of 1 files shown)


DocFogel:
I was mentoring a student and the initial solution missed this case. I realized it from reading the code, so I deviced the test and verified that the code failed. 
The student added yet another test where the whole matrix was actually full of saddle-points, but I think this test and the existing one, checking for multiples in a column, covers that scenario too.

rpottsoh:
@DocFogel  thanks for the background information.  Nice find!  I have gone ahead and borrowed your initial comment from your exercism/python#1427 comment so others that may come across this PR in the future will have a better idea behind the PR's purpose.

rpottsoh:
I think this is ready for merging, but I will hold off at least another day, to allow anyone else a chance to provide any input they might like to give.  If someone else wants to merge sooner that is OK too.

PR #1275, opened 2018-07-26
Allergies: add testcase for checking "allergicTo" 

rachelhornung:
One of the solutions I mentored returned True for all allergies that had a  score lower than the total test score. This test should catch this.

### DIFF
--- a/exercises/allergies/canonical-data.json
+++ b/exercises/allergies/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "allergies",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "cases": [
     {
       "description": "allergicTo",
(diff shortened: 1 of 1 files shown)


exercises/allergies/canonical-data.json
```diff
@@ -66,6 +66,31 @@
               "result": false
             }
           ]
+        },
+        {
+          "description": "allergic to strawberries but not peanuts",
```
rpottsoh: I am wondering if it wouldn't be enough to just modify `"allergic to eggs in addition to other stuff"`.  It seems to me that the new case is attempting to test for the same thing, and doing a better job of it apparently.  I am curious to read what others think.  Generally speaking I think this is a good improvement to the exercise. :+1:   Thanks!

rachelhornung:
I agree that allergic to eggs in addition to other stuff with a test for peanuts: false would be sufficient.
I opted for an additional test case since here I want to make sure that the test identified not only multiple allergies, but that some allergies were not diagnosed. Which wouldn't be reflecte by "allergic to eggs in addition to other stuff". But technically, just modifying the existing test case would catch the same problem

rpottsoh:
We'll give this a couple of days.  If no has a strong opinion one way or the other I'll merge this as is.  Stay tuned....

rpottsoh:
@rachelhornung thanks for contributing to this exercise, to make it a little bit better! 👍   I am going to go ahead and merge this.

PR #1282, opened 2018-08-01
bob: improve test descriptions re: absence of letters

Cohen-Carlisle:
I ran across a solution that passed the tests but violated the spirit of
the tests. The student was checking for shouting by checking for:
* the absence of lower case letters AND
* (a string ending in `!` OR
* a string without numbers)

This means that "1,2,3!" would be shouting, but "JUST 1 NUMBER"
wouldn't, which seems to violate the spirit of tests titled:
* shouting numbers
* shouting with no exclamation mark
---
Should a test be added for this case, or is changing test descriptions to be more clear enough?

### DIFF
--- a/exercises/bob/canonical-data.json
+++ b/exercises/bob/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "bob",
-  "version": "1.2.0",
+  "version": "1.2.1",
   "cases": [
     {
       "description": "stating something",
@@ -83,7 +83,7 @@
       "expected": "Whoa, chill out!"
     },
     {
-      "description": "only numbers",
+      "description": "no letters",
       "property": "response",
       "input": {
         "heyBob": "1, 2, 3"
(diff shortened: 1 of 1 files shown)


rpottsoh:
The version will need to be changed to 1.3.0 if you end up adding new cases.

Cohen-Carlisle:
I could easily go either way. I just wanted to see what other people thought about adding new test cases given the same solution would still technically pass the tests. The test names should clue students in now that it's not the "correct" way, but not stop you from using the same logic.

rpottsoh:
@Cohen-Carlisle it appears as though you are able to merge this.  The commits should squashed before the PR is merged.

Cohen-Carlisle:
I actually cannot merge as I'm a member of exercism/alumni, but I've squashed the commits.
Maybe I should get reinstated soon, though :laughing:.

cmccandless:
@Cohen-Carlisle got you covered!

PR #1285, opened 2018-08-04
little schemer: add new problem specs

FTLam11:
Hi, this is a new problem request based around basic recursion. The main inspiration for this exercise comes from the book Little Schemer. Looking forward to feedback. Thanks!

### DIFF
--- a/exercises/little-schemer/canonical-data.json
+++ b/exercises/little-schemer/canonical-data.json
@@ -1,0 +1,215 @@
+{
+  "exercise": "little-schemer",
+  "version": "1.0.0",
+  "comments": [
+    "Tests are ordered to allow sequential build up of primitives",
+    "for use in writing the final recursive function lat?."
+  ],
+  "cases": [
+    {
+      "description": "Test null? of a list",
+      "cases": [
+        {
+          "description": "null of non-empty list",
+          "property": "boolean",
+          "input": {
+            "function": "null?",
+            "list": [1]
+          },
+          "expected": false
+        },
+        {
+          "description": "null of empty list",
+          "property": "boolean",
+          "input": {
+            "function": "null?",
+            "list": []
+          },
+          "expected": true
+        },
+        {
+          "description": "null of non-list",
+          "property": "exception",
+          "input": {
+            "function": "null?",
+            "list": "'Zang'"
+          },
+          "comments": [
+            "The primitive null? is defined only for lists."
+          ],
+          "expected": -1
+        }
+      ]
+    },
+    {
+      "description": "Test car of a list",
+      "cases": [
+        {
+          "description": "car of simple list",
+          "property": "equality",
+          "input": {
+            "function": "car",
+            "list": [1]
+          },
+          "expected": 1
+        },
+        {
+          "description": "car of small list",
+          "property": "equality",
+          "input": {
+            "function": "car",
+            "list": ["Immaculate", 23, "G.O.A.T"]
+          },
+          "expected": "'Immaculate'"
+        },
+        {
+          "description": "car of nested list",
+          "property": "equality",
+          "input": {
+            "function": "car",
+            "list": [[4, 5], 6, 7, [8, 9]]
+          },
+          "expected": [4, 5]
+        },
+        {
+          "description": "car of null list",
+          "property": "exception",
+          "input": {
+            "function": "car",
+            "list": []
+          },
+          "comments": [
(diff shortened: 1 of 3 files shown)


exercises/little-schemer/canonical-data.json
```diff
@@ -0,0 +1,215 @@
...
+        },
+        {
+          "description": "null of non-list",
+          "property": "exception",
+          "input": {
+            "function": "null?",
+            "list": "'Zang'"
+          },
+          "comments": [
+            "The primitive null? is defined only for lists."
```
Insti: Don't test invalid input error conditions

exercises/little-schemer/canonical-data.json
```diff
@@ -0,0 +1,215 @@
...
+            "list": [1]
+          },
+          "expected": 1
+        },
+        {
+          "description": "car of small list",
+          "property": "equality",
+          "input": {
+            "function": "car",
+            "list": ["Immaculate", 23, "G.O.A.T"]
```
Insti: Either use all strings or numbers in the list, we're not testing lists.

Insti:
You also need the problem description and metadata files.

exercises/little-schemer/canonical-data.json
```diff
@@ -0,0 +1,215 @@
...
+  ],
+  "cases": [
+    {
+      "description": "Test null? of a list",
+      "cases": [
+        {
+          "description": "null of non-empty list",
+          "property": "boolean",
+          "input": {
+            "function": "null?",
```
Insti: Those from languages where `?` is not a valid identifier character will probably object to this.

Insti:
While I appreciate the effort that has gone into this, I don't think this is a good exercise to add, it's a bit too specific algorithm.

See also: https://github.com/exercism/problem-specifications/issues/761
And: https://github.com/exercism/discussions/issues/124

(Unless these policies have changed with V2?)

petertseng:
I infer that the ultimate goal of this exercise is to write the function to determine whether a given list is a list of atoms. I infer this using a few things:

A comment in the JSON file:

> sequential build up of primitives for use in writing the final recursive function lat?.

In the description:

>  to be *used together* in defining a **recursive operation** `lat?` (short for list of atoms).

Given this is the case, I would suggest at least two things:

### Exercise name

I'd suggest the `little-schemer` change to something indicative of the exercise content, such as `list-of-atoms`.

The reason is that the name `little-schemer` will be automatically converted to "Little Schemer" and displayed on https://exercism.io/my/tracks/track-name-here , and it would be useful to be able to know at a glance what the exercise is about from there.

This will make more sense to those who have not read that particular book. Even for those who have read that particular book, if the book has more than this particular exercise then the more specific name also serves to disambiguate.

### Intermediate functions

Testing the intermediate functions (`null?`, `car`, `cdr`, `atom?`) goes against the current idea that exercises should not be prescriptive (https://github.com/exercism/discussions/issues/44). If following that idea, the description would only mention the `lat?` function and none of the others, and the tests would also only test `lat?`. After all, who's to say `lat?` can only be implemented that way? There could be other ways and it would be valuable to see them all.

exercises/little-schemer/canonical-data.json
```diff
@@ -0,0 +1,215 @@
...
+    },
+    {
+      "description": "Test atom of a list",
+      "cases": [
+        {
+          "description": "atom of a list",
+          "property": "boolean",
+          "input": {
+            "function": "atom?",
+            "list": ["Immaculate", 23]
```
petertseng: since it is expected that inputs to `atom?` may be things other than lists, I would say to use something reflective of that fact. Perhaps `input`, or if that's deemed too redundant (having an `input` nested inside `input`) then `value`.

FTLam11:
@Insti @petertseng Thank you both for all the helpful feedback. The linked discussions were insightful. I'll rethink this exercise through and see if it's still worthwhile.

ErikSchierboom:
> While I appreciate the effort that has gone into this, I don't think this is a good exercise to add, it's a bit too specific algorithm.

I agree with @Insti here. The exercise also feels like it has some overlap with the existing `list-ops` exercise.

Insti:
I'll close this since it's been a while with no activity.
Thanks for your work on this @FTLam11 
Feel free to re-open or make a new PR if you find a way to make this exercise more Exercism-like.

PR #1288, opened 2018-08-07
saddle-points: Add tests for non-square matrices (NxM)

### LINKED ISSUE #1276: saddle-points: Clarify it's NxN matrices or add NxM test cases
In the [saddle_points](https://github.com/exercism/python/tree/master/exercises/saddle-points) exercise definition there's no indication that the saddle point definition should only cover NxN matrices.

That's why I added tests in my solution to cover NxM as well like so:

```python
def test_non_square(self):
    matrix = [[3, 1, 3], [3, 2, 4]]
    self.assertEqual(saddle_points(matrix), set([(0, 2), (0, 0)]))
 ```

It has been suggested to me that I create an issue or PR to address this but I'm not entirely sure what option I should pick.  

The alternative would've been to leave the tests cases as is and specify the constraints as NxN but the added complexity doesn't seem to be that much of a burden.

### PR DESCRIPTION
**Pull request by Alexhans**
The saddle-points exercise made no indication that it was just for NxN matrices but tests for NxM matrices were non-existent.  

I added the tests and made the description more explicit in relation to non-square matrices.

closes #1276

### CODE REVIEW

File: exercises/saddle-points/description.md [hunk 1]
```diff
@@ -21,5 +21,7 @@ A matrix may have zero or more saddle points.
 Your code should be able to provide the (possibly empty) list of all the
 saddle points for any given matrix.
 
+The matrix can have different a number of rows and columns (Non square).
```
cmccandless: Should be:
The matrix can have *a* different number of rows and columns (non-square).

exercises/saddle-points/canonical-data.json
```diff
@@ -117,6 +117,68 @@
...
+        "expected": [
+            {
+                "row": 0,
+                "column": 1
+            },
+            {
+                "row": 0,
+                "column": 3
+            }
+        ]
```
rpottsoh: A cosmetic request.  Please adjust your indentation spacing to conform to the other test cases.

Alexhans:
Hi, @cmccandless 

I made the change you requested by amending the commit and forcing the push.  I could've made another commit but then I probably would've had to squash the commits for a cleaner history and still have to use force again.

Was this the right approach?  It seems problematic in the sense that we can no longer see the file you were requesting me to change.

exercises/saddle-points/canonical-data.json
```diff
@@ -117,6 +117,68 @@
...
     }
+
   ]
-}
\ No newline at end of file
+}
```
rpottsoh: Technically this is an unrelated change.  If you can undo this that would be great, or move to a separate commit.  Watch out for this going forward.  This won't prevent me from accepting.

cmccandless:
@Alexhans 

Generally, force-pushing is not the best practice (but it does have its uses).

The best approach in this case is usually to go ahead and simply make another commit. There is a "Squash and merge" option for maintainers to keep the `master` branch commit history clean, so contributors generally don't have to worry about squashing their own PRs.

exercises/saddle-points/canonical-data.json
```diff
@@ -117,6 +117,68 @@
...
+                "row": 0,
+                "column": 1
+            },
+            {
+                "row": 0,
+                "column": 3
+            }
+        ]
     }
+
```
rpottsoh: blank line not needed.

Alexhans:
I forgot to change credentials.  I'm going to have to force update with the proper ones.

rpottsoh:
@Alexhans another good habit to get in to would be to create a branch within your fork that contains the changes of your PR, rather than working off your master.  Don't worry about this for this PR.

I am not sure what credentials you need to change.

Alexhans:
>

Alexhans:
> @Alexhans another good habit to get in to would be to create a branch within your fork that contains the changes of your PR, rather than working off your master. Don't worry about this for this PR.

Yes.  That sounds like a good idea.  Not trying to rush things either.

> I am not sure what credentials you need to change.

The issue was committing with the wrong user.email/user.name (I have 2 aliases to set up my local config and absentmindledly used the wrong one after cloning my fork).  

Thanks for the patience.  I'll take these things into consideration for future pull requests.

rpottsoh:
Overall you have done a great job. :+1:    We are only nit picking some pretty minor things.   I see that you have closed this PR.  Are you intending to open another with your other set of credentials?

Alexhans:
@rpottsoh:  Oh, no.  This one is the one I intend to merge.  I've fixed the user.name and user.email on the commits already.

rpottsoh:
@Alexhans is there significance to the order of expected coordinates in case `Can identify saddle points in a non square matrix`?  I am working on an example solution which computes the correct coordinates for the two saddle points.  However, the points found are in (0,0), (0,2) order, opposite as you have indicated in the testdata.

cmccandless:
If the order of the points is actually relevant, the list of points ought to be sorted (probably in ascending order).

rpottsoh:
The order of the coordinate pairs is opposite that of the other tests that have more than one saddle-point.  There is nothing stated in the README or in the testcase about order.  To me the order shouldn't be relevant.  I think in most cases a solution is going to start at (0,0) when looking for valid points.  

As long as the number of actual points agrees with the number of expected points and each actual point matches each expected point, that should be enough.

- Should the docs be updated to specify ascending order?
This is likely the least disruptive to testsuite design.
- Should the docs be updated to specify any order?
This would require the testsuites to compare each actual pair against every expected pair and be satisfied with equality regardless of position of each in a list / array structure.
- Should this conversation be moved to a new issue?

cmccandless:
> Should this conversation be moved to a new issue?

I think so, yes.

rpottsoh:
This conversation has been moved to #1294.

PR #1289, opened 2018-08-08
markdown: Extra tests to ensure markdown symbols are not incorrectly interpreted

grantwinney:
I recently reviewed a nicely refactored solution for the markdown exercise, but I noticed that while the original code passes for the tests I'm suggesting, the student's newly refactored code did not. It's not the student's fault, since the existing tests all passed, but I think these tests may be helpful.

I suggested these tests to the student, who was able to rework the code in a second iteration and make them pass.

My reasoning was that it's entirely reasonable to have a header where the text has a hash in it (like `# My C# Header` and that shouldn't be interpreted as a H2 header.

What does everyone think? If it's too much, or gets away from the goals of this exercise too much, I'm fine with leaving them out.

### DIFF
--- a/exercises/markdown/canonical-data.json
+++ b/exercises/markdown/canonical-data.json
@@ -76,6 +76,30 @@
         "markdown": "# Header!\n* __Bold Item__\n* _Italic Item_"
       },
       "expected": "<h1>Header!</h1><ul><li><strong>Bold Item</strong></li><li><em>Italic Item</em></li></ul>"
+    },
+    {
+      "description": "with markdown symbols in the header text that should not be interpreted",
+      "property": "parse",
+      "input": {
+        "markdown": "# This is a header with # and *** in the text"
+      },
+      "expected": "<h1>This is a header with # and *** in the text</h1>"
+    },
+    {
+      "description": "with markdown symbols in the list item text that should not be interpreted",
+      "property": "parse",
+      "input": {
+        "markdown": "* Item 1 with a # in the text\n* Item 2 with *** in the text"
+      },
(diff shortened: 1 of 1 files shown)


exercises/markdown/canonical-data.json
```diff
@@ -76,6 +76,30 @@
...
+      "property": "parse",
+      "input": {
+        "markdown": "This is a paragraph with # and *** in the text"
+      },
+      "expected": "<p>This is a paragraph with # and *** in the text</p>"
     }
   ]
-}
\ No newline at end of file
+}
```
rpottsoh: Some may criticize this as it is a change unrelated to the purpose of commit 70ffff5ee8d2cee582f240def607ab212ba36624.  This sort of thing happens pretty frequently, especially if using the GitHub editor.  I usually bring it to peoples attention but usually don't require it to be fixed.

grantwinney: @rpottsoh Shoot... I'm going to close this PR and fix the version number and.. maybe whatever happened with the newline.

rpottsoh:
You don't have to.  The PR will get squashed when it is merged to keep the timeline clean.

exercises/markdown/canonical-data.json
grantwinney: Thanks for the heads up. I opened a new one, and didn't use the github editor.
https://github.com/exercism/problem-specifications/pull/1291

grantwinney:
@rpottsoh At first, I tried to re-use this one, thinking it would merge all subsequent changes into this existing PR, and just show two commits. Instead, it ended up opening two separate PRs, which was a little confusing. Oh well, it gave me a chance to clone my fork to avoid the weird newline issue with the end of the file. 🤷‍♂️

PR #1291, opened 2018-08-08
markdown: Extra tests to ensure markdown symbols are not incorrectly interpreted

grantwinney:
I recently reviewed a nicely refactored solution for the markdown exercise, but I noticed that while the original code passes for the tests I'm suggesting, the student's newly refactored code did not. It's not the student's fault, since the existing tests all passed, but I think these additional ones may be helpful.

I suggested these tests to the student, who was able to rework the code in a second iteration and make them pass.

My reasoning was that it's entirely reasonable to have a header where the text has a hash in it (like `# My C# Header` and that shouldn't be interpreted as an H2 header.

What does everyone think? If it's too much, or gets away from the goals of this exercise too much, I'm fine with leaving them out.

### DIFF
--- a/exercises/markdown/canonical-data.json
+++ b/exercises/markdown/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "markdown",
-  "version": "1.2.0",
+  "version": "1.3.0",
   "comments": [
     "Markdown is a shorthand for creating HTML from text strings."
   ],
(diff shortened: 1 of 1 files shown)


petertseng:
If I may ask a question about the `***` contained in these new cases

There are no cases that show that `***` works standalone, so a parser that doesn't handle `***` at all doesn't have to do any extra handling here. The only extra handling required is that of `#`.

So, I think it makes the most sense if either one of the two happens:

* Adding a case where `***` converts to an `<hr/>` (I think that's what it converts to)
* Removing `***` from these new cases

grantwinney:
@petertseng Hm, maybe having three asterisks is confusing since that _is_ valid markdown, and it's not what I intended to test. What I was trying to make sure was that a `#` or `*` in the middle of the text wasn't inadvertently interpreted as a header or list item.

Maybe it'd be more to-the-point if I had used a single asterisk?

```csharp
    {
        "description": "with markdown symbols in the header text that should not be interpreted",
        "property": "parse",
        "input": {
            "markdown": "# This is a header with # and * in the text"
        },
        "expected": "<h1>This is a header with # and * in the text</h1>"
    },
    {
        "description": "with markdown symbols in the list item text that should not be interpreted",
        "property": "parse",
        "input": {
            "markdown": "* Item 1 with a # in the text\n* Item 2 with * in the text"
        },
        "expected": "<ul><li>Item 1 with a # in the text</li><li>Item 2 with * in the text</li></ul>"
    },
    {
        "description": "with markdown symbols in the paragraph text that should not be interpreted",
        "property": "parse",
        "input": {
            "markdown": "This is a paragraph with # and * in the text"
        },
        "expected": "<p>This is a paragraph with # and * in the text</p>"
    }
```

petertseng:
Indeed, if the intent to test that it's not a list item, I think a single asterisk achieves the goal better.

PR #1292, opened 2018-08-08
collatz-conjecture: fix capitalized description

guygastineau:
It seemed like this description shouldn't be capitalized so that it would fit with the style of the other cases in this problem.


guygastineau:
Any chance this will get merged by sometime tomorrow?

I don't mean to be pushy, but we have an open PR waiting on this change in the bash track 😀

guygastineau:
Thank you @rpottsoh !

PR #1295, opened 2018-08-10
affine-cipher: fixed expected policy for errors

guygastineau:
It was brought to my attention while implementing a generator in the Ruby track that I did not follow the proper convention in the cases that return an error.

### DIFF
--- a/exercises/affine-cipher/canonical-data.json
+++ b/exercises/affine-cipher/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "affine-cipher",
-  "version": "1.1.0",
+  "version": "2.0.0",
   "comments": [
     "The test are divided into two groups: ",
     "* Encoding from English to affine cipher",
--- a/exercises/affine-cipher/canonical-data.json
+++ b/exercises/affine-cipher/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "affine-cipher",
-  "version": "1.0.0",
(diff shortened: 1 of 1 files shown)


exercises/affine-cipher/canonical-data.json
```diff
@@ -1,6 +1,6 @@
 {
   "exercise": "affine-cipher",
-  "version": "1.0.0",
+  "version": "1.1.0",
```
cmccandless: This should probably be 2.0.0 instead.

From https://github.com/exercism/problem-specifications#major-version-changes:
```Markdown
MAJOR version changes

The MAJOR version should be changed when the test suite is modified in a fundamentally incompatible way.

There are examples of changes requiring a MAJOR version change:

    A new "property" (test type).
    Renaming a "property".
    Insertion, deletion or renaming of keys in the test data object.
    Changing the type of one of the test data keys.

MAJOR changes should be expected to break even well-behaved test generators.
```

guygastineau: Okay, thank you.
I wasn't sure which one it should be, but you are very right.  It would break any generator's logic.

guygastineau:
Thanks @rpottsoh

guygastineau:
If anyone wants to merge this today, there is a PR in the ruby track awaiting this PR.

Thanks.

guygastineau:
Thank you.

PR #1297, opened 2018-08-15
book-store: Add test case to thwart list-order greedy solutions

### LINKED ISSUE #1215: book-store: missing test case?
Hi.
I was working on the book-store problem (python track) and realised that there are a few cases that were not covered in [canonical-data](https://github.com/exercism/problem-specifications/blob/master/exercises/book-store/canonical-data.json)
What if the set of books wasnt't reassigned and sorted by occurrence and looks like [1, 2, 3, 3, 4, 4, 5, 5]? I couldn't find any restriction on this in specification.
The majority of existing solutions (at least on python track) and even [example.py](https://github.com/exercism/python/blob/master/exercises/book-store/example.py) don't handle this case, but pass all other existing tests.

For them:
[1, 1, 2, 2, 3, 3, 4, 5] gives 5120 and that is the right answer, but
[1, 2, 3, 3, 4, 4, 5, 5] (the same case, but different naming) gives 5160, which seems to be wrong

valrus:
The current "two groups of four" test allows solutions to pass if they add books to groups in the order in which they appear in the list, which doesn't work in general. I saw two solutions in a row which pass the tests as-is but fail this new test, because adding books 2 and 3 to the first group of four will prevent making a second one.

fixes #1215

### DIFF
--- a/exercises/book-store/canonical-data.json
+++ b/exercises/book-store/canonical-data.json
@@ -87,7 +87,7 @@
         },
         {
          "property": "total",
-         "description": "Two groups of four is cheaper than group of five plus group of three, regardless of order",
+         "description": "Two groups of four is cheaper than group of five and three",
          "comments": ["Suggested grouping, [[1,2,4,5],[1,3,4,5]]. This differs from the other 'two groups of four' test in that it will fail for solutions that add books to groups in the order in which they appear in the list."],
          "input": {
            "basket": [1,1,2,3,4,4,5,5]
--- a/exercises/book-store/canonical-data.json
+++ b/exercises/book-store/canonical-data.json
@@ -87,7 +87,7 @@
         },
         {
          "property": "total",
-         "description": "Two groups of four is cheaper than group of five and three",
+         "description": "Two groups of four is cheaper than groups of five and three",
          "comments": ["Suggested grouping, [[1,2,4,5],[1,3,4,5]]. This differs from the other 'two groups of four' test in that it will fail for solutions that add books to groups in the order in which they appear in the list."],
          "input": {
            "basket": [1,1,2,3,4,4,5,5]
--- a/exercises/book-store/canonical-data.json
+++ b/exercises/book-store/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "book-store",
-  "version": "1.3.0",
+  "version": "1.4.0",
   "cases": [
     {
       "description": "Return the total basket price after applying the best discount.",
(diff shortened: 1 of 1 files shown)


exercises/book-store/canonical-data.json
```diff
@@ -85,6 +85,15 @@
          },
          "expected": 5120
         },
+        {
+         "property": "total",
+         "description": "Two groups of four is cheaper than group of five plus group of three, regardless of order",
```
ErikSchierboom: I would suggest to re-use the terminology of the last test case:

"Two groups of four are cheaper than two groups of five and three"

Do we need the "regardless of order" addition? I feel this is better suited for the comment, in which it already is explained.

rpottsoh:
Is this attempting to fix #1215?

exercises/book-store/canonical-data.json
rpottsoh: > Do we need the "regardless of order" addition? I feel this is better suited for the comment, in which it already is explained.

I agree.  `regardless of order` implies that we know this to be true.  There are no tests for every possible combination to confirm this accusation to be true.

valrus:
Yes, that looks like the same thing as I intended to address here.

exercises/book-store/canonical-data.json
valrus: Scaled back the description per your comments.

exercises/book-store/canonical-data.json
```diff
@@ -85,6 +85,15 @@
          },
          "expected": 5120
         },
+        {
+         "property": "total",
+         "description": "Two groups of four is cheaper than group of five and three",
```
ErikSchierboom: I think `than group` should be `than groups`

valrus: Oops. Will fix tonight

rpottsoh: The test case above this one has the same grammatical mistake.  It should probably be fixed in a separate PR and would bump version by `?.?.1`.

valrus: OK, fixed this here, but not in the other test case.

ErikSchierboom: That's good, it should be fixed in a separate PR.

rpottsoh:
I think enough time has elapsed and seeing how there are two approvals I am going to go ahead and merge this PR.  Thanks @valrus for submitting this PR, congrats on your first PR here being merged. 👍

PR #1301, opened 2018-08-17
anagram: Change property to be more descriptive

Stargator:
I find `anagrams` to be a generic name for a method or function. I think the property should be description of either the action or the result.

Since the property is about returning a list of anagrams found of a given word, I think `findAnagrams` would be alright.

### DIFF
--- a/exercises/anagram/canonical-data.json
+++ b/exercises/anagram/canonical-data.json
@@ -1,10 +1,10 @@
 {
   "exercise": "anagram",
-  "version": "1.2.1",
+  "version": "1.3.0",
   "cases": [
     {
       "description": "no matches",
-      "property": "anagrams",
+      "property": "findAnagrams",
       "input": {
         "subject": "diaper",
         "candidates": ["hello", "world", "zombies", "pants"]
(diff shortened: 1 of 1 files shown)


exercises/anagram/canonical-data.json
```diff
@@ -1,10 +1,10 @@
 {
   "exercise": "anagram",
-  "version": "1.2.1",
+  "version": "1.3.1",
```
rpottsoh: Should be 1.3.0 I think.

rpottsoh:
`findAnagrams` is fine to me. :+1:

macta:
Having these kinds of names will make it much easier for many tracks to auto generate test routes. Thanks

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

ErikSchierboom:
I'm with @petertseng on this, where I also slightly prefer `anagrams` over `findAnagrams`, although I find it hard to articulate why. Don't let that deter you from merging this, as my preference is not strong.

petertseng:
I was about to say "but the anagrams exist independent of any action; no verb is required for the anagrams to exist". However, I then refreshed my memory of the exercise, and the implementation needs to pick out the anagrams from a list of candidates, so that can be argued that that needs an action.

So, my preference for `anagrams` over `findAnagrams` is now downgraded to 1/10

PR #1308, opened 2018-08-27
isogram: adding testcase with same first and last character

wiredancer:
There are very talented students who exploit every loophole of the test suite. This guy forgot to look at the first and the last character, and still managed to pass all existing tests! (I like the originality of comparing strchr() with strrchr(), though)
```
    for (size_t i = 1; i < len - 1; i++)
    {
        // If the fist occurence of a character is not the last,
        // that's not an isogram.
        if (strchr(str, str[i]) != strrchr(str, str[i]))
        {
            return false;
        }
    }
```

### DIFF
--- a/exercises/isogram/canonical-data.json
+++ b/exercises/isogram/canonical-data.json
@@ -103,7 +103,7 @@
           "description": "same first and last characters",
           "property": "isIsogram",
           "input": {
-            "phrase": "Angola"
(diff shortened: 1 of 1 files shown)


wiredancer:
I totally agree. Changed to all-lowercase.

petertseng:
That's great, thank you

Issue #1298, opened 2018-08-17
simple-cipher: validating input strings


xarxziux:
I've already raised this as a separate issue in exercism/ecmascript#445, but have just discovered that I should be raising it here.

The unit tests for the Simple Cipher include testing for an error if the cipher is all caps, or all digits, but this is in adequate for the description.  Part of the instructions state:  "Let's make your substitution cipher a little more fault tolerant by ... ensuring that the key contains only lowercase letters".

Unfortunately this is not checked for, and I've seen one case where the submission passed the tests but did not conform to the description.  The addition of tests for spaces or punctuation marks in the key would have failed the submission.

The specification needs to be updated to include these kind of checks.

rpottsoh:
[description](https://github.com/exercism/problem-specifications/blob/master/exercises/simple-cipher/description.md)
[testdata](https://github.com/exercism/problem-specifications/blob/master/exercises/simple-cipher/canonical-data.json)

rpottsoh:
@xarxziux thanks for bringing this issue to light.  Would you be willing to submit a PR that fixes this issue?

xarxziux:
Willing: yes, able: not at the moment.  If you're OK with waiting a week or so, then I'll put something up.

rpottsoh:
There is no rush.

Insti:
I'm in favour of removing the case checking from the description rather than adding it to the tests. This exercise is about simple cipher not string validation.

See also: https://github.com/exercism/problem-specifications/issues/523

Insti:
See also: https://github.com/exercism/problem-specifications/issues/902

rpottsoh:
#1346 has been merged and #1345 closed, which should now close this issue as well.

PR #1316, opened 2018-09-01
Simple-Cipher: Validating Input Strings

xarxziux:
Expand unit tests for validating input strings as per issue #1298.
Three tests added covering:
- Keys containing a single upper case letter.
- Keys containing a space.
- Keys containing a punctuation mark.

### DIFF
--- a/exercises/simple-cipher/canonical-data.json
+++ b/exercises/simple-cipher/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "simple-cipher",
-  "version": "1.1.0",
+  "version": "1.2.0",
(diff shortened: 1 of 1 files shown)


rpottsoh:
consideration for #902 should be made before this PR is merged.  #1298 also discussed an easier solution of updating the description.md.

Insti:
While I think that these tests are well thought out and well implemented, they move the exercise in the wrong direction. So I am not in favor of merging them into this exercise.

See https://github.com/exercism/problem-specifications/pull/1346#issuecomment-428366558 for an alternate suggestion.

PR #1318, opened 2018-09-04
Pangram: Diacritics and ligatures

leNEKO:
What about checking if it can handle diacritics and ligatures ? (é è ç œ ß …)

That could seem secondary, but that is a must for lot of languages that use more characters than the plain ASCII (French, Deutsch, Spanish, …)

What do you think ?

### DIFF
--- a/exercises/pangram/canonical-data.json
+++ b/exercises/pangram/canonical-data.json
@@ -90,6 +90,14 @@
             "sentence": "the quick brown fox jumps over with lazy FX"
           },
           "expected": false
+        },
+        {
+          "description": "can handle diacritics and ligatures",
+          "property": "isPangram",
+          "input": {
+            "sentence": "àbçdfghîjklmnpqrßtÛvWxyzœ"
+          },
+          "expected": true
         }
       ]
     }


petertseng:
In similar vein, we used to have test with umlaut, and a test with Cyrillic. They were removed in https://github.com/exercism/problem-specifications/pull/440 because https://github.com/exercism/problem-specifications/issues/428. Same situation or different situation?

Please prefix PR title and commit message with exercise being affected, so it is possible to tell at a glance from `git log` and PR list https://github.com/exercism/problem-specifications/pulls . We have so many exercises that this becomes necessary to prefix.

leNEKO:
Oh that is strictly the same situation than in #428 …
I have read the discussion and can understand the PITA argument that lead to this decision as i undergo these issues on a day basis :)

In fact we can suggest this to students while mentoring their solutions, if we feel that is valuable, 
but i wonder if we can set canonical test as "extra/optionnal/bonus" ?

petertseng:
> i wonder if we can set canonical test as "extra/optionnal/bonus"

Good question. I don't know what the standard is, but `git grep -i optional` and `git grep -i bonus` do show some interesting results, such as

* https://github.com/exercism/problem-specifications/blob/master/exercises/complex-numbers/canonical-data.json
* https://github.com/exercism/problem-specifications/blob/master/exercises/diffie-hellman/canonical-data.json
* https://github.com/exercism/problem-specifications/blob/master/exercises/robot-simulator/canonical-data.json

PR #1321, opened 2018-09-10
Acronym: Add test for edge case

marionebl:
Add test case for input where, when splitting via the specified delimiter characters, empty strings in the resulting "word" list might trigger out of bound errors.


sshine:
Good job, @marionebl!

rpottsoh:
Seeings how it has been 72 hours since this PR was created and there are two approvals I am going to go ahead and merge this.  Thanks @marionebl :+1:

PR #1322, opened 2018-09-10
Grains: Fixed canonical data to have standard error indicator

sdublish:
Fixes https://github.com/exercism/problem-specifications/issues/1309. 

Also, modified the comments so that it reflects the current state of how the error is indicated.


rpottsoh:
@sdublish thanks for helping out! :+1:

Issue #1320, opened 2018-09-08
phone-number: test cases could be more thorough


LyleCharlesScott:
I noticed when fixing some problems with the JavaScript test cases for phone-number that there is room for improvement, specifically around 11-digit numbers starting with 1. There are tests covering 11-digit numbers that pass and fail based on the first digit being 1, but the tests do not cover whether the rest of the number fits the NXX-NXX-XXXX format 
(N >= 2 && N <= 9) && (X >= 0 && X <= 9). 

Numbers like 1-123-456-7890 and 1-223-156-7890 are invalid and should fail, but would pass the test cases if the code returns 11-digit numbers starting with 1 as valid without further checking the string.

rpottsoh:
@LyleCharlesScott would you be willing to open a PR that would add test cases to the [testdata](https://github.com/exercism/problem-specifications/blob/master/exercises/phone-number/canonical-data.json) to cover this issue?

LyleCharlesScott:
I updated the javascript track; I can send a pull request for the canonical data also :-)

sshine:
@LyleCharlesScott Let me know if you still plan to move your improvements from [ac0ec013...](https://github.com/exercism/javascript/commit/ac0ec01300b6a4079814b030647ea75caf1cbfa6) into canonical data. :1st_place_medal:

PR #1325, opened 2018-09-12
phone-number: invalid when 12 digits and starting with "1" "1"

sshine:
This should catch solutions with a validation function that calls itself recursively when the first digit is '1'.

### DIFF
--- a/exercises/phone-number/canonical-data.json
+++ b/exercises/phone-number/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "phone-number",
-  "version": "1.4.0",
+  "version": "1.5.0",
   "cases": [
     {
       "description": "Cleanup user-entered phone numbers",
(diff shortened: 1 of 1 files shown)


rpottsoh:
@sshine I like this change you've made.  #1320 would like to make more changes that are very similar to what you have done so far.  Would you be willing to make a few more changes?  Please don't feel obligated.

sshine:
I encountered this one because of a bug in my own first solution. "Solving for 11 digits is like checking for '1' and solving for 10 digits!" Except I actually checked for N digits as long as the first N-10 digits were '1's, heh.)

rpottsoh:
@sshine it has been 4 days....  I am going to close this PR in favor of #1327.  Thanks for your efforts.

sshine:
How about a comment that this is non-mandatory for such powerful languages? Surely the vast majority of languages use strings. Even Haskell.

sshine:
It doesn't have a 12-digit number that begins with two 1s. This test catches when the preceding 1 is stripped recursively and what remains is a valid 10-digit number because more than one 1 is eaten.

sshine:
I retract that as I didn't look careful enough.

The 11-digit test you describe is sufficient.

PR #1330, opened 2018-09-18
rest-api: clarify list ordering

### LINKED ISSUE #1304: rest-api: sorting of user list
from @karelvancamp (https://github.com/exercism/exercism/issues/4293#issue-)353830321
> It's not defined in the Instruction Methods if and how the users should be sorted.

    {"users":<List of User objects for >}

> Though it does seem to require alphabetic sort based on object.name to pass the tests.

cmccandless:
Resolves #1304


cmccandless:
I'll give this another 24 hours, and then will merge if no one else has any comments.

PR #1333, opened 2018-09-20
robot-simulator: remove ambiguity from test data

rpottsoh:
Several cases do not have completely clear "expected" results documented.  Some of the exercises did not indicate what compass direction was related to "x" or "y", those cases were removed as follow on cases did clearly indicate compass direction association.  Also cases that only changed direction, without advancing, were not checking that the expected position was still the same as the input position.

### DIFF
--- a/exercises/robot-simulator/canonical-data.json
+++ b/exercises/robot-simulator/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "robot-simulator",
-  "version": "2.2.0",
+  "version": "2.3.0",
   "comments": [
     "Some tests have two expectations: one for the position, one for the direction",
     "Optionally, you can also test",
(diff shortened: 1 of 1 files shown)


exercises/robot-simulator/canonical-data.json
```diff
@@ -209,20 +207,6 @@
...
-          "property": "advance",
-          "input": {
-            "position": {
-              "x": 0,
-              "y": 0
-            },
-            "direction": "north"
-          },
-          "expected": {
-            "direction": "north"
```
rpottsoh: I removed this exercise and another just like it for a couple of reasons.  First, I think they are redundant and two, I cannot infer if `x` or `y` should be changing.  It is clearly decided in other `advance` cases which variable (`x` or `y`) should be changing as the writer of the JSON made the decision.  (X for East/West and Y for North/South).

Issue #1311, opened 2018-08-30
Exercises with non-standard error indicators


Insti:
There are several exercises with non-standard error indicators.

The [canonical data documentation says](https://github.com/exercism/problem-specifications#test-data-format-canonical-datajson)

> If an error is expected (because the input is invalid, or any other reason), the value at "expected" should be an object containing exactly one property, "error", whose value is a string. 

I started creating issues for these but found there were a few more, so I've grouped them all together here.

**Todo:**

Update the `canonical-data.json` for each exercise to use the correct error format.

Exercises:
- [ ] grains   (Issue: #1309  PR: )
- [ ] largest-series-product    (Issue: #1310  PR:)
- [ ] ocr-numbers (PR: )
- [ ] phone-number (PR: )
- [ ] wordy (PR: )
- [ ] queen-attack (PR: )

See also: [hamming canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/hamming/canonical-data.json) for another example of how to do things the right way.

petertseng:
Anything in https://github.com/petertseng/exercism-problem-specifications that has to pass a custom `error` to `verify` must be using a non-standard error (Since the default `error` is the standard).

So, that'd be:

```
$ git grep 'error: ->' | cut -d/ -f2            (08-30 19:35)
binary
change
forth
grains
largest-series-product
ocr-numbers
pascals-triangle
rna-transcription
say
trinary
variable-length-quantity
wordy
```

Note that that is a *sufficient* condition but not a *necessary* condition, so the omission of any exercise in this list does not give it a free pass (particularly, please still do pay attention to phone-number and queen-attack even though they are not in my list but are in the list in the issue description)

petertseng:
I strike `rna-transcription` from my list; it apparently doesn't do errors anymore.

rpottsoh:
Is #902 worth considering again?  Do all these exercises need to be testing for invalid input?  Is looking for invalid input germane to the exercise, if not can those cases be removed?

Insti:
@petertseng nice, you've picked up some I didn't find (probably because the Ruby track doesn't have generators for them.)

@rpottsoh probably. I'm strongly in favor of removing invalid input related tests.

petertseng:
One may wish to discuss `binary-search` which uses `-1` to signal "searched-for item is not in array", and ask whether that is a non-standard error indicator. I don't consider searching for an nonexistent item an error, so I would oppose changing this to use the `error` object.

[...]

PR #1336, opened 2018-09-21
pascals-triangle: implement error object

rpottsoh:
per #1311

### DIFF
--- a/exercises/pascals-triangle/canonical-data.json
+++ b/exercises/pascals-triangle/canonical-data.json
@@ -1,10 +1,10 @@
 {
   "exercise": "pascals-triangle",
-  "version": "1.3.0",
+  "version": "1.4.0",
   "comments": [
     "Expectations are represented here as an array of arrays.",
     "How you represent this idiomatically in your language is up to you.",
-    "An expectation of -1 indicates some sort of failure should occur"
+    "An expectation of 'error' indicates some sort of failure should occur."
   ],
   "cases": [
     {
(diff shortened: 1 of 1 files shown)


exercises/pascals-triangle/canonical-data.json
```diff
@@ -1,6 +1,6 @@
 {
   "exercise": "pascals-triangle",
-  "version": "1.3.0",
+  "version": "1.4.0",
   "comments": [
     "Expectations are represented here as an array of arrays.",
     "How you represent this idiomatically in your language is up to you.",
```
Insti: Comment needs fixing.

exercises/pascals-triangle/canonical-data.json
```diff
@@ -80,7 +80,7 @@
           "input": {
             "count": -1
           },
-          "expected": -1
+          "expected": {"error": "count must be positive"}
```
Insti: Out of scope for this PR, but (opinion) this test should be removed.

rpottsoh: @Insti your meaning that, in your opinion, this case should be removed because it deals with invalid input?  I agree with that opinion.  If I am understanding you correctly I will leave this as it is.  Another PR can be opened to to remove this case.

Insti: @rpottsoh you understand me and I agree with you. :smile:

PR #1337, opened 2018-09-21
implement error object on multiple exercises

rpottsoh:
per #1311
phone-number
say
variable-length-quantity
queen-attack
forth

### DIFF
--- a/exercises/say/canonical-data.json
+++ b/exercises/say/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "say",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "comments": [
     "Here -1 is used as expected value to indicate that the",
     "input value is out of the range described in the exercise."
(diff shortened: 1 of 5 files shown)


exercises/say/canonical-data.json
```diff
@@ -1,6 +1,6 @@
 {
   "exercise": "say",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "comments": [
     "Here -1 is used as expected value to indicate that the",
```
sdublish: Would it make sense to update the comments to reflect the new error object?

Insti:
Multiple changes in 1 PR makes reviewing harder.
Make sure you check all the comments at the start of the files.

exercises/say/canonical-data.json
rpottsoh: Yes it would.  :smile:   Thanks for pointing it out.

rpottsoh:
> Multiple changes in 1 PR makes reviewing harder.

Sorry about that.  I hadn't considered that.

Issue #1312, opened 2018-08-31
binary-search: Is searching for a nonexistent item an error?


petertseng:
Notice that https://github.com/exercism/problem-specifications/blob/master/exercises/binary-search/canonical-data.json uses `-1` to indicate "searched-for item is not in the array". The question: Should this be considered an error? If so, `binary-search` is covered by https://github.com/exercism/problem-specifications/issues/1311 .

I consider searching for a nonexistent item an occurrence that the implementation should expect to routinely happen -- in a language with exceptions, for example, I would not recommend the use of an exception for this. A language with optionals would use its `None` for this case, and use `Some(index)` when there actually is an index.

However, I'll refrain from commenting on whether this should be represented in JSON as a `-1` sentinel or an `error` object or `null`.

In the absence of any other preference, I would suggest to keep it as `-1`, and close this issue without changes.

Insti:
I agree it is not an error.

> I would suggest to keep it as -1, and close this issue without changes.

I'm fine with this.

coriolinus:
The correct answer to this question is strongly dependent on the idoms of language being used. Some off the top of my head:

- C, PHP: `-1`
- go: `0, false` tuple
- rust: `None` from the `Option` enum
- Haskell: `Nothing` from the `Maybe` monad
- python: `KeyError` exception
- bash: `""`

The problem, to me, is that it is possible to return `-1` to indicate this condition in all of these languages, but it is unidiomatic in the majority of them. If we use a non-error `-1` to indicate this condition in the canonical data, it can be interpreted as a requirement that languages implement that, despite it being unidiomatic. 

For that reason, I prefer to represent it as an error; this more clearly signals to track maintainers that special handling and rewriting may be required. This doesn't preclude tracks from using a magic value in languages where that is idiomatic, and it allows tracks with better error handling than magic values to proceed with their native idioms.

PR #1338, opened 2018-09-21
binary-search: Use error object instead of sentinel -1

petertseng:
binary-search 1.2.0

As discussed in
https://github.com/exercism/problem-specifications/issues/1312

Although -1 is a sentinel value, using this sentinel value is not the
usual course of action in some languages. In using an error object, we
avoid giving the wrong idea that we are requiring the use of the
sentinel value.

This error value was defined in
https://github.com/exercism/problem-specifications/issues/401

Of course, languages that wish to use a sentinel value may continue to
do so; this commit is not intended to decree that sentinel values are
forbidden.

Neither is this commit decreeing that all languages must represent this
condition as an error; it is simply a declaration that this condition
bears enough consideration that we'll represent it with a different
type.

Closes https://github.com/exercism/problem-specifications/issues/1312
Checks the related box in https://github.com/exercism/problem-specifications/issues/1311

### DIFF
--- a/exercises/binary-search/canonical-data.json
+++ b/exercises/binary-search/canonical-data.json
@@ -1,8 +1,8 @@
 {
   "exercise": "binary-search",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "comments": [
-    "Here -1 is used to indicate that the value is not included in the array.",
+    "The error object is used to indicate that the value is not included in the array.",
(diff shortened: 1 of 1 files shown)


petertseng:
Thanks! My plan will be to merge it in about 48 hours if I don't hear reason I should not.

Issue #1313, opened 2018-08-31
change: Which, if any, input/coin combinations are errors?


petertseng:
Notice that https://github.com/exercism/problem-specifications/blob/master/exercises/change/canonical-data.json uses `-1` to indicate "cannot make the target amount using the coins given". The question: Should this be considered an error? If so, `change` is covered by https://github.com/exercism/problem-specifications/issues/1311.

I consider searching for a negative target an occurrence that an implementation should not be expected to handle. So I think using an `error` object is reasonable here. We can discuss implementations that should handle negative targets with negative coins, but I think it can be considered out of scope for the exercise. But you can see my previous comments https://github.com/exercism/problem-specifications/issues/336#issuecomment-280231149 on it.

I consider searching for a non-negative but unattainable target an occurrence that the implementation should expect to routinely happen -- in a language with exceptions, for example, I would not recommend the use of an exception for this. A language with optionals would use its `None` for this case, and use `Some(coins)` when there actually is a coin combination.

I would normally refrain from commenting on whether this should be represented in JSON as a `-1` sentinel or an `error` object or `null`.

But, as I say in https://github.com/exercism/problem-specifications/issues/905#issuecomment-330184896 , I also wouldn't encourage languages with *both* optionals and errors (Rust's `Option` and `Result`, or Haskell's `Maybe` and `Either`) to use both of those. So in the target language, I'd suggest just using one.

So this leads me to think they should all be expressed as an `error` object.

In the absence of any other preference, I would suggest to say that all of these should be represented by an `error` object, so this should be handled as part of https://github.com/exercism/problem-specifications/issues/1311. In that case, we can close this issue, for https://github.com/exercism/problem-specifications/issues/1311 will remind us to change `change`'s JSON appropriately.

petertseng:
EBWODP:

> I'll refrain from commenting

That's incorrect. I commented. So, I amend that to "I would normally refrain"

Insti:
```
    {
      "description": "no coins make 0 change",
      "property": "findFewestCoins",
      "input": {
        "coins": [1, 5, 10, 21, 25],
        "target": 0
      },
      "expected": []
    },
    {
      "description": "error testing for change smaller than the smallest of coins",
      "property": "findFewestCoins",
      "input": {
        "coins": [5, 10],
        "target": 3
      },
      "expected": -1
},
```
I don't understand why an error even needs to be considered here.
These cases should be equivalent.
The expected value should be an empty array.

[...]

Issue #905, opened 2017-09-17
Discussion: Getting rid of null and error in canonical-data


Vankog:
It seems we found quite a deep topic to talk about.

There have been some concerns that show, that the canonical data should not have `null` and the `"expected": "error"` structures, because there are quite some inconsistent language concepts in the tracks.

E.g. @NobbZ said in #902:
>> null input,

> In some languages not possible at all, and in my opinion, we should assume some safetynet in the early languages. Also those languages that constantly deal with null-errors should are free to add them as necessary.

--------------

>> I am not aware how many languages do not have a null-concept or something alike

> Haskell, Rust, OCaml, Erlang, Elixir, maybe other.
> 
> Even those that do have a null value do treat it different in their idioms…
> 
> In go it seems as if null is often used synonym with the corresponding zero-value, especially when dealing with arrays, slices or strings.
> 
> In other languages I've seen it as "please insert your default value here" when calling a function or the presence of a computation error/argument error when returned from a function.
> 
> So in other languages we do not even have strict typing to enforce an input string, shall we therefore do create a canonical test that throws when given an integer when strings are expected?
> 
> In some languages we could decide to even use the typesystem of the language to keep out invalid input and create a datatype date Nucleotide = A | C | G | T. Again this would make much of the tests obselete…
> 
> Therefore, as I said, the canonical data should only contain a small limited set of test data which deals with correct input and expectations, while handling errornous input should be in the responsibility of the track.
> 
> Another reason why it should be in the tracks responsibility are different idioms and possibilities of error signaling. In go we have a multi-return and return errors as a value when they occur or nil if not. In Haskell, Rust, OCaml, Elixir, Erlang we have the habbit of returning Maybe, Either, Result, Some, :ok, and :error tuples or whatnot. In Java and Ruby we throw or raise exceptions (which we do in erlang/elixir as well if we consider an error irrecoverable). In C we return a magic value which is documented (I hope) and also do sometimes even some global state which gives numerical and human readable info about the error.

-------------

also @stevejb71 said the same independently here: https://github.com/exercism/problem-specifications/pull/895#discussion_r139301743
> Some languages do not have exceptions, or would return an error type for invalid inputs.

So I have a feeling we have to purge these two concepts from the canonical data and from the schema.

I found "error" in the following exercises:
bowling
collatz-conjecture
hamming
nucleotide-count
palindrome-products
perfect-numbers

I found `null` in the following exercises:
word-search
variable-length-quantity
two-fer
[...]

PR #1340, opened 2018-09-21
change: Use error object instead of sentinel -1

petertseng:
change 1.3.0

As proposed in
https://github.com/exercism/problem-specifications/issues/1313
https://github.com/exercism/problem-specifications/issues/905#issuecomment-330184896

In contrast to the proposal in
https://github.com/exercism/problem-specifications/issues/336#issuecomment-280231149

Although -1 is a sentinel value, the sentinel value had been overloaded
in this JSON file to mean two separate conditions:

* "Can't make target" (ostensibly, we might be able to make the target
  with a different set of coins)
* "Target is negative" (no matter what coins we provide, this target is
  always invalid)

To make clear that these two conditions are different, we use an error
object describing each.

This error object was defined in
https://github.com/exercism/problem-specifications/issues/401

Note that this commit is not a decree that all languages must represent
these conditions as errors; languages should continue to represent the
conditions in the usual way for that language. It is simply a
declaration that these two conditions bear enough consideration that
we'll represent them with a different type and also clearly
differentiate between the two.

Closes https://github.com/exercism/problem-specifications/issues/1313
Checks the related box in https://github.com/exercism/problem-specifications/issues/1311

### DIFF
--- a/exercises/change/canonical-data.json
+++ b/exercises/change/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "change",
-  "version": "1.2.0",
+  "version": "1.3.0",
   "comments": [
     "Given an infinite supply of coins with different values, ",
     "find the smallest number of coins needed to make a desired ",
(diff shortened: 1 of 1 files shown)


petertseng:
Thanks! My plan will be to merge it in about 48 hours if I don't hear reason I should not.

PR #1345, opened 2018-10-02
Simple-Cipher: Added additional test case

darthbob88:
There are more invalid keys for this exercise than just UPPER-CASE and numer1c.

### DIFF
--- a/exercises/simple-cipher/canonical-data.json
+++ b/exercises/simple-cipher/canonical-data.json
@@ -134,6 +134,14 @@
           },
           "expected": { "error": "Bad key" }
         },
+         {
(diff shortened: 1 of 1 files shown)


rpottsoh:
closing in favor of #1316.

`"Throws an error with a key containing a punctuation mark"` in #1316 I think covers the proposed case in this PR.  Furthermore I think  this [comment](https://github.com/exercism/problem-specifications/issues/1298#issuecomment-415665248) along with #902. need to be considered.  It seems that this can be resolved more easily by updating the wording in [description.md](https://github.com/exercism/problem-specifications/blob/master/exercises/simple-cipher/description.md)

Issue #523, opened 2017-01-30
luhn: It's about checksums.


Insti:
The main purpose of this problem is doing the checksum calculation so checking that the input string is valid should not be part of the requirements.

Test strings in the canonical-data.json file should all be well formed and only contain numbers. (Although I could be convinced that adding spaces for readability is a good idea.)

gibfahn:
I find it quite helpful to have to parse and filter input, it makes the exercise seem more like actually writing a function that might actually be used rather than just implementing an algorithm. (At least for me) the idea is to practice writing realistic code in different languages to get a feel for them, so I appreciate extra touches such as handling dodgy input.

rootulp:
I think there are enough practice problems that require parsing and filtering input. I agree with @Insti  that the focus should be on calculating the checksum.

stkent:
There's a definite tension here: parsing is well-covered by Exercism, but skipping it in any individual exercise is certainly likely to reduce the "real-world applicability." I'd be in favor of trimming down to assume valid input, and continuing to test validity (vs. returning the actual checksum value; not sure if that was part of the original suggestion here or not, so figured I'd clarify).

gibfahn:
Having been through the exercises for Rust, and then spending some time making the tests fit the description in the README, I found this specific test had quite an interesting filtering section. The solution should only allow digits or ` ` (the space character), not any other whitespace, but should also require you to strip out the whitespace before applying the algorithm.

The actual implementation by comparison was quite mundane.

I haven't written a lot of exercises, so I don't have much perspective on that, just giving my feeling as someone going through all the tests to practice a language.

>and continuing to test validity (vs. returning the actual checksum value; not sure if that was part of the original suggestion here or not, so figured I'd clarify).

I don't _think_ anyone suggested removing the `result % 10 == 0` part of the exercise.

rbasso:
In the current form, this exercises forces us to return `false` on invalid strings (which I don't like too much), so we have to check the input. Changing it, so that all the input strings are clean, would create this scenario:

- If the solution doesn't check the input, when given invalid input, it will either:
    - return `true` and/or `false` to some invalid inputs.
    - generate an error or halt, so it is a *partial function*.
- If it checks the input, it can, reasonably:
    - return `false` on every invalid input, keeping the current behavior.
    -  generate an error, so it is a *partial function*.

If the exercise is changed as proposed, given the above options, I would either:
  - check the input, to generate an useful error message.
[...]

PR #1346, opened 2018-10-02
simple-cipher: remove checks for invalid input

rpottsoh:
- Mention of invalid input removed from README.
- Removed cases that test for invalid input.

These changes follow on the discussion that has taken place in  #902 and #523.  There may be some other examples regarding the discussion that I am missing.  Feel free to point them out.

This PR is intended to be an alternate solution to #1298 to replace #1316.

### DIFF
--- a/exercises/simple-cipher/canonical-data.json
+++ b/exercises/simple-cipher/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "simple-cipher",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "comments":
     ["Some of the strings used in this file are symbolic and do not represent their literal value. They are:",
       "cipher.key - Represents the Cipher key",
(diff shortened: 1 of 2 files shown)


Insti:
I like this change.

Perhaps we (@xarxziux?) can make a new exercise called 'key-validation' or something that combines these tests with the tests that they added in #1346 since string validation is still an important thing to learn.

xarxziux:
Do you mean an advanced version of this exercise (not-so-simple-cipher?), or an entirely separate exercise that only performs general data validation?

I think having a tiered version of these kind of exercises would be good in principle - whether that is in the form of a separate, more advanced exercise, or optional test cases.  I'm just not sure that the format of this site really supports either.  Is there an accepted way to declare optional or advanced test cases in the spec?

rpottsoh:
A new exercise could be a side exercise and it would either require that the student have already completed this exercise in order to have the code necessary for ciphering or the new exercise could provide a stub that contained code for ciphering.  The student would be tasked with writing the code to protect the ciphering logic from bad input.  If the ciphering logic is provided then the new exercise would be independent of this exercise...  I am not sure which way is the best to go.

xarxziux:
On another coding site, Codewars, there are a number of exercises that form a series.  You first take on a relatively easy problem.  Once you've solved that, then there's a more difficult version so solve and so on.  This site doesn't really have that concept, but maybe it would not be a bad idea to do something like that.

The description on this exercise already has three stages.  If we are making the requirements easier, instead of making the unit tests harder, maybe we should look at serialising the exercise.  Instead of steps 1 to 3, split it into two or three separate exercises with increasing difficulty.  I'm not sure there's enough material in the current exercise to warrant three exercises, but I'm sure there are extra complexities and restrictions that can be made on the final stage to make it work.

rpottsoh:
I am not opposed to this idea.  I have already been thinking about it for a
new exercise specific to the Delphi track.  The exercise is for writing a
calculator.  The first half of the exercise will deal with the business
logic of dealing with performing the calculations; handled in a class....
The follow on exercise will see the student creating a UI to interface with
the class they wrote in the previous exercise.  This allows the student a
chance to work with Delphi's nice UI creation capabilities while at the
same time promoting loose coupling between the business logic on the UI.  I
have not worked on this in MANY months.  The idea could certainly be
applied to other tracks.

Insti:
I meant an entirely new and separate exercise that just performs data validation.

It should work standalone, but would be ok if it is part of a series.

PR #1351, opened 2018-10-04
sum-of-multiples: test added, multiples of 3 or 5 up to 20

lotharschulz:
follow up of https://github.com/exercism/kotlin/pull/235

This PR adds a test about an interesting case:

Given a set of 2 items `setOf(3, 5)` and the natural number of _20_ as threshold the result can be **78** or **93**. This is because: 15 is part of the interim results for both, `3` and `5`.
I believe 15 should be counted only once for the final result, so I am a fan of the final result **78**. In case you'd count 15 twice, you'd end up with a final result of **93**

This debug screenshot based on a kotlin solution hopefully helps to understand my point of view:

screenshot_2018-10-02_08-33-36

All other existing test cases passed in my tests with both implementations. The additional test case I propose shall make sure only one implementation as described above is valid and thus shape the expected overall result.

### DIFF
--- a/exercises/sum-of-multiples/canonical-data.json
+++ b/exercises/sum-of-multiples/canonical-data.json
@@ -39,7 +39,7 @@
       "expected": 23
     },
     {
-      "description": "multiples of 3 or 5 up to 20 with distinct multiples of 3 or 5",
+      "description": "distinct multiples of 3 or 5 up to 20",
       "property": "sum",
       "input": {
         "factors": [3, 5],
@@ -48,7 +48,7 @@
       "expected": 78
     },
     {
-      "description": "multiples of 3 or 5 up to 100",
+      "description": "distinct multiples of 3 or 5 up to 100",
       "property": "sum",
       "input": {
         "factors": [3, 5],
@@ -57,7 +57,7 @@
       "expected": 2318
     },
     {
-      "description": "multiples of 3 or 5 up to 1000",
+      "description": "distinct multiples of 3 or 5 up to 1000",
       "property": "sum",
       "input": {
         "factors": [3, 5],
@@ -75,7 +75,7 @@
       "expected": 51
     },
     {
-      "description": "multiples of 4 or 6 up to 15",
+      "description": "distinct multiples of 4 or 6 up to 15",
       "property": "sum",
       "input": {
         "factors": [4, 6],
@@ -84,7 +84,7 @@
       "expected": 30
     },
     {
-      "description": "multiples of 5, 6 or 8 up to 150",
+      "description": "distinct multiples of 5, 6 or 8 up to 150",
       "property": "sum",
       "input": {
         "factors": [5, 6, 8],
@@ -93,7 +93,7 @@
       "expected": 4419
     },
     {
-      "description": "multiples of 5 or 25 up to 51",
+      "description": "distinct multiples of 5 or 25 up to 51",
       "property": "sum",
       "input": {
         "factors": [5, 25],
@@ -102,7 +102,7 @@
       "expected": 275
     },
     {
-      "description": "multiples of 43 or 47 up to 10000",
+      "description": "distinct multiples of 43 or 47 up to 10000",
       "property": "sum",
       "input": {
         "factors": [43, 47],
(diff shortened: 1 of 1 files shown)


exercises/sum-of-multiples/canonical-data.json
```diff
@@ -38,6 +38,15 @@
       },
       "expected": 23
     },
+    {
+      "description": "multiples of 3 or 5 up to 20",
+      "property": "sum",
+      "input": {
+        "factors": [3, 5],
+        "limit": 10
```
petertseng: if the limit is to be 10, this case is the same as the above case which also has `[3, 5]` up to 10. please clarify how this case is different from the above case.

exercises/sum-of-multiples/canonical-data.json
```diff
@@ -38,6 +38,15 @@
       },
       "expected": 23
     },
+    {
+      "description": "multiples of 3 or 5 up to 20",
```
petertseng: It would be better if the description would say why this case is important to test; that will help both students of Exercism who might want to understand why their implementation does not pass this test.

Here, you want to test the case where a single number is a multiple of two of the given factors. That should be made clear in the description.

The https://github.com/exercism/problem-specifications#test-data-format-canonical-datajson supports this by stating "description should [...] explain [...] why each case is there. For example, what kinds of implementation mistakes might this case help us find?"

petertseng: EBWODP: Incomplete sentence.

> help both students of Exercism who might want to understand why their implementation does not pass this test

"Both" implies two element, but in the sentence Iw rote, I provided one element.

Should add: "and track maintainers who are evaluating whether to keep this test or not." to serve as the second element.

petertseng:
It is true that an implementation that double-counts 15 (15 being a multiple of both 3 and 5) would get 93 instead of 78.

Such an implementation will also get the following answers for existing cases:

4. multiples of 3 or 5 up to 100: Got 2633 instead of 2318
5. multiples of 3 or 5 up to 1000: Got 266333 instead of 233168
7. multiples of 4 or 6 up to 15: Got 42 instead of 30
8. multiples of 5, 6 or 8 up to 150: Got 5343 instead of 4419
9. multiples of 5 or 25 up to 51: Got 350 instead of 275
10. multiples of 43 or 47 up to 10000: Got 2223370 instead of 2203160

However, I trust you when you say that you have an implementation that passes all existing cases except for multiples of 3 or 5 up to 20. Just need more details to describe how it manages to pass the six cases above, but fails the new case added

(For reference, the implementation I described that double-counts numbers can be seen at https://github.com/petertseng/exercism-problem-specifications/blob/verify/exercises/sum-of-multiples/verify.rb#L38-L50 , as can be seen it is marked as `should_fail` and already does fail the above existing tests

NobbZ:
> Given a set of 2 items setOf(3, 5) and the natural number of 20 as threshold the result can be 78 or 93. This is because: 15 is part of the interim results for both, 3 and 5.

The description.md clearly states with exactly the example you gave, that the result should be 78.

Also without looking at your code, I'm pretty sure there was already a test looking for this behaviour, but maybe I am wrong here, as I'm on mobile.

lotharschulz:
@NobbZ @petertseng @rpottsoh: thanks for your reviews.

@petertseng: I did retest with all the existing case for counting distinct factor multiples. Existing test cases cover that indeed as you describe it in https://github.com/exercism/problem-specifications/pull/1351#issuecomment-427194283. I can't explain in retrospective why I missed that.

I agree with you that descriptions could be more detailed e.g. `distinct multiples of 3 or 5 up to 20`

This PR is probably not needed anymore. WDYT?

petertseng:
Hmm, all right. Given that the case of distinct multiples is already covered, and the fact that https://github.com/exercism/docs/blob/master/language-tracks/exercises/anatomy/test-suites.md says test cases should introduce "a single new requirement", it is true that we don't need `factors: [3, 5] limit: 20` case for that.

Is there another reason we might want to add the test? Note that it is in https://github.com/exercism/problem-specifications/blob/master/exercises/sum-of-multiples/description.md

Might consider removing some tests if there is too much overlap between them.

It's true that descriptions of existing test cases could be improved as well; that can be done separately from adding this new case.

lotharschulz:
> Is there another reason we might want to add the test?

not anymore

PR #1352, opened 2018-10-05
triangle: More combinations of non-isosceles triangles

sshine:
A wrong implementation of `is_isosceles` might look like:

    let is_isosceles a b c =
      is_triangle a b c &&
        a = b || b = c || a = c

when the intention was:

    let is_isosceles a b c =
      is_triangle a b c &&
        (a = b || b = c || a = c)

The current test suite doesn't catch that because of order.


sshine:
I considered that, too, and thought it was verbose. I'll merge this then.

PR #1356, opened 2018-10-06
grep: conflicting flags

### LINKED ISSUE #1256: grep: conflicting flags
I just completed the grep exercise, and noticed that the tests didn't cover cases where flags conflict. E.g., what should happen in the following cases?
1. `-n` and `-l`
2. `-x` and `-v`

Some experimentation with grep shows that
1. `-n` takes precedence over `-l`
2. checks that the entire line does not match the pattern

The tests `oneFileNoMatchesVariousFlags()` and `multipleFilesNoMatchesVariousFlags()` contain the flags `-n`, `-l`, `-x`, and `-i`, but don't actually test the situations described above because there are no matches. 

I did make sure that my solution produces the behavior described above (http://exercism.io/submissions/74bcc15ff83b4a00aeb7a919e0f1158d) although certainly I may have missed other conflicting flag combinations.

It might be helpful to either include tests that indicate how to resolve conflicting flags, or the README should indicate that this level of detail is not necessary. What do you think?

(Originally posted on the Java track and was encouraged to move it here.)

yvincent:
resolves #1256.

### DIFF
--- a/exercises/grep/canonical-data.json
+++ b/exercises/grep/canonical-data.json
@@ -230,7 +230,7 @@
             "His wrath pernicious, who ten thousand woes",
             "Caused to Achaia's host, sent many a soul",
             "And Heroes gave (so stood the will of Jove)",
-            "To dogs and to all ravening fowls a prey",
+            "To dogs and to all ravening fowls a prey,",
             "When fierce dispute had separated once",
             "The noble Chief Achilles from the son",
             "Of Atreus, Agamemnon, King of men."
--- a/exercises/grep/canonical-data.json
+++ b/exercises/grep/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "grep",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "comments": [
     " JSON doesn't allow for multi-line strings, so all   ",
     " outputs are presented here as arrays of strings.    ",
@@ -204,6 +204,37 @@
             "files": ["iliad.txt"]
           },
           "expected": []
+        },
+        {
+          "description": "One file, one match, file flag takes precedence over line flag",
+          "property": "grep",
+          "input": {
+            "pattern": "ten",
+            "flags": ["-n", "-l"],
+            "files": ["iliad.txt"]
+          },
+          "expected": [
+            "iliad.txt"
+          ]
+        },
+        {
+          "description": "One file, several matches, inverted and match entire lines flags",
+          "property": "grep",
+          "input": {
+            "pattern": "Illustrious into Ades premature,",
+            "flags": ["-x", "-v"],
+            "files": ["iliad.txt"]
+          },
+          "expected": [
+            "Achilles sing, O Goddess! Peleus' son;",
+            "His wrath pernicious, who ten thousand woes",
+            "Caused to Achaia's host, sent many a soul",
+            "And Heroes gave (so stood the will of Jove)",
+            "To dogs and to all ravening fowls a prey",
+            "When fierce dispute had separated once",
+            "The noble Chief Achilles from the son",
+            "Of Atreus, Agamemnon, King of men."
+          ]
         }
       ]
     },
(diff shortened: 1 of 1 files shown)


exercises/grep/canonical-data.json
```diff
@@ -204,6 +204,37 @@
...
+        {
+          "description": "One file, one match, line number and file flags",
+          "property": "grep",
+          "input": {
+            "pattern": "ten",
+            "flags": ["-n", "-l"],
+            "files": ["iliad.txt"]
+          },
+          "expected": [
+            "2:His wrath pernicious, who ten thousand woes"
```
petertseng: is there a desire to match the behavior of `grep`? If so, it appears that this output should be `iliad.txt`, as I find that `-l` takes precedence over `-n` in following experiments:

```
$ grep -n -l hello <(echo hello)                                                              (10-08 07:58)
/proc/self/fd/11
$ grep -l -n hello <(echo hello)                                                              (10-08 07:58)
/proc/self/fd/11
```

exercises/grep/canonical-data.json
```diff
@@ -204,6 +204,37 @@
...
+          ]
+        },
+        {
+          "description": "One file, several matches, inverted and match entire lines flags",
+          "property": "grep",
+          "input": {
+            "pattern": "Illustrious into Ades premature",
```
petertseng: this should have a comma inside the string. Otherwise, since the flag to match entire lines and the line in the file is `"Illustrious into Ades premature,"` the line will be included in the output.

exercises/grep/canonical-data.json
```diff
@@ -332,6 +363,54 @@
...
+        {
+          "description": "Multiple files, several matches, line number and file flags",
+          "property": "grep",
+          "input": {
+            "pattern": "who",
+            "flags": ["-n", "-l"],
+            "files": ["iliad.txt", "midsummer-night.txt", "paradise-lost.txt"]
+          },
+          "expected": [
+            "2:His wrath pernicious, who ten thousand woes",
```
petertseng: if wanting to match the behaviour of `grep`, it seems this needs to be `["iliad.txt", "midsummer-night.txt", "paradise-lost.txt"]`, as seen in the following experiment

```
$ grep -n -l hello <(echo "hello\nworld") <(echo "a\nhello") <(echo "hello\nhello")
/proc/self/fd/11
/proc/self/fd/12
/proc/self/fd/13
```

exercises/grep/canonical-data.json
```diff
@@ -332,6 +363,54 @@
...
+            "2:Of that Forbidden Tree, whose mortal tast",
+            "8:That Shepherd, who first taught the chosen Seed"
+          ]
+        },
+        {
+          "description": "Multiple files, several matches, inverted and match entire lines flags",
+          "property": "grep",
+          "input": {
+            "pattern": "Illustrious into Ades premature",
```
petertseng: needs a comma, otherwise it does not match the entire line, as in https://github.com/exercism/problem-specifications/pull/1356/files#r223276145

petertseng:
about precedence of `-l` over `-n`, here are the grep versions I tested with that exhibited the behaviour I described:

```
$ grep --version
grep (BSD grep) 2.5.1-FreeBSD
```

and

```
$ grep --version
grep (GNU grep) 3.1
```

any grep versions with any different behaviours?

exercises/grep/canonical-data.json
```diff
@@ -204,6 +204,37 @@
             "files": ["iliad.txt"]
           },
           "expected": []
+        },
+        {
+          "description": "One file, one match, line number and file flags",
```
petertseng: I propose that this description should explicitly say which of the two flags takes precedence.

I propose this because it makes the intent of the test easier to understand for a reader of the description.

exercises/grep/canonical-data.json
yvincent: Yes you are right. The issue describes the opposite behavior and I didn't check. I will update the test cases accordingly.

exercises/grep/canonical-data.json
yvincent: Exactly, well spotted!

exercises/grep/canonical-data.json
yvincent: Yes, as mentioned before I followed the behavior described in the issue which was incorrect. However this should be updated to `["iliad.txt", "paradise-lost.txt"]` since midsummer-night.txt doesn't contain the string "who". I'll update it as well.

exercises/grep/canonical-data.json
yvincent: Yes, I'll fix that one too.

exercises/grep/canonical-data.json
yvincent: It makes perfect sense. I'll update the test cases with an explicit description.

exercises/grep/canonical-data.json
```diff
@@ -204,6 +204,37 @@
...
+            "pattern": "Illustrious into Ades premature,",
+            "flags": ["-x", "-v"],
+            "files": ["iliad.txt"]
+          },
+          "expected": [
+            "Achilles sing, O Goddess! Peleus' son;",
+            "His wrath pernicious, who ten thousand woes",
+            "Caused to Achaia's host, sent many a soul",
+            "And Heroes gave (so stood the will of Jove)",
+            "To dogs and to all ravening fowls a prey",
```
petertseng: missing comma at end of this line

exercises/grep/canonical-data.json
```diff
@@ -332,6 +363,53 @@
...
+        {
+          "description": "Multiple files, several matches, inverted and match entire lines flags",
+          "property": "grep",
+          "input": {
+            "pattern": "Illustrious into Ades premature,",
+            "flags": ["-x", "-v"],
+            "files": ["iliad.txt", "midsummer-night.txt", "paradise-lost.txt"]
+          },
+          "expected": [
+            "Achilles sing, O Goddess! Peleus' son;",
```
petertseng: notice that in the above `multiple files` tests, the lines are each prefixed with the file they appear in.

I believe the same should happen here, as demonstrated in an experiment I share below:

```
$ grep -x -v hello <(echo "hello\nworld") <(echo "a\nhello") <(echo "hello\nhello")
/proc/self/fd/11:world
/proc/self/fd/12:a
```

petertseng:
Yup, this the new cases are matching `grep`'s output now, just need that comma at the end of the `To dogs and to all ravening fowls a prey` line I pointed out now.

yvincent:
Sorry about the missing comma, I've made the change in the wrong branch and then forgot to redo it in the correct branch. It should be all good now. Thanks!

rstockbridge:
Thank you @petertseng and @yvincent for addressing this issue!

PR #1343, opened 2018-09-30
robot-simulator: remove intermediate methods from test data

rpottsoh:
resolves #1279

PR #1357, opened 2018-10-06
robot-simulator: fix repeat case name conflict

rpottsoh:
Case that I added in #1343 introduced scenario described in README.  However the description of the case was already being used by the last case....


jpreese:
Thanks for this @rpottsoh. This actually just came up in the csharp repository (we were running into conflicts when updating the robot-simulator).

PR #1358, opened 2018-10-06
simple-cipher: Add a test for mixed-case key

tekerson:
The [problem specification](https://github.com/exercism/problem-specifications/blob/69a64641e21a50dc8d037e55c1404cf20074ee70/exercises/simple-cipher/description.md#step-3) asserts that "the key contains only lowercase letters", but the test data only tests the key is "not all uppercase".

I have already added this test to the [TypeScript](https://github.com/exercism/typescript/pull/219) and [JavaScript](https://github.com/exercism/javascript/pull/491) tracks. It was suggested it should be added to the canonical test data too.

### DIFF
--- a/exercises/simple-cipher/canonical-data.json
+++ b/exercises/simple-cipher/canonical-data.json
@@ -127,6 +127,14 @@
           "expected": { "error": "Bad key" }
         },
         {
+          "description": "Throws an error with a mixed-case key",
+          "property": "new",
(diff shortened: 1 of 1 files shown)


rpottsoh:
Closing in favor of #1316.  Thanks @tekerson for submitting this PR, however #1316 already has a test case that covers mixed case input keys.

PR #1365, opened 2018-10-11
pythagorean-triplet: Add missing constraints to description

TheBestJohn:
Added missing constraint to the Pythagorean triplet problem that is outlined in the linked Project Euler problem.

### DIFF
--- a/exercises/pythagorean-triplet/description.md
+++ b/exercises/pythagorean-triplet/description.md
@@ -5,6 +5,12 @@ which,
 a**2 + b**2 = c**2
 ```
 
+and such that,
+
+```text
+a < b < c
+```
+
 For example,
 
 ```text


coriolinus:
I now plan to wait until Monday the 15th. This is to give other maintainers the chance to look at this and make comments and request changes if desired. If there are no objections, I plan to merge this on that date. I do not object if another maintainer wants to merge this before then.

TheBestJohn:
> I now plan to wait until Monday the 15th. This is to give other maintainers the chance to look at this and make comments and request changes if desired. If there are no objections, I plan to merge this on that date. I do not object if another maintainer wants to merge this before then.

Ah yes I should have stated that I looked into the data and found no conflicts when submitting the pull request. 

also @rpottsoh no problem :) I just figured that it would definitely affect the performance of generated answers. Every result I bruteforce where a>b, or b>c is a calculation that didn't need to be done.

As @petertseng says a*b*c order does not matter however if we were tasked with identifying `a` `b` and `c` then `a` and `b` would have 2 answers as currently stated.

PR #1366, opened 2018-10-12
scale-generator: update test properties: one property per signature

coriolinus:
Many tracks use automatic or semiautomatic tools to assist in
implementing an exercise for that track. These tools are very
useful, but depend on certain properties of the canonical data.

One such property is that all tests with the same "property"
field should have the same signature: the same keys in the
"input" object and the same type of "expected" object.

The scale-generator exercise used to violate that property:
all tests used the "pitches" property. This PR brings it into
compliance by separating the "chromatic" tests from the
"interval" tests. These new names, in addition to aiding
test generators, are more descriptive than "pitches".

### DIFF
--- a/exercises/scale-generator/canonical-data.json
+++ b/exercises/scale-generator/canonical-data.json
@@ -11,7 +11,7 @@
       "cases": [
         {
           "description": "Chromatic scale with sharps",
-          "property": "pitches",
+          "property": "chromatic",
           "input": {
             "tonic": "C"
           },
@@ -19,7 +19,7 @@
         },
         {
           "description": "Chromatic scale with flats",
-          "property": "pitches",
+          "property": "chromatic",
           "input": {
             "tonic": "F"
           },
@@ -39,7 +39,7 @@
           "comments": [
             "The simplest major scale, with no sharps or flats."
           ],
-          "property": "pitches",
+          "property": "interval",
           "input": {
             "tonic": "C",
             "intervals": "MMmMMMm"
@@ -48,7 +48,7 @@
         },
         {
           "description": "Major scale with sharps",
-          "property": "pitches",
+          "property": "interval",
           "input": {
             "tonic": "G",
             "intervals": "MMmMMMm"
@@ -57,7 +57,7 @@
         },
         {
           "description": "Major scale with flats",
-          "property": "pitches",
+          "property": "interval",
           "input": {
             "tonic": "F",
             "intervals": "MMmMMMm"
@@ -66,7 +66,7 @@
         },
         {
           "description": "Minor scale with sharps",
-          "property": "pitches",
+          "property": "interval",
           "input": {
             "tonic": "f#",
             "intervals": "MmMMmMM"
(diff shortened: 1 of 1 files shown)


coriolinus:
I interpret the presence of the chromatic tests as a test of student understanding of music theory: chromatic tests are equivalent to interval tests with an interval of `"mmmmmmmmmmmm"`. 

As a general policy I'm not a fan of exercises requiring external domain knowledge of students. At the same time, I recently implemented this exercise for the Rust track and didn't even notice that it required external domain knowledge.

I suggest that we pick from among the following options:

- decide that the knowledge of what constitutes a chromatic scale is general knowledge we can expect of the population, and therefore no action is required
- decide that the knowledge of what constitutes a chromatic scale is easily googleable, and therefore no action is required
- decide that knowledge of chromatic scales is domain knowledge, and therefore that the chromatic-property tests should be removed

rpottsoh:
I am not fully versed on this exercise so bare with me.  `Interval` sounds to vague to me.  How do I know if we are dealing with whole steps or half steps.  Chromatic is clear to me, always half steps I believe.  Diatonic is the only term I find that seems to be counter to chromatic.  However, a diatonic scale can consist of a combination of whole and half steps (for 7 steps in all, vs 12 for chromatic).  I don't know that I am really helping, more just "thinking out loud".

coriolinus:
The interval defines whether we are talking about whole steps, half steps, or extended steps. In the notation of the canonical data, `m` indicates a half step, `M` indicates a whole step, and `A` indicates an extended (1.5) step. A scale consists of some number of intervals such that the sum of steps equals 12.

petertseng:
> I interpret the presence of the chromatic tests as a test of student understanding of music theory: chromatic tests are equivalent to interval tests with an interval of "mmmmmmmmmmmm".

I agree about the equivalence. I want to add that another possibility for why the chromatic tests are included is to make sure that the function under test can actually produce all the possible notes. This property is possible to test piecemeal with a bunch of other tests, but it's harder to verify that, in sum, all notes had been tested.

Therefore, I propose that if both of the following statements are true:

* We desire to remove the chromatic tests in their current form
* We agree that it's still important to test that all notes can be generated

then we should also add tests that explicitly specify intervals `mmmmmmmmmmmm`.

petertseng:
> I interpret the presence of the chromatic tests as a test of student understanding of music theory: chromatic tests are equivalent to interval tests with an interval of "mmmmmmmmmmmm".

I'd also consider the possibility of:

Explicitly state in the README the need to be able to produce a chromatic scale.

I do think that the README already provides an explanation of what a chromatic scale is, though there's always the possibility that it could be clearer.

coriolinus:
Ok. I have updated the version; according to the version policy, changing the name of a property is a major change.

I have created https://github.com/exercism/problem-specifications/issues/1372 so we can continue discussion about what to do about chromatic tests there.

Nobody has yet formally approved this PR, so I will not yet merge it. On Thursday, it will have been open for seven days. If nobody objects before then, I intend to merge this then. I would be grateful if someone else approved and/or merged this before then.

PR #1374, opened 2018-10-17
Trying to simplify the Twofer exercise specification

uzilan:
As a mentor, I often see solutions of two-fer where the mentee simply didn't understand the exercise. Sometimes they just check for the names **Alice** and **Bob** and return a string with these names. If the given name is neither, they return the string with **you** instead. See for example here: https://exercism.io/mentor/solutions/93be9a7ec7014a38920063a683963ffa

I believe this problem stem from the fact that the description is unclear. This PR contains my suggestion for making it a bit clearer by being more thorough and by giving some examples.

### DIFF
--- a/exercises/two-fer/description.md
+++ b/exercises/two-fer/description.md
@@ -1,11 +1,24 @@
 `Two-fer` or `2-fer` is short for two for one. One for you and one for me.
 
+Given a name, return a string with the message:
+
 ```text
-"One for X, one for me."
+One for X, one for me.
 ```
 
-When X is a name or "you".
+Where X is the given name.
 
-If the given name is "Alice", the result should be "One for Alice, one for me."
-If no name is given, the result should be "One for you, one for me."
+However, if the name is missing, return the string:
 
+```text
+One for you, one for me.
+```
+
+Here are some examples:
+
+|Name    | String to return 
+|:------:|:-----------------: 
+|Alice   | One for Alice, one for me. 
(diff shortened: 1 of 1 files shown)


rpottsoh:
Only mentors for the track you considering will be able to view the link you have posted.  Please consider posting the code into a comment so we all can see it.

rpottsoh:
I would say that this change clarifies the specification, but doesn't simplify the explanation.  It was pretty simple already.  :smile:

uzilan:
Here's the example solution I was referring to:
```kotlin
class Twofer {
    String twofer(String name) {
        if (name == "Alice") {
    	    name = "Alice";
        } else if (name == "Bob") {
    	    name = "Bob";
        } else {
    	    name = "you";
        }
    	return "One for " + name + ", one for me.";
    }   
}
```

uzilan:
Thank you too. So, how do the different tracks get the updated version? Do we need a PR in every track?

rpottsoh:
It is up to each track to update their README for this exercise.  Hopefully the track maintainers are monitoring this REPO and have noticed this PR merging.  The tracks applying the changes you have submitted is completely voluntary.  The tracks are not required to follow what happens in this repo to the letter, but most do.  Others may add their own twists to the exercises that work well with their particular language.

If there is a particular track where you would like to see you changes incorporated I suggest creating an issue in that track's repo and site this PR in your comment.

PR #1375, opened 2018-10-21
Acronym: New test case for apostrophe

alanvardy:
The current solution for Acronym doesn't account for apostrophes, it is a more rare case for acronyms but an edge case nonetheless. Is this useful?

I added a single test case and incremented the version number.

Thank you!

### DIFF
--- a/exercises/acronym/canonical-data.json
+++ b/exercises/acronym/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "acronym",
-  "version": "1.5.1",
+  "version": "1.6.0",
   "cases": [
     {
       "description": "Abbreviate a phrase",


alanvardy:
Corrected to 1.6.0! Thank you.

jpreese:
My only concern with this would be that since there's already a case for a space and a case for a hyphen, adding a third case doesn't seem to bring a lot of value in my opinion. 

Adding a second case forces the developer to do more than simply calling .split() with a single delimiter. Once you can do two, adding 3..4.. etc shouldn't be that different.

alanvardy:
I am bring this forward after an issue encountered on the Ruby track. When using the regex `/\b\w/` or `/\b[a-zA-Z]/` it would also grab the s at the end of Haley's etc. Because Ruby involves the use of scan() and regex rather than split() I think this could be useful.

F3PiX:
Thanks @alanvardy for your suggestion!

However. Can we keep this on hold? We're in the process of rearranging the Ruby track, and adding Acronym as a core exercise is part of that. We're monitoring how it works. I'll add this issue to the evaluation. 

That said, I agree with @jpreese that adding one extra case doesn't seem to bring much more value to the goal of the exercise. The extra case for the hyphen makes it clear that a simple `split` won't do. Another extra case of the apostroph seems a bit arbitrary - why not underscores, quotes and any other edge cases?

alanvardy:
Absolutely!

I perused through an online acronym dictionary and found plenty of hyphens, the occasional apostrophe and no underscores or quotes. I suppose I am thinking of what valid inputs are not covered, rather than being concerned with all inputs.

Nonetheless I thank you all for your time and the efforts you are all making! I will keep my fingers crossed for the review.

F3PiX:
@rpottsoh Could this be interesting for other tracks? 
If not, I'd rather close this PR. The issue is added to my evaluation notes of Acronym in Ruby, so we can reopen it if needed. 
Is that okay with you?

rpottsoh:
> @rpottsoh Could this be interesting for other tracks?

I think this is an interesting edge case to add.  In other cases that contain punctuation, the punctuation is essentially "treated" like a space.  The `'` cannot be treated in this manner as it is not separating two words.  For that reason I think it is worthwhile to test for apostrophe for the very reason I think @alanvardy is considering.

There is no reason to rush to merge this.  If you would like to wait @F3PiX, that is fine.  Just signal when ready.

If the group consensus is to close then I wont stand in the way.  I'm just a participant here like the rest of you. 😄

rpottsoh:
> @rpottsoh Could this be interesting for other tracks?

I think this is an interesting edge case to add.  In other cases that contain punctuation, the punctuation is essentially "treated" like a space.  The `'` cannot be treated in this manner as it is not separating two words.  For that reason I think it is worthwhile to test for apostrophe for the very reason I think @alanvardy is considering.

There is no reason to rush to merge this.  If you would like to wait @F3PiX, that is fine.  Just signal when ready.

If the group consensus is to close then I wont stand in the way.  I'm just a participant here like the rest of you. 😄

rpottsoh:
> @rpottsoh Could this be interesting for other tracks?

I think this is an interesting edge case to add.  In other cases that contain punctuation, the punctuation is essentially "treated" like a space.  The `'` cannot be treated in this manner as it is not separating two words.  For that reason I think it is worthwhile to test for apostrophe for the very reason I think @alanvardy is considering.

There is no reason to rush to merge this.  If you would like to wait @F3PiX, that is fine.  Just signal when ready.

If the group consensus is to close then I wont stand in the way.  I'm just a participant here like the rest of you. 😄

F3PiX:
Thanks all for your thoughts! 
I'm closing this for now, and we can reopen it if need to.

F3PiX:
Thanks all for your thoughts! 
I'm closing this for now, and we can reopen it if need to.

F3PiX:
Thanks all for your thoughts! 
I'm closing this for now, and we can reopen it if need to.

F3PiX:
Thanks all for your thoughts! 
I'm closing this for now, and we can reopen it if need to.

F3PiX:
Thanks all for your thoughts! 
I'm closing this for now, and we can reopen it if need to.

rpottsoh:
reopening appears to no be possible.  The repo that opened this PR has been deleted.

alanvardy:
This is my bad, when Github went down I was trying to create a new repository without success. I deleted a couple of repositories in case I was over a limit. One of those was this fork ;(. Wine and Github don't seem to mix.

How can I fix this?

rpottsoh:
I am not sure what your options might be.  Restoring a deleted branch is documented [here](https://gist.github.com/jbgo/1944238) and in other places I am sure.  Not sure how to restore a deleted repo.

In lieu of restoring your fork and restoring the ability to reopen this PR, you are probably faced with opening another PR.  

@F3PiX closed this PR.  I think it could have remained open until it was decided that the change to the test data was truly not desired.  I don't fully understand why the activities of the Ruby track are of consequence to what happens in this repo.   Tracks are allowed to update their exercise test suites and READMEs when they see fit to do so.

F3PiX:
I don't mind reopening, but it's gone.

rpottsoh:
@alanvardy got a little too hasty the other day during the GitHub outage and ended deleting his fork.  He's more than welcome to open another PR.   I think there is merit to the case he proposed.

alanvardy:
I'll submit another identical (and less transient!) PR later today and will reference in this thread. Thank you all for all your consideration!

alanvardy:
New PR created! #1377 .

PR #1377, opened 2018-10-23
Acronym: New test case for apostrophe

alanvardy:
This is a redo of PR #1375, to be held until later review!

The current solution for Acronym doesn't account for apostrophes, it is a more rare case for acronyms but an edge case nonetheless. Is this useful?

I added a single test case and incremented the version number.

### DIFF
--- a/exercises/acronym/canonical-data.json
+++ b/exercises/acronym/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "acronym",
-  "version": "1.5.0",
+  "version": "1.6.0",
   "cases": [
     {
       "description": "Abbreviate a phrase",
@@ -60,6 +60,14 @@
             "phrase": "Something - I made up from thin air"
           },
           "expected": "SIMUFTA"
+        },
+        {
+          "description": "apostrophes",
+          "property": "abbreviate",
+          "input": {
+            "phrase": "Haley's Comet"
+          },
+          "expected": "HC"
         }
       ]
     }
--- a/exercises/acronym/canonical-data.json
+++ b/exercises/acronym/canonical-data.json
@@ -65,7 +65,7 @@
           "description": "apostrophes",
           "property": "abbreviate",
           "input": {
-            "phrase": "Haley's Comet"
+            "phrase": "Halley's Comet"
           },
           "expected": "HC"
         }


exercises/acronym/canonical-data.json
```diff
@@ -60,6 +60,14 @@
             "phrase": "Something - I made up from thin air"
           },
           "expected": "SIMUFTA"
+        },
+        {
+          "description": "apostrophes",
+          "property": "abbreviate",
+          "input": {
+            "phrase": "Haley's Comet"
```
petertseng: If I may -

If this is to be representative of a real-world phrase, then I would suggest using `Halley's Comet` here with the extra L. 

However, if all resemblances to real-world phenomena are purely coincidental, then as you were.

rpottsoh:
Thanks for fixing spelling mistake.  Good eyes @petertseng.

alanvardy:
Wow, I'm shocked that I missed that. The website I orginally looked it up on misspelled Halley's and I never checked any other sources. Thanks @petertseng

rpottsoh:
@F3PiX let me know when it is OK to merge this.  Recall you originally wanted to hold off due to some organizational changes taking place in the Ruby track.

rpottsoh:
I would just like to add that some students like to use regex, like this `[A-Z]+[a-z]*|[a-z]+` for example to "find the words" in the input phrase.  Apostrophes trip'em up.

F3PiX:
I'm afraid I was confusing the Ruby tests with the canonical data, so the Ruby restructuring has nothing to do with merging this. Sorry for the confusing :-) 

> some students like to use regex, like this [A-Z]+[a-z]*|[a-z]+ for example to "find the words" 

That's already taken care of by the test in line 44, with the hyphen as a non-space, non-word character?

rpottsoh:
> That's already taken care of by the test in line 44, with the hyphen as a non-space, non-word character?

If you apply that regex to `Halley's Comet` it results in `Halley s Comet`.  Some students solutions then return `HSC` as the acronym instead of the correct acronym of `HC`.

coriolinus:
It should be "OMFR", because the glottal stop apostrophe in "O'Malley" is
different from the possessive apostrophe at its end.

That's a bit advanced for this exercise, though; I'd recommend against
adding that test case.

rpottsoh:
> That's a bit advanced for this exercise, though; I'd recommend against

I agree.  It came to mind while I was thinking about this exercise and I found it interesting.

rpottsoh:
As long as there are no objections or is no further discussion I will likely merge this PR tomorrow (Sunday).  If someone else decides to merge this in the mean time, please squash while merging.

rpottsoh:
Thanks for working on this @alanvardy. 👍

PR #1378, opened 2018-10-24
high-scores: New exercise

pgaspar:
An introductory Array-focused exercise me and @F3PiX worked on.

In the Ruby track, we're lacking an easy exercise with an array as a list and some list manipulations.  
This exercise is meant to fill that gap, and may be interesting for other tracks to implement as well.
We paid extra attention to the sequence of the tests, so the code is building up nicely, and at the end there should be some nice mentoring opportunities left.
It's set up for incremental progress, where students can both learn new things and implement lessons learned;  in the Ruby track, this will be the first exercise where they work with multiple methods, and the string interpolation that they know from TwoFer is here too, but now they have to prevent double spaces.   

We created the Ruby implementation (https://github.com/exercism/ruby/pull/886) at the same time, so we were probably influenced by it when creating this generic specification - please help us improve it :+1:

We've also submitted a logo request [here](https://github.com/exercism/website-icons/issues/15).

### DIFF
--- a/exercises/high-scores/canonical-data.json
+++ b/exercises/high-scores/canonical-data.json
@@ -1,0 +1,107 @@
+{
+  "exercise": "high-scores",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "List of scores",
+      "property": "scores",
+      "input": {
+        "scores": [30, 50, 20, 70]
+      },
+      "expected": [30, 50, 20, 70]
+    },
+    {
+      "description": "Latest score",
+      "property": "latest",
+      "input": {
+        "scores": [100, 0, 90, 30]
+      },
+      "expected": 30
+    },
+    {
+      "description": "Highest score",
+      "property": "highest",
+      "input": {
+        "scores": [40, 100, 70]
+      },
+      "expected": 100
+    },
+    {
+      "description": "Top 3 scores",
+      "cases": [
+        {
+          "description": "Personal bests",
+          "property": "top",
+          "input": {
+            "scores": [50, 30, 10]
+          },
+          "expected": [50, 30, 10]
+        },
+        {
+          "description": "Personal bests highest to lowest",
+          "property": "top",
+          "input": {
+            "scores": [20, 10, 30]
+          },
+          "expected": [30, 20, 10]
+        },
+        {
+          "description": "Personal bests when there is a tie",
+          "property": "top",
+          "input": {
+            "scores": [40, 20, 40, 30]
+          },
+          "expected": [40, 40, 30]
+        },
+        {
+          "description": "Personal bests when there are less than 3",
+          "property": "top",
+          "input": {
+            "scores": [30, 70]
+          },
+          "expected": [70, 30]
+        },
+        {
+          "description": "Personal bests when there is only one",
+          "property": "top",
+          "input": {
+            "scores": [40]
+          },
+          "expected": [40]
+        },
+        {
+          "description": "Personal bests from a long list",
+          "property": "top",
+          "input": {
+            "scores": [10, 30, 90, 30, 100, 20, 10, 0, 30, 40, 40, 70, 70]
+          },
+          "expected": [100, 90, 70]
+        }
+      ]
+    },
+    {
+      "description": "Message for new personal best",
+      "property": "report",
+      "input": {
+        "scores": [20, 40, 0, 30, 70]
+      },
+      "expected": "Your latest score was 70. That's your personal best!"
+    },
+    {
+      "description": "Message when latest score is not the highest score",
+      "property": "report",
+      "input": {
+        "scores": [20, 100, 0, 30, 70]
+      },
+      "expected": "Your latest score was 70. That's 30 short of your personal best!"
+    },
+    {
+      "description": "Message for repeated personal best",
+      "property": "report",
+      "input": {
+        "scores": [20, 70, 50, 70, 30]
+      },
+      "expected": "Your latest score was 30. That's 40 short of your personal best!"
+    }
+  ]
+}
--- a/exercises/high-scores/description.md
+++ b/exercises/high-scores/description.md
@@ -1,0 +1,3 @@
+Manage a game player's High Score list.
+
+Your task is to build a high-score component of the classic Frogger game, one of the highest selling and addictive games of all time, and a classic of the arcade era. Your task is to write methods that return the highest score from the list, the last added score, the three highest scores, and a report on the difference between the last and the highest scores. 
--- a/exercises/high-scores/metadata.yml
+++ b/exercises/high-scores/metadata.yml
@@ -1,0 +1,3 @@
+---
+blurb: "Manage a player's High Score list"
+source: "Tribute to the eighties' arcade game Frogger"


ErikSchierboom:
Love the idea for a simpler array exercise! This is something that the C# also lacks.

exercises/high-scores/description.md
```diff
@@ -0,0 +1,3 @@
+Manage a game player's High Score list.
+
+The high score list knows about the highest scores, the last added score and other info useful to a game player.
```
Insti: I'd like to see more description of what the task actually is.

exercises/high-scores/canonical-data.json
```diff
@@ -0,0 +1,91 @@
...
+        }
+      ]
+    },
+    {
+      "description": "Returns the goal when highest is latest",
+      "property": "goal",
+      "input": {
+        "scores": [2, 4, 0, 3, 7]
+      },
+      "expected": "Your latest score was 7. That's your personal best!"
```
Insti: Something about this bugs me. I think it is that this property returns a string while the others are arrays.

Insti: Maybe if the property was something like `message` or `goal_report` that made it a bit clearer that we will be expecting a string rather than a numeric value.

exercises/high-scores/canonical-data.json
```diff
@@ -0,0 +1,91 @@
...
+  "exercise": "high-scores",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "Returns the list of scores",
+      "property": "scores",
+      "input": {
+        "scores": [3, 5, 2, 7]
+      },
+      "expected": [3, 5, 2, 7]
```
Insti: These scores seem low for 'high' scores.
Perhaps you could help this by just putting a line in the readme that says something like "Scores are usually in millions but we're leaving that bit off." or "It's really hard to get Frogs across the road so 7 is a good score!"

exercises/high-scores/canonical-data.json
```diff
@@ -0,0 +1,91 @@
...
+          "expected": [5, 3, 1]
+        },
+        {
+          "description": "Returns personal bests highest to lowest",
+          "property": "top",
+          "input": {
+            "scores": [2, 1, 3]
+          },
+          "expected": [3, 2, 1]
+        },
```
Insti: Do we need a test with tied scores?

exercises/high-scores/canonical-data.json
Insti: `encouraging_message`?

exercises/high-scores/canonical-data.json
```diff
@@ -0,0 +1,91 @@
...
+      "expected": "Your latest score was 7. That's your personal best!"
+    },
+    {
+      "description": "Returns the goal when not highest score",
+      "property": "goal",
+      "input": {
+        "scores": [2, 10, 0, 3, 7]
+      },
+      "expected": "Your latest score was 7. That's 3 short of your personal best!"
+    }
```
Insti: Might be nice to make a distinction between a new personal best and a tied personal best.

exercises/high-scores/canonical-data.json
Insti: Does it even need to be a string?
Could we make it a boolean? `personal_best`

What is it you're trying to help teach?

exercises/high-scores/description.md
F3PiX: Thanks @insti, for the review, much appreciated! 
How about this for the description:
```
The high score list knows about the highest scores, the last added score and other info useful to a game player.
Your task is to write methods that return the highest score from the list, the last added score, the three highest scores, and a report on the difference between the last and the highest scores. 
```

exercises/high-scores/canonical-data.json
pgaspar: Thanks for the review @Insti :+1:.

As for your question: maybe. I assume you're asking because without it those cases will be ambiguous for students, right? If that's the case, yes we can specify that :+1:

exercises/high-scores/canonical-data.json
pgaspar: I had a good laugh with "It's really hard to get Frogs across the road so 7 is a good score!" 😄 However the exercise only refers Frogger in the `source` bit, which is relatively hidden.

exercises/high-scores/description.md
Insti: It's a step in the right direction.

It's nice to provide a little more context and "color"

Have look at the description for exercises like `book-store` or `kindergarten-garden` or `two-bucket`

exercises/high-scores/canonical-data.json
pgaspar: Sounds good, thank you. FYI we're planning on returning `[4, 4, 3]` for `[4, 2, 4, 3]`. So yeah, it's a bit prescriptive to not using sets, but hopefully that's ok.

exercises/high-scores/canonical-data.json
pgaspar: We're going to add a new test to specify what happens in case of a tie :+1: We don't want to add too many cases here so it's just going to return the same as the personal best.

exercises/high-scores/canonical-data.json
Insti: > I'd rather not add it to the Readme.

That's fine :heart:
(It is also fine if you don't want any of the "color" I was suggesting in another comment and and just want to retain a short to the point readme.)

exercises/high-scores/canonical-data.json
Insti: What about making all the return values interpolated strings?

pgaspar:
> In Frogger, every step is worth 10points

I completely ignored that, but that's a great catch - thanks @iHiD. I'll add a zero for now.

As for the idea of making the Frogger theme more prominent and the exercise extension, I'll defer to @F3PiX who originally thought up the exercise 👍

They're interesting ideas.

exercises/high-scores/canonical-data.json
pgaspar: This has now been added. Check "Returns personal bests when there is a tie".

exercises/high-scores/canonical-data.json
pgaspar: I've pushed the new test. Check "Returns report when there is a tie", please.

Insti:
None of my suggestions should be considered blockers. It's would be fine to merge this and address tweaks through future pull requests.

exercises/high-scores/canonical-data.json
```diff
@@ -80,24 +80,24 @@
...
     },
     {
-      "description": "Returns the goal when highest is latest",
-      "property": "goal",
+      "description": "Returns report when highest is latest",
+      "property": "report",
       "input": {
         "scores": [2, 4, 0, 3, 7]
       },
       "expected": "Your latest score was 7. That's your personal best!"
```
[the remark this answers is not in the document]
pgaspar: I believe we wanted to keep the number of possibilities on the `report` method relatively short (as in 2 😅). @F3PiX may have more context to that.

exercises/high-scores/canonical-data.json
```diff
@@ -80,24 +80,24 @@
...
     },
     {
-      "description": "Returns the goal when there is a tie",
-      "property": "goal",
+      "description": "Returns report when there is a tie",
```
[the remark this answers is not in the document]
pgaspar: Interesting - would you have the descriptions just like that or with "Returns" before? I looked at other examples in the repo and saw both styles. I was following the "spec" style, in which typically there's an "it" automatically added before each test. So:

> it returns message when latest score is the high score

But it probably makes more sense to go with the "test" style:

> test message when latest score is the high score

Is there guidance in the documentation as to which style we should use? From the example in the readme it seems it's more of the latter.

Does it make sense for you to remove the "Returns"s in other descriptions too?

exercises/high-scores/canonical-data.json
```diff
@@ -0,0 +1,107 @@
...
+    {
+      "description": "Returns report when there is a tie",
+      "property": "report",
+      "input": {
+        "scores": [20, 40, 70, 30, 70]
+      },
+      "expected": "Your latest score was 70. That's your personal best!"
+    },
+    {
+      "description": "Returns report when not highest score",
```
ErikSchierboom: To me, the description could maybe be a bit clearer:

```suggestion
      "description": "Returns report when highest is not latest"",
```

exercises/high-scores/canonical-data.json
```diff
@@ -0,0 +1,107 @@
...
+    {
+      "description": "Returns report when highest is latest",
+      "property": "report",
+      "input": {
+        "scores": [20, 40, 0, 30, 70]
+      },
+      "expected": "Your latest score was 70. That's your personal best!"
+    },
+    {
+      "description": "Returns report when there is a tie",
```
ErikSchierboom: I think that the current input should probably be changed, as the way it is now, I think most implementations that pass the previous test case ("Returns report when highest is latest) will also automatically pass this test. So I would suggest to either remove this test (my preference), or else change the input to have the tied high score not being the last value.

exercises/high-scores/canonical-data.json
pgaspar: Thanks for the suggestion!

F3PiX:
@ihid Great idea for the string formatting. I'm going to use that in a next exercise, that's slightly more advanced, and where it's a really really great feature, that just finishes it off 😍 I feel it's out of scope, at least stretching the scope, of Frogger. 

@insti @ihid I added some extra Frogger colour to the description, without stretching it too much, because 'hi score' is actually the only thing this exercise and Frogger has in common.

exercises/high-scores/canonical-data.json
pgaspar: > most implementations that pass the previous test case will also automatically pass this test.

I understand @ErikSchierboom's point here and I think it's a good point.

> or else change the input to have the tied high score not being the last value

I'll do this.

exercises/high-scores/canonical-data.json
```diff
@@ -0,0 +1,107 @@
...
+    {
+      "description": "List of scores",
+      "property": "scores",
+      "input": {
+        "scores": [30, 50, 20, 70]
+      },
+      "expected": [30, 50, 20, 70]
+    },
+    {
+      "description": "Lastest score",
```
ErikSchierboom: Lastest -> Latest

pgaspar:
Folks, I've squashed everything down to a single commit. Thank you for your reviews so far!

F3PiX:
Thanks all for your inputs. Can we merge this?

Insti:
I can merge it. Thanks everyone :heart:

F3PiX:
THANKS All !!

pgaspar:
Thank you!

PR #1376, opened 2018-10-22
fix broken link to Code of Conduct document

rpottsoh:
fixes #1364

PR #1381, opened 2018-10-27
Update URL for Exercism Code of Conduct link.

houhoulis:
See https://github.com/exercism/exercism/issues/4441.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,7 +64,7 @@ email Katrina at kytrinyx@exercism.io.
 ## Code of Conduct
 
 Help us keep Exercism welcoming. Please read and abide by the
-[Code of Conduct](https://github.com/exercism/exercism.io/blob/master/CODE_OF_CONDUCT.md).
+[Code of Conduct](https://exercism.io/code-of-conduct).
 
 ## Overview
 


CONTRIBUTING.md
```diff
@@ -64,7 +64,7 @@ email Katrina at kytrinyx@exercism.io.
 ## Code of Conduct
 
 Help us keep Exercism welcoming. Please read and abide by the
-[Code of Conduct](https://github.com/exercism/exercism.io/blob/master/CODE_OF_CONDUCT.md).
+[Code of Conduct](https://exercism.io/code-of-conduct).
```
rpottsoh: I am not sure this is the best way to link code of conduct to this document.  I am not sure  that CONTRIBUTING.md is accessible via the site so it is probably not a good idea to link to it (the site).  Please see #1376.

rpottsoh: Typically one document links to another.  I don't think it is typical for a document to link directly to the Exercism site.
My main purpose in bringing this up is that I don't think this repo's CONTRIBUTING.md is accessible via the Exercism site so it makes more sense to me for the document to link directly with the code of conduct document in the exercism/website-copy repo.  I am not sure this was considered.  If it is still desired that this document should be linked directly to https://exercism.io/code-of-conduct then #1376 needs to be closed in favor of #1381.  I realize it is the same document and just two different links to take the user to the same place ultimately.

Perhaps stated another way.  If I am navigating a repo within GitHub I am not expecting to click a link in an Exercism md file to find myself in the Exercism site, I expect to find myself still in GitHub in the repo where the md files lives that I clicked the link on a moment ago.  Conversely if I am in the Exercism site and click a link I do not expect to wind up in GitHub, unless the text of the link I click eludes to that.

This is just a matter of perception for me, nothing more.

Thoughts @iHiD?

iHiD: @rpottsoh Thanks for your thoughts.

So the canonical location for this is https://exercism.io/code-of-confuct so I'd like everything to point there. If that ever does change then we have 302 control over that URL. However, if we change our GitHub structure (which is quite likely over time - this situation being an example) then we have to update things everywhere and can't 302, which is a pain.

So I would rather have things like to the exercism.io website for the Code of Conduct.

Does that make sense?

rpottsoh:
Yes, makes sense.  I am going to go ahead and merge this.

rpottsoh:
Thanks for working on this @houhoulis.

houhoulis:
No problem!  Thanks, @rpottsoh

PR #1334, opened 2018-09-21
wordy: implement "error" object

rpottsoh:
per #1311 proper error handling object incorporated into last two test cases.

PR #1383, opened 2018-11-04
wordy: test that various syntax errors are in fact syntax errors

petertseng:
By explicitly testing these we are making sure that the implementation
fails in an appropriate way, rather than failing in an inappropriate way
or erroneously giving an answer.

For example, in languages using an optional for their answer, ensuring
that [`None` is returned instead of panicking/aborting][1].

[1]: https://github.com/exercism/rust/pull/704#discussion_r230593072

This is beneficial because the problem is partially about building a
parser (even if very rudimentary), and it's common for parsers to have
to deal with errors like this. We may otherwise see some unfortunately
fragile parsers.

~~Intentional merge conflict for the same reason as #1382~~

### DIFF
--- a/exercises/wordy/canonical-data.json
+++ b/exercises/wordy/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "wordy",
-  "version": "1.2.0",
+  "version": "1.3.0",
   "comments": [
     "The tests that expect an 'error' should be implemented to raise",
     "an error, or indicate a failure. Implement this in a way that",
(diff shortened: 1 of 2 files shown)


exercises/wordy/description.md
```diff
@@ -41,6 +41,14 @@ left-to-right, _ignoring the typical order of operations._
 
 15  (i.e. not 9)
 
+## Iteration 4 — Errors
+
+The parser should reject:
+
+* Unsupported operations ("What is 52 cubed?")
+* Non-math questions ("Who is the President of the United States")
```
petertseng: Even if this PR is rejected, I'd suggest adding these lines in any case!

exercises/wordy/canonical-data.json
```diff
@@ -134,6 +134,38 @@
         "question": "Who is the President of the United States?"
       },
       "expected": false
+    },
+    {
+      "description": "reject two operations in a row",
+      "property": "answer",
+      "input": {
+        "question": "What is 1 plus plus 2?"
```
petertseng: let's ignore the fact that in some languages, `1 + + 2` is a valid expression that evaluates to `3`. The reason for the validity is that the second `+` sign is a unary plus. We are probably assuming in these word problems that the `plus` refers only to the binary operation.

exercises/wordy/canonical-data.json
```diff
@@ -134,6 +134,38 @@
         "question": "Who is the President of the United States?"
       },
       "expected": false
```
coriolinus: I think all instances of `"expected": false` should be converted to syntax errors in this PR.

petertseng: :+1: 

will happen when rebasing on top of https://github.com/exercism/problem-specifications/pull/1334

petertseng:
I'm going to add another case that I think will be useful to test: `"What is 1 plus?"`

petertseng:
With regard to https://github.com/exercism/problem-specifications/issues/902 - one very good argument against testing invalid inputs there was that we can imagine there is an additional layer above the problem that takes on the job of filtering out any invalid inputs.

For problems like `nucleotide-count` this makes sense because this filtering can be done sort of independently of the meat of the problem.

For a problem like `wordy`, it seems to me that something that filters for invalid inputs would probably need to parse the input, which is part of the meat of the problem here. Thus, to test these syntax errors is consistent with the ideas of https://github.com/exercism/problem-specifications/issues/902.

petertseng:
Since I have an approval, if no further comments are forthcoming, one should expect me to merge this when 120 hours have passed since the PR was submitted.

PR #1384, opened 2018-11-05
Add explanation for the MUST NOT warning

F3PiX:
and how to solve if you did it anyway

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -235,7 +235,7 @@ A problem must have a unique slug. This slug is used as
 * Do the same as when [porting an
   exercise](#porting-an-exercise-to-another-language-track).
   Reference the PR in problem-specifications if it hasn't been merged yet,
-  this **must not** be merged until the exercism/problem-specifications PR is merged.
+  this **must not** be merged until the exercism/problem-specifications PR is merged. (That's to prevent problems when regenerating all the test. If you do merge the track PR before the problem-specifications PR is merged, make sure to regenerate the tests for the track once again.)  
 
 ## Track Anatomy
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -235,7 +235,7 @@ A problem must have a unique slug. This slug is used as
 * Do the same as when [porting an
   exercise](#porting-an-exercise-to-another-language-track).
   Reference the PR in problem-specifications if it hasn't been merged yet,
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -235,7 +235,7 @@ A problem must have a unique slug. This slug is used as
 * Do the same as when [porting an
   exercise](#porting-an-exercise-to-another-language-track).
   Reference the PR in problem-specifications if it hasn't been merged yet,
-  this **must not** be merged until the exercism/problem-specifications PR is merged.
+  this **must not** be merged until the exercism/problem-specifications PR is merged. (That's to prevent problems when regenerating all the test. If you do merge the track PR before the problem-specifications PR is merged, make sure to regenerate the tests for the track once again.)
```
rpottsoh: @F3PiX should `all the test.` be `all the tests.`?

petertseng:
I'm not convinced about a "must not" anymore, not since https://github.com/exercism/meta/issues/15 . Since the website and API don't use problem-specifications to generate READMEs anymore, I don't see a "must".

I guess it would be slightly inconvenient if the exercise will want a non-default capitalisation/punctuation, because https://github.com/exercism/exercism/issues/4183 tells us that is one place where problem-specifications is being used. Without getting the problem-specifications PR merged, you'd have to have a track-specific metadata.yml if you wanted to have a non-default capitalisation/punctuation.

It's true that tracks with test generators will look at problem-specifications, since they want to insert a version number. But since this doesn't apply to every track, I also don't see it as a "must".

If there were no other observable effects of merging high-score into Ruby too early, then I suggest making it suggested instead of "must". The suggestion is based on the fact that if changes are requested to the README and/or tests in problem-specifications after the track PR is merged, you don't want to have to make a second PR to the track, now do you?

rpottsoh:
@F3PiX just a friendly nudge.

rpottsoh:
I was curious and did some digging to see when `must not` was added.  It has been there since the beginning essentially (added 11/30/15).

F3PiX:
How about this: if the change I proposed is making the readme a bit better, merge this first. Further improvements can be done in a new PR.
If my change is not helping at all, close this PR and someone with more insight in the matter opens a new PR to make it right.

petertseng:
I agree that's a very reasonable course of action!

I agree that further improvements such as that I mentioned deserve their separate PR! I volunteer.

To me it seems that the *specific* symptoms noted in this PR are not generally applicable to all tracks, *but* that the presence of track-specific tools does merit a mention. This is why I think this PR as it currently stands falls under the category of:

> If my change is not helping at all, close this PR

and that would be what I recommend

F3PiX:
Thanks @petertseng for the clear reply! I'll close this. Let me know if you'd like me to open an issue and assign it to you.

PR #1387, opened 2018-11-07
matrix: Switch to canonical notation of matrix indices

SaschaMann:
Matrix indices are usually defined so that (1, 1) is the top left entry, not (0, 0). See for example any linear algebra textbook or [wikipedia](https://en.wikipedia.org/wiki/Matrix_(mathematics)#/media/File:Matrix.svg):

wikipedia.

Even in programming, this notation is more commonly used in the context of matrices, not 2d arrays, (e.g. in Matlab, Fortran, Julia, ...) than a 0-based matrix notation.

It's less confusing for a student if the canonical definition and notation is used, so this PR changes the `canonical-data` and `description` accordingly.

### DIFF
--- a/exercises/matrix/canonical-data.json
+++ b/exercises/matrix/canonical-data.json
@@ -1,13 +1,13 @@
 {
   "exercise": "matrix",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "cases": [
     {
       "description": "extract row from one number matrix",
       "property": "row",
       "input": {
         "string": "1",
-        "index": 0
+        "index": 1
       },
       "expected": [1]
     },
@@ -16,7 +16,7 @@
       "property": "row",
       "input": {
         "string": "1 2\n3 4",
-        "index": 1
+        "index": 2
       },
       "expected": [3, 4]
     },
@@ -25,7 +25,7 @@
       "property": "row",
       "input": {
         "string": "1 2\n10 20",
-        "index": 1
+        "index": 2
       },
       "expected": [10, 20]
     },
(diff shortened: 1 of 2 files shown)


rpottsoh:
@petertseng curious if you have an opinion one way or the other on this.  I have pondered it from time to time over the past couple of weeks and have decided I don't have a real string opinion one way or the other.

petertseng:
it does not matter to me, therefore I have done nothing with this PR

sshine:
I would assume that the simplest explanation is the one that uses the indexing rules of the language in use. Since the languages you refer to (Matlab, Fortran, Julia, ...) index by 1, this would be true for them.

Changing indexing from 0 to 1 would come as a benefit to those tracks. I would assume they're a minority by numbers, but on the other hand, matrix-based exercises may play a larger role in these tracks.

rpottsoh:
@sshine actually I don't think @SaschaMann's change is to make the exercise more tailored to Matlab, Fortran, and Julia.  I think that may just be a side affect.  Some languages use 0, 1 , n based arrays...  The goal is just to use a language agnostic way of indicating matrices and indexes that more closely resembles what one may have learned from text books in math class.

SaschaMann:
> The goal is just to use a language agnostic way of indicating matrices and indexes that more closely resembles what one may have learned from text books in math class.

Indeed, this is the goal.

(Sorry, I had notifications sent to the wrong email, so I couldn't respond earlier)

rpottsoh:
Seeing no comments against merging or for more changes I am going to go ahead and merge.

PR #1392, opened 2018-11-11
bracket-push: Add test case v1.5.0

### LINKED ISSUE #1386: bracket-push: add test case with too many closing brackets
I would like to suggest an additional test case, where there are more closing brackets than opening brackets. 

For example: `"[]]"` should return `False`

 I came across a stack based solution that passed all current tests, but would fail this test. In general, students could fail this test with a stack based solution if they do not treat an empty stack as a failure, when a bracket is expected to be there. I think it would make sense to add this test case. I would be happy to hear what others think.

omer-g:
Add test case for too many closing brackets. Without the extra
closing bracket the string is legal, so this may fail certain
stack based solutions that pass the other tests.

Resolves #1386

### DIFF
--- a/exercises/bracket-push/canonical-data.json
+++ b/exercises/bracket-push/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "bracket-push",
-  "version": "1.4.0",
+  "version": "1.5.0",
   "cases": [
     {
       "description": "paired square brackets",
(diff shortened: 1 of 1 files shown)


petertseng:
EBWODP: In case anyone was interested, this will be the fifth `should_fail: true` impl for bracket-push

Issue #1211, opened 2018-03-24
pythagorean-triplet: improve readme and specification


kytrinyx:
@yawpitch submitted this over in https://github.com/exercism/exercism.io/issues/3743

---- 

I'm working on the Python track specifically, but I notice that across many of the language tracks the tests and examples do not reflect the description of the problem:

> Pythagorean Triplet
> A Pythagorean triplet is a set of three natural numbers, {a, b, c}, for which,
> `a**2 + b**2 = c**2`
> For example,
> `3**2 + 4**2 = 9 + 16 = 25 = 5**2`.
> There exists exactly one Pythagorean triplet for which a + b + c = 1000.
> Find the product `a * b * c`.

In the Python, Rust, and Haskell tracks, for instance, there are no tests whatsoever related to the "find the product `a*b*c`" statement, and all three require different constructs than what's implied by the problem description and title. The Javascript track, however, does require that the work produced can calculate the product.

petertseng:
Since we made different design choices than Project Euler and, as a result of the choices, we are not limited to validating a single numerical answer, I suggest we do away with finding the product.

There are choices for parameterisation:

* Only find the single triplet where a + b + c = 1000
* Find triplets where a + b + c = N
* Find triplets where a + b + c < N
* Find triplets where a < N and b < N and c < N
* I"m sure one can find more options if you just look through http://exercism.io/contribute/canonical-data/pythagorean-triplet but I'm not going to do that.
* Decide the problem is not a good fit for exercism and decide to deprecate it

"Only find the single triplet where a + b + c = 1000" is my least-preferred option. I do not see a need for me to express an ordering between any other options at this time.

petertseng:
EBWODP

> I suggest we do away with finding the product.

I suggest test suites simply check the values of `a`, `b` and `c` that a test case will specify.

petertseng:
Hey there, just editing the initial issue description to put backticks around the asterisks, otherwise something like `a*b*c` turns into a*b*c and `a**2 + b**2 = c**2` turns into a**2 + b**2 = c**2 which is not what we want to show in this issue (it would decrease understanding)

felix91gr:
> What test cases shall we recommend?

Checking the solution is much easier if we can ask to be given a, b and c.

petertseng:
Update on status of this issue: The canonical tests are created in https://github.com/exercism/problem-specifications/pull/1332.

[...]

PR #1395, opened 2018-11-12
pythagorean-triplet: Describe problem with variable sum N

petertseng:
This form of the problem was ratified in the canonical data
https://github.com/exercism/problem-specifications/pull/1332

Closes https://github.com/exercism/problem-specifications/issues/1211

### DIFF
--- a/exercises/pythagorean-triplet/description.md
+++ b/exercises/pythagorean-triplet/description.md
@@ -17,6 +17,6 @@ For example,
 3**2 + 4**2 = 9 + 16 = 25 = 5**2.
 ```
 
-There exists exactly one Pythagorean triplet for which a + b + c = 1000.
+Given an input integer N, find all Pythagorean triplets for which `a + b + c = N`.
 
-Find the product a * b * c.
+For example, with N = 1000, there is exactly one Pythagorean triplet for which `a + b + c = 1000`: `{200, 375, 425}`.


exercises/pythagorean-triplet/description.md
```diff
@@ -17,6 +17,6 @@ For example,
 3**2 + 4**2 = 9 + 16 = 25 = 5**2.
 ```
 
-There exists exactly one Pythagorean triplet for which a + b + c = 1000.
+Given an input integer N, find all Pythagorean triplets for which `a + b + c = N`.
```
petertseng: Debatable: whether to explicitly note here that "none" is a possible answer. For example:

> all Pythagorean triplets for which `a + b + c = N` (possibly zero triplets)

I chose not to since…

* The additional clarification is not necessary for this to be a true statement. Zero triplets is a valid result for the word "all".
* Even if a student were to miss the possibility of zero triplets, it is highly unlikely to be a fatal error requiring a rewrite of the entire implementation upon discovery. The student has already been dealing with a variable number of triplets because the test case for zero triplets comes after the test cases for multiple triplets. Even in a language whose type system prevents the return type from having zero elements, in all such languages I know, making that change is a mere one-character change (change a + to a * or something similarly easy)

petertseng: EBWODP:

> in all such languages I know

I'll leave it less ambiguous as to how many languages I know and instead say "in the [only such language](https://ceylon-lang.org/documentation/1.3/tour/sequences/) I know"

petertseng:
Given it's been a week, everyone else has missed the boat if they wanted to get on at all.

PR #1397, opened 2018-11-15
dnd-character: add new exercise

sshine:
Dungeons & Dragons character generation involves rolling and picking dice.

This seems like a nice and easy introduction to (pseudo-)random generators.

There were some uncertainties with regards to how to express non-unit properties in canonical data.

Also, I'm not sure if the `[Troll]` reference was correctly formatted in Markdown.

### DIFF
--- a/exercises/dnd-character/canonical-data.json
+++ b/exercises/dnd-character/canonical-data.json
@@ -1,0 +1,176 @@
+{
+  "exercise": "dnd-character",
+  "version": "1.0.0",
+  "comments": [
+    "The random generator 'ability' can be property-tested.",
+    "The pure function 'modifier' can be tested totally.",
+    "The random generator 'character' should return some",
+    "kind of structure with six random abilities and a",
+    "hitpoint score. This structure can be property-tested",
+    "in the same way as 'ability' and 'modifier'.",
+    "",
+    "It may be fun to provide a 'name' property for characters",
+    "that may or may not be randomly generated but not tested.",
+    "",
+    "Many programming languages feature property-based test",
+    "frameworks. For a language-agnostic introduction to",
+    "property-based testing, see:",
+    "",
+    "https://hypothesis.works/articles/what-is-property-based-testing/"
+  ],
+  "cases": [
+    {
+      "description": "ability modifier",
+      "cases": [
+        {
+          "description": "ability modifier for score 3 is -4",
+          "property": "modifier",
+          "input": {
+            "score": 3
+          },
+          "expected": -4
+        },
+        {
+          "description": "ability modifier for score 4 is -3",
+          "property": "modifier",
+          "input": {
+            "score": 4
+          },
+          "expected": -3
+        },
+        {
+          "description": "ability modifier for score 5 is -3",
+          "property": "modifier",
+          "input": {
+            "score": 5
+          },
+          "expected": -3
+        },
+        {
+          "description": "ability modifier for score 6 is -2",
+          "property": "modifier",
+          "input": {
+            "score": 6
+          },
+          "expected": -2
+        },
+        {
+          "description": "ability modifier for score 7 is -2",
+          "property": "modifier",
+          "input": {
+            "score": 7
+          },
+          "expected": -2
+        },
+        {
+          "description": "ability modifier for score 8 is -1",
+          "property": "modifier",
+          "input": {
+            "score": 8
+          },
+          "expected": -1
+        },
+        {
+          "description": "ability modifier for score 9 is -1",
+          "property": "modifier",
+          "input": {
+            "score": 9
+          },
+          "expected": -1
+        },
+        {
+          "description": "ability modifier for score 10 is 0",
+          "property": "modifier",
+          "input": {
+            "score": 10
+          },
+          "expected": 0
+        },
+        {
+          "description": "ability modifier for score 11 is 0",
+          "property": "modifier",
+          "input": {
+            "score": 11
+          },
+          "expected": 0
+        },
+        {
+          "description": "ability modifier for score 12 is +1",
+          "property": "modifier",
+          "input": {
+            "score": 12
+          },
+          "expected": 1
+        },
+        {
+          "description": "ability modifier for score 13 is +1",
+          "property": "modifier",
+          "input": {
+            "score": 13
+          },
+          "expected": 1
+        },
+        {
+          "description": "ability modifier for score 14 is +2",
+          "property": "modifier",
+          "input": {
+            "score": 14
+          },
+          "expected": 2
+        },
+        {
+          "description": "ability modifier for score 15 is +2",
+          "property": "modifier",
+          "input": {
+            "score": 15
+          },
+          "expected": 2
+        },
+        {
+          "description": "ability modifier for score 16 is +3",
+          "property": "modifier",
+          "input": {
+            "score": 16
+          },
+          "expected": 3
+        },
+        {
+          "description": "ability modifier for score 17 is +3",
+          "property": "modifier",
+          "input": {
+            "score": 17
+          },
+          "expected": 3
+        },
+        {
+          "description": "ability modifier for score 18 is +4",
+          "property": "modifier",
+          "input": {
+            "score": 18
+          },
+          "expected": 4
+        }
+      ]
+    },
+    {
+      "description": "random ability is within range",
+      "property": "ability",
+      "input": {},
+      "expected": "score >= 3 && score <= 18"
+    },
+    {
+      "description": "random character is valid",
+      "property": "character",
+      "input": {},
+      "expected": {
+        "strength": "strength >= 3 && strength <= 18",
+        "dexterity": "dexterity >= 3 && dexterity <= 18",
+        "constitution": "constitution >= 3 && constitution <= 18",
+        "intelligence": "intelligence >= 3 && intelligence <= 18",
+        "wisdom": "wisdom >= 3 && wisdom <= 18",
+        "charisma": "charisma >= 3 && charisma <= 18",
+        "hitpoints": "hitpoints == 10 + modifier(constitution)"
+      }
+    }
+  ]
+}
--- a/exercises/dnd-character/description.md
+++ b/exercises/dnd-character/description.md
@@ -1,0 +1,31 @@
+For a game of [Dungeons & Dragons][DND], each player starts by generating a
+character they can play with. This character has, among other things, six
+abilities; strength, dexterity, constitution, intelligence, wisdom and
+charisma. These six abilities have scores that are determined randomly. You
+do this by rolling four 6-sided dice and record the sum of the largest three
+dice. You do this six times, once for each ability.
+
+Your character's initial hitpoints are 10 + your character's constitution
+modifier. You find your character's constitution modifier by subtracting 10
+from your character's constitution, divide by 2 and round down.
+
+Write a random character generator that follows the rules above.
+
+For example, the six throws of four dice may look like:
+
+* 5, 3, 1, 6: You discard the 1 and sum 5 + 3 + 6 = 14, which you assign to strength.
+* 3, 2, 5, 3: You discard the 2 and sum 3 + 5 + 3 = 11, which you assign to dexterity.
+* 1, 1, 1, 1: You discard the 1 and sum 1 + 1 + 1 = 3, which you assign to constitution.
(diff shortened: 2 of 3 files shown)


exercises/dnd-character/description.md
```diff
@@ -0,0 +1,30 @@
+For a game of Dungeons & Dragons, you generate your character's six ability
```
ErikSchierboom: Maybe also introduce the concept of a character, listing the various properties (strength, etc.).

ErikSchierboom: I was wondering if we should perhaps link to the Wiki page of D&D: https://en.wikipedia.org/wiki/Dungeons_%26_Dragons, which might help people not familiar with D&D to become acquainted with it.

exercises/dnd-character/canonical-data.json
```diff
@@ -0,0 +1,170 @@
+{
+  "exercise": "dnd-character",
+  "version": "1.0.0",
+  "comments": [
+    "The random generator 'ability' can be property-tested.",
```
ErikSchierboom: It might be useful to link to an introduction of property-based testing, as that is a relatively unknown concept.

exercises/dnd-character/canonical-data.json
```diff
@@ -0,0 +1,170 @@
...
+        "dexterity": "dexterity >= 3 && dexterity <= 18",
+        "constitution": "constitution >= 3 && constitution <= 18",
+        "intelligence": "intelligence >= 3 && intelligence <= 18",
+        "wisdom": "wisdom >= 3 && wisdom <= 18",
+        "charisma": "charisma >= 3 && charisma <= 18",
+        "hitpoints": "hitpoints == 10 + modifier(constitution)"
+      }
+    },
+    {
+      "description": "the ability modifier",
```
ErikSchierboom: This is just a personal preference of mine, but in general I try to keep descriptions as terse as possible. You could consider removing "the" in both this description and in its child case descriptions. The would change a test from `"the ability modifier for the score 3 is -4"` to `"ability modifier for score 3 is -4"`. But feel free to ignore.

exercises/dnd-character/description.md
sshine: OK.

sshine: Good idea.

exercises/dnd-character/canonical-data.json
sshine: Good idea. Fortunately there is https://hypothesis.works/articles/what-is-property-based-testing/ and Hypothesis is a cross-language property-based test framework, so if the (language-agnostic) article is not enough, it may even be the case for several language track that they want to use Hypothesis specifically.

For students on the Haskell track where property-based testing has come up as a subject, I've heard once or twice that people had prior experience with Hypothesis in, I think, Python.

sshine: Would it make sense to link to this article in the description as well? Even though we expect to provide a sufficient test suite for each track, students as well as track implementers may be interested in knowing how such functions are tested, too.

exercises/dnd-character/canonical-data.json
sshine: Thanks.

sshine:
I've addressed your changes, @ErikSchierboom.

What do you think about the new introduction?

exercises/dnd-character/canonical-data.json
ErikSchierboom: I think that including it in the description is a good idea. I also like the article.

exercises/dnd-character/canonical-data.json
```diff
@@ -0,0 +1,176 @@
...
+        "dexterity": "dexterity >= 3 && dexterity <= 18",
+        "constitution": "constitution >= 3 && constitution <= 18",
+        "intelligence": "intelligence >= 3 && intelligence <= 18",
+        "wisdom": "wisdom >= 3 && wisdom <= 18",
+        "charisma": "charisma >= 3 && charisma <= 18",
+        "hitpoints": "hitpoints == 10 + modifier(constitution)"
+      }
+    },
+    {
+      "description": "ability modifier",
```
ErikSchierboom: I was thinking that perhaps the "ability modifier" tests should be listed before the "ability" and "character" tests. I have two reasons for this:

- It is easier for students to start with implementing a "regular" function.
- The "character" test builds on "modifier" property, and it makes sense to introduce dependencies before using them.

exercises/dnd-character/description.md
```diff
@@ -0,0 +1,34 @@
+For a game of [Dungeons & Dragons][DND], each player starts by generating a
+character they can play with. This character has, among other things, six
+abilities; strength, dexterity, constitution, intelligence, wisdom and
```
ErikSchierboom:
```suggestion
abilities: strength, dexterity, constitution, intelligence, wisdom and
```

exercises/dnd-character/metadata.yml
```diff
@@ -0,0 +1,4 @@
+---
+blurb: "Randomly generate Dungeons & Dragons characters"
+source: "Simon Shine, Erik Schierboom"
```
ErikSchierboom: While I appreciate the reference, this is really all your work, so feel free to remove my name.

exercises/dnd-character/canonical-data.json
sshine: OK, good argument. If you think that the pure `modifier` function should be implemented before the two random generators `ability` and `character`, I'll list them first. It's just that the tests for `ability` seem to flood canonical-data.json.

The current order is based on the same reasoning as you apply wrt. dependencies.

exercises/dnd-character/metadata.yml
sshine: Oh, but I wouldn't have suggested this as an exercise at all; it was just an example.

And your current feedback has more than warranted it. I'd like to think of it as a joint effort. :-)

exercises/dnd-character/canonical-data.json
sshine: Hmm, now that I've tried, I don't think that I will. I think different languages have very different approaches to testing randomness. In Perl at work I sometimes use [`Test::MockRandom`](https://metacpan.org/pod/Test::MockRandom), and I can't find an equally language-agnostic article on mocking and dependency inversion.

So a cheap alternative is to let students inspect whatever tests implementers decide to make, and perhaps think about good linkable resources on how it's being tested and spread these later.

exercises/dnd-character/description.md
```diff
@@ -0,0 +1,34 @@
...
+do this by rolling four 6-sided dice and record the sum of the highest three
+dice.  You do this six times, once for each ability.
+
+Your character's initial hitpoints are 10 + your character's constitution
+modifier. You find your character's constitution modifier by subtracting 10
+from your character's constitution, divide by 2 and round down.
+
+Write a random character generator that follows the rules above.
+
+Normally you would assign each dice roll to an ability freely, but for this
```
coriolinus: "dice" is always plural. The singular is "die".
```suggestion
Normally you would assign each die roll to an ability freely, but for this
```

exercises/dnd-character/description.md
```diff
@@ -0,0 +1,34 @@
...
+exercise, assign the dice rolls randomly among the six abilities.
+
+For example, the six dice rolls may look like:
```
coriolinus:
```suggestion
For example, the 24 dice rolls may look like:
```

exercises/dnd-character/description.md
sshine: Is it also called a die roll if there is more than one die?

exercises/dnd-character/description.md
sshine: I think this would be confusing and ambiguous. You're not allowed to combine your 24 dice freely, and you roll them in six groups of four.

Also, shouldn't it be "die rolls" here, too, then?

If I'm rolling four dice six times, is it six die rolls or six dice rolls? It is not conceptually 24 die rolls, since the "largest 3 of 4" rule apply to groups of dice.

sshine: I could call it "six throws of four dice"?

exercises/dnd-character/description.md
coriolinus: Yes, I'm pretty sure.

exercises/dnd-character/description.md
coriolinus: Heh, six throws of four dice does solve the ambiguity. Otherwise, I think you're right: it would be die rolls here also.

sshine: Thanks. :)

ErikSchierboom: I like the six throws of four dice suggestion

exercises/dnd-character/description.md
```diff
@@ -0,0 +1,34 @@
+For a game of [Dungeons & Dragons][DND], each player starts by generating a
+character they can play with. This character has, among other things, six
+abilities; strength, dexterity, constitution, intelligence, wisdom and
+charisma. These six abilities have scores that are determined randomly. You
+do this by rolling four 6-sided dice and record the sum of the highest three
+dice.  You do this six times, once for each ability.
```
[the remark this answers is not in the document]
sshine: Thanks.

exercises/dnd-character/canonical-data.json
```diff
@@ -0,0 +1,176 @@
...
+          "property": "modifier",
+          "input": {
+            "score": 18
+          },
+          "expected": 4
+        }
+      ]
+    },
+    {
+      "description": "random ability is within range",
```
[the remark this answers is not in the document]
sshine: But do language tracks actually implement [randomness tests](https://en.wikipedia.org/wiki/Randomness_tests)?

I can see that the [C# track](https://github.com/exercism/csharp/blob/master/exercises/diffie-hellman/DiffieHellmanTest.cs#L21) checks if the output is distinct between runs. If we don't expect to test the ephemeral quality of randomness, how about we just let it be up to the mentor to validate if the student initialized the random generator in the right way?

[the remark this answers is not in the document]
sshine: Since `ability` can only output values in the range 3-18, duplicates are quite likely.

The C# Diffie-Hellman "unique implies random" property relies on the odds 1 : 10^*d* for *d* digits in the private key; I don't know how much *d* is in this exercise, but the point with the exercise is that it's pretty big in practice.

I don't want to make it too hard for track implementers to write tests.

I think what might happen in a lot of cases is that the RNG can be mocked.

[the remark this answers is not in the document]
coriolinus: Real statistical randomness tests are non-trivial to implement. I can't speak for other tracks, but in the Rust track, the [`diffie-hellman` tests](https://github.com/exercism/rust/blob/master/exercises/diffie-hellman/tests/diffie-hellman.rs) just kind of punted on randomness checking. 

Range tests are more plausible to request of non-mathematician test implementers.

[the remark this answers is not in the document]
sshine: I think we should defer this addition to a separate issue of testing whether randomness is used as expected, since it may cover more than one exercise (and in particular also the diffie-hellman exercise).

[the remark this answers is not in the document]
ErikSchierboom: Agreed, those are very valid points. Let's defer the discussion then.

sshine:
Since this has been waiting for a while, I'll merge it now.

Thanks for all the responses!

ErikSchierboom:
Congrats on getting a new exercise added!

rpottsoh:
Nice job guys!

cmccandless:
Has any track implemented this yet? Python is starting to, and the issue I'm immediately noticing is this: how do you force a solution to "roll 4 dice" for an ability instead of `rand(3, 18)` (pseudocode, assume both ends inclusive)? Thoughts?

coriolinus:
Real randomness tests are kind of hard. Doing this the mathematically
rigorous way would involve a few million iterations; at that scale, the
mathematical properties of the generated numbers will be reliably
measurably different from the mathematical properties of `rand(3, 18)`.

That's too big for a unit test, though. A more reasonable solution is to
add a note to mentors that they should pay particular attention to the way
the numbers are generated. It's human-mediated, but much more tractable in
the short term.

sshine:
Yes, it isn't the goal of the property test to check the probability distribution experimentally. But the example pitfall you gave is likely to pop up in most tracks. Perhaps this similarity can be shared between mentor notes.

There is a PR on the Haskell track, to answer your other question.

PR #1399, opened 2018-11-17
binary-search: test for when l and r bounds cross

Cohen-Carlisle:
this test can catch a solution that explicitly checks for an empty array
but fails to check if the bounds cross

plus test description tweak

### DIFF
--- a/exercises/binary-search/canonical-data.json
+++ b/exercises/binary-search/canonical-data.json
@@ -80,7 +80,7 @@
       "expected": {"error": "value not in array"}
     },
     {
-      "description": "a value smaller than the array's smallest value is not included",
+      "description": "a value smaller than the array's smallest value is not found",
       "property": "find",
       "input": {
         "array": [1, 3, 4, 6, 8, 9, 11],
@@ -89,7 +89,7 @@
       "expected": {"error": "value not in array"}
     },
     {
-      "description": "a value larger than the array's largest value is not included",
+      "description": "a value larger than the array's largest value is not found",
       "property": "find",
       "input": {
         "array": [1, 3, 4, 6, 8, 9, 11],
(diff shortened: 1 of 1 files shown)


petertseng:
I think I found a solution that meets the description you speak of.

```ruby
def naive_bs(arr, val)
  return nil if arr.empty?

  subfind = ->(left, right) {
    return arr[left] == val ? left : nil if right == left + 1

    middle = (right - 1 - left) / 2 + left
    at_middle = arr[middle]

    if right <= left
      raise TestFailure, "This solution is too naive to handle right #{right} <= left #{left}"
    elsif at_middle == val
      middle
    elsif val < at_middle
      subfind[left, middle]
    else
      subfind[middle + 1, right]
    end
  }
  subfind[0, arr.size]
end
```

Cohen-Carlisle:
Hi @petertseng ! It looks like you figured this one out but I'll post the write up I worked up anyway.

I'll use notation similar to that used on [wikipedia](https://en.wikipedia.org/wiki/Binary_search_algorithm). This test will produce a condition in which the left bound (`l`) becomes greater than the right bound (`r`). This example solution will currently pass the tests, but will result in an error under the new test case:
```
defmodule BinarySearch do
  def search(a, t) do
    if a == {} do
      :not_found
    else
      do_search(a, t, 0, tuple_size(a) - 1)
    end
  end

  defp do_search(a, t, l, r) do
    cond do
      l == r and am != t ->
        :not_found

      am < t ->
        do_search(a, t, m + 1, r)

      am > t ->
        do_search(a, t, l, m - 1)

      true ->
        {:ok, m}
    end
  end
end
```

Here is why:
Iteration 1:
`l` is initialized at 0 and `r` at 0. `m` is calculated to be 0 and the value there is `1`. Because the target value (`t`) is less than `am`, the third branch of the `cond` is taken.

Iteration 2: 
`l` is still 0 but `r` is now `-1` since it now has the value `m - 1` from the previous call, where `m` was 0. This will lead to `m` being calculated as `-1`, which causes an error when trying to retrieve the value at the negative index.

A correct solution checks for `l > r` and returns `:not_found` at that point, which avoids this error.

PR #1400, opened 2018-11-18
clock: Add test case for exactly negative sixty minutes.

tmccombs:
I encountered a solution that passed all provided test cases but would have failed this test.


tmccombs:
added

petertseng:
I added `clock:` to the PR title, I hope you can add it to the commit message before merging. So that it is possible to understand what exercise it refers to when just looking through `git log`

PR #1401, opened 2018-11-18
wordy: Test question `What is?` because it's too short

petertseng:
wordy 1.5.0

In keeping with [1.3.0][], we want to test that these fail in an
expected way rather than an unexpected way.

[1.3.0]: https://github.com/exercism/problem-specifications/pull/1383

For example, imagine a language that uses an optional as the answer. We
would want to make sure an implementation written in that language would
return `None`, rather than (say) accessing an out-of-bounds index due to
always assuming that the inputs will have at least three space-separated
elements (of the form "What is X...?").

### DIFF
--- a/exercises/wordy/canonical-data.json
+++ b/exercises/wordy/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "wordy",
-  "version": "1.4.0",
+  "version": "1.4.1",
   "comments": [
     "The tests that expect an 'error' should be implemented to raise",
     "an error, or indicate a failure. Implement this in a way that",
@@ -144,7 +144,7 @@
       "expected": {"error": "unknown operation"}
     },
     {
-      "description": "reject incomplete problem",
+      "description": "reject problem missing an operand",
       "property": "answer",
       "input": {
         "question": "What is 1 plus?"
--- a/exercises/wordy/canonical-data.json
+++ b/exercises/wordy/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "wordy",
-  "version": "1.4.1",
+  "version": "1.5.0",
   "comments": [
     "The tests that expect an 'error' should be implemented to raise",
     "an error, or indicate a failure. Implement this in a way that",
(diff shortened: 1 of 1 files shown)


exercises/wordy/canonical-data.json
```diff
@@ -151,6 +151,22 @@
...
+    {
+      "description": "reject unstarted problem",
+      "property": "answer",
+      "input": {
+        "question": "What is?"
+      },
+      "expected": {"error": "syntax error"}
+    },
+    {
+      "description": "reject empty string",
```
rpottsoh: Is this test perhaps unnecessary because of #902?

petertseng: Part of me wants to say "but then who will guard against people assuming that the input is always at least two words!" but I guess part of the question is what assumptions we're at liberty to make. Here's my deal. I'm going to leave that case for later. If there's a desire to add it, it'll be in 1.6.0

rpottsoh: Perhaps we need a case that asks a valid question but doesn't start with "What is".  "What is" shouldn't influence the outcome.
```
5 plus 6 is?
```
should work the same as:
```
What is 6 plus 5?
```
and
```
6 plus 2 plus 3?
```
For another PR perhaps......

petertseng: These proposals, along with the empty string, do bring up interesting questions of how far we want to take the parsing aspect in this problem.

The arithmetic portion is not *too* complicated, but does bring up interesting questions of how to translate each operation from its verbal form to its code form. So there could be some room to do more parsing.

I have not yet decided what I would suggest in this realm.

exercises/wordy/canonical-data.json
```diff
@@ -151,6 +151,14 @@
       },
       "expected": {"error": "syntax error"}
     },
+    {
+      "description": "reject unstarted problem",
```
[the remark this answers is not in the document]
petertseng: if I call it unfinished, I worry that the description sounds too similar to `incomplete problem` above (since I see `unfinished` and `incomplete` to have too similar meanings). 

If we take for granted that "What is" is a common feature of all questions, then I see it making sense since they didn't even start saying anything else

But if this assumption gets broken in a later PR that adds questions that don't start with "What is", then this description won't make sense!!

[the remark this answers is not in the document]
rpottsoh: Perhaps `ambiguous` instead of `unfinished`.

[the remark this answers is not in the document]
petertseng: I think the way out will be to say exactly in what way the two problems are incomplete. I have now made a proposal.

petertseng:
All right, thanks! Everyone else I'd say you have about 24 hours to get a word in.

PR #1402, opened 2018-11-19
high-scores: multiple improvements

F3PiX:
These updates address issues that were raised after the first implementation (#1378 )

### DIFF
--- a/exercises/high-scores/canonical-data.json
+++ b/exercises/high-scores/canonical-data.json
@@ -1,6 +1,14 @@
 {
   "exercise": "high-scores",
   "version": "2.0.0",
+  "comments": [
+    "This is meant to be an easy exercise to practise: ",
+    "* arrays as simple lists",
+    "* instantiating a class",
+    "Consider adding a track specific recommendation in the track's hint.md.",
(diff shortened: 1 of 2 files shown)


rpottsoh:
Are there any issues or PRs that document the impetus behind these changes?

F3PiX:
@rpottsoh I know for sure that the naming things came up in discussions about the initial implementation here and in the Ruby track, but I couldn't pinpoint them.

The comment added is a solution for the fact that the important information of the goal of the exercise is buried in the Ruby Track hint.md, and I feel that that kind of information needs to be available to the maintainers. To make it easier if and how the exercise fits in their tracks.

rpottsoh:
I see.  I just wanted to make sure there wasn't something that should be documented here because of how the initial comment is worded.
> These updates address issues that were raised after the first implementation

I am fine with the changes, they seem reasonable.  I just was not understanding why they were needed.  :smile:

F3PiX:
> I just was not understanding why they were needed. 😄

You're right, I should've documented it better, but honestly, I lost track of them.  😱 😜 
Thanks!

F3PiX:
Thanks @ErikSchierboom , @rpottsoh !! 
cc @pgaspar

PR #1408, opened 2018-11-27
 travis: Check that changed JSON files have their versions changed

petertseng:
A common mistake


petertseng:
```
$ sh bin/check_versions
This PR changed exercises/affine-cipher/canonical-data.json, so you should change the version as well.
You can see the versioning policy at https://github.com/exercism/problem-specifications#test-data-versioning
The command "sh bin/check_versions" exited with 1.
```

bin/check_versions
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+
+need_version=0
+
+for json_file in $(git diff --name-only master HEAD | grep 'exercises/.*json' | sort | uniq); do
```
sshine: Perhaps `grep -E '^exercises/.*\.json$'` to avoid conflicts with future exercise names that contain the word "json".

There's also `sort -u`.

bin/check_versions
```diff
@@ -0,0 +1,18 @@
...
+  old_version=$(git show master:$json_file | jq .version)
+  new_version=$(jq .version < $json_file)
```
sshine: It seems that `jq` supports

    jq .version $json_file

cmccandless:
Pending the suggestions by @sshine (particularly for the `grep` pattern as that is the only one that is a functional change), looks good to me.

rpottsoh:
> I'm not versed enough in bash to comment on the code, but functionality wise this is a great addition I feel.

I would like to echo @ErikSchierboom's sentiment.  Thanks for opening this PR @Petertseng.  This is something that has been needed for a long time. :+1:

bin/check_versions
petertseng: would you believe it, I completely missed `-u` when reading the manual. how did that happen...

so, is there a need for the `-E` in the proposed `grep` invocation? On the various compuers I have, `echo "exercises/a/test.json\nexercises/a/test.json.backup\nexercises/a-json-exercise/readme.md" | grep '^exercises/.*\.json$'` does the right thing (only `exercises/a/test.json`). Is there a different implementation of `grep` where the `-E` would be required, some other reason we'd want to include it, etc.?

bin/check_versions
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+
+need_version=0
+
+for json_file in $(git diff --name-only master HEAD | grep '^exercises/.*\.json$' | sort --unique); do
```
petertseng: it occurs to me:

* the `sort`, although it does not affect correctness, may help us make sense of the output if the output is alphabetised
* the `uniq` or `--unique` neither affects correctness nor do I expect it to ever change the output. I cargo-culted from https://github.com/exercism/rust/blob/master/_test/ensure-readmes-are-updated.sh#L15 (there is a subtle pun there), but a `cut` operation is done there, whereas there is no `cut` here. If we assume that `git diff` only outputs each file once, then it is useless. I suppose it can be kept despite its uselessness, in case `git diff`'s behaviour ever changes, but I don't see why it ever would

cmccandless: > the sort, although it does not affect correctness, may help us make sense of the output if the output is alphabetised

Since the output is directly from `git diff`, which will be pre-sorted, sort can be removed.

> the uniq or --unique neither affects correctness nor do I expect it to ever change the output.

Agreed. It is unnecessary and can also be removed.

petertseng: Upon consideration, I've decided that in this case "can be removed" implies "should be removed". If people assume things are there for a reason and try to find a reason for the `sort` and `uniq`, they will be at a loss, and I don't want to inflict that on others.

petertseng:
Nobody has asked the important question: "What happens if I want to make a change to the JSON files that should NOT result in a version change? For example, a formatting change"

There are a few possible solutions.

1. Use https://docs.travis-ci.com/user/environment-variables/#default-environment-variables to see if a special string like "I have confirmed that this change to the JSON files does not need a corresponding version change" appears in a commit message, and if so, skip this check.
2. in the PR in question, remove the version check, demonstrate that Travis CI passes, then add it back, and merge despite the subsequent CI failure - highly recommended to squash so that the removal doesn't appear in history

I'm not convinced that I'm ready to implement the first plan in this PR. maybe later.

cmccandless:
> 1. Use docs.travis-ci.com/user/environment-variables/#default-environment-variables to see if a special string like "I have confirmed that this change to the JSON files does not need a corresponding version change" appears in a commit message, and if so, skip this check.
> ...
> I'm not convinced that I'm ready to implement the first plan in this PR. maybe later.

I don't think implementing this plan in this PR would be difficult or unreasonable. If you change the shebang to point to `bash` instead of `sh`, the check becomes:

```Bash
#!/bin/bash
OVERRIDE_MESSAGE='I have confirmed that this change to the JSON files does not need a corresponding version change'
#...
if [ "$old_version" = "$new_version" ] && [[ $TRAVIS_COMMIT_MESSAGE != *"$OVERRIDE_MESSAGE"* ]]; then
    echo "This PR changed $json_file, so you should change the version as well."
    need_version=1
fi
```

petertseng:
Very well. I have slightly modified the idea. The commit message only needs to be checked once since it won't change. So we just do it right at the top.

I would normally just let the following terminal session count as a test of this, but I'm still going to test in Travis nevertheless:

```
$ env TRAVIS_COMMIT_MESSAGE="hello" sh bin/check_versions
$ env TRAVIS_COMMIT_MESSAGE="I have confirmed that this change to the JSON files does not need a corresponding version change" sh bin/check_versions
You have overridden the version check, so we assume you know what you are doing and will not run this check.
```

petertseng:
One thing to note: The string is a little long, such that it does not follow certain rules for column widths in commit messages... including that of my editor's auto wrap. So perhaps "I have confirmed that no JSON version change is needed" instead. Hopefully I have not attracted too much bike shedding.

bin/check_versions
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
+
+override_message="I have confirmed that no JSON version change is needed"
+# We could use TRAVIS_COMMIT_MESSAGE,
+# but that has not been found to be reliably set.
+commit_message=$(git log --format=%B -n 1 HEAD)
+
+# If commit message contains the override message:
+if [ -n "$commit_message" ] && [ -z "${commit_message##*$override_message*}" ]; then
```
petertseng: note: reference for that `${a##b}`: https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

petertseng: I deemed the `-n $commit_message` unfortunately necessary, because `[ -z "$a##anything" ]` is true (string is empty) if `$a` is empty.

bin/check_versions
```diff
@@ -0,0 +1,33 @@
+#!/bin/sh
+
+override_message="I have confirmed that no JSON version change is needed"
```
cmccandless: I have no problem with this wording.

I think perhaps the only thing left is to document this check and the override process in the repository README? Any good reasons not to do this?

petertseng:
right then! expect me to be merging sometime soon. It will be squashed, you wouldn't want all those test commits and associated reversions to go in the history.

bin/check_versions
```diff
@@ -0,0 +1,33 @@
...
+commit_message=$(git log --format=%B -n 1 HEAD)
```
petertseng: unneeded line which I'll remove.

PR #1269, opened 2018-07-19
sgf-parsing: add canonical-data.json

cmccandless:
Resolves #585

PR #1410, opened 2018-11-28
sgf-parsing: Added an explicit 'children' field in the 'node without properties' test case

ZapAnton:
Based on [this comment](https://github.com/exercism/problem-specifications/pull/1269#discussion_r206646675) I gathered, that it is OK to add an explicit field

### DIFF
--- a/exercises/sgf-parsing/canonical-data.json
+++ b/exercises/sgf-parsing/canonical-data.json
@@ -1,6 +1,6 @@
 {
     "exercise": "sgf-parsing",
-    "version": "1.0.0",
+    "version": "1.1.0",
     "cases": [
         {
             "description": "empty input",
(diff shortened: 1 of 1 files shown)


exercises/sgf-parsing/canonical-data.json
```diff
@@ -1,6 +1,6 @@
 {
     "exercise": "sgf-parsing",
-    "version": "1.0.0",
+    "version": "1.0.1",
```
rpottsoh: This should be 1.1.0 I believe.

ZapAnton:
@rpottsoh anything else in this PR that needs to be changed?

rpottsoh:
Thanks @ZapAnton!

PR #1412, opened 2018-12-01
hamming: remove redundant test cases

### LINKED ISSUE #1403: hamming: reduce number of test cases
If one examines the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/hamming/canonical-data.json) of the `hamming` exercise, one finds that it has many different test cases. However, I would argue that many test cases don't add any value and go against our unwritten rule that new tests cases should require changes to the code to make them pass. In other words, at some point, the implementation will  pass the first number of tests, and then most likely also automatically pass all the subsequent tests (except for the error handling ones).

IMHO the fix to this is to remove (most of) the unneeded test cases. What exactly the test cases are that should be removed is up for discussion.

ErikSchierboom:
Closes #1403

### DIFF
--- a/exercises/hamming/canonical-data.json
+++ b/exercises/hamming/canonical-data.json
@@ -1,6 +1,6 @@
 {
 "exercise": "hamming",
-"version": "2.1.1",
+"version": "2.2.0",
   "comments": [
     "Language implementations vary on the issue of unequal length strands.",
     "A language may elect to simplify this task by only presenting equal",
@@ -21,7 +21,7 @@
       "expected": 0
     },
     {
-      "description": "identical strands",
+      "description": "single letter identical strands",
       "property": "distance",
       "input": {
         "strand1": "A",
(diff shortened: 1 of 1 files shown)


exercises/hamming/canonical-data.json
```diff
@@ -30,97 +30,25 @@
...
-      "description": "small distance in long strands",
-      "property": "distance",
-      "input": {
-        "strand1": "ACCAGGG",
-        "strand2": "ACTATGG"
-      },
-      "expected": 2
-    },
-    {
-      "description": "non-unique character in first strand",
```
petertseng: I have never understood what kinds of implementations this case was meant to detect, and https://github.com/exercism/problem-specifications/pull/51 which added it did not provide the rationale.

However, I suppose in its original form (`AGA` vs `AGG`), it would find implementations that would checked whether the *set* of letters is the same (for example, here, declare that the two strings have distance 0 because both of them contain at least one A and at least one G).

If that is so, https://github.com/exercism/problem-specifications/pull/953 lost that ability. Since nobody has come forth in the year that it's been gone and requested that it be changed back, that's a fine indication that that was of no concern to anyone, and thus should be removed.

PR #1414, opened 2018-12-04
sgf-parsing: add test case for multiple properties

cmccandless:
@ RE: @elyashiv in https://github.com/exercism/python/pull/1615#issuecomment-444123399

This test case is mentioned in the README, but is not yet present in the canonical data.


cmccandless:
If there are no further comments in 24 hours, this may be merged. I myself will not be available to merge over the weekend, but any other member is free to merge after that time.

PR #1415, opened 2018-12-06
Isogram: test a string containing digits

elyashiv:
I actually had a solution submitted with a bug in this area (out of bounds of an array).

### DIFF
--- a/exercises/isogram/canonical-data.json
+++ b/exercises/isogram/canonical-data.json
@@ -3,7 +3,7 @@
   "comments": [
     "An isogram is a word or phrase without a repeating letter."
   ],
-  "version": "1.7.0",
+  "version": "1.8.0",
   "cases": [
     {
       "description": "Check if the given string is an isogram",
@@ -114,6 +114,14 @@
             "phrase": "angola"
           },
           "expected": false
+        },
+        {
+          "description": "string contains duplicate numbers",
+          "property": "isIsogram",
+          "input":{
+            "phrase": "123123abc"
+          },
+          "expected": true
         }
       ]
     }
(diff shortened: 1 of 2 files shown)


rpottsoh:
I guess I am not for or against this change, but wonder if it might be better to update [decription.md](https://github.com/exercism/problem-specifications/blob/c514bf3bd0/exercises/isogram/description.md) to contain another example of an acceptable isogram instead of adding a test?

for example:
Examples of isograms:

- lumberjacks
- background
- downstream
- six-year-old

could become:

Examples of isograms:

- lumberjacks
- background
- downstream
- six-year-old
- 123123abc

I suppose it could _also_ be acceptable to change description.md as well.   I'm just thinking out loud....

elyashiv:
I think that there should be a test for every case, but I do agree that this should be mentioned in the description. I updated the PR.

sshine:
I don't think this test adds value to the exercise. Numbers only become a corner-case by extending the test to expect them.

I would rather not see input sanitization become a mandatory part of every string-related exercise, but rather have tracks make this choice depending on how the exercise is used.

rpottsoh:
Do we have issues with the clarification added to description.md or are we opposed to all of the proposed changes?  I remain on the fence.

sshine:
Before the addition of digits, the assignment works on the subset `[A-Za-z -]` with `[ -]` being "ignore characters". Little is gained by extending this set arbitrarily but not completely.

@elyashiv: Could this bug be triggered by excessive occurrence of `[ -]`?

ErikSchierboom:
> Little is gained by extending this set arbitrarily but not completely: You make the exercise longer but not harder.

I agree with this sentiment.

kotp:
Closing as indicated in comments by @bitfield and @elyashiv .

PR #1416, opened 2018-12-12
dnd-character: add test for one-time ability init

cmccandless:
I came across an implementation that simply randomized each property whenever it was called, something like:

```
class Character():
    def strength():
        return randint(3, 18)
```

These values should be calculated only once each. Checking one twice should verify that this is the case.


sshine:
Interesting!

And good catch!

PR #1417, opened 2018-12-12
dnd-character: Add title to metadata

sshine:
Currently, the exercise shows up as 'Dnd Character' when 'DND Character' would be more appropriate.

I don't need to bump any version number for this change, right?


cmccandless:
> I'll let you decide this, but you may find that DnD is more common than DND.

What about 'D&D'?

sshine:
'D&D' is even better, but I wasn't sure if it'd work in all circumstances, '&' being a scary character on the internet.

sshine:
Also, I haven't played DnD / D&D since 1st edition when I was 11, so I'm no authority any more. :-D

cmccandless:
> but I wasn't sure if it'd work in all circumstances, '&' being a scary character on the internet.

That's... a really good point. I guess we'll find out if that breaks anything on the site.

PR #1418, opened 2018-12-14
Leap: Remove recommendation to DIY in description

sshine:
Leap-year functions don't generally occur in standard libraries.

Is there a single example of a language we have that does this?

If there is one, then that would be the right way to solve this exercise in that language. Presumably this exercise is primarily placed at the top of core tracks. We should not encourage to re-implement standard library functions at that point. But this is a hypothetical argument.

### DIFF
--- a/exercises/leap/description.md
+++ b/exercises/leap/description.md
@@ -11,9 +11,6 @@ on every year that is evenly divisible by 4
 For example, 1997 is not a leap year, but 1996 is.  1900 is not a leap
 year, but 2000 is.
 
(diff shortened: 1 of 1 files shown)


sshine:
Would it make sense to have that comment in the Ruby README only? Maybe even with a link to the function you're not allowed to use.

sshine:
> I don't mind that we keep DIY exercises in the track, as long as they're not core.

That's exactly why I tagged you; I've seen you make that point before and wanted to make that come across. :)

sshine:
Can I merge this?

PR #1420, opened 2018-12-22
luhn: non-digit at end is invalid

bitfield:
From [luhn: add test case with non-digit at the end](https://github.com/exercism/go/pull/1186):

> Problem: I had a student only include the digits before the first invalid character. There was no test that failed his solution.

> Solution: Added a test that should return false but does not if the student takes only the numbers up to first non-digit.


rpottsoh:
Let me know if you need me to merge.

PR #1424, opened 2018-12-30
Created New Exercise: Reforestation

fidelcoria:
The task is to rebuild binary trees using pre-order and in-order traversals of the original tree.
Possible tags for this exercise include: recursion, binary trees, and tree traversals.

### DIFF
--- a/exercises/reforestation/canonical-data.json
+++ b/exercises/reforestation/canonical-data.json
@@ -1,0 +1,93 @@
+{
+    "exercise": "reforestation",
+    "version": "1.0.0",
+    "cases": [
+      {
+        "description": "Base case: empty tree",
+        "property": "treeTraversals",
+        "input": {
+          "preorder": [],
+          "inorder": []
+        },
+        "expected": {}
+      },
+      {
+        "description": "1 more step than base case",
+        "property": "treeTraversals",
+        "input": {
+            "preorder": ["a"],
+            "inorder": ["a"]
+        },
+        "expected": {"v":"a", "l": {}, "r": {}}
+      },
+      {
+        "description": "Core functionality w/ realistic size traversal",
+        "property": "treeTraversals",
+        "input": {
+            "preorder": ["a","i","x","f","r"],
+            "inorder": ["i","a","f","x","r"]
+        },
+        "expected": {"v": "a",
+            "l": {"v": "i", "l": {}, "r": {}},
+            "r": {"v": "x",
+                "l": {"v": "f", "l": {}, "r": {}},
+                "r": {"v": "r", "l": {}, "r": {}}
+            }
+        }
+      },
+      {
+        "description": "Traversals must be fully consumed. No extra values.",
+        "property": "treeTraversals",
+        "input": {
+          "preorder": ["a", "b"],
+          "inorder": ["b", "a", "r"]
+        },
+        "expected": {"error": "invalid traversal"}
+      },
+      {
+          "description": "Reject inconsistent traversals of same length",
+          "property": "treeTraversals",
+          "input": {
+              "preorder": ["x","y","z"],
+              "inorder": ["a","b","c"]
+          },
+          "expected": {"error": "traversals are not consistent"}
+      },
+      {
+        "description": "BonusPoints: generate multiple trees",
+        "property": "treeTraversals",
+        "input": {
+            "preorder": ["p","r","o","l","o","g"],
+            "inorder": ["o","l","o","r","p","g"]
+        },
+        "expected": [
+            {"v":"p",
+                "l": {"v": "r",
+                    "l": {"v": "o",
+                        "l": {"v": "l",
+                            "l": {"v": "o", "l": {}, "r": {}},
+                            "r": {}
+                        },
+                        "r": {}
+                    },
+                    "r": {}
+                },
+                "r": {"v": "g", "l": {}, "r": {}}
+            },
+            {"v": "p",
+                "l": {"v": "r",
+                    "l": {"v": "o",
+                        "l": {},
+                        "r": {"v": "l",
+                            "l": {},
+                            "r": {"v": "o", "l": {}, "r": {}}
+                        }
+                    },
+                    "r": {}
+                },
+                "r": {"v": "g", "l": {}, "r": {}}
+            }
+        ]
+      }
+    ]
+  }
--- a/exercises/reforestation/description.md
+++ b/exercises/reforestation/description.md
@@ -1,0 +1,22 @@
+A binary tree with no repeated elements can be represented uniquely by its
+pre-order and in-order traversals (post-order and in-order also work).
+
+Tree traversals start at the root of the tree and can be defined recursively.
+
+Pre-order traversal of this tree is [a, i, x, f, r]     |
+    take the node value                                 |    a
+    then pre-order left subtree (recursive)             |   / \
+    and finally pre-order right subtree (recursive)     |  i   x
+In-order traversal of this tree is [i, a, f, x, r]      |     / \
+    in-order left subtree (recursive)                   |    f   r
+    then take the node value                            |
+    and finally in-order right subtree (recursive)      |
+
+Implement an algorithm to rebuild a tree from its pre-order and in-order
+traversals.
+
+Bonus Points: 
+Consider the case when the traversals contain repeated elements. There may be
+multiple possible trees. Find all of them.
+
+Save the trees!
--- a/exercises/reforestation/metadata.yaml
+++ b/exercises/reforestation/metadata.yaml
@@ -1,0 +1,3 @@
+---
+title: "Reforestation"
+blurb: "Rebuild binary trees from pre-order and in-order traversals."


sshine:
Would it make sense to phrase the exercise with a theme, so that it sounds less like a hunt for an existing algorithm that one can mindlessly transcribe from pseudocode?

As in, if you recognize the general pattern or somehow extract it, that's perfect, but not what's primarily encouraged.

fidelcoria:
Yup, this is the first place for discussion..

1.  True. "retree" was my other option. "tree decompression" might work, too.

2. Do you mean difficult to read? The definitions are from my mind, but I did read them in a text book.

3. I know of one algorithm which I am sure can be implemented in different ways. 

My original intention was to add this exercise to the Prolog track. In Prolog you can implement the algorithm directly or use DCGs. Currently there are no exercises that encourage using DCGs so I thought I should add one. This exercise lets Declarative Prolog show off its best qualities, but also some of its shortcomings (i.e. not purely declarative). In short DCGs would be the theme I suggest.

The Exercism docs recommend creating an exercise in this repo first. For most languages I can see this being simply "transcribing pseudocode." But in Prolog it could be useful.

Is it allowable to create an exercise specifically for a language track without a problem-specifications entry?

sshine:
1. "Retree" is also a good, creative name. I kind of like it better because it sounds like "retry", but other than that, the only reason why I want something other than "Reforestation" is that I'd like an exercise with forests. :-)

2. Yes, difficult to figure out how to solve. I suspect. I don't know what would be the right amount of guidance, though. The "transcribing pseudocode" critique is mostly if there's really one way to do it, and the average student ends up googling it (which I suspect is more likely when the exercise is phrased in the same way as the solutions online). An example I can think of is "difference-of-squares".

Wrt. language-specific tracks, yes, I think so. See e.g. the [Ballerina](https://github.com/exercism/ballerina/tree/master/exercises) track's exercises. But I really don't think it is a good approach in general. Prolog *might* be an exception because it's borderline not-general-purpose.

In this case, for example, I'd welcome more tree-based exercises in the ML-style tracks, so we can share the exercise design process.

One last comment on this exercise text proposal: Either remove the "you can generate multiple results" paragraph from the description, or add a test case that demonstrates multiple results. I would remove the paragraph and add a "hint" (extra section of the README on the language track) and a track-specific test for the multiple results: The exercise without duplicates are already difficult and interesting, and the "multiple solutions as a side-effect" is a Prolog thing. :)

exercises/reforestation/canonical-data.json
```diff
@@ -0,0 +1,93 @@
...
+        "description": "1 more step than base case",
+        "property": "treeTraversals",
+        "input": {
+            "preorder": ["a"],
+            "inorder": ["a"]
+        },
+        "expected": {"v":"a", "l": {}, "r": {}}
+      },
+      {
+        "description": "Core functionality w/ realistic size traversal",
```
rpottsoh:
```suggestion
        "description": "Core functionality with realistic size traversal",
```
This may make things a little easier for some tracks using test generators to not have to cope with abbreviations.

exercises/reforestation/canonical-data.json
```diff
@@ -0,0 +1,93 @@
+{
+    "exercise": "reforestation",
+    "version": "1.0.0",
+    "cases": [
+      {
+        "description": "Base case: empty tree",
```
ErikSchierboom: Does the "Base case: " prefix add much value over just describing this as "Empty tree"?

exercises/reforestation/canonical-data.json
```diff
@@ -0,0 +1,93 @@
...
+        "property": "treeTraversals",
+        "input": {
+          "preorder": [],
+          "inorder": []
+        },
+        "expected": {}
+      },
+      {
+        "description": "1 more step than base case",
```
ErikSchierboom: This description refers to a previous case. Perhaps it would be simpler if this would be named "Tree with single node" or something like that, without having to refer to another test case.

exercises/reforestation/canonical-data.json
```diff
@@ -0,0 +1,93 @@
+{
+    "exercise": "reforestation",
+    "version": "1.0.0",
+    "cases": [
+      {
+        "description": "Base case: empty tree",
+        "property": "treeTraversals",
```
ErikSchierboom: I'm not sure about the property name. Yes one has to do a tree traversal, but that is the method and not the goal I think. Maybe something like "treeFromTraversals"?

exercises/reforestation/canonical-data.json
```diff
@@ -0,0 +1,93 @@
...
+        }
+      },
+      {
+        "description": "Traversals must be fully consumed. No extra values.",
+        "property": "treeTraversals",
+        "input": {
+          "preorder": ["a", "b"],
+          "inorder": ["b", "a", "r"]
+        },
+        "expected": {"error": "invalid traversal"}
```
ErikSchierboom: Isn't the error that the traversals array are of incorrect length? If so, maybe the error message could reflect this.

exercises/reforestation/canonical-data.json
```diff
@@ -0,0 +1,93 @@
...
+          "description": "Reject inconsistent traversals of same length",
+          "property": "treeTraversals",
+          "input": {
+              "preorder": ["x","y","z"],
+              "inorder": ["a","b","c"]
+          },
+          "expected": {"error": "traversals are not consistent"}
+      },
+      {
+        "description": "BonusPoints: generate multiple trees",
```
ErikSchierboom: I don't really like the "BonusPoints: " prefix. Also, it is hard to create a test generator for the current test data, as this test case expects an array to be returned for the `treeTraversals` property, whereas the other test cases return an object and not an array. I think there are a couple of options:

- Have all test cases return an array (which would have length equal to one for the previous test cases)
- Name the property differently for this test case
- Remove this additional test case

exercises/reforestation/canonical-data.json
```diff
@@ -0,0 +1,93 @@
...
+                "l": {"v": "f", "l": {}, "r": {}},
+                "r": {"v": "r", "l": {}, "r": {}}
+            }
+        }
+      },
+      {
+        "description": "Traversals must be fully consumed. No extra values.",
+        "property": "treeTraversals",
+        "input": {
+          "preorder": ["a", "b"],
```
ErikSchierboom: There is a bit of inconsistency in the formatting of arrays. Here, `, ` is used to separate the elements of the array, whereas other test cases use `,`. I believe this version (that uses a space) is what most other exercises use, but I'm not completely certain.

exercises/reforestation/canonical-data.json
fidelcoria: Okay. Will rename. I have the feeling most exercises start of with only one test case enabled and let the exerciser enable them as they go. Personally I just activate all of them.

exercises/reforestation/canonical-data.json
fidelcoria: I chose the property name to be idiomatic with Prolog (tree_traversals/3). I like your suggestion.

exercises/reforestation/canonical-data.json
fidelcoria: okay

exercises/reforestation/canonical-data.json
fidelcoria: This is the second suggestion to remove this test case. I have been convinced to remove it.

exercises/reforestation/canonical-data.json
fidelcoria: Ah. Linter didn't catch that, neither did I.

fidelcoria:
@sshine I have been thinking about how to address your suggestions. I tried my best. "Bonus points" section of the README and the corresponding test were removed. I have added a little story to make the exercise less dry. Finally I added a small note at the end. I think it is just the right size hint. Big enough to encourage exerciser not to google, but small enough not to spoil the exercise.

@ErikSchierboom do you resolve conversations or do I? I'm not familiar with etiquette for this feature.

exercises/reforestation/canonical-data.json
sshine: I think you can add it as a Prolog-specific extension to the exercise, since I believe multiple results are generated somewhat automatically in case of duplicates if you choose the right approach?

I'd add this as a hint; you can always add a track-specific test, too.

(I assume that Prolog doesn't have test-generating scripts, so the overhead of adding custom tests in combination with deriving unit tests from canonical data need not be addressed in your track's case.)

exercises/retree/canonical-data.json
```diff
@@ -0,0 +1,57 @@
...
+        "expected": {"v": "a",
+            "l": {"v": "i", "l": {}, "r": {}},
+            "r": {"v": "x",
+                "l": {"v": "f", "l": {}, "r": {}},
+                "r": {"v": "r", "l": {}, "r": {}}
+            }
+        }
+      },
+      {
+        "description": "Traversals must be fully consumed. No extra values.",
```
ErikSchierboom: Maybe keep this consistent with the description of the next, invalid test case:

```suggestion
        "description": "Reject traversals of different length",
```

exercises/retree/canonical-data.json
```diff
@@ -0,0 +1,57 @@
...
+        "description": "Tree with one item",
+        "property": "treeFromTraversals",
+        "input": {
+            "preorder": ["a"],
+            "inorder": ["a"]
+        },
+        "expected": {"v":"a", "l": {}, "r": {}}
+      },
+      {
+        "description": "Core functionality with realistic size traversal",
```
ErikSchierboom: Maybe simplify this to:

```suggestion
        "description": "Tree with many items",
```

or 

```suggestion
        "description": "Large tree",
```

exercises/retree/description.md
```diff
@@ -0,0 +1,25 @@
+Imagine you need to transmit a binary tree to a satellite approaching Alpha
+Centauri and you have limited bandwidth. Since the tree has no repeating
+items it can be uniquely represented by its pre-order and in-order traversals.
+
+Write the software for the satellite to rebuild the tree from the traversals.
+
+A pre-order traversal reads the value of the current node before (hence "pre")
```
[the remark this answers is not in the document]
fidelcoria: "Visits" sounds like the node value is ignored, while "reads" communicated the value is used

fidelcoria:
Added a test for unique items since I don't think it adds too much complexity. And it helps point out subtle details.

exercises/retree/canonical-data.json
```diff
@@ -0,0 +1,66 @@
...
+        "description": "Reject traversals of different length",
+        "property": "treeFromTraversals",
+        "input": {
+          "preorder": ["a", "b"],
+          "inorder": ["b", "a", "r"]
+        },
+        "expected": {"error": "traversals must have the same length"}
+      },
+      {
+          "description": "Reject inconsistent traversals of same length",
```
ErikSchierboom: I think the indentation of this test case is off (too much to the right).

exercises/retree/description.md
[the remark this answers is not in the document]
sshine: OK.

exercises/retree/canonical-data.json
sshine: Other than that and we're ready to merge.

We should have a linting rule for this.

ErikSchierboom: > We should have a linting rule for this.

Yes, we should!

exercises/retree/canonical-data.json
```diff
@@ -0,0 +1,66 @@
+{
+    "exercise": "retree",
```
sshine:
```suggestion
  "exercise": "retree",
```
Since GitHub does not yet support multi-line suggestions, perhaps you can view this as a token of @ErikSchierboom's request to indent the file consistently with two spaces. This appears to be the final request for change before we can merge the exercise.

petertseng:
Fun exercise. Figuring out how to return all the possibilities of a traversal with duplicates was interesting as well (well, at its heart it's just a `concatMap`/`flat_map` of possibilities of the single-tree case)

Note that the number of possible trees of the input where the preorder and inorder traversals consist of N instances of `"a"` is https://oeis.org/A000108.

This comment has nothing to do with whether this exercise should be approved or not, so I will mark it as off-topic after posting it.

exercises/retree/canonical-data.json
```diff
@@ -1,66 +1,66 @@
...
-            "inorder": ["i", "a", "f", "x", "r"]
-        },
-        "expected": {"v": "a",
-            "l": {"v": "i", "l": {}, "r": {}},
-            "r": {"v": "x",
-                "l": {"v": "f", "l": {}, "r": {}},
-                "r": {"v": "r", "l": {}, "r": {}}
-            }
-        }
+      "expected": {"v":"a", "l": {}, "r": {}}
```
petertseng: don't forget this one, where there is missing a space after the colon in `"v":"a"`

(manually found, so I may have missed others)

sshine:
A tangential pun: If we are ever to colonize the stars, eventually we'll want to send send trees in serialized form for terraforming purposes. :-D

exercises/retree/canonical-data.json
ErikSchierboom: Ah yes, good catch. @fidelcoria could you perhaps also fix this?

fidelcoria: Not a problem. there were no other similar occurrences (checked with regex ":[^\s]")

sshine:
I look forward to implementing this exercise for Haskell, Ocaml and SML. :-)

ErikSchierboom:
Great work @fidelcoria! Thanks for being patient with us reviewers :)

fidelcoria:
Thank you! This is my first Open Source contribution. This was a good experience!

ErikSchierboom:
> This is my first Open Source contribution.

You should be proud of yourself then! This is high quality first pull request.

PR #1426, opened 2019-01-04
scale-generator: remove the last interval from test data input

### LINKED ISSUE #1425: scale-generator: test data inconsistency
I think scale-generator test data has error in part of expected result specification and input interval specification.
In many cases input interval contains 7 chars, that means 7 intervals between 2 notes. SO that means that expected result should contain 8 notes in it. But currently expected result contains only 7 notes.

Example:
```
"intervals": "MMmMMMm"
"expected": [ "G", "A", "B", "C", "D", "E", "F#" ]
```
If we replace each comma with interval char then we get following
```
"intervals": "MMmMMMm"
"expected": [ "G" -M- "A" -M- "B" -m- "C" -M- "D" -M- "E" -M- "F#" ]
```

The last 'm' is not for any note from expected result. Or we need to add G into expected result.

So I think there are 2 options only:
1) add one more note to the expected result list
2) remove last interval char from input interval description

aivanov-oneinc:
Fix #1425

### DIFF
--- a/exercises/scale-generator/canonical-data.json
+++ b/exercises/scale-generator/canonical-data.json
@@ -42,7 +42,7 @@
           "property": "interval",
           "input": {
             "tonic": "C",
-            "intervals": "MMmMMMm"
+            "intervals": "MMmMMM"
           },
           "expected": [ "C", "D", "E", "F", "G", "A", "B" ]
         },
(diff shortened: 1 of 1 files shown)


petertseng:
If you are suggesting this to be merged,

* please use a word in https://help.github.com/articles/closing-issues-using-keywords/
* please make sure Travis CI passes before it can be merged

PR #1427, opened 2019-01-04
palindrome-products: remove inconsistencies from canonical data

juhlig:
Finding no palindromes in the given range should not be an error, but result in an empty result. The descriptions of the two tests in question even state "_empty_ result for...", for that matter.

The two tests where the range is invalid (`min>max`) produce a bad error message, stating "invalid input: `min` is X and `max` is Y", which is something of a "go figure". Instead, they should specify the expectation that has been violated, namely that `min` must be smaller or equal to `max`

### DIFF
--- a/exercises/palindrome-products/canonical-data.json
+++ b/exercises/palindrome-products/canonical-data.json
@@ -1,6 +1,6 @@
 {
     "exercise": "palindrome-products",
-    "version": "1.1.0",
+    "version": "1.2.0",
     "cases": [{
             "description": "finds the smallest palindrome from single digit factors",
             "property": "smallest",
@@ -121,9 +121,7 @@
               "min": 1002,
               "max": 1003
             },
-            "expected": {
-                "error": "no palindrome with factors in the range 1002 to 1003"
-            }
+            "expected": []
         },
         {
             "description": "empty result for largest if no palindrome in the range",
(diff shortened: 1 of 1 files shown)


exercises/palindrome-products/canonical-data.json
```diff
@@ -121,9 +121,7 @@
               "min": 1002,
               "max": 1003
             },
-            "expected": {
-                "error": "no palindrome with factors in the range 1002 to 1003"
-            }
+            "expected": []
```
petertseng: As it stands, this doesn't have the same type as the cases that do result in palindromes. This proposal has an **array**. Previous cases have an **object** with an integer `value` and array `factors`.

Using an array here is hiding the fact that we have no `value`, and I predict that not being explicit about this will confuse readers of this file.

Therefore, I suggest if we are going to go this route, it should be:

```json
{
    "value": null,
    "factors": []
}
```

exercises/palindrome-products/canonical-data.json
```diff
@@ -144,7 +140,7 @@
               "max": 1
             },
             "expected": {
-                "error": "invalid input: min is 10000 and max is 1"
```
petertseng: I agree that the old `error` was not useful, and that `min must be <= max` is a significant improvement

This change can be in a separate commit from the change about the cases resulting in no factors, since these two changes are dealing with completely different things. I might go so far as to say they *should* be in separate commits.

juhlig: You are right in both cases, so I will close this PR for now and open two separate ones instead.

PR #1428, opened 2019-01-04
difference-of-squares: hint at formulas

### LINKED ISSUE #1421: difference-of-squares: hint at formulas
The 'Difference of Squares' exercise basically involves computing the sum of the first N integers, and the sum of the first N squares. Although there are standard formulas for both of these, I've noticed that students almost never use them. Virtually without exception, every solution I see uses loops.

Clearly, it's not difficult to find the information you need to solve this problem efficiently; as soon as I hint "Might there be a quicker way to calculate this?", the students almost immediately resubmits the solution using the formulas.

It's not that I'd reasonably expect anyone to remember the formulas from high school; we're learning programming here, not math. My point is that no one seems to think of _looking_ for a formula in the first place. Maybe they think it would be cheating?

Perhaps a hint in the README along these lines would help:

> You are not expected to discover an efficient solution to this yourself from first principles; research is allowed, indeed, encouraged. Finding the best algorithm for the problem is a key skill in software engineering.

bitfield:
The 'Difference of Squares' exercise basically involves computing the sum of the first N integers, and the sum of the first N squares. Although there are standard formulas for both of these, I've noticed that students almost never use them. Virtually without exception, every solution I see uses loops.

This PR adds a hint to the problem description as follows:

> You are not expected to discover an efficient solution to this yourself from first principles; research is allowed, indeed, encouraged. Finding the best algorithm for the problem is a key skill in software engineering.

Fixes #1421 .

### DIFF
--- a/exercises/difference-of-squares/description.md
+++ b/exercises/difference-of-squares/description.md
@@ -9,3 +9,7 @@ The sum of the squares of the first ten natural numbers is
 Hence the difference between the square of the sum of the first
 ten natural numbers and the sum of the squares of the first ten
 natural numbers is 3025 - 385 = 2640.
+
+You are not expected to discover an efficient solution to this yourself from
+first principles; research is allowed, indeed, encouraged. Finding the best
+algorithm for the problem is a key skill in software engineering.


rpottsoh:
I would allow some time to transpire before merging.  Give others a chance to weigh in.

sshine:
I like the change.

PR #1429, opened 2019-01-04
saddle-point: use unambiguous coordinates

### LINKED ISSUE #313: Saddle Point: Uses y, x coordinate values, rather than x, y
> So say you have a matrix like so:

```
    0  1  2
  |---------
0 | 9  8  7
1 | 5  3  2     <--- saddle point at (1,0)
2 | 6  6  7
```

> It has a saddle point at (1, 0).

It appears that the saddle point at 1, 0 is 8 (though the arrow distracts from this coordinates given) which goes against the "unambiguous definition above" as stated in the readme, though 8 is not following the rules of what it means to be a saddle by the definition given.

The point at (0, 1) however, which is 5 does fulfil this definition.  And matches the grid "pointer".

**Edit:** The change, then, may be to change the coordinate shown as (1,0) to (0,1).

bitfield:
The description for 'Saddle Point' illustrates what a saddle point is by showing an example matrix:

```text
    0  1  2
  |---------
0 | 9  8  7
1 | 5  3  2     <--- saddle point at (1,0)
2 | 6  6  7
```

Because there are at least two completely reasonable coordinate systems which could apply here, `(1,0)` is ambiguous (does it mean "column 0, row 1", which is in fact correct, or "row 1, column 0"?)

This PR makes it unambiguous:

```text
    0  1  2
  |---------
0 | 9  8  7
1 | 5  3  2     <--- saddle point at column 0, row 1, with value 5
2 | 6  6  7
```

Fixes #313.

This is also a good complement to the test data, which specifies:

```json
      "expected": [
        {
          "row": 1,
          "column": 0
        }
```

### DIFF
--- a/exercises/saddle-points/canonical-data.json
+++ b/exercises/saddle-points/canonical-data.json
@@ -1,8 +1,8 @@
 {
   "exercise": "saddle-points",
-  "version": "1.3.0",
+  "version": "1.4.0",
   "comments": [
-    "Matrix rows and columns are 0-indexed."
+    "Matrix rows and columns are 1-indexed."
   ],
   "cases": [
     {
@@ -20,8 +20,8 @@
       },
       "expected": [
         {
-          "row": 1,
-          "column": 0
+          "row": 2,
+          "column": 1
         }
       ]
     },
@@ -59,16 +59,16 @@
       },
       "expected": [
         {
-          "row": 0,
-          "column": 1
-        },
-        {
           "row": 1,
-          "column": 1
+          "column": 2
         },
         {
           "row": 2,
-          "column": 1
+          "column": 2
+        },
+        {
+          "row": 3,
+          "column": 2
         }
       ]
     },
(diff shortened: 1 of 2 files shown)


rpottsoh:
What I see looks fine but I am wondering if everything addressed in #313 is fixed here.  This [comment](https://github.com/exercism/problem-specifications/issues/313#issuecomment-236803469) stuck out to me when I quickly scanned #313 just now.  If you decide to pursue re-basing the coordinate system to be 1 based the changes will spill over into the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/saddle-points/canonical-data.json).

rpottsoh:
Matrix was recently updated to follow a more canonical notation for matrix indices.  See #1387.

bitfield:
Fine. As far I can see, the issue of coordinate _base_ is orthogonal to the issue of coordinate _ordering_ (that's why I didn't address it in this PR). I'm happy to update it to follow the Matrix example.

exercises/saddle-points/canonical-data.json
```diff
@@ -151,12 +151,12 @@
...
         {
-          "row": 1,
-          "column": 0
+          "row": 2,
+          "column": 1
         },
         {
-          "row": 3,
-          "column": 0
+          "row": 2,
```
petertseng: 4.

petertseng: https://github.com/exercism/problem-specifications/issues/1433

PR #1430, opened 2019-01-05
palindrome-products: improve canonical data

juhlig:
The two tests where the range is invalid (`min>max`) produce a bad error message, stating "invalid input: `min` is X and `max` is Y", which is something of a "go figure". Instead, they should specify the expectation that has been violated, namely that `min` must be smaller or equal to `max`.
(Addressed by commit dc69569)

Finding no palindromes in the given range should not be an error, but result in an empty result. The descriptions of the two tests in question even state "_empty_ result for...", for that matter.
(Addressed by commit a4871b4)


sshine:
I'll wait a little with merging in case Peter has more to say.

exercises/palindrome-products/canonical-data.json
```diff
@@ -155,8 +157,8 @@
...
             },
             "expected": {
-                "error": "invalid input: min is 2 and max is 1"
+                "error": "min must be <= max"
             }
         }
     ]
-}
\ No newline at end of file
+}
```
rpottsoh: Very minor nit pick.  This change (probably unintentional) is not related and should not be included.

rpottsoh:
Besides my nit picky comment I think this looks good.  Thanks for making these changes @juhlig!

petertseng:
The changes being proposed: I think the meaningful error message is crucial to have. I am neutral on the empty result change (as long as it's clear that there is no result, I don't really care how it's represented in JSON)

There are very good explanations for the changes being proposed in the PR description. I think that these changes should go in the individual commit messages as well - no harm in including them there, right? Even though the one-liner is already understandable, why not include the info in `git log`?

I also think each commit message should be prefixed with `palindrome-products:` just as the PR title is

petertseng:
> I am neutral on

so if others want it, so be it

sshine:
`null` as a generic, empty result seems fine to me.

exercises/palindrome-products/canonical-data.json
juhlig: Ok, it is removed

juhlig:
@petertseng: i prefixed the commit messages with "palindrome-products" and added a body text, much what is given in the PR description.

PR #1434, opened 2019-01-06
saddle-points: fix a Row index

### LINKED ISSUE #1433: saddle-points: test data wrong
https://github.com/exercism/problem-specifications/blob/00d8e30e7592e23c676742c62444b16c53eea124/exercises/saddle-points/canonical-data.json#L158

that needs to be a 4

rpottsoh:
Bumped version.

fixes #1433


petertseng:
EBWODP: 

> and it's not 

that should be "and it's now"

PR #1436, opened 2019-01-08
Acronym: add underscore test case

link2xt:
Many regular expression libraries have a way to detect word boundaries,
but their definition of word characters includes underscores.

`\b` and `\w` metacharacters were designed to detect programming language keywords,
and it is a common mistake to use them to match words in natural languages.

### DIFF
--- a/exercises/acronym/canonical-data.json
+++ b/exercises/acronym/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "acronym",
-  "version": "1.6.0",
+  "version": "1.7.0",
   "cases": [
     {
       "description": "Abbreviate a phrase",
@@ -68,6 +68,14 @@
             "phrase": "Halley's Comet"
           },
           "expected": "HC"
+        },
+        {
+          "description": "underscore emphasis",
+          "property": "abbreviate",
+          "input": {
+            "phrase": "The Road _Not_ Taken"
+          },
+          "expected": "TRNT"
         }
       ]
     }


rpottsoh:
@link2xt thanks for opening the PR.  The Travis CI build failed.  Please check the details and correct the issue that it is reporting.  let me know if you have any questions.

sshine:
Excellent!

Should the acronym of "Foo_Bar" be "FB" and not "F"? And should the acronym of "_Foo" be well-defined?

This does not appear to be a clarification of the exercise, but an extension of it. And it seems to test two different things. So it would warrant two test cases and a major version bump.

I am personally content with not having them, but I'm not seeing a lot of solutions to this exercise. I think your suggestion to teach that `\w` in regex includes `_` is one that benefits many tracks.

And I think that if the acronym of "Foo-Bar" is "FB", you can make the case for `_`. Is that what you're doing?

petertseng:
> `\b` and `\w` metacharacters were designed to detect programming language keywords,
> and it is a common mistake to use them to match words in natural languages.

This has subtly pointed out that up to this point, all inputs have been written in a form that one might see in natural language. I think FOO_BAR is something I would not expect to see in natural language (but I could be convinced about `_BAZ_` if we assume that someone is trying to emulate underlining a word). So FOO_BAR would be a departure from existing test cases for this reason. Be prepared to accept this departure.

The above comment must not be read as either variant of "I {support, oppose} adding `FOO_BAR`".

link2xt:
> This has subtly pointed out that up to this point, all inputs have been written in a form that one might see in natural language.

I guess I will change this to look like markdown-emphasized word then.

link2xt:
@sshine
> Should the acronym of "Foo_Bar" be "FB" and not "F"?

I guess it is better not to test it then, because it is ambiguous, just like "FooBar", which is also not tested. But `_foo_` should not be treated different from `*foo*` simply because your language has regex library.

PR #1438, opened 2019-01-09
rotational-cipher: Added negative shiftKey test cases

ZapAnton:
New test cases ask students to handle negative shift keys, which makes the exercise a little more challenging and could increase learning.

### DIFF
--- a/exercises/rotational-cipher/canonical-data.json
+++ b/exercises/rotational-cipher/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "rotational-cipher",
-  "version": "1.2.0",
+  "version": "1.3.0",
   "comments": [
     "The tests are a series of rotation tests: "
   ],
--- a/exercises/rotational-cipher/canonical-data.json
+++ b/exercises/rotational-cipher/canonical-data.json
@@ -54,6 +54,24 @@
           "expected": "a"
         },
         {
+          "description": "rotate m by negative 1",
+          "property": "rotate",
+          "input": {
+            "text": "m",
+            "shiftKey": -1
+          },
+          "expected": "l"
+        },
+        {
+          "description": "rotate letters by negative 26",
+          "property": "rotate",
+          "input": {
+            "text": "omg",
+            "shiftKey": -26
+          },
+          "expected": "omg"
+        },
+        {
           "description": "rotate capital letters",
           "property": "rotate",
           "input": {


ErikSchierboom:
Thanks for the PR! I must admit though, that I'm not entirely sure if there is much benefit in the added test cases. Maybe someone has a different opinion?

ZapAnton:
@ErikSchierboom for some rationale you may refer to the discussion in https://github.com/exercism/rust/issues/782

ErikSchierboom:
Thanks for the link! It does give some context on why this might be desirable. I'll hold off until more people have chipped in with their thoughts.

ErikSchierboom:
@petertseng @rpottsoh Could I ask you for your opinions?

rpottsoh:
I have been thinking about this and I guess I am not opposed to the additional tests.  However, a negative shift is not covered as an example in README.  Should [description.md](https://github.com/exercism/problem-specifications/blob/master/exercises/rotational-cipher/description.md) be updated to include a negative example?

ZapAnton:
@rpottsoh how about these changes to the README:

- After the 'The letter is shifted for...' sentence add the following sentence:
```text
Notice, however, that you are not restricted to the positive-only keys. The keys from `-26` to `-1` are also valid.
```

- Add the following example in the 'Examples' section:
```text
ROT-1 `m` gives `l`
```

ErikSchierboom:
I've given this some more thought, and I'm a bit on the fence here. I understand that the Rust track wants an exercise to introduce unsigned types, but it also makes this exercise a tiny bit more difficult for students to implement. In general, exercises defined in this repository have a tendency to become _more_ complex, ultimately leading to less simple exercises and many complex ones. 

In this particular case, I'm okay with the change (though not necessarily in favor of it), but I did want to make above remark.

rpottsoh:
@ErikSchierboom do you think `description.md` should also be updated?

ErikSchierboom:
@rpottsoh I do!

ZapAnton:
@rpottsoh , in that case how about moving the additional README part in the separate 'Bonus task' section? For example:
___
### Bonus task
For an additional challenge, you may want to take a look at the last two test cases,  `rotate m by negative 1` and `rotate letters by negative 26`. These tests expect you solution to work with the negative cipher keys from `-1` to `-26`, which, technically, are also valid keys. For example ROT-1 from  `m` gives `l`. Note, that these test cases are optional and you are free to skip them, if you have problems with your solution.
___

This way we isolate the additional test cases from the rest of the README, which allows us to remove it in the future, if the need arises.

ErikSchierboom:
> Edit: This also implies, that the cases in the canonical data are placed at the end, so not to confuse track maintainers too much.

I don't think this will work, as there a quite some languages that auto-generate the exercises. Whether they are listed at the end or not does not make any real difference, expect that one could theoretically exclude the last two test cases. Also, in my experience people will often not read the README after having done the first couple of exercises, and just start coding right away. The fact that the last two test cases are bonus exercises seems to me to be fairly confusing if they are included as regular test cases.

My preferred solution would be to not have these additional test cases added to the canonical data, and to have each track that wants more test cases add them manually.

@F3PiX @petertseng any thoughts here?

ZapAnton:
Alternatively, the new test cases could be switched to a new property (for instance `rotateNegative`) with the `comments` field that would contains something along the lines of 'This is a bonus property cases. Implement them when you see fit.'

F3PiX:
> My preferred solution would be to not have these additional test cases added to the canonical data, and to have each track that wants more test cases add them manually.

That sounds good to me.

rpottsoh:
@ErikSchierboom's suggestion I think is the most sensible. 

Any tracks that are interested in adding cases like these may do so.  Those tracks may also choose to indicate in the test suite via comments or whatever mechanism is convenient, that these two cases are bonuses.

ErikSchierboom:
@ZapAnton I agree. Note that I really do appreciate you taking the time to create this PR, it's just that in this particular case I'm not in favor of this specific change. That said, each track is of course free to add particular tests themselves. This is something that I've done in the C# track too, where I added some additional tests that made more sense there, and would not necessarily benefit other tracks.

Once again, thanks for the PR and I'm sorry the end result may be disappointing to you.

PR #1440, opened 2019-01-10
rest-api - add new test "lender owes borrower same as new loan"

### LINKED ISSUE #1439: rest-api: Missing test case of lender owes borrower same as new loan
rest-api is missing a test case when the previous lender becomes the borrower and borrows the same amount as the current balance thus zeroing the balances and removing the respective owes and owed_by for both parties.  This exercises some necessary code for this case that was previously not covered. Have already tested this and will submit a pull-request to add the new test.

turnkey-commerce:
Resolves #1439 by adding a test case when the previous lender becomes the borrower and borrows the same amount as the current balance thus zeroing the balances and removing the respective owes and owed_by for both parties.   This exercises some necessary code for this case that was previously not covered.  In zeroing balances it is self consistent with the previous two exercises that change balances and whom is owed/owes.

### DIFF
--- a/exercises/rest-api/canonical-data.json
+++ b/exercises/rest-api/canonical-data.json
@@ -1,6 +1,6 @@
 {
     "exercise": "rest-api",
-    "version": "1.0.2",
+    "version": "1.0.3",
     "comments": [
       "The state of the API database before the request is represented",
       "by the input->database object. Your track may determine how this",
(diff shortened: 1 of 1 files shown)


petertseng:
Although the PR desciption contains the string "Resolves issue #1439" as recommended by https://help.github.com/articles/closing-issues-using-keywords/ , it will not automatically close the issue (compare to https://github.com/exercism/problem-specifications/pull/1392 - the word "Resolves" is underlined but is not underlined in this PR), so I suggest putting that string "Resolves issue #1439" on its own line

petertseng:
your PR description includes a good description of the benefits of including this case. your commit message should do the same. Please amend the commit message to do so.

If this is not done, whoever merges this PR please amend the commit message appropriately.

rpottsoh:
> it will not automatically close the issue 

@petertseng I noticed this too and fixed it, then noticed your comment.  Oh well.

exercises/rest-api/canonical-data.json
```diff
@@ -1,6 +1,6 @@
 {
     "exercise": "rest-api",
-    "version": "1.0.2",
+    "version": "1.0.3",
```
rpottsoh:
```suggestion
    "version": "1.1.0",
```

I agree with @petertseng's assessment.

rpottsoh:
At this point the PR should also be squashed along with @petertseng's recommendation to amend the commit message.

exercises/rest-api/canonical-data.json
ErikSchierboom: Same here.

turnkey-commerce:
Thanks for the reviews, I have made the requested change.

PR #1441, opened 2019-01-13
space-age: restrict seconds to fit within 32-bit int range

### LINKED ISSUE #1437: space-age: size of numbers
When working on restructuring the C# track, I found that `space-age` was an ideal candidate for a core exercise that introduces floating-point arithmetic (note: "mathy" exercises are not allowed as core exercises, so `darts` was ruled out). However, I then found out that some of the input values were so large that they didn't fit in C#'s basic `int` type (which max size is `2147483647` (`2^231-1`), which is four bytes). This means that in C#, we will have to introduce not only the floating-point `double` type to the student, but also to bigger numeric types (`long` or `ulong`). For the C# track, this is a problem, as a core exercise should preferrably only introduce a single new concept. By the way, I assume that there will be more tracks for which the default number type has a constrained range similar to the C# one's.

Now I know that is of course very specific to C#, but I was wondering if maybe this exercise is actually teaching two things: working with floating point numbers and working with very large numbers. If people feel there is nothing wrong with the current canonical data, feel free to let me know. I just wanted to put it out there to see if one would be up for modifying `space-age` to use smaller inputs.

Note that these are four input values which value exceeds this maximum value:

`8210123456`
`3210123456`
`3000000000`
`2329871239`

ErikSchierboom:
Closes #1437

### DIFF
--- a/exercises/space-age/canonical-data.json
+++ b/exercises/space-age/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "space-age",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "cases": [
     {
       "description": "age on Earth",
@@ -34,9 +34,9 @@
       "property": "age",
       "input": {
         "planet": "Mars",
-        "seconds": 2329871239
+        "seconds": 2129871239
       },
-      "expected": 39.25
+      "expected": 35.88
     },
     {
       "description": "age on Jupiter",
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
It's been 72 hours and there have been no objections, so I'll merge this.

rpottsoh:
Sort of a posthumous :+1: (since it is already merged).  Thanks for making the change.

PR #1442, opened 2019-01-14
REST-API: Update expected user balance of last test

davidelettieri:
User balance of test "lender owes borrower same as new loan" is not formatted as the previous cases. For example "lender has negative balance" has an expected user balance of zero formatted as "0.0" as all other values while the last test has zero formatted as "0".

### DIFF
--- a/exercises/rest-api/canonical-data.json
+++ b/exercises/rest-api/canonical-data.json
@@ -396,13 +396,13 @@
                                 "name": "Adam",
                                 "owes": {},
                                 "owed_by": {},
-                                "balance": 0
+                                "balance": 0.0
                             },
                             {
                                 "name": "Bob",
                                 "owes": {},
                                 "owed_by": {},
-                                "balance": 0
+                                "balance": 0.0
                             }
                         ]
                     }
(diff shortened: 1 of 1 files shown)


exercises/rest-api/canonical-data.json
```diff
@@ -396,18 +396,18 @@
...
                             }
                         ]
                     }
                 }
             ]
         }
     ]
-}
\ No newline at end of file
+}
```
rpottsoh: A small nit pick, but this change is not related to the other change you made earlier in the file.

rpottsoh:
Oh, and make sure you clear up the issue that Travis CI is having.

exercises/rest-api/canonical-data.json
ErikSchierboom: I've created an issue for this, as it is a common issue with PR's created using the GitHub interface: https://github.com/exercism/problem-specifications/issues/1443

ErikSchierboom:
> Oh, @davidelettieri you should look into and see what the issue is that Travis CI is having.

The issue is that the version number should be incremented.

ErikSchierboom:
Looking great!

PR #1446, opened 2019-01-21
word-count: check counting of empty words

tehsphinx:
Ran across the following solution on the go track that will count empty words on some inputs:

```
package wordcount

import (
	"strings"
	"unicode"
)

// Frequency is the structure of words frequency
type Frequency map[string]int

// WordCount counts the occurrences of each word in a given phrase
func WordCount(input string) Frequency {
	counter := Frequency{}
	phrase := strings.FieldsFunc(input, split)

	for _, word := range phrase {
		word = normalize(word)
		counter[word]++
	}

	return counter
}

func normalize(word string) string {
	word = strings.ToLower(word)

	return strings.TrimFunc(word, func(r rune) bool {
		return !unicode.IsLetter(r) && !unicode.IsNumber(r)
	})
}

func split(r rune) bool {
	return r == ' ' || r == ','
}
```

The added test catches that behaviour.


exercises/word-count/canonical-data.json
```diff
@@ -142,6 +142,18 @@
...
+      "expected": {
+        "one": 1,
+        "two": 1,
+        "three": 1
+      }
     }
   ]
-}
\ No newline at end of file
+}
```
rpottsoh: unrelated change.

tehsphinx: I changed the file here on github. It seems i adds that automatically. I tried to remove it. I can't. Should be there anyway.

Smarticles101: Okay, so how about a commit that adds the newline and then the commit that adds the test. at least that way, it's separate in the commit history

tehsphinx: added separate commit for line ending

rpottsoh:
> also need to update "version", see Travis CI output.

See https://github.com/exercism/problem-specifications#test-data-versioning

PR #1447, opened 2019-01-22
Rename personalTop to personalTopThree

iHiD:
In Ruby the test for HighScores#personalTop is 
```ruby
  def test_personal_top
    scores = [50, 30, 10]
    expected = [50, 30, 10]
    assert_equal expected, HighScores.new(scores).personal_top
  end
```

I solve exercises purely from the tests, and this test didn't give me any information about what the method actually does. I think the method should be called "personalTopThree" and the input for the first test should be an array with greater than three numbers.

### DIFF
--- a/exercises/high-scores/canonical-data.json
+++ b/exercises/high-scores/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "high-scores",
-  "version": "2.0.0",
+  "version": "3.0.0",
   "comments": [
     "This is meant to be an easy exercise to practise: ",
     "* arrays as simple lists",
--- a/exercises/high-scores/canonical-data.json
+++ b/exercises/high-scores/canonical-data.json
@@ -38,16 +38,16 @@
       "description": "Top 3 scores",
       "cases": [
         {
-          "description": "Personal top",
-          "property": "personalTop",
+          "description": "Personal top three from a list of scores",
+          "property": "personalTopThree",
           "input": {
-            "scores": [50, 30, 10]
+            "scores": [10, 30, 90, 30, 100, 20, 10, 0, 30, 40, 40, 70, 70]
           },
-          "expected": [50, 30, 10]
+          "expected": [100, 90, 70]
         },
         {
           "description": "Personal top highest to lowest",
-          "property": "personalTop",
+          "property": "personalTopThree",
           "input": {
             "scores": [20, 10, 30]
           },
@@ -55,7 +55,7 @@
         },
         {
           "description": "Personal top when there is a tie",
-          "property": "personalTop",
+          "property": "personalTopThree",
           "input": {
             "scores": [40, 20, 40, 30]
           },
@@ -63,7 +63,7 @@
         },
         {
           "description": "Personal top when there are less than 3",
-          "property": "personalTop",
+          "property": "personalTopThree",
           "input": {
             "scores": [30, 70]
           },
@@ -71,19 +71,11 @@
         },
         {
           "description": "Personal top when there is only one",
-          "property": "personalTop",
+          "property": "personalTopThree",
           "input": {
             "scores": [40]
           },
           "expected": [40]
-        },
-        {
-          "description": "Personal top from a long list",
-          "property": "personalTop",
-          "input": {
-            "scores": [10, 30, 90, 30, 100, 20, 10, 0, 30, 40, 40, 70, 70]
-          },
-          "expected": [100, 90, 70]
         }
       ]
     },


exercises/high-scores/canonical-data.json
```diff
@@ -1,6 +1,6 @@
 {
   "exercise": "high-scores",
-  "version": "2.0.0",
+  "version": "3.0.0",
   "comments": [
     "This is meant to be an easy exercise to practise: ",
```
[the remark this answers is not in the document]
rpottsoh: `practise` is a verb in British English and `practice` is a noun. https://www.google.com/search?q=practise+vs+practice&rlz=1C1GCEU_enUS821US821&oq=practise&aqs=chrome.2.69i57j0l5.5811j0j1&sourceid=chrome&ie=UTF-8

I think trying to make everything American English (`practice` is a noun/verb) is a pretty monumental expectation.  I can tell what the misspelled word is trying to say :wink:

exercises/high-scores/canonical-data.json
```diff
@@ -38,52 +38,44 @@
       "description": "Top 3 scores",
       "cases": [
         {
-          "description": "Personal top",
-          "property": "personalTop",
+          "description": "Personal top three from a list of scores",
```
rpottsoh: Why does this description have to make a point about the top three when I don't see any of the following cases making the same distinction?

rpottsoh:
For example:
```
        {
          "description": "Personal top when there is a tie",
          "property": "personalTop",
          "input": {
            "thisMany": 3,
            "scores": [40, 20, 40, 30]
          },
          "expected": [40, 40, 30]
        },
```
I don't know that I like `thisMany` as a field name....  But hopefully the intent is understood.

iHiD:
@rpottsoh  I'm totally fine with having a paramater too. I think that's a nicer solution, but it also feels like a bigger change. But I'd :+1: that instead if you want to PR that and people agree.

exercises/high-scores/canonical-data.json
iHiD: Yeah. To explain what the point of the method is.

rpottsoh:
I don't mind making the change, if there is interest.

F3PiX:
The issues raised are definitively valuable points.

However, I'm looking at this issue from a special point of view, that colours my opinion. This exercise was designed for its role in the Track Restructurings. And I'm monitoring its effect, both for students and for mentors, mainly in the Ruby track. To properly evaluate, I'd prefer to change as little as possible. So, changing the name of the method I don't mind. The other changes, to the tests and the default value, I'd rather save until we have had a proper evaluation of the role of the exercise in the track, rather than on the individual 'characteristics' of the exercise. 

In that regard, the biggest issue that I notice with the High Score exercise is the `report` method. In Ruby it turns out that it is hard to mentor in the way it was intended to, and it doesn't work as I hoped. For the same reasons as listed above, I didn't propose that change yet.

How about we add the points about the tests and the parameter as points to the evaluation, and take it from there?  Would that work?

iHiD:
@F3PiX Make sense to me!

rpottsoh:
> How about we add the points about the tests and the parameter as points to the evaluation, and take it from there? Would that work?

All sounds good to me @F3PiX.

ErikSchierboom:
If I am reading the comments correctly, it appears like everybody agrees that the suggested change is desired, is that correct? If so, we can merge this PR and move the open discussion to a new issue I think.

iHiD:
@ErikSchierboom Thanks. I think so too, but maybe @pgaspar could confirm he is happy?

pgaspar:
@iHiD to be clear, we’re talking about the currently proposed change right? `personalTopThree` and the test tweaks. If that’s the case, sounds good to me.

@F3PiX’s comments regarding the test order seem to be blocking though.

F3PiX:
Here's my thinking: I'd rather just change the name, not even the order of the tests. 
My understanding was we had consensus for that, TLDR: https://github.com/exercism/problem-specifications/pull/1447#issuecomment-456615968, 
Meaning: I thought we were waiting for a revert of the test-order. Did I misread that?

F3PiX:
> I solve exercises without reading the README, just TDDing them, and I found it difficult to understand with the tests in the existing order.

Aha cool! My concers was we were making TDDing harder, if we're not, I'm totally okay with the new order. 
Thanks for clarifying.

iHiD:
That makes sense. Thanks, @F3PiX.

I've just merged this but then realised I have no idea if there's actually other steps that need taking to get it live. Do tracks manually regenerate their test suites now?

rpottsoh:
> Do tracks manually regenerate their test suites now?

Yes....  and done.  😃

pgaspar:
I can open a PR on the Ruby track tomorrow if nobody does it first.

PR #1448, opened 2019-01-22
darts: augment test suite

juhlig:
I was just mentoring an exercise where the students solution treated the target as a square (as opposed to a circle), and was surprised that it nevertheless passed the tests.

I found that they only (except for one, that accidentially works still) use coordinates where one of them is `0`, which also works if the solution calculates scores as if the target was a square.

I also noticed that, while the target is supposed to be a circle centered at `(0, 0)`, so that there are valid target areas with negative coordinates, there are no test cases that provide negative coordinates at all.

### DIFF
--- a/exercises/darts/canonical-data.json
+++ b/exercises/darts/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "darts",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "cases": [
     {
       "description": "Return the correct amount earned by a dart landing in a given point in the target problem.",
@@ -9,8 +9,8 @@
          "property": "score",
          "description": "A dart lands outside the target",
          "input": {
-           "x": 15.3,
-           "y": 13.2
+           "x": -9,
+           "y": 9
          },
          "expected": 0
         },
@@ -18,35 +18,53 @@
          "property": "score",
          "description": "A dart lands just in the border of the target",
          "input": {
-           "x": 10,
-           "y": 0
+           "x": 0,
+           "y": 10
          },
          "expected": 1
         },
         {
          "property": "score",
-         "description": "A dart lands in the middle circle",
+         "description": "A dart lands in the outer circle",
          "input": {
-           "x": 3,
-           "y": 3.7
+           "x": 4,
+           "y": 4
+         },
+         "expected": 1
+        },
+        {
+         "property": "score",
+         "description": "A dart lands right in the border between outer and middle circles",
+         "input": {
+           "x": 5,
+           "y": 0
          },
          "expected": 5
         },
         {
          "property": "score",
-         "description": "A dart lands right in the border between outside and middle circles",
+         "description": "A dart lands in the middle circle",
          "input": {
-           "x": 0,
-           "y": 5
+           "x": 0.8,
+           "y": -0.8
          },
          "expected": 5
+        },
+        {
+         "property": "score",
+         "description": "A dart lands right in the border between middle and inner circles",
+         "input": {
+           "x": 0,
+           "y": -1
+         },
+         "expected": 10
         },
         {
          "property": "score",
          "description": "A dart lands in the inner circle",
          "input": {
-           "x": 0,
-           "y": 0
+           "x": -0.1,
+           "y": -0.1
          },
          "expected": 10
         }


juhlig:
The tests already combine coordinate signedness and target area/shape into one. There is a test for well away from the target, outside but near the target, and then, roughly, for each ring, it tests a hit on the very border, in the middle of the ring at a zero-coordinate, in middle of the ring at a 45° angle, and in the ring but near the border to another ring at a 45° angle (this is where tests would fail if the target was treated as a square instead of a circle in a solution). The coordinates "rotate" around at the same time, starting at negative x and positive y, to positive x and positive y, to positive x and negative y, and so on.

I'm not sure which to drop. The way-outdside-the-target test could go. Other obvious candidates are those for the middle of the ring at a zero-coordinate, they serve no definite purpose, but there are only 2 of those. Next could be those at the middle of a ring at a 45° angle, there are 2 of those as well. The tests for hits at the very border should stay, I would say, as should the tests for hits near the border to another ring. And the center hit should be there as well, since where is the fun in a dart match if you don't have one of those ;).

ErikSchierboom:
The way I perhaps would go about this is to modify some of the existing test cases to use negative values. E.g. one you use `-3` here: https://github.com/exercism/problem-specifications/blob/master/exercises/darts/canonical-data.json#L30. This would probably mean also adding one new test cases, but I would really rather keep the number of test cases fairly small.

juhlig:
Ok, can do. But what actually prompted me to make this PR was that the existing test cases allowed for the target to be treated as a square (instead of a circle, as a dart target normally is, and is also mentioned in the instructions), and still pass. So, what I would do now, to keep the number of test cases small, is use the existing test cases with different coordinates, namely negative/positive mixed at points that will pass if the target is treated as a circle but fail if treated as a square. Would that be ok?

ErikSchierboom:
> [...] use the existing test cases with different coordinates, namely negative/positive mixed at points that will pass if the target is treated as a circle but fail if treated as a square

To me, that is a great solution. I'd be curious to hear what @rpottsoh or @petertseng think about the best approach?

rpottsoh:
Alright....  I agree that negative values for X and Y be introduced.  I think of a dart board more as a Polar system more than a Cartesian system.  Since this is a mathy exercise I got to thinking that maybe it would be more interesting if the coordinates were Polar... There would not be any negative values....  A quick search lead me [here](https://www.mathsisfun.com/polar-cartesian-coordinates.html).  A nice explanation of converting from Cartesian to Polar and back....  Since we are thinking of a Cartesian system then we obviously should be considering the possibility of X and/or Y being negative as the coordinates locate a thrown dart having landed in one of the 4 quadrants of the Cartesian system.....

I am not quite sure at this point if it is better to add tests, alter existing tests, or both....  Perhaps as long as there is at least one case for each of the 4 quadrants would be a good enough change?

ErikSchierboom:
>  Perhaps as long as there is at least one case for each of the 4 quadrants would be a good enough change?

I agree! @juhlig do you think that is do-able?

juhlig:
I don't fully agree. There should at least be tests for hits at the very borders between the rings (border hit counts as a hit to the more inward ring)), as this ensures the proper usage of `<`and `>` vs `>=` and `=<`.

ErikSchierboom:
> There should at least be tests for hits at the very borders between the rings

Alright. Perhaps you can add those as new test cases then?

exercises/darts/canonical-data.json
```diff
@@ -9,44 +9,62 @@
          "property": "score",
          "description": "A dart lands outside the target",
          "input": {
-           "x": 15.3,
-           "y": 13.2
+           "x": -9,
+           "y": 9
```
rpottsoh: This almost had me thinking it was inside the target.  I calculate the distance from the center to be  ~12.7 so checks out for this case.  👍

exercises/darts/canonical-data.json
```diff
@@ -9,44 +9,62 @@
...
+         "expected": 1
+        },
+        {
+         "property": "score",
+         "description": "A dart lands right in the border between outer and middle circles",
+         "input": {
+           "x": 5,
+           "y": 0
+         },
+         "expected": 5
```
petertseng: note that here, the dart is treated as being in the smaller circle (here, the middle one) rather than the larger one (outer)

exercises/darts/canonical-data.json
```diff
@@ -9,44 +9,62 @@
...
         {
          "property": "score",
-         "description": "A dart lands right in the border between outside and middle circles",
+         "description": "A dart lands right in the border between middle and inner circles",
          "input": {
            "x": 0,
-           "y": 5
+           "y": -1
          },
          "expected": 5
```
petertseng: here, the dart is treated as being in the larger circle (middle) rather than the smaller (inner)

ErikSchierboom:
> the treatment of darts landing on borders doesn't seem consistent, based on what I have noted. I request an explanation of the rules, or a change to be consistent

@juhlig @petertseng raises some valid points. Could you perhaps look at his comments?

exercises/darts/canonical-data.json
juhlig: You are right, this should count as a hit to the inner circle. Fixed in my last commit.

ErikSchierboom:
@petertseng Could you see if your comments have been addressed?

petertseng:
I already ghave.

ErikSchierboom:
Ah sorry, I didn't see your name in the list of approvers.

petertseng:
it is a little too easy to miss it, since as of about 1-2 days ago dismissals still do not send a notification

ErikSchierboom:
Merged. Thanks @juhlig!

PR #1450, opened 2019-02-01
Hamming: Add a test case to avoid wrong recursion solution

tqa236:
Add an extra test case about one empty strand based on the discussion in exercism/haskell#796 to avoid wrong recursion solution.

### DIFF
--- a/exercises/hamming/canonical-data.json
+++ b/exercises/hamming/canonical-data.json
@@ -81,7 +81,7 @@
         "strand1": "",
         "strand2": "G"
       },
-      "expected": {"error": "left and right strands must be of equal length"}
+      "expected": {"error": "left strand must not be empty"}
     },
     {
       "description": "disallow right empty strand",
--- a/exercises/hamming/canonical-data.json
+++ b/exercises/hamming/canonical-data.json
@@ -90,7 +90,7 @@
         "strand1": "G",
         "strand2": ""
       },
-      "expected": {"error": "left and right strands must be of equal length"}
+      "expected": {"error": "right strand must not be empty"}
     }
   ]
 }
--- a/exercises/hamming/canonical-data.json
+++ b/exercises/hamming/canonical-data.json
@@ -75,13 +75,22 @@
       "expected": {"error": "left and right strands must be of equal length"}
     },
     {
-  "description": "disallow only one empty strand",
-  "property": "distance",
-  "input": {
-    "strand1": "",
-    "strand2": "G"
-  },
-  "expected": {"error": "left and right strands must be of equal length"}
-}
+      "description": "disallow left empty strand",
+      "property": "distance",
+      "input": {
+        "strand1": "",
+        "strand2": "G"
+      },
+      "expected": {"error": "left and right strands must be of equal length"}
+    },
+    {
+      "description": "disallow right empty strand",
+      "property": "distance",
+      "input": {
+        "strand1": "G",
+        "strand2": ""
+      },
+      "expected": {"error": "left and right strands must be of equal length"}
+    }
   ]
 }
--- a/exercises/hamming/canonical-data.json
+++ b/exercises/hamming/canonical-data.json
@@ -1,6 +1,6 @@
 {
 "exercise": "hamming",
-"version": "2.2.0",
+"version": "2.3.0",
   "comments": [
     "Language implementations vary on the issue of unequal length strands.",
     "A language may elect to simplify this task by only presenting equal",
@@ -73,6 +73,15 @@
         "strand2": "AGTG"
       },
       "expected": {"error": "left and right strands must be of equal length"}
-    }
+    },
+    {
+  "description": "disallow only one empty strand",
+  "property": "distance",
+  "input": {
+    "strand1": "",
+    "strand2": "G"
+  },
+  "expected": {"error": "left and right strands must be of equal length"}
+}
   ]
 }


cmccandless:
What's interesting about this test case is that is far more likely to fail in Haskell (and other languages that support pattern matching on function parameters) than other languages (that don't). I was trying to think of a solution similar to the example provided in exercism/haskell#796 without using pattern matching... it seems a bit contrived, but possible?

```Python
def distance(strandA, strandB):
    if len(strandA) == 0 and len(strandB) == 0:
        return 0
    headA = strandA.pop(0)  # This would raise an unexpected IndexError in the new test case
    headB = strandB.pop(0)
    if len(strandA) != len(strandB):
        raise ValueError()
    return (1 if headA != headB else 0) + distance(strandA, strandB)
```

Still a good test case though!

I wonder: is test case `disallow second strand longer` now redundant?

rpottsoh:
> I wonder: is test case disallow second strand longer now redundant?

Or `disallow first strand longer` for that matter.  Help me understand how these two cases are not enough?

From the example code in @cmccandless comment:

Why is the length equality of the strands _not_ checked first before anything else?   I am not a FP person so I would like to understand better what is happening in Haskell that makes  what looks like a redundant change necessary.  Is this an issue that really affects all tracks?

exercises/hamming/canonical-data.json
```diff
@@ -73,6 +73,15 @@
...
+    },
+    {
+  "description": "disallow only one empty strand",
+  "property": "distance",
+  "input": {
+    "strand1": "",
+    "strand2": "G"
+  },
+  "expected": {"error": "left and right strands must be of equal length"}
+}
```
rpottsoh: indentation of lines 76 through 85 should match indentation of cases above.

rpottsoh:
I think the other two cases should remain, even though see seem redundant.  They are not redundant to each other.  Perhaps the new case should be stated this way:
```
"description": "disallow left empty strand",
"property": "distance",
  "input": {
    "strand1": "",
    "strand2": "G"
  },
"expected": {"error": "left and right strands must be of equal length"}
```
then also add:
```
"description": "disallow right empty strand",
"property": "distance",
  "input": {
    "strand1": "G",
    "strand2": ""
  },
"expected": {"error": "left and right strands must be of equal length"}
```
We could also, optionally, change the error message to be more informative.  ie. `left strand must not be empty` and `right strand must not be empty`

tqa236:
Thank you for your review and suggestion. Should I push a new commit to this pull request or wait for some kind of final decisions?

rpottsoh:
There is no harm in waiting.  It would be interesting to get @petertseng's and @ErikSchierboom's opinions.

sshine:
Sure,

1. Fix whitespace
2. Add the second test case
3. Name them as

ErikSchierboom:
Although I think this particular problem is not that likely to occur in other languages, I don't much mind the addition of the two new test cases. I would go with @rpottsoh' suggestions: adding two new test cases and giving them an error message stating that the input must not be empty.

sshine:
It is just as likely to occur in Standard ML and Ocaml. So I encouraged @tqa236 to submit this here first.

I would also encourage a lower entry bar for new contributors. :-)

ErikSchierboom:
> I would also encourage a lower entry bar for new contributors. :-)

I'm not entirely sure I understand what you mean by this comment :)

Just to be sure, I think the appropriate action here is the list you mentioned earlier:

> Fix whitespace
> Add the second test case
> Name them as @rpottsoh suggested

Would that work for you @tqa236?

tqa236:
It’s ok for me. I will do those 3 tasks you mentioned. Please wait for a few days

tqa236:
I updated as you said. Please review it.

exercises/hamming/canonical-data.json
```diff
@@ -73,6 +73,24 @@
...
       "expected": {"error": "left and right strands must be of equal length"}
+    },
+    {
+      "description": "disallow left empty strand",
+      "property": "distance",
+      "input": {
+        "strand1": "",
+        "strand2": "G"
+      },
+      "expected": {"error": "left and right strands must be of equal length"}
```
rpottsoh:
```suggestion
      "expected": {"error": "left strand must not be empty"}
```

exercises/hamming/canonical-data.json
```diff
@@ -73,6 +73,24 @@
...
+    },
+    {
+      "description": "disallow right empty strand",
+      "property": "distance",
+      "input": {
+        "strand1": "G",
+        "strand2": ""
+      },
+      "expected": {"error": "left and right strands must be of equal length"}
```
rpottsoh:
```suggestion
      "expected": {"error": "right strand must not be empty"}
```

rpottsoh:
@ErikSchierboom still OK with PR?

ErikSchierboom:
@rpottsoh 👍

ErikSchierboom:
Thanks @tqa236!

sshine:
> > I would also encourage a lower entry bar for new contributors. :-)

> I'm not entirely sure I understand what you mean by this comment :)

Sorry I never got back to this, @ErikSchierboom.

I thought that the discussion of keeping/merging redundant cases might evolve so that it encompassed the original issue, discouraging @tqa236 from finalizing the change on the Haskell track. Similarly, in the hope of a returning contributor, I brought Hamming on the Haskell track up to 2.2.0 in exercism/haskell#797 while waiting for this PR.

I'm sorry if this left an impression of bullying or arguing "too many cooks" when we're generally *happy* to see people participate in the problem-specifications issues. (I haven't been able to keep up with this participation myself.)

ErikSchierboom:
Right, thanks for the explanation! I was just unsure how to interpret said statement, no harm done :)

PR #1453, opened 2019-02-04
list-ops: add test case for reversing nested lists

cmccandless:
The following implementation of `reverse` passes all existing tests but fails the new test:

```Python
# Passes all existing tests for append()
def append(base_list, to_append):
    if isinstance(to_append, list):
        for x in to_append:
            base_list.append(x)
    else:
        base_list.append(to_append)
    return base_list

# Passes existing test for length()
def length(base_list):
    return foldl(lambda a, b: a + 1, base_list, 0)

# Passes existing tests for foldr()
def foldr(function, base_list, acc):
    while length(xs):
        acc = function(base_list.pop(), acc)
    return acc

def reverse(base_list):
    return foldr(lambda b, a: append(a, b), base_list, [])
```

### DIFF
--- a/exercises/list-ops/canonical-data.json
+++ b/exercises/list-ops/canonical-data.json
@@ -212,7 +212,7 @@
           "description": "empty list",
           "property": "reverse",
           "input": {
-            "list": []-patch-1
+            "list": []
           },
           "expected": []
         },
--- a/exercises/list-ops/canonical-data.json
+++ b/exercises/list-ops/canonical-data.json
@@ -225,7 +225,7 @@
           "expected": [7, 5, 3, 1]
         },
         {
-          "description": "list of nested lists",
+          "description": "list of lists",
           "property": "reverse",
           "input": {
             "list": [[1, 2], [3], [], [4, 5, 6]]
--- a/exercises/list-ops/canonical-data.json
+++ b/exercises/list-ops/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "list-ops",
-  "version": "2.3.0",
+  "version": "2.4.0",
   "comments": [
     "Though there are no specifications here for dealing with large lists,",
     "implementers may add tests for handling large lists to ensure that the",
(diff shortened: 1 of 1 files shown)


cmccandless:
Above implementation produces `[4, [5, 6], 2, 3, 1, 0]` instead of expected.

cmccandless:
> this case is the first case proposed where the list is heterogeneous

The input could be modified to the following and achieve the same coverage:
```JSON
"input": {
  "list": [[1, 2], [3], [], [4, 5, 6]]
},
"expected": [[4, 5, 6], [], [3], [1, 2]]
```

This input matches the other existing `list of lists` test case.

petertseng:
> The input could be modified to the following and achieve the same coverage:
> `[[1, 2], [3], [], [4, 5, 6]]`

I recommend doing so, so as to be able to share with languages with homogeneous lists

cmccandless:
@petertseng done.

exercises/list-ops/canonical-data.json
```diff
@@ -223,6 +223,14 @@
             "list": [1, 3, 5, 7]
           },
           "expected": [7, 5, 3, 1]
+        },
+        {
+          "description": "list of nested lists",
```
petertseng: I dispute any assertion that the input is a list of nested lists, because each element of the list is not a nested list (there are no additional levels of lists in each element of the input list). So it seems that `list of lists` is a better name.

I also refer to the standard https://github.com/exercism/problem-specifications/blob/master/README.md#test-data-format-canonical-datajson "the description should not simply explain **what** each case is (that is redundant information) but also **why** each case is there." , and suggest that the description mention something about not flattening or reversing any inner lists.

cmccandless: Will `list of lists is not flattened` suffice?

petertseng: :+1:

cmccandless:
This may be merged in 24 hours by anyone with merge permissions unless further changes are requested.

cmccandless:
EBWODP: please squash merge.

ErikSchierboom:
Squashed and merged. Thanks @cmccandless!

PR #1460, opened 2019-02-12
simple-cipher: test that messages longer than the key can be decoded

paparomeo:
I just come across a solution for the "Simple Cipher" exercise from a student, which could encode messages longer than the key but failed to _decode_ messages longer than the key, and yet, because we haven't a test to cover this, all tests would pass giving an illusion of correctness.

This pull request adds a test to also check that a message longer than the key can be decoded.

PR #1462, opened 2019-02-15
simple-cipher: Fix property name and major bump

paparomeo:
#1460 erroneously named a property `plaintext`. This fixes it to `ciphertext`.

Also the version bump is made a major bump according to versioning guidelines.


ErikSchierboom:
Thanks @paparomeo!

PR #1465, opened 2019-02-23
leap: Add test case for year divisible by 2 but not by 4

G-Rath:
This adds a new test that will fail if students use `year % 2 === 0` instead of `year % 4 === 0`.

### DIFF
--- a/exercises/leap/canonical-data.json
+++ b/exercises/leap/canonical-data.json
@@ -1,12 +1,20 @@
 {
   "exercise": "leap",
-  "version": "1.4.0",
+  "version": "1.5.0",
   "cases": [
     {
       "description": "year not divisible by 4: common year",
       "property": "leapYear",
       "input": {
         "year": 2015
+      },
+      "expected": false
+    },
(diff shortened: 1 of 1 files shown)


bitfield:
Nice one, @G-Rath! Thanks a lot!

petertseng:
An implementation that solely does `year % 2 == 0` would already give incorrect answers on two existing cases:

```
== anti-verify 2 only leapYear ==
2. year divisible by 100, not divisible by 400: common year: Got true instead of false
4. year divisible by 200, not divisible by 400: common year: Got true instead of false
```

Therefore, this case is primarily useful for adding additional clarity before involving divisibility by 100 or 400. Therefore, if this new case is added, it has to be before any cases involving 100 or 400. And this PR has indeed placed it before them.

An implementation that does `year % 2 == 0 && (year % 100 != 0 || year % 400 == 0)` will give correct answers on all existing cases and an incorrect answer on only this new case, so this new case potentially adds coverage for those implementations.

Issue #1458, opened 2019-02-10
Proposal for new exercise: ResistorColor (Resistor Series #1)


F3PiX:
Resistors have color bands. Each color represents a number. Resistor can have up to 6 bands, and together they describe the resistance value in Ohms, together with the tolerance and the temperature coefficient.  

![image](https://user-images.githubusercontent.com/6314687/52533780-3eb5d300-2d39-11e9-8f69-5a726735c294.png)

We're planning to have a series of Resistor-* exercises, that are building up to decoding 4-6 band resistor color codes.

1) The new exercise ResistorColor
 The first and most basic one is just reading the (first) band's color from an array and return its numeric value. 
![schermafbeelding 2019-02-10 om 13 15 39](https://user-images.githubusercontent.com/6314687/52533577-0791f280-2d36-11e9-92a3-9556cb5a1885.png)

See PR #1457 

2) This is supposed to be part of a series of Resistor related exercises, in ascending complexity:
- Reading the first two bands and combine them to a number 
- Reading TriBand, and returning a value in Ω (Ohms) or K Ω 
- Reading 4 to 6 bands, returning the value in Ohms, the tolerance and the temperature coefficient. In those resistors, the meaning of the 3rd and 4th band is depending on the total number of bands. 
Coming soon!

rpottsoh:
1. Will you be getting into combining resistors in series and/or in parallel?
1. Will the solution of one exercise be required  before proceeding to the next?
1. Will this first exercise be recommended as a core exercise and then each of the follow-ons as a side?

Looking forward to learning more about what the forward looking plan might be.

F3PiX:
@rpottsoh :-))))
1) Not even thought of it. Great idea! 
2) My thinking is: not in the first exercises. But in the more advanced ones, they can be, yes. Currently only core exercises can unlock other, so it also depends on how a track implements them. But for instance in Ruby, including one in the other can be a nice showcase for mixins.
3) Also, depending on the track. For instance for Ruby, the second one is more appropriate as a core exercise than the first one. I can imagine that it's good to have the second one as a core, if you want to have it included. On the other hand, it's also possible that the student writes the code to be mixed in at the moment that they need it in the later exercise. In general, I don't think we should force rules on how tracks implement such a series, but you're right, it would be nice to offer a optional plan. 👌  

Thoughts?

rpottsoh:
> 2\. including one in the other can be a nice showcase for mixins

[...]

PR #1466, opened 2019-02-24
resistor-colors: add new exercise

### LINKED ISSUE #1464: Resistor series # 2 : ResistorSignificants
The ResistorColor exercise is meant as [the start of a series of resistor decoding exercises](https://github.com/exercism/problem-specifications/issues/1458). 
Here's number 2: the Resistor Significants exercise.

For this exercise, a given input represents the colors of the first 2 bands of a resistor, for instance `['brown', 'black']` and should return the combined number value `10`.

This are real life examples of such color combinations: 
image
translated to tests:
- brown and black -> 10
- orange and orange -> 33, 
- blue and grey -> 68
- yellow and violet -> 47 

About the naming:
The two first bands of resistors are also referenced to as 'significant figures'. The other bands have a different meaning (number of zeros, percentages).

ErikSchierboom:
Closes #1464.


exercises/resistor-significants/description.md
```diff
@@ -0,0 +1,18 @@
...
+- Orange: 3
+- Yellow: 4
+- Green: 5
+- Blue: 6
+- Violet: 7
+- Grey: 8
+- White: 9
+
+Combining the numbers of the first two bands gives one its significant value. As an example, if the first band is blue (6) and the second band is grey (8), the significant value is 68. 
+
```
F3PiX:
```suggestion
Combining the numbers of the first two bands gives one the significant figures of the value in Ohm.
```
And in the example: the value is `68`

F3PiX: Or: the value to be returned is `68` . (To avoid the problem of what we should call this part. I'm afraid significant value may be misleading or confusing for its meaning in statistics. )

exercises/resistor-significants/description.md
```diff
@@ -0,0 +1,18 @@
+Resistors have color coded bands, where each color maps to a number. The first 2 bands of a resistor have a simple encoding scheme: each color maps to a single number.
```
F3PiX:
```suggestion
Resistors have color coded bands that encode their value in Ohm. The bands have a simple encoding scheme: each color maps to a single number. The first 2 bands represent the significant figures of a resistor's value in Ohm. 
[Significant figures](https://en.wikipedia.org/wiki/Significant_figures) or significant digits is a concept used in measurements to indicate the numbers without trailing zero's.   
```

exercises/resistor-significants/canonical-data.json
```diff
@@ -0,0 +1,42 @@
+{
+  "exercise": "resistor-significants",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "Brown and black",
+      "property": "significantValue",
```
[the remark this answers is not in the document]
F3PiX: Sorry, I didn't know that when I wrote the issue! I updated the issue accordingly.

ErikSchierboom:
@F3PiX I've updated the PR. I am wondering if people won't expect an array/list for the expected value, as the property "significantFigures" does seem to imply that. But maybe they won't :)

exercises/resistor-significants/description.md
```diff
@@ -13,6 +14,6 @@ These colors are encoded as follows:
 - Grey: 8
 - White: 9
 
-Combining the numbers of the first two bands gives one its significant value. As an example, if the first band is blue (6) and the second band is grey (8), the significant value is 68. 
+Combining the numbers of the first two bands gives one the significant figures of its value in Ohm. As an example, if the first band is blue (6) and the second band is grey (8), the combined significant figures in Ohm equal the number 68.
```
F3PiX: 👍 Better, thanks. I think 'in Ohm' can go too, to make it even more clear.

exercises/resistor-significants/description.md
```diff
@@ -13,6 +14,6 @@ These colors are encoded as follows:
...
```
F3PiX: 👍 Better, thanks. I think 'in Ohm' can go too, to make it even more clear.

rpottsoh: Ohm should be Ohms.

F3PiX: I googled that ;-) But I trust you way more than google.

F3PiX:
@ErikSchierboom I see what you mean re: the property. 🤔 Is `Value` a better name then? 
`ResistorSignificants#value`  makes sense to me.

ErikSchierboom:
I like `value`!

exercises/resistor-significants/description.md
ErikSchierboom: @rpottsoh should we use Ohms everywhere?

F3PiX:
😍

exercises/resistor-significants/description.md
rpottsoh: Think of Ohms as cookies.  Conversationally you wouldn't say you have 5 cookie, but rather 5 cookies.  How many cookies in a dozen?....  Having singular Ohm in your sentence(s) may not be technically wrong, but it "sounds" odd when spoken out loud.  How many Ohms is the resistor?  It is 5 Ohms or it is 1 Ohm or it is 1/2 [an] Ohm.  Use plural when the resistor is more than 0 and <= 1.  Then there is a twist, when asking a question....  Do you have a 5 Ohm resistor I can borrow?  Ohm is singular.

rpottsoh:
What Topic(s) are you envisioning for this exercise?

ErikSchierboom:
> What Topic(s) are you envisioning for this exercise?

It depends a bit on the language involved, but what I envision is that students will now have to learn how to combine to individual numbers into a single number. There are several ways to do that, so it can be quite interesting.

iHiD:
Could you hold off merging this until @kytrinyx has had chance to take a look pls?

ErikSchierboom:
Yes please wait. I think I may have missed part of the original plan for the series.

ErikSchierboom:
I've updated the PR. What changed is that instead of passing the colors as separate arguments, they are now passes as an array. This keeps them in line with the array-theme of the series.

exercises/resistor-significants/description.md
cmccandless: > Then there is a twist, when asking a question.... Do you have a 5 Ohm resistor I can borrow? Ohm is singular.

:+1: Like a 5-cookie jar.

rpottsoh:
I have been reading over the comments here and wonder if perhaps `colorCode` might not be a better property name than `value`?  Perhaps you are holding back using `colorCode` for when the entire code for a given resistor is to be decoded?

exercises/resistor-significants/description.md
rpottsoh: Ha, thanks for keeping the analogy going.  :smile:

ErikSchierboom:
> Perhaps you are holding back using `colorCode` for when the entire code for a given resistor is to be decoded?

Actually, we already used it: https://github.com/exercism/problem-specifications/blob/master/exercises/resistor-color/canonical-data.json :)

kytrinyx:
I'm a bit late to the party here, and I am sorry for delaying the process.

I'm going to take a really big step backwards here, and not talk about this resistor exercise, but rather I'd like to talk about something which @iHiD and @nicolechalmers and I have been mulling over for some time now.

To start, I'd ask that you read this issue: https://github.com/exercism/problem-specifications/issues/1470.

To summarize: we think that one fundamental change that we could make to Exercism to make the site more engaging is to ensure that each exercise is defined by a story. This helps add color and an element of liveliness (fun?) to the exercise, but much more than that, it helps avoid a situation where an exercise feels overwhelming or foreign or "too hard" (even when it is actually simple). We've found that sometimes the words that we use, or the way that we describe things can make something seem much harder than they actually are.

So for the resistor exercises: what does it look like to strip away all the surrounding details from the technical problem? At it's very core, what is each of the resistor exercises asking you to do?

ErikSchierboom:
> At it's very core, what is each of the resistor exercises asking you to do?

I think @F3PiX will probably know this better than me, but I figure we could perhaps write this exercise from the viewpoint of a mechanic or something like that? Someone that sees a resistor, and then has to try and figure out its use, or something like that.

F3PiX:
First, I'm a big fan of adding stories to exercises, so 💯 .

Then, as so many tracks are really suffering by a lack of easy exercises like this, especially this easiest iterator, I'd vote for not holding up this PR because of the story. We could easily work on the story-telling after this exercise is launched, we can leave the [exercise's issue](#1466) open for that or reopen the [series introduction issue](https://github.com/exercism/problem-specifications/issues/1458), and delete this exercise as soon as there is a better story telling alternative. 

What do you think @kytrinyx?

kytrinyx:
To be honest, I find this exercise _really_ difficult to understand. I don't know anything about resistors, and reading through this I am having a bit of anxiety trying to get past the unfamiliar jargon to figure out what I actually would have to do to solve it.

F3PiX:
That's not good!! 
I'm moving this conversation to the issue. I tried to block the PR for merging, but I can be overruled ;-) 
Anyway, we're not merging until we get the story sorted out.

ErikSchierboom:
Sure. Let me know what I need to change and I'll make the adjustments. Maybe we can simplify this by replacing the resistors with some sort of color coded bracelet? Just thinking out loud.

F3PiX:
Haha, I had the same thought, but made it [necklaces](https://github.com/exercism/problem-specifications/issues/1464#issuecomment-467632170), not bracelets ;-)

kytrinyx:
❤️ I already love the jewelry direction. I'll discuss over in the other issue.

kytrinyx:
Actually, having read @F3PiX's alternate description that just "un-technifies" the resistor thing a bit, is exactly the right amount of change we need to get this merged. That way we can think about the story in more depth as a completely separate issue.

exercises/resistor-significants/description.md
```diff
@@ -0,0 +1,19 @@
...
+- Orange: 3
+- Yellow: 4
+- Green: 5
+- Blue: 6
+- Violet: 7
+- Grey: 8
+- White: 9
+
+Combining the numbers of the first two bands gives one the significant figures of its value in Ohms. As an example, if the first band is blue (6) and the second band is grey (8), the combined significant figures' value is the number 68. 
+
```
F3PiX:
```suggestion
If you would build anything with a Raspberry Pi or an Arduino, one of the first components you need are resistors. In this exercise, it doesn't matter what a resistor is or does, the only thing that matters here is that they come with different values. And that they are tiny. So tiny that it's hard to read such a value if it were printed on the thing. To solve that, fabricators use color codes that reveals their value. Which is great, but it makes it hard to see the actual value.
You are going to solve this with a program that you can feed two colors, and it will translate the color to a number. For instance "brown" and "black" are code for `1` and `0`, and together they represent the value `10`.

The band colors are encoded as follows:

Black: 0
Brown: 1
Red: 2
Orange: 3
Yellow: 4
Green: 5
Blue: 6
Violet: 7
Grey: 8
White: 9

```
So, I removed all the tech talk ;-) including the part about the Ohms value. I feel it's important to not suggest that we're decoding Ohms value, because that would confuse people who do know about resistors.

exercises/resistor-significants/description.md
```diff
@@ -0,0 +1,19 @@
...
+More information on the color encoding of resistors can be found in the [Electronic color code Wikipedia article](https://en.wikipedia.org/wiki/Electronic_color_code)
```
F3PiX: And shall we take this out too? It's not helpful in solving the exercise.

exercises/resistor-significants/metadata.yml
```diff
@@ -0,0 +1,4 @@
+---
+blurb: "Convert the colors of a resistor's first two bands to the significant figures of its value in Ohms."
```
F3PiX:
```suggestion
blurb: "Convert color codes, as used on resistors, to a numeric value."
```
Or leave the resistor out here?

F3PiX:
The one thing left is the name. Significants has no significance any more.

iHiD:
Maud - that's so much more like what I feel exercises should be like. Thank you!!

exercises/resistor-significants/metadata.yml
kytrinyx: I think it's fine to leave it in here.

exercises/resistor-significants/description.md
kytrinyx: Yeah, let's take it out. People can google it if they want to!

exercises/resistor-significants/description.md
kytrinyx: I agree! Good choice.

kytrinyx:
For the name, we could go with something super basic like "color codes" maybe?

rpottsoh:
I don't know anything about bracelets or necklaces....  Resistors was easy...  😉   But I am content to see where this all goes.  The journey can be as much fun as the destination, or more than.

ErikSchierboom:
Great discussion, and I like the new direction! I've applied the review comments and pushed a new version. There are two open things:

- What should be the new name of the exercise? @kytrinyx suggested `color-codes`, which I like. What do you think @F3PiX?
- We should apply the lessons of this discussion to our existing `resistor-color` exercise (including a rename probably). Right?

F3PiX:
> What should be the new name of the exercise? @kytrinyx suggested `color-codes`, which I like.

I like it too, except for the fact that it’s not recognizable as part of the series. Unless we take ‘Colorcode’ as identifier for the series: SingleColorCode, ColorCodePairs, TriColorCode, etc. 
Alternatively: ResistorColorCode for `this one, ResistorValue for the next one.
I like the name with Resistor a bit better, but I find it more important that it’s identifyable as part of the series.

> We should apply the lessons of this discussion to our existing `resistor-color` exercise (including a rename probably). Right?

Right! I’ll open an issue.

exercises/resistor-significants/description.md
```diff
@@ -0,0 +1,16 @@
+If you would build anything with a Raspberry Pi or an Arduino, one of the first components you need are resistors. In this exercise, it doesn't matter what a resistor is or does, the only thing that matters is that they come with different values. And that they are tiny. So tiny, that it's hard to read such a value if it were printed on the thing. To solve that, fabricators use color codes that reveal their value. Which is great, but it makes it hard to see the actual value.
```
rpottsoh: I have some thoughts on rewording this but I don't have the time at the moment.  This evening I'll revisit this and update this comment...

rpottsoh: I wasn't paying very careful attention to what comes after what I just wrote.  Some of it may be redundant.

ErikSchierboom: @F3PiX Could you look at the suggested changes?

F3PiX: Thanks @rpottsoh ! 
It's not so easy to distinguish the separate changes in texts with the Suggested Changes feature.  What I see, there's some corrections of mistakes (thanks!). There's also some level of detail added that I'd rather keep out. Let me see how I can work the changes in.

F3PiX:
@rpottsoh My thinking is: for people who _know_ resistors, the concept of 'colors' will create an image in their minds of color _bands_; for people who don't know resistors, it's not relevant to solve this exercise whether it's bands, spots, solids or whatever, so we don't make them think about it. 
Similarly, I avoid to use the word 'value' in the part of the description that tells what the two colors return other than a 'number'.  So that people who do know resistors will not think: 'Pff, you can't tell the value of a resistor by 2 color bands only'.
Hope this works!

exercises/resistor-significants/description.md
```diff
@@ -0,0 +1,16 @@
+If you would build anything with a Raspberry Pi or an Arduino, one of the components you will likely use are resistors. In this exercise, it doesn't matter what a resistor is or how it works, the only thing that matters is that a resistor has a 'value'. And that they are small. So small, that it's hard to read such a value if it were to be actually printed on it. To solve that, fabricators use color codes that denotes the resistor's value. Which is great, but it makes it hard to see the actual value.
```
iHiD: I've tweaked this below. The main changes:
-  I've restructured it to make it flow a little more clearly, and to fill in a few blanks/unknowns that I had as I read it (I found myself googling about resistors to clarify things). 
- I've removed the "or an Arduino" because I don't think we need both and Raspberry Pis are more well know and - I changed the example to brown and green, because the brown/black `10` thing made me think of binary and confused me :)
- I changed "fabricators" to "manufacturers" as I think it's a more common word.
- I specified "resistance value" because I found myself asking "what/why does this have a value" whereas I could immediately visualise "resistance value" as a tangible thing.

```
If you want to build something using a Raspberry Pi, you'll probably use *resistors*. For this exercise, you need to know two things about them:
1. Each resistor has a resistance value. 
2. Resistors are small - so small in fact that if you printed the resistance value on them, it would be impossible to read.  

To get around this problem, manufacturers print color-coded bands onto the resistors to denote their resistance values. Each band acts as a digit of a number. For example, if they printed a brown strip (value 1) followed by a green strip (value 5), the resistance value of the resistor would be 15.

In this exercise, you are going to create a helpful program so that you don't have to remember the values of the bands. The program will take two colors as input, and output the correct resistance value.

The band colors are encoded as follows:

- Black: 0
- Brown: 1
- Red: 2
- Orange: 3
- Yellow: 4
- Green: 5
- Blue: 6
- Violet: 7
- Grey: 8
- White: 9
```

Would appreciate everyone's thoughts - especially @kytrinyx and @f3pix. I've not done it as a "suggestion" so I'll leave you to manually put it in the PR if you like ti.

iHiD:
As a sidenote, we're planning on removing anything from the `README.MD` other than the contents of `description.md`. I mention this because atm, the text I wrote makes the README a lot longer (just through extra formatting), which might feel overwhelming with all the other stuff that comes below in the README, but as we're planning on dropping that, spreading things out is good.

I also just noticed Maud's comments about Ryan's comments (this can get confusing sometimes!). I suspect I've repeated some of Ryan's points in my suggestion, which Maud had already disapproved for being factually incorrect (eg. `the resistance value of the resistor would be 15`). This is sort of a general worry I have about having exercises where there is a lot of external knowledge required to know what is correct or not-correct, because we want to be factual, but also we give enough detail that we're not just using abstract terms (like "value") which made me just have more questions. It's a tricky balance to get right! It's much easier in that regard to make a story up that has no factual grounding.

Anyway, I apologise if I have inadvertently added to the noise rather than providing more signal :)

exercises/resistor-significants/description.md
kytrinyx: This looks good to me! (I know nothing about resistors)

rpottsoh: These suggestions overall seem fine to me.

> I changed "fabricators" to "manufacturers" as I think it's a more common word.

How about `makers` instead?

Personally if I didn't know anything about resistors having some basic facts included is helpful, adds some color to the exercise (I know, but pun wasn't intended) , makes the "thing" less nebulous even if the minor information isn't directly relateable it helps me to focus what my mind should be considering while solving the problem; as long as the problem description goes on to indicate which information is essential for completing the exercise.

rpottsoh:
> 
Anyway, I apologise if I have inadvertently added to the noise rather than providing more signal :)

As long as we stay above the `noise floor` all will be fine.

F3PiX:
Updated with the latest proposals!

exercises/resistor-significants/canonical-data.json
```diff
@@ -0,0 +1,38 @@
+{
+  "exercise": "resistor-significants",
```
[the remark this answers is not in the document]
ErikSchierboom: I think I like `resistor-colors` best. It is a very subtle change from `resistor-color`, but it does represent the exercise best I feel.

exercises/resistor-significants/description.md
```diff
@@ -0,0 +1,20 @@
+If you want to build something using a Raspberry Pi, you'll probably use _resistors_. For this exercise, you need to know two things about them:
+
+* Each resistor has a resistance value.
+* Resistors are small - so small in fact that if you printed the resistance value on them, it would be hard to read.
+To get around this problem, manufacturers print color-coded bands onto the resistors to denote their resistance values. Each band acts as a digit of a number. For example, if they printed a brown strip (value 1) followed by a green strip (value 5), it would translate to the number 15.
```
rpottsoh: Are they bands or strips of color?

rpottsoh: I like what I am reading.  :+1:

ErikSchierboom:
@rpottsoh @F3PiX I've renamed the exercise to `resistor-colors`. Could you take a final look before we merge (note: it should be squashed)?

exercises/resistor-colors/description.md
```diff
@@ -0,0 +1,20 @@
+If you want to build something using a Raspberry Pi, you'll probably use _resistors_. For this exercise, you need to know two things about them:
+
+* Each resistor has a resistance value.
+* Resistors are small - so small in fact that if you printed the resistance value on them, it would be hard to read.
+To get around this problem, manufacturers print color-coded bands onto the resistors to denote their resistance values. Each band acts as a digit of a number. For example, if they printed a brown strip (value 1) followed by a green strip (value 5), it would translate to the number 15.
```
rpottsoh: I think I pointed this out earlier.  But I see that color bands are also called strips.  Why not just call them bands throughout?
```suggestion
To get around this problem, manufacturers print color-coded bands onto the resistors to denote their resistance values. Each band acts as a digit of a number. For example, if they printed a brown band (value 1) followed by a green band (value 5), it would translate to the number 15.
```

rpottsoh:
I've posted a comment regarding the description but otherwise everything looks good to me too.  :+1:

exercises/resistor-colors/description.md
ErikSchierboom: @F3PiX what do you think about this suggestion?

F3PiX: As @rpottsoh's approved the PR without it, let's keep it as is.

ErikSchierboom: @F3PiX I don't see @rpottsoh as having approved though.

F3PiX: Ooh! @rpottsoh ? Am I jumping to conclusions?

ErikSchierboom:
> I've posted a comment regarding the description but otherwise everything looks good to me too

I've applied the comment. Anything else?

rpottsoh:
LGTM.  Thanks @ErikSchierboom

ErikSchierboom:
Squashed and merged! Thanks for the reviews everybody!

rpottsoh:
I apologize if I was too vague on my approval (or lack of it) as it related to my suggested change regarding calling bands strips.  I felt strongly enough about the suggested change to mention it, but not strongly enough to require it be changed before giving approval (in GitHub terms).  Thanks for making the change!

ErikSchierboom:
No problem!

sshine:
@kytrinyx wrote:
> I find this exercise really difficult to understand

So do I. But mostly because the exercise does not describe what problem to solve, only the problem domain.

@iHiD wrote:
> we're planning on removing anything from the README.MD other than the contents of description.md

Do you mean stuff like how to get started with the track, extra exercise hints, etc.?

---

And now for my question, since @tejasbubane just ported this exercise to Haskell in exercism/haskell#808:

If `value ["black"] == 0`, then what is `value []`? Is that an error? Also 0?

F3PiX:
> And now for my question, since @tejasbubane just ported this exercise to Haskell in [exercism/haskell#808](https://github.com/exercism/haskell/pull/808):
> 
> If `value ["black"] == 0`, then what is `value []`? Is that an error? Also 0?

`0` is valid. 
In case it helps: There's always two colors passed in in the tests, and they return a integer composed out of 2 digits.

sshine:
So `[]`, `["black"]`, `["black", "black"]` etc. are synonymous. and their `value` is 0.

The tests don't specify what happens to empty input.

The resistor illustrations at #1464 all have three colors.

I can see from googling around a little that there can be many color bands, and that in practice you wouldn't have a resistor without colors, since you'd call that a wire. So I suppose that tracks are free to interpret `[]` either as not well-defined, or as a zero-resistor.

ErikSchierboom:
> The tests don't specify what happens to empty input.

This is correct. There is no error handling defined in the tests. Tracks are free to handle errors as they like, possibly adding test cases.

> The resistor illustrations at #1464 all have three colors.

I believe there is special significance for the first two colors, isn't there @F3PiX?

F3PiX:
Value 0 would also be a wire .

> tracks are free to interpret [] either as not well-defined, or as a zero-resistor.

Just keep in mind that we deliberately left that out, as this is meant to be nr 2 in a series with increasing complexity. The input validation would be an extra layer that we intend to add to nr 3. (And nr 3 is also going to calculate the value of all three color bands.)

This post for the previous one, nr 1, tells a bit more: https://github.com/exercism/problem-specifications/issues/1458

cmccandless:
> I believe there is special significance for the first two colors, isn't there @F3PiX?

This question wasn't directed at me, but I can answer it: on a 3-band resistor, the 3rd band is a multiplier on the number created by the first 2. So technically, the 3rd band is the one with special significance, not the first 2.

ErikSchierboom:
Right, that makes sense. Thanks for explaining!

PR #1469, opened 2019-02-26
collatz-conjecture: add test for float input and clarify error messages

juhlig:
The test suite had no check for float inputs. If an implementation does not ensure it is working with (positive) integers, depending on language, it may run forever or crash.

Furthermore, the error message "only positive numbers are allowed" was in that sense misleading, as numbers _include_ floats.

### DIFF
--- a/exercises/collatz-conjecture/canonical-data.json
+++ b/exercises/collatz-conjecture/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "collatz-conjecture",
-  "version": "1.2.1",
+  "version": "1.3.0",
   "cases": [
     {
       "description": "zero steps for one",
@@ -40,7 +40,7 @@
       "input": {
         "number": 0
       },
-      "expected": { "error": "Only positive numbers are allowed" }
+      "expected": { "error": "Only positive integers are allowed" }
     },
     {
       "description": "negative value is an error",
@@ -48,7 +48,15 @@
       "input": {
         "number": -15
       },
-      "expected": { "error": "Only positive numbers are allowed" }
+      "expected": { "error": "Only positive integers are allowed" }
+    },
+    {
+      "description": "float value is an error",
+      "property": "steps",
+      "input": {
+        "number": 1.5
+      },
+      "expected": { "error": "Only positive integers are allowed" }
     }
   ]
 }


exercises/collatz-conjecture/canonical-data.json
```diff
@@ -40,15 +40,23 @@
...
         "number": -15
       },
-      "expected": { "error": "Only positive numbers are allowed" }
+      "expected": { "error": "Only positive integers are allowed" }
+    },
+    {
+      "description": "float value is an error",
+      "property": "steps",
+      "input": {
+        "number": 1.5
```
coriolinus: Adding this test breaks type inference for test generators for typed languages which distinguish between integer and floating point types. 

Agree that it's a good idea for such a test to exist, in cases where the type system can't help you. However, from the perspective of someone maintaining a language with a strong type system, it's nicer when the tooling can do the right thing without human intervention.

I believe it to be an exercism principle that students shouldn't have to perform input validation; they can trust that they will get valid inputs, except in cases where input validation is a major portion of the intent of the exercise. In my opinion, input validation is not the point of this exercise, and therefore, we should not add tests checking students' input validation.

NobbZ: Yeah, in general we try to assume a typesafe language.

If though a tracks language allow passing in arbitrary types, we encourage maintainers to add additional tests local to their track.

juhlig: > Adding this test breaks type inference for test generators for typed languages which distinguish between integer and floating point types.
> 
> Agree that it's a good idea for such a test to exist, in cases where the type system can't help you. However, from the perspective of someone maintaining a language with a strong type system, it's nicer when the tooling can do the right thing without human intervention.
> 
> I believe it to be an exercism principle that students shouldn't have to perform input validation; they can trust that they will get valid inputs, except in cases where input validation is a major portion of the intent of the exercise. In my opinion, input validation is not the point of this exercise, and therefore, we should not add tests checking students' input validation.

In that case, this would mean removing current tests for zero and negative inputs. This would be ok with me, also. I'm just saying that, if we have tests for input validity in this exercise, they should contain a test for floats regardless of a specific language.

NobbZ: > In that case, this would mean removing current tests for zero and negative inputs.

I'm with you about negative inputs, as in most typed languages you can use an unsigned integer and are good to go, still, unless you are using Idris, Agda or some other theorem proover or dependently typed language, you won't get rid of the `0`.

So we have to make some assumptions. And these assumptions for integers are regularly, that we have a single integer type that is unbound but checked, and we can not use a different type for the same function.

This way Rust, C, go and many others could decide to use unsigned integers and drop the negative input test while keeping the `0`-test.

Languages that are checked and which have an integer type that is described as above can just use the tests as they are.

Non-statically typed languages though, might need to add some tests for further error checking or simply drop them depending on the position in the track.

I'd not overload an exercise which is mainly about doing some work recusrively with error handling.

Instead, we could pre-generate the stub with a typeguard as a hint.

juhlig: Ah, no, maybe I expressed myself badly. Or I'm confused. Probably both ;)

I wasn't concerned about what error cases a language might implicitly rule out (like floats in typed languages), and which would still be possible (like zeroes). I was merely pondering @coriolinus last paragraph, which in a nutshell says that a student shouldn't have to perform input validation and can trust they will get valid inputs, unless input validation is (a major part of) the intent of the exercise. Since input validation is IMO _not_ part of the intent of the exercise, doesn't that mean that the tests for negative/zero input should be done away with, as they _do_ impose input validation on the student?

ErikSchierboom: > Since input validation is IMO _not_ part of the intent of the exercise, doesn't that mean that the tests for negative/zero input should be done away with, as they _do_ impose input validation on the student?

I'm in favor of doing this.

> I'd not overload an exercise which is mainly about doing some work recusrively with error handling.

I agree.

ErikSchierboom:
Reading the comments, I think I see some agreement on not merging this PR, and letting tracks decide for themselves if they would value adding the suggested test cases. Is that correct?

ErikSchierboom:
As there has been no response for some time, I'm gonna go with what I feel is the prevailing sentiment here, and close this issue. @juhlig thanks a lot for creating the PR, even though the end result is not what you'd hope for. You are of course completely free to modify the test suite for the track of your choice.

juhlig:
@ErikSchierboom no problem :) I was away on vacation, thus the silence on my part.

ErikSchierboom:
Cool! Hope your vacation was nice.

PR #1474, opened 2019-03-04
pig-latin: Add test case for word starting with `xr`

G-Rath:
In the README it has the following note for Rule 1:

> Please note that "xr" and "yt" at the beginning of a word make vowel sounds (e.g. "xray" -> "xrayay", "yttria" -> "yttriaay").

This isn't actually tested anywhere! Now it is 😄 

I've not added a test for `yt`, but don't mind if people want that too :)

### DIFF
--- a/exercises/pig-latin/canonical-data.json
+++ b/exercises/pig-latin/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "pig-latin",
-  "version": "1.2.0",
+  "version": "1.3.0",
   "cases": [
     {
       "description": "ay is added to words that start with vowels",
@@ -52,6 +52,14 @@
             "phrase": "equal"
           },
           "expected": "equalay"
+        },
+        {
+          "description": "word beginning with xr",
+          "property": "translate",
+          "input": {
+            "phrase": "xray"
+          },
+          "expected": "xrayay"
         }
       ]
     },


petertseng:
> This isn't actually tested anywhere

https://github.com/exercism/problem-specifications/blob/18875ece112eb46c7227e7e1745432ed16030ddd/exercises/pig-latin/canonical-data.json#L160-L165

> I've not added a test for yt

https://github.com/exercism/problem-specifications/blob/18875ece112eb46c7227e7e1745432ed16030ddd/exercises/pig-latin/canonical-data.json#L152-L157

G-Rath:
Well, that's embarrassing... 😨

It seems then the TS version of this exercise hasn't got the new tests... 😂

marnen:
Hmm, now I wonder if we should change the README. No English word (that I can think of) begins “xr”: “X-ray” has a hyphen in it.

G-Rath:
I was wondering about that - it seems they're [all computer-related terms](https://www.thefreedictionary.com/words-that-start-with-xr) :joy:

Does pig latin say anything about what to do with words that have hyphens...?

marnen:
Good question. I would expect to ignore but preserve the hyphen (“x-rayay”?).

kotp:
Another seemingly popular rule is to do "exay-ayray".

PR #1477, opened 2019-03-12
rational-numbers: add abs test case for den < 0

SaschaMann:
Currently there's only a test for a negative num, but not a negative den.

### DIFF
--- a/exercises/rational-numbers/canonical-data.json
+++ b/exercises/rational-numbers/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "rational-numbers",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "comments": [
     " The canonical data assumes mathematically correct real   ",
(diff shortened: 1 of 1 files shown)


petertseng:
Will there also be a case where both numerator and denominator are negative?

Or will such a case not exist because there is a guarantee (even if implicit) that the numerators and denominators of all inputs for `abs` will not share a common factor?

Note that merging is not blocked on receiving any particular answer, so one should not wait for me to acknowledge any answer given. One should merge without waiting for my ack.

SaschaMann:
I'll add a test case where both are negative.

> Or will such a case not exist because there is a guarantee (even if implicit) that the numerators and denominators of all inputs for abs will not share a common factor?

Most implementations I've seen (not as student submissions but in general) "normalize" the number on construction, so it wouldn't matter. However, it doesn't hurt to explicitely test abs will work if that doesn't happen, imo.

PR #1479, opened 2019-03-15
TOPICS.txt: add operator_overloading

rpottsoh:
A specific form of polymorphism see https://en.wikipedia.org/wiki/Operator_overloading

### DIFF
--- a/TOPICS.txt
+++ b/TOPICS.txt
@@ -10,6 +10,7 @@ generics
 inheritance
 interfaces
 loops
+operator_overloading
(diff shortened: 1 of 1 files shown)


cmccandless:
Common candidates for this new topic:
- `clock`
- `list-ops`
- `complex-numbers` (this is the biggest one)
- [insert here]

rpottsoh:
I amusing operator overloading for the implementation of `rational-numbers` for Delphi.

Issue #1198, opened 2018-02-23
Too verbose descriptions in the canonical_data for crypto-square -> failing Travis build


vlzware:
When building tests for crypto-square in the 'C' track, naming the functions after the descriptions in the [canonical_data](https://github.com/exercism/problem-specifications/blob/master/exercises/crypto-square/canonical-data.json) leads to funny behavior when using indent.sh which leads to failing Travis build.

By funny behavior I mean running indent.sh on the same file leads to changes every time, which gets detected in Travis with the ```git diff``` check.

[Here is the link](https://github.com/exercism/c/pull/274#issuecomment-367941634) to the pull request where this was discovered.

wolf99:
I've given some suggestions for function name length reduction in the issue on the track, but it may well be worth continuing the discussion here in the general case - i.e.how to deal with super long test function names indicated by canonical data, where the implementation language values terseness and line column count limits.

rpottsoh:
@vlzware or @wolf99 is still an issue that needs to be pursued?

wolf99:
The test name lengths can be significantly reduced by removing the description of the result and only describint the test itself.

Compare, for example, to `all-your-base`.  The test descriptions there mention `input base is zero`, rather than `test that when the input base is zero it results in an error that the input must be larger`. (https://github.com/exercism/problem-specifications/blob/main/exercises/all-your-base/canonical-data.json#L148)

Similarly the tests here could be changed from the likes of `54 character plaintext results in 7 chunks, the last two with trailing spaces` can be reduced to `54 character plaintext` as this is the test case (the rest of the text describing the specifics of the result). 

The fact that the expected result is specifically to have multiple chunks with two trailing spaces at the end _could_ be added in the `"comments"` field, if it is needed at all (personally I think the fact is evident from the `"expected"` field, but it is subjective).

ErikSchierboom:
@vlzware @wolf99 Is this still worth pursuing? I wouldn't mind the test descriptions to be shortened. Maybe we could use the `comments` field to describe the result?

PR #1480, opened 2019-03-16
Luhn: Update testcase to avoid wrong solution 

tqa236:
The change might seem trivia at first. It isn't. Here's a solution that passes all the test in Java
```
import java.util.List;
import java.util.stream.Collectors;
import java.util.stream.IntStream;

class LuhnValidator {
  public boolean isValid(String candidate) {
    String number = candidate.replaceAll("\\s+", "");
    if ("0".equals(number)) return false;
    List<Integer> digits = number.chars().map(d -> d - 48).boxed().collect(Collectors.toList());
    return IntStream.range(0, digits.size())
                .map(
                    n ->
                        (n % 2 == digits.size() % 2)
                            ? (2 * digits.get(n) - 9 * (digits.get(n) >= 5 ? 1 : 0))
                            : digits.get(n))
                .sum()
            % 10
        == 0;
  }
}
```

In this solution, after I remove all the whitespaces, I convert all characters to the ascii number and subtract them to 48 (the ascii code of 0) to get the digits. Then I check with the list I receive and it passes all test cases. I didn't remove invalid characters at all.

The problem is that the ascii code of the invalid characters automatically make the valid number becomes invalid, so no need to remove them.

With the new test case, the sum is 90, and it should still be false because if the invalid character.


petertseng:
I also don't have access to running Java so I'll have to rely on others' help for the answer to the question: What does it do to `:9` from https://github.com/exercism/problem-specifications/pull/1246 ? `:` has value 58, which becomes 10. Doubled and becomes 20, subtract 9 to get 11, add 9 to get 20, so it would do the same. Any differences in behaviour of the supplied code such that it already rejects `:9` ?

coriolinus:
Using the previous Python def of `isValid`:

```python
>>> isValid(':9')
True
```

Given that the expected value of this case is `False`, that case should already exclude this solution, assuming always that the python implementation produces results identical to the Java one.

tqa236:
It seems that this solution is already be ruled out by that test case. My mistake.
My original solution that passes all the test is a little different, it uses integer division. I forgot to recheck it.
```
(2 * digits.get(n) - 9 * (digits.get(n) / 5))
 // ? (2 * digits.get(n) - 9 * (digits.get(n) >= 5 ? 1 : 0))
```
Given that the range of `2*digits` is from 0 to 18, these 2 formulas are both correct.
I agree that we will need a stronger test suite to completely rules out this type of solution (like randomly generate 1000 tests for example). I just propose a quick fix for now.

petertseng:
That means the interesting thing about these two cases `:9` and `055b 444 285` is whether the non-digit character occupies a doubled position (yes in the former, no in the latter). That implies it would be valuable to state this fact in both of their respective test descriptions. And perhaps bring them next to each other. Thus, I recommend this.

I won't stand in the way if the decision is instead to ignore this recommendation and merge this PR as it currently stands.

tqa236:
I agree that we should update the description to better describe the test, do you have any suggestions? By the way, is it possible to add a comment in a test about the intent of that test? It is not easy to recognize just by looking at the file. Just in case there's a wrong solution with a similar approach that passes, we know we should update the test suite to a more rigorous version.

petertseng:
> I agree that we should update the description to better describe the test, do you have any suggestions? 

I have a hard time finding a short description for the cases. Even though it is not a requirement that it be short, https://github.com/exercism/problem-specifications/issues/1198 seems to imply there are benefits.

I'll try:

* Doubled non-digit isn't treated as its ascii value
* Non-doubled non-digit isn't treated as its ascii value

Open question, do we need to mention "ascii value offset from the value of 0" or is "ascii value" sufficient

> is it possible to add a comment in a test about the intent of that test

it is possible. see an example in https://github.com/exercism/problem-specifications/blob/d28d6a5f40bb19f6a5954b7bda3430f4089aa680/exercises/largest-series-product/canonical-data.json#L107-L120

see the formal spec in https://github.com/exercism/problem-specifications/blob/d28d6a5f40bb19f6a5954b7bda3430f4089aa680/canonical-schema.json#L44 and https://github.com/exercism/problem-specifications/blob/d28d6a5f40bb19f6a5954b7bda3430f4089aa680/canonical-schema.json#L62-L67

tqa236:
I updated the descriptions based on your suggestion. I also grouped them together to reflect the fact that they address the same problem.

exercises/luhn/canonical-data.json
```diff
@@ -123,7 +115,23 @@
       "expected": true
     },
     {
-      "description": "strings with non-digits is invalid",
+      "comments": [
+        "Convert non-digits to their ascii value and then offset them by 48 sometimes accidently return the correct result.",
+        "This test is designed to avoid that solution."
+      ],
+      "description": "using ascii value for doubled non-digit isn't allow",
```
petertseng: I think for grammar, `allow` should be `allowed` here, and in the other test case as well

exercises/luhn/canonical-data.json
```diff
@@ -123,7 +115,23 @@
       "expected": true
     },
     {
-      "description": "strings with non-digits is invalid",
+      "comments": [
+        "Convert non-digits to their ascii value and then offset them by 48 sometimes accidently return the correct result.",
```
petertseng: spelling `accidently` -> `accidentally`

petertseng: "accidentally return the correct result" might be easy to misinterpret, or require careful thinking before a reader can determine whether the sentence is true. Can I suggest "accidentally declare an invalid string to be valid" so that requires less brainpower? my feeble mind was having a hard time figuring things out

tqa236:
Thank you for the review. I made a new commit based on your suggestion.

exercises/luhn/canonical-data.json
```diff
@@ -123,7 +115,23 @@
       "expected": true
     },
     {
-      "description": "strings with non-digits is invalid",
+      "comments": [
+        "Convert non-digits to their ascii values and then offset them by 48 sometimes accidentally declare an invalid string to be valid.",
+        "This test is designed to avoid that solution."
+      ],
+      "description": "using ascii value for doubled non-digit isn't allowed",
```
petertseng: can you double-check me on this one? I think in this case, the `b` is in a **non**-doubled position, and in the latter case the `:` is in a doubled position. So these two test case descriptions should be switched, right?

Or did I get them mixed up

tqa236:
You're correct. I fix that in the new commit.

Issue #1473, opened 2019-03-01
Proposal: Add canonical data "identifier" field to separate concern from test "description"


bencoman:
[Edit:] Some languages don't individually identify tests and use the canonical data "description" field to generate comments into their tests code.  Other languages individually identify tests, but all that is available for generating a test-name identifiers is the "description" field.   This results in some extremely long test-method-names.  e.g. For `robot-simulator` Pharo has...
`test18_WhereRTurnRightLTurnLeftAndAAdvanceTheRobotCanFollowASeriesOfInstructionsAndEndUpWithTheCorrectPositionAndDirectionInstructionsToMoveEastAndNorth`

Can an optional `identifier` field be introduced for cases where the description doesn't make a good identifier?

For example, for the [Bowling exercise](https://github.com/exercism/problem-specifications/blob/master/exercises/bowling/canonical-data.json#L207)... 
```
{
    "description": "Bonus rolls for a strike in the last frame must be rolled before score can be calculated",
    "property": "score",
    "input": {
      "previousRolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10]
}
```
generating tests for different tracks produces the following...

CSharp: 
public void Bonus_rolls_for_a_strike_in_the_last_frame_must_be_rolled_before_score_can_be_calculated()
https://github.com/exercism/csharp/blob/master/exercises/bowling/BowlingTest.cs#L250

Delphi:
procedure BowlingTests.Bonus_rolls_for_a_strike_in_the_last_frame_must_be_rolled_before_score_can_be_calculated;
https://github.com/exercism/delphi/blob/master/exercises/bowling/uBowlingTests.pas#L324

FSharp:
let \`\`Both bonus rolls for a strike in the last frame must be rolled before score can be calculated`` () =
https://github.com/exercism/fsharp/blob/master/exercises/bowling/BowlingTest.fs#L177

Adding a optional `identifier` like below would be useful. 
```
{
    "identifier": "roll_bonus_balls_before_scoring"
    "description": "Bonus rolls for a strike in the last frame must be rolled before score can be calculated",
    "property": "score",
    "input": {
      "previousRolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10]
}
```

bencoman:
Further to this, it seems that `description` ends up trying to serve two purposes:
[...]

PR #1483, opened 2019-03-21
triangle: Add clarity to test descriptions

Mehni:
Proposed fix to https://github.com/exercism/csharp/issues/1253

The current [Test] names confused me when I was doing the exercise. It was also difficult to see which tests failed, because of long method names combined with a small window with the test results.

This PR changes the "Returns_true_if_the_triangle_is_equilateral_false_if_any_side_is_unequal" style in favour of a pattern that follows "equilateral_triangle_returns_false_if_any_side_is_unequal".

### DIFF
--- a/exercises/triangle/canonical-data.json
+++ b/exercises/triangle/canonical-data.json
@@ -111,7 +111,7 @@
           "expected": false
         },
         {
-          "description": "triangle inequality violation (1)",
+          "description": "first triangle inequality violation",
           "property": "isosceles",
           "input": {
             "sides": [1, 1, 3]
@@ -119,7 +119,7 @@
           "expected": false
         },
         {
-          "description": "triangle inequality violation (2)",
+          "description": "second triangle inequality violation",
           "property": "isosceles",
           "input": {
             "sides": [1, 3, 1]
@@ -127,7 +127,7 @@
           "expected": false
         },
         {
-          "description": "triangle inequality violation (3)",
+          "description": "third triangle inequality violation",
           "property": "isosceles",
           "input": {
             "sides": [3, 1, 1]
--- a/exercises/triangle/canonical-data.json
+++ b/exercises/triangle/canonical-data.json
@@ -9,7 +9,7 @@
     " triangles have zero area, however.                                    ",
     " They're degenerate triangles with all three vertices collinear.       ",
     " (In contrast, we will test (0, 0, 0, Illegal), as it is a point)      ",
-    " The tests assert properities of the triangle are true or false.       ",
+    " The tests assert properties of the triangle are true or false.        ",
     " See: https://github.com/exercism/problem-specifications/issues/379 for disscussion  ",
     " of this approach                                                      ",
     " How you handle invalid triangles is up to you. These tests suggest a  ",
@@ -19,10 +19,10 @@
   ],
   "cases": [
     {
-      "description": "returns true if the triangle is equilateral",
+      "description": "equilateral triangle",
       "cases": [
         {
-          "description": "true if all sides are equal",
+          "description": "returns true if all sides are equal",
           "property": "equilateral",
           "input": {
             "sides": [2, 2, 2]
@@ -30,7 +30,7 @@
           "expected": true
         },
         {
-          "description": "false if any side is unequal",
+          "description": "returns false if any side is unequal",
           "property": "equilateral",
           "input": {
             "sides": [2, 3, 2]
@@ -38,7 +38,7 @@
           "expected": false
         },
         {
-          "description": "false if no sides are equal",
+          "description": "returns false if no sides are equal",
           "property": "equilateral",
           "input": {
             "sides": [5, 4, 6]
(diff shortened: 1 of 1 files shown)


Mehni:
Some things I considered but didn't do:

- Change the version. Reading the test-data-versioning, would this be a major change?
- Technically the "sides can be floats" test uses doubles for the C# track. Since this is likely language specific, I'm not certain what the correct naming would be.

rpottsoh:
> Change the version. Reading the test-data-versioning, would this be a major change?

The changes I believe constitute a [patch version change](https://github.com/exercism/problem-specifications#patch-version-changes).  I think [bin/check_versions](https://github.com/exercism/problem-specifications/blob/master/bin/check_versions) should have taken exception to the version not being changed and caused Travis CI to fail.  @petertseng did `check_versions` perform as expected in this instance?

> Technically the "sides can be floats" test uses doubles for the C# track. Since this is likely language specific, I'm not certain what the correct naming would be.

I think `floats` is OK.  Sometimes the term `real` is used as well as a more language agnostic way to infer that a value not be an integer, or a string, etc.  Perhaps a more concise way to make the statement could be `sides can be of a floating point type`.

SaschaMann:
With #1473 in mind, perhaps the descriptions could be shortened a bit, e.g. `all sides are equal` instead of `returns true if all sides are equal`? The returned value is right below it anyway.

(Just a thought, not objecting to the proposed change here if others approve it)

SaschaMann:
Closed and reopened to trigger a Travis build.

Mehni:
@rpottsoh from what you posted, wouldn't it be [a major version](https://github.com/exercism/problem-specifications#major-version-changes) since it renames a "property" (test type)? Or am I misunderstanding the nomenclature, and does it fall under "key formatting"?

@SaschaMann while "the returned value is right below it" is true when looking at the Test.cs itself, the test methods themselves aren't in the same screen as the test results (which is when you're most interested in expected vs actual). Coming from the C#-track, the IDE has the test results in a different pane. If I do the recommend way of testing `dotnet test`, I get a CLI with just the results.

SaschaMann:
> @SaschaMann while "the returned value is right below it" is true when looking at the Test.cs itself, the test methods themselves aren't in the same screen as the test results (which is when you're most interested in expected vs actual). Coming from the C#-track, the IDE has the test results in a different pane. If I do the recommend way of testing dotnet test, I get a CLI with just the results.

That seems like an issue with dotnet tooling, not one that the common test spec needs to solve. I referred to that issue because shortening descriptions seemed to be a common wish in it and, imo, `return true if` (or `true if`) doesn't add a lot of info to the test data. The vast majority of test cases on this repo don't mention the expected result in the description.
But as I said, I leave it up to others to decide on this, just wanted to bring it up.

Mehni:
I've re-read the documentation and changing the `description` of tests is indeed a patch increment. Thanks for your patience, @rpottsoh.

I've also removed the "expected result" to bring this more in line with the other exercises. Thanks for your feedback, @SaschaMann.

exercises/triangle/canonical-data.json
```diff
@@ -103,31 +103,31 @@
...
+          "description": "no sides are equal",
           "property": "isosceles",
           "input": {
             "sides": [2, 3, 4]
           },
           "expected": false
         },
         {
-          "description": "Sides that violate triangle inequality are not isosceles, even if two are equal (1)",
+          "description": "triangle inequality violation (1)",
```
rpottsoh:
```suggestion
          "description": "first triangle inequality violation",
```

exercises/triangle/canonical-data.json
```diff
@@ -103,31 +103,31 @@
...
           "property": "isosceles",
           "input": {
             "sides": [1, 1, 3]
           },
           "expected": false
         },
         {
-          "description": "Sides that violate triangle inequality are not isosceles, even if two are equal (2)",
+          "description": "triangle inequality violation (2)",
```
rpottsoh:
```suggestion
          "description": "second triangle inequality violation",
```

exercises/triangle/canonical-data.json
```diff
@@ -103,31 +103,31 @@
...
           "property": "isosceles",
           "input": {
             "sides": [1, 3, 1]
           },
           "expected": false
         },
         {
-          "description": "Sides that violate triangle inequality are not isosceles, even if two are equal (3)",
+          "description": "triangle inequality violation (3)",
```
rpottsoh:
```suggestion
          "description": "third triangle inequality violation",
```

rpottsoh:
> @rpottsoh from what you posted, wouldn't it be a major version since it renames a "property" (test type)? Or am I misunderstanding the nomenclature, and does it fall under "key formatting"?

@Mehni I think this still falls under `patch`, all of your changes only involve descriptions and comments.

Mehni:
ofc 😄 

I think Exercism is one of the best ways to learn coding and wanted to contribute.

rpottsoh:
If this PR is merged it should be squashed.  This PR should remain open for a day or two to allow others a chance to comment and offer suggestions.  If it is still open on March 24th I'll Squash and merge it.

PR #1485, opened 2019-03-22
ETL: Modified canonical description to ease test suite generation

macta:
Small change to the json spec that is easier for code generation, and clearly shows that you are calling the etl function with 1 parameter (which has "1"..."x" tile arrays)

### DIFF
--- a/exercises/etl/canonical-data.json
+++ b/exercises/etl/canonical-data.json
@@ -1,77 +1,76 @@
 {
   "exercise": "etl",
-  "version": "1.0.0",
+  "version": "1.0.1",
+  "comments": [
+    "Transforms a set of scrabble data previously indexed by the",
+    "tile score to a set of data indexed by the tile letter",
+    "Note:  The expected input data for these tests should have",
+    "integer keys (not stringified numbers as shown in the JSON below",
+    "Unless the language prohibits that, please implement these tests",
+    "such that keys are integers. e.g. in JavaScript, it might look ",
+    "like `transform( { 1: ['A'] } );`"
+  ],
   "cases": [
     {
-      "comments": [
-        "Note:  The expected input data for these tests should have",
-        "integer keys (not stringified numbers as shown in the JSON below",
-        "Unless the language prohibits that, please implement these tests",
-        "such that keys are integers. e.g. in JavaScript, it might look ",
-        "like `transform( { 1: ['A'] } );`"
-      ],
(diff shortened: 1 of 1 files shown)


macta:
Also - the use of a nested case was redundant, they were essentially all top level.

exercises/etl/canonical-data.json
```diff
@@ -1,77 +1,76 @@
...
-            "p":  3, "q": 10, "r": 1, "s": 1, "t": 1,
-            "u":  1, "v":  4, "w": 4, "x": 8, "y": 4,
-            "z": 10
-          }
-        }
-      ]
+      "description": "transforms a single letter",
+      "property": "transform",
+      "input": {
+        "data" : {"1": ["A"]}
```
rpottsoh: How does inserting "data" between "input" and the actual data make things easier?

macta: Strictly be definition (and if automatically generating the tests), each input translates as a new parameter to a function - so the previous data is confusing - you end up with a function that takes :

Transform(input1: x, input2: y, input 3: z etc)... which is not the intent of the exercise.

Contrast this with other exercises where the function genuinely has these parameters - eg

calcumateArea(width:5, length: 10).

The last test for this exercise drives it home. It’s inconsistent with other exercises as previously defined. And if you look at how other languages have manually translated it - they all pass a dictionary of the input.

macta: @ErikSchierboom , I'm wondering if you have any comment on this one too - as presumably you might have hit this inconsistency when generating exercises from the spec too? I notice that Etl in C# also calls a single function and passes a single dictionary with "1" and "2" as keys.

ErikSchierboom: @macta You are correct that in principle, it would be slightly easier to have the inputs grouped into a single parameter.

petertseng:
> Also - the use of a nested case was redundant, they were essentially all top level.

The "Also" there implies to me that this PR should be two commits, one for removing the extra nesting and one for adding `"data"`. However, it appears that others are happy to review it as one commit, so I see I am not productively contributing to this conversation and will stop harping on this immediately.

exercises/etl/canonical-data.json
```diff
@@ -1,77 +1,76 @@
 {
   "exercise": "etl",
-  "version": "1.0.0",
+  "version": "1.0.1",
```
SleeplessByte: This is not the correct next version. Changing the input like this is not backwards compatible, so should probably be "2.0.0".

petertseng:
this was done in the linked PR

PR #1488, opened 2019-03-26
darts: add tests to disallow absolute addition

yawpitch:
The current tests allow an erroneous-but-passing solution where the sum of the absolute values of the coordinates of the dart are compared to the radius thresholds (ie return 5 if `abs(x) + abs(y)` is greater than 1 but less than or equal to 5), rather than the calculated distance of the dart from the origin. Thanks to [grwkremilek](https://exercism.io/tracks/python/exercises/darts/solutions/fd550c5405d64d84be3312566b300e1b) for the find.

### DIFF
--- a/exercises/darts/canonical-data.json
+++ b/exercises/darts/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "darts",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "cases": [
     {
       "description": "Return the correct amount earned by a dart landing in a given point in the target problem.",
--- a/exercises/darts/canonical-data.json
+++ b/exercises/darts/canonical-data.json
@@ -67,6 +67,33 @@
            "y": -0.1
          },
          "expected": 10
+        },
+        {
+         "property": "score",
+         "description": "A dart whose coordinates sum to > 1 but whose radius to origin is <= 1 is scored in the inner circle",
+         "input": {
+           "x": 0.4,
+           "y": 0.8
+         },
+         "expected": 10
+        },
+        {
+         "property": "score",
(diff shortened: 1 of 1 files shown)


rpottsoh:
Update [versioning](https://github.com/exercism/problem-specifications#test-data-versioning) reminder.

rpottsoh:
The square root of the sum of the squares of *`x`* and *`y`* yields the correct distance from origin of the dart board.  Thanks for adding these tests I think they will certainly catch solutions that are not calculating the distance correctly.

petertseng:
Since it came up today.

My approval is conditional on commits being squashed. (It doesn't matter to me whether the submitter squashes them or the merger squashes them, as long as someone does)

yawpitch:
Agreed, but in this case I rather assume that’s for the merger.

ErikSchierboom:
Squashed and merged. Thanks @yawpitch!

rpottsoh:
Thinking more about this change....  I still agree it was a good change, but am now wondering would it have been better to just update 

- `"A dart lands in the outer circle"` with x = 4 and y = 8, 
- `"A dart lands in the middle circle"` with x = 2 and y = 4, and 
- `"A dart lands in the inner circle"` with x = 0.4 and y = 0.8.

It would then not be necessary to add three more tests.

yawpitch:
I had considered doing that, but the problem is that someone could revert that change or otherwise modify those values in the future and inadvertently and silently put the tests back into a place where they have a gap in coverage for certain values of x and y. The only mechanism for flagging that the specific chosen _points_ are significant -- as opposed to them being arbitrary points that fall within the chosen _rings_ -- is the "description" field, which isn't exactly robust, so by making them distinct tests they both close the gap and make it harder to inadvertently re-open it, hopefully without introducing any significant overhead.

rpottsoh:
Excellent point.  Thanks for clarifying.

PR #1490, opened 2019-03-27
markdown: unordered lists with surrounding content

yawpitch:
Adds a tests for unordered lists with surrounding content; many solutions I've seen are not properly closing an unordered list when there's content below the list's final item.

### DIFF
--- a/exercises/markdown/canonical-data.json
+++ b/exercises/markdown/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "markdown",
-  "version": "1.3.0",
+  "version": "1.4.0",
   "comments": [
     "Markdown is a shorthand for creating HTML from text strings."
   ],
@@ -100,6 +100,14 @@
             "markdown": "This is a paragraph with # and * in the text"
         },
         "expected": "<p>This is a paragraph with # and * in the text</p>"
+    },
+    {
+        "description": "unordered lists close properly with preceding and following lines",
+        "property": "parse",
+        "input": {
+            "markdown": "# Start a list\n* Item 1\n* Item 2\nEnd a list"
+        },
+        "expected": "<h1>Start a list</h1><ul><li>Item 1</li><li>Item 2</li></ul><p>End a list</p>"
     }
   ]
 }


yawpitch:
IIRC, basic Markdown doesn’t itself specify a blank line at the end of an unordered list, so I don’t think so, but since we’re sort of doing an opinionated subset of Markdown functionality (the spec does require that list items with a new line between them get wrapped in `<p>` tags), I’m not entirely sure.

petertseng:
So, as you can see in https://github.com/exercism/problem-specifications/pull/1491 GitHub's Markdown renderer would put that "End a list" as part of the list element. I also see the same behaviour in https://parsedown.org/demo

Although I muse that it would be easier to implement it the other way ("End a list" not part of the list element), since making it part of the list element seems to require me to keep state in my parser from line to line, which it currently does not.

This comment is solely meant to be informational in nature. It does not suggest any specific course of action to be taken in this PR.

yawpitch:
Yeah, this is the problem of the ambiguity in the original spec... IIRC the original markdown.pl didn’t require the white space.

Perhaps we should ensure the tests adhere to [CommomMark](https://commonmark.org/), at least for the subset of the spec we’re asking for in the tests?

petertseng:
**I will not object to this PR being merged in its current form**.

It is possible to write a parser that passes both the newly-added case and all previous cases, without resorting to anything that smells suspicious (a strange special case, or anything similar to that); https://github.com/petertseng/exercism-problem-specifications/blob/verify/exercises/markdown/verify.rb serves as one existence proof of the above statement.

I will leave it to dealer's choice (the choice of whoever wants to do the work) on whether additional action should be taken before or after merging. Here are some questions that may help decide that, all of which I have no opinion on.

I see that https://github.com/exercism/problem-specifications/blob/master/exercises/markdown/description.md cites https://guides.github.com/features/mastering-markdown/ . That cited page does not specify, so we can agree that it is ambiguous.

Would adding a second newline after `Item 2` make the newly-added case less ambiguous? Is it good for these tests to be unambiguous?

One might observe that, for the purposes of what the exercise might teach to a student, it is not of paramount importance whether this implementation matches any particular spec or existing implementation, however having these qualities may make it easier for a student to complete a submission (potentially more resources available).

yawpitch:
A quick survey of online Markdown parsers seems to indicate a general preference for a newline after a list. That said, several seem to want an extra newline _before_ a list -- definitely not stated in Gruber's original spec -- and the treatment of extra newlines _between_ list items seems to be inconsistent. All of this is because of ambiguity around how to handle list items that span multiple lines, a problem we've rather chosen to ignore in our example implementation.

Since the author of the language didn't seem concerned with resolving the ambiguity, and because the ambiguity really stems from a form of list item we've not chosen to require support for, I'd say we can fall on the question wherever we'd like.

Personally I'd tend to say let's keep the implementation costs low and maximize learning value ... requiring the extra trailing newline adds (some) complexity to the solution, but I can see no way in which it increase the educational benefit of the exercise over not having that complexity.

So my vote's for single newline.

petertseng:
I deem that an appropriate amount of time has passed for those who would have opinions to express them. Thus, I conclude that the combination of these three: 1) An Approval 2) my "I'm OK with anything", and 3) a vote for the single newline, we should merge as is. I'll add a short note to this effect in the commit message. Thanks for this interesting test.

rpottsoh:
Would there be any merit in adding this case?  Related to the case added by this PR.

```json
   {
        "description": "two separated unordered lists",
        "property": "parse",
        "input": {
            "markdown": "* First Item\n* Second Item\nOrdinary Text\n* new list item"
        },
        "expected": "<ul><li>First Item</li><li>Second Item</li></ul><p>Ordinary Text</p><ul><li>new list item</li></ul>"
    }
```

This was recently suggested to me by one of my students.  They found it useful to help double check the RegEx they are using so I am not sure how useful this test might be in a language agnostic sort of way.

cmccandless:
> Would there be any merit in adding this case?

The merit I can immediately think of is that solutions that make use of a static list/array for holding list elements would possibly fail on this. I think it's worth opening an issue or PR to discuss it.

Issue #1470, opened 2019-02-26
Tricky Triplets - a storified replacement for Two Fer


kytrinyx:
**Background**

When we redesigned Exercism we gave each exercise its own icon. This made a huge difference. The tracks feel a lot more fun and approachable.

A common complaint about exercises is that they feel overwhelming or unapproachable or hard to understand. Often this is because there's too much jargon in the description. Other times it's because the domain is foreign to the reader.

Lately @iHiD, @nicolechalmers, and I have been discussing how we'd like every single exercise on Exercism to be fun and approachable, and we've been discussing how we might accomplish this.

**Suggested solution**

The approach that we think has a lot of merit is to ensure that each exercise has

1. an appealing icon
2. an inviting title
3. an engaging storyline

We aren't going to go all out just yet and mandate that every single exercise be transformed into a story, however we would very much like to talk about stories for any new exercises that are introduced.

Also, we are open to people proposing stories to either tweak or replace an existing exercise.

There will be a balance to strike with stories. We want to hit the goldilocks zone of "not to little" and "not too much". E.g. "two fer" is not quite enough, but we also don't want to have to invent a whole new alternate universe with magical laws and a complex cast of characters.

Since this can feel really abstract to talk about without an example, we looked at what it could mean to turn the two-fer exercise into a story, and it boils down to this:

- strip away all the details
- determine what the technical challenge is

For two-fer, if you strip everything away, the fundamental challenge is to format a simple string with user-provided input, and provide a default/fallback for when the user did not provide a value.

It doesn't matter what the string actually says, what the values are, or what the default value is. That said, we decided to stick with the basic "one for {x, you}, one for me" that two fer already has.

Here's the story that we came up with, as a starting point:

----

**Tricky Triplets**

One year ago, Lori gave birth to three beautiful daughters - Toa, Layla, and Adra - **identical** triplets. She hasn't slept much since!

Lori recently started introducing her daughters to solid food and is using the time-old tradition of eating a bit herself to encourage them: "One for Toa, one for me!", substituting the name of the baby in question, of course.

Recently, her parents came to visit, and they love helping out with the triplets. Unfortunately, they can't always tell them apart.

Your task is to provide a safety net for whichever adult happens to be feeding the triplets. When they can remember the name, they'll use it, but when they can't, they'll just say "you" instead.

In coding terms, you're going to write a short program which can be called in two ways, to either `encourage` a specific baby by name, or to `encourage` whichever baby is being fed at the time.

[...]

PR #1503, opened 2019-04-06
Reverse String: Add emoji bonus to description

ccouzens:
I've added an emoji bonus to the description of reverse string.
This introduces Unicode. Knowing that there are different ways to encode
text, and that not all characters are the same size is important.

I've made it a bonus because it's not straight forward in every
language.

### DIFF
--- a/exercises/reverse-string/canonical-data.json
+++ b/exercises/reverse-string/canonical-data.json
@@ -1,9 +1,9 @@
 {
   "exercise": "reverse-string",
-  "version": "1.1.0",      
+  "version": "1.2.0",
   "comments": [
-        "If property based testing tools are available, a good property to test is reversing a string twice: reverse(reverse(string)) == string"
-      ],
+    "If property based testing tools are available, a good property to test is reversing a string twice: reverse(reverse(string)) == string"
+  ],
   "cases": [
     {
       "description": "an empty string",
@@ -31,7 +31,7 @@
     },
     {
       "description": "a sentence with punctuation",
-      "property": "reverse",  
+      "property": "reverse",
       "input": {
         "value": "I'm hungry!"
       },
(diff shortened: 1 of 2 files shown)


ccouzens:
Maybe this would be better as a whole new exercise- Unicode Reverse String.

This way languages that don't make Unicode convenient can easily skip it.

Let me know your thoughts

coriolinus:
The Rust track already has two track-specific tests like this one:

https://github.com/exercism/rust/blob/42166418fc8370bf2a3d2a89774d5adda439e870/exercises/reverse-string/tests/reverse-string.rs#L49-L62

The first of these tests wide characters, and is part of the basic test suite. This is because in rust it is easy to reverse a string of characters regardless of width.

The second is feature-locked, because it involves reversing on a grapheme basis, which is more complicated. By default this test is excluded from the tests run; it only runs if you run the test suite with a particular flag. It's meant to give students who are interested some more insight into how unicode works.

ccouzens:
I purposely didn't mention multiple characters making up a single grapheme and picked emojis where this wasn't a concern. Rust has a good 3rd party library for this, but I don't think that's true for a lot of languages.

---

I think there are 3 types of languages to consider:

1. Languages like Rust where it's almost difficult not to correctly handle unicode characters.
2. Languages like Java or Javascript where the initial solution likely doesn't handle unicode characters correctly, but can be made to work without too much effort.
3. Languages like C where the standard library doesn't help with unicode. (forgive me if I've mis-characterized C)

I think the bonus text is helpful for students of # 1 and # 2. But I worry that students of # 3 might get frustrated trying to make it work.

For this reason, I'm beginning to think within the problem specifications it may be better to make Unicode reverse string into a separate exercise.

coriolinus:
I agree with your categorization of languages. I think that the right level of unicode depends on which type a language is.

For type 1 languages, I think that wide chars such as the emoji you used or the kanji Rust uses should be part of the basic test suite, and bonuses should introduce something new, such as multi-character graphemes. The language does a lot to make things easy, so let's show the students how easy it really is.

For type 2 languages, I agree with you: a bonus test with wide characters should be part of the normal suite, but reversing multi-character graphemes might be too much unless there exists a good external library, so it should be optional and track-specific. 

For type 3 languages, a wide character test might be part of the bonus tests, but doesn't need to be part of the standard test suite.

In all these cases, the bonus exercise introduces a level of difficulty above and beyond the standard test suite. However, because the difficult bonus for type 3 languages is appropriately part of the standard test suite for type 1 languages, it is very difficult to add reasonable test cases in a track-independent way.

Possibly the best way to handle this is a detailed comment in the canonical data talking about the categorization of languages and giving some sample inputs and outputs both for wide characters and multi-char graphemes. That way each track could implement idiomatic tests according to its categorization.

ccouzens:
> Possibly the best way to handle this is a detailed comment in the canonical data talking about the categorization of languages and giving some sample inputs and outputs both for wide characters and multi-char graphemes. That way each track could implement idiomatic tests according to its categorization.

@coriolinus, have I done what you had in mind in my latest commit?

NobbZ:
Hi @ccouzens!

I welcome your PR, but some years ago we agreed upon a "no unicode" rule, since unicode handling in some languages is below par.

But we also agreed upon that extra exercises might get created that do unicode.

Even as an optional exercise, unicode might be problematic, as unicode might confuse generators of languages that do not deal very well with unicode, which we also want to avoid.

ErikSchierboom:
Oh and as @NobbZ said, we do welcome PR's!

SleeplessByte:
Unicode exercises are on the list of soon to be added!

cmccandless:
I like the idea of separate exercises for handling unicode. That way, tracks that don't have good unicode support can mark the exercise as `"foregone": true` and move on instead of having to mess with custom generator handling.

ccouzens:
Thanks everyone 🙂

> What do you think @ccouzens? Are you okay with waiting for/helping with a separate Unicode exercise?

I like that idea. I hope to have time tomorrow to move these changes into a new unicode-reverse-string exercise. There would be other advantages beyond the technical ones, such as being able to use the Readme to give a brief overview of unicode.

ErikSchierboom:
> I hope to have time tomorrow to move these changes into a new unicode-reverse-string exercise.

Could you perhaps open an issue first? We are trying to give all (new) exercises a background story, so a little discussion on the specific story first could help. For reference, see #1470.

ccouzens:
@ErikSchierboom 

> Could you perhaps open an issue first? We are trying to give all (new) exercises a background story, so a little discussion on the specific story first could help.

I've opened #1507. I don't think there was much scope for a story with reversing strings, so I've made the challenge about truncating them to 5 characters.

ccouzens:
I'm going to close this pull request, because it looks like I'll be creating a different exercise instead #1507

PR #1504, opened 2019-04-06
Correct Queen Attack test values

TeaDrivenDev:
There is a flaw in the Queen Attack test suite that causes the most naive implementation that makes the tests green to actually be incomplete.

The tests for "third diagonal" (2, 2 and 1, 1) and "fourth diagonal" (2, 2 and 5, 5) actually test for the same diagonal (A1 to H8) and are fulfilled by the condition `row1 = column1 && row2 = column2`. This does not cover the parallels of that diagonal, however, so if there was an additional test for (0, 6) and (1, 7) that would still be red.

This is fixed by replacing the redundant "fourth diagonal" test values with said (0, 6) and (1, 7), which will require the condition to be changed to `row1 - column1 = row2 - column2`, thus covering all diagonals from "top left" to "bottom right".

That said, the tests for "first diagonal" and "second diagonal" are actually also covering the same diagonal (the one from "bottom left" to "top right" that includes 2, 2), but this PR is not addressing that.

### DIFF
--- a/exercises/queen-attack/canonical-data.json
+++ b/exercises/queen-attack/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "queen-attack",
-  "version": "2.2.0",
+  "version": "2.2.1",
   "comments": [
     "Testing invalid positions will vary by language. The expected",
     "value of 'error' is there to indicate some sort of failure should",
@@ -204,14 +204,14 @@
           "input": {
             "white_queen": {
               "position": {
-                "row": 2,
-                "column": 2
+                "row": 0,
+                "column": 6
               }
             },
             "black_queen": {
               "position": {
-                "row": 5,
-                "column": 5
+                "row": 1,
+                "column": 7
               }
             }
           },
--- a/exercises/queen-attack/canonical-data.json
+++ b/exercises/queen-attack/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "queen-attack",
-  "version": "2.2.1",
+  "version": "2.3.0",
   "comments": [
     "Testing invalid positions will vary by language. The expected",
     "value of 'error' is there to indicate some sort of failure should",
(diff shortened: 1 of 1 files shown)


cmccandless:
I disagree. A solution that uses the difference of the positions would be testing against "third diagonal" `(2,2) - (1,1) = (1,1)`, versus "fourth diagonal" `(2,2) - (5,5) = (-3,-3)`. For such solutions, each test case verifies against a different edge case.

Not to say that such solutions are quite optimal, but they are completely valid, if implemented correctly. Keeping both cases ensures that solutions like this handle both edge cases.

cmccandless:
My apologies; I misread your initial statement (not enough coffee yet).

However, I do stand by my assertion that there needs to be a case where the second queen is closer to the bottom-left corner of the board than the first. What about switching the new suggested values?

TeaDrivenDev:
Makes sense; updated.

exercises/queen-attack/canonical-data.json
```diff
@@ -1,6 +1,6 @@
 {
   "exercise": "queen-attack",
-  "version": "2.2.0",
+  "version": "2.2.1",
```
cmccandless: RE: [Test Data Versioning](https://github.com/exercism/problem-specifications#test-data-versioning) guidelines:

> MINOR version changes
> The MINOR version should change when you add functionality in a backwards-compatible manner, make non-breaking changes that alter the meaning of the test suite, make previously passing solutions possibly fail, or failing solutions fail at a different spot.
>
> There are examples of changes requiring a MINOR version change:
> - Adding or deleting test cases.
> - **Changing the test cases inputs and/or outputs.**
> - Changing the test cases ordering.
>
> MINOR changes would never break well-designed test generators, because the test-generation logic remains exactly the same.

```suggestion
  "version": "2.3.0",
```

SleeplessByte:
:rocket: lgtm and with 3 approvals I think this is good to go.

PR #1505, opened 2019-04-06
Zero is an Armstrong number and requires specific testing

paparomeo:
[`0` is also an Armstrong number](https://en.wikipedia.org/wiki/Narcissistic_number) and needs to be tested for explicitly if students are using a base 10 logarithm to calculate the number of digits for which `0` is a case that needs special handling.

### DIFF
--- a/exercises/armstrong-numbers/canonical-data.json
+++ b/exercises/armstrong-numbers/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "armstrong-numbers",
-  "version": "2.0.0",
+  "version": "1.1.0",
   "cases": [
     {
       "description": "Zero is an Armstrong number",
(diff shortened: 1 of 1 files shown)


exercises/armstrong-numbers/canonical-data.json
```diff
@@ -1,7 +1,15 @@
 {
   "exercise": "armstrong-numbers",
-  "version": "1.0.0",
+  "version": "2.0.0",
```
cmccandless: This should be `1.1.0`

paparomeo: Indeed! I misread [_"Major version changes"_](https://github.com/exercism/problem-specifications#major-version-changes) assuming "new property" was a new test case and skimmed over "Minor version changes" where this was made clear. Thanks!

paparomeo:
@cmccandless: Version bump corrected to minor version. Thanks!

PR #1508, opened 2019-04-11
yacht: Add "No pairs but not a big straight" test case

Mehni:
simple implementations checking only uniques will otherwise pass the test suite.

I submitted a solution with

```
case YachtCategory.BigStraight:
    return dice.OrderBy(x => x).Distinct().Count() == 5 && dice.Max() == 6 ? 30 : 0;
```

which incorrectly identifies 6, 5, 4, 3, 1 as a big straight.

### DIFF
--- a/exercises/yacht/canonical-data.json
+++ b/exercises/yacht/canonical-data.json
@@ -1,6 +1,6 @@
 {
     "exercise":"yacht",
-    "version":"1.1.0",
+    "version":"1.2.0",
     "comments":[
         "The dice are represented always as a list of exactly five integers",
         "with values between 1 and 6 inclusive. The category is an string.",
@@ -235,6 +235,15 @@
             "input":{
                 "dice":[6, 5, 4, 3, 2],
                 "category":"little straight"
+            },
+            "expected":0
+        },
+        {
+            "description":"Only consecutive numbers form a straight",
+            "property":"score",
+            "input":{
+                "dice":[6, 5, 4, 3, 1],
+                "category":"big straight"
             },
             "expected":0
         },
--- a/exercises/yacht/canonical-data.json
+++ b/exercises/yacht/canonical-data.json
@@ -239,7 +239,7 @@
             "expected":0
         },
         {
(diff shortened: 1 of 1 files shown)


exercises/yacht/canonical-data.json
```diff
@@ -238,6 +238,15 @@
             },
             "expected":0
         },
+        {
+            "description":"Only consecutive numbers form a straight",
```
rpottsoh: Shouldn't this say `big straight`?
```suggestion
            "description":"Only consecutive numbers form a big straight",
```
The input depicts a small straight I believe.

I otherwise like this change.

petertseng: Since the input doesn't depict a straight at all, and mistaken implementations may have interprted `6, 5, 4, 3, 1` as either a big or small straight, I think we don't want to say "big" in the description.

If there is a concrete improvement to be made, then here is my observation:

Reading this description "only consecutive numbers form a straight" some might say "okay this is an interesting general description of any of a large set of test cases. What does it say about THIS test case?"

So, to answer the question of "what are the implications of this description for THIS test case", I would say the description should be "Non-consecutive numbers do not form a straight".

petertseng: EBWODP:

> as either a big or small straight, I think we don't want to say "big" in the description.

Oh, actually the category is in the desciption.

And this case is already present in the little straight category, so this PR just adds it to the big too. Okay.

My description suggestion still stands, just the rationale for it is different :)

petertseng: EBWODP:

> Oh, actually the category is in the description.

s/description/input/

Mehni: Let me add another suggestion to the list then. There's already a `No pairs but not a little straight` exercise which solves the same edge case for little straight.

Instead of "Only consecutive numbers form a straight", it could be `No pairs but not a big straight`. That improves consistency between the various tests.

Mehni:
Apologies for the delay. I hope you all agree this new test name is more in line with the rest of the test suite.

PR #1510, opened 2019-04-14
Darts: Flatten sub-cases to single level - to simplify code generation

macta:


exercises/darts/canonical-data.json
```diff
@@ -1,101 +1,99 @@
 {
   "exercise": "darts",
-  "version": "1.2.0",
+  "version": "1.2.1",
```
SleeplessByte: This is a breaking change. Should be 2.0.0

macta: I don't think there is anything for me to do, as you've increased the version number right (and I'll remember for future that while functionaly equivalent - its structurally different)

SleeplessByte: I've not increased it?

It really should be 2.0.0.

macta: Ah yes - momentary mental freeze, will resubmit

SleeplessByte:
Thank you @macta.

I've merged this as we had three (3) approvals.

PR #1511, opened 2019-04-18
space-age: forces a second branch for each planet

michaelglass:
previously, space age could hard code the test results for each planet. With two examples per planet, it makes more sense to calculate the results

(started here: https://github.com/exercism/elm/pull/250)

### DIFF
--- a/exercises/space-age/canonical-data.json
+++ b/exercises/space-age/canonical-data.json
@@ -39,7 +39,7 @@
       "expected": 8.3
     },
     {
-      "description": "6 Earth years  in Venus years",
+      "description": "6 Earth years in Venus years",
       "property": "age",
       "input": {
         "planet": "Venus",
--- a/exercises/space-age/canonical-data.json
+++ b/exercises/space-age/canonical-data.json
@@ -18,7 +18,7 @@
         "planet": "Earth",
         "seconds": 63115200
       },
-      "expected": 2
+      "expected": 2.0
     },
     {
       "description": "67.6 Earth years in Mercury years",
--- a/exercises/space-age/canonical-data.json
+++ b/exercises/space-age/canonical-data.json
@@ -3,15 +3,6 @@
   "version": "1.3.0",
   "cases": [
     {
-      "description": "age on Earth",
-      "property": "age",
-      "input": {
-        "planet": "Earth",
-        "seconds": 1000000000
-      },
-      "expected": 31.69
-    },
-    {
       "description": "age on Mercury",
       "property": "age",
       "input": {
@@ -19,69 +10,6 @@
         "seconds": 2134835688
       },
       "expected": 280.88
-    },
-    {
-      "description": "age on Venus",
-      "property": "age",
-      "input": {
-        "planet": "Venus",
-        "seconds": 189839836
-      },
-      "expected": 9.78
-    },
-    {
-      "description": "age on Mars",
-      "property": "age",
-      "input": {
-        "planet": "Mars",
-        "seconds": 2129871239
-      },
-      "expected": 35.88
-    },
-    {
-      "description": "age on Jupiter",
-      "property": "age",
-      "input": {
-        "planet": "Jupiter",
-        "seconds": 901876382
-      },
-      "expected": 2.41
-    },
-    {
-      "description": "age on Saturn",
-      "property": "age",
-      "input": {
-        "planet": "Saturn",
-        "seconds": 2000000000
-      },
-      "expected": 2.15
-    },
-    {
-      "description": "age on Uranus",
-      "property": "age",
-      "input": {
-        "planet": "Uranus",
-        "seconds": 1210123456
-      },
-      "expected": 0.46
-    },
-    {
-      "description": "age on Neptune",
-      "property": "age",
-      "input": {
-        "planet": "Neptune",
-        "seconds": 1821023456
-      },
-      "expected": 0.35
-    },
-    {
-      "description": "2 earth years in earth years",
-      "property": "age",
-      "input": {
-        "planet": "Earth",
-        "seconds": 63072000
-      },
-      "expected": 2
     },
     {
       "description": "2 earth years in mercury years",
@@ -93,6 +21,15 @@
       "expected": 8.3
     },
     {
+      "description": "age on Venus",
+      "property": "age",
+      "input": {
+        "planet": "Venus",
+        "seconds": 189839836
+      },
+      "expected": 9.78
+    },
+    {
       "description": "2 earth years in venus years",
       "property": "age",
       "input": {
@@ -100,6 +37,33 @@
         "seconds": 63072000
       },
       "expected": 3.25
+    },
+    {
+      "description": "age on Earth",
+      "property": "age",
+      "input": {
+        "planet": "Earth",
+        "seconds": 1000000000
+      },
+      "expected": 31.69
+    },
+    {
+      "description": "2 earth years in earth years",
+      "property": "age",
+      "input": {
+        "planet": "Earth",
+        "seconds": 63072000
+      },
+      "expected": 2
+    },
+    {
+      "description": "age on Mars",
+      "property": "age",
+      "input": {
+        "planet": "Mars",
+        "seconds": 2129871239
+      },
+      "expected": 35.88
     },
     {
       "description": "2 earth years on mars",
@@ -111,6 +75,15 @@
       "expected": 1.06
     },
     {
+      "description": "age on Jupiter",
+      "property": "age",
+      "input": {
+        "planet": "Jupiter",
+        "seconds": 901876382
+      },
+      "expected": 2.41
+    },
+    {
       "description": "2 earth years on jupiter",
       "property": "age",
       "input": {
@@ -118,6 +91,15 @@
         "seconds": 63072000
       },
       "expected": 0.17
+    },
+    {
+      "description": "age on Saturn",
+      "property": "age",
+      "input": {
+        "planet": "Saturn",
+        "seconds": 2000000000
+      },
+      "expected": 2.15
     },
     {
       "description": "2 earth years on saturn",
@@ -129,6 +111,15 @@
       "expected": 0.07
     },
     {
+      "description": "age on Uranus",
+      "property": "age",
+      "input": {
+        "planet": "Uranus",
+        "seconds": 1210123456
+      },
+      "expected": 0.46
+    },
+    {
       "description": "2 earth years on uranus",
       "property": "age",
       "input": {
@@ -136,6 +127,15 @@
         "seconds": 63072000
       },
       "expected": 0.02
+    },
+    {
+      "description": "age on Neptune",
+      "property": "age",
+      "input": {
+        "planet": "Neptune",
+        "seconds": 1821023456
+      },
+      "expected": 0.35
     },
     {
       "description": "2 earth years on neptune",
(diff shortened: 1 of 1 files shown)


michaelglass:
@cmccandless ordered by planet

exercises/space-age/canonical-data.json
```diff
@@ -3,22 +3,22 @@
   "version": "1.3.0",
   "cases": [
     {
-      "description": "age on Earth",
+      "description": "age on Mercury",
```
cmccandless: I like the new `x earth years in mercury years` naming scheme. Perhaps we should apply the same scheme to the existing tests?

exercises/space-age/canonical-data.json
```diff
@@ -3,22 +3,22 @@
   "version": "1.3.0",
```
cmccandless: Tests are being added, renamed, or removed, so we'll need to bump the minor version number.

`"version": "1.4.0",`

michaelglass: bump it twice?

cmccandless: Ah, no, don't do that. I didn't realize I was only looking at the changes for a single commit, not the whole PR. You're good already.

exercises/space-age/canonical-data.json
michaelglass: reworked everything as 2 and 20 earth years

michaelglass:
@cmccandless addressed your two tick boxes

exercises/space-age/canonical-data.json
```diff
@@ -66,13 +120,31 @@
...
+    },
+    {
+      "description": "2 Earth years in Neptune years",
+      "property": "age",
+      "input": {
+        "planet": "Neptune",
+        "seconds": 63115200
+      },
+      "expected": 0.01
+    },
```
cmccandless: Comma after last array item is invalid.

```suggestion
    }
```

michaelglass: I've never seen this, "commit suggestion" thing. Neat!

exercises/space-age/canonical-data.json
```diff
@@ -21,7 +30,16 @@
...
+      "description": "2 Earth years in Mercury years",
+      "property": "age",
+      "input": {
+        "planet": "Mercury",
+        "seconds": 63115200
+      },
+      "expected": 8.3
+    },
+    {
+      "description": "6 Earth years  in Venus years",
```
petertseng: it would probably be good to remove the extra space between `years` and `in`, for the purpose of consistency (no other case has this extra space)

exercises/space-age/canonical-data.json
```diff
@@ -12,7 +12,16 @@
...
     },
     {
-      "description": "age on Mercury",
+      "description": "2 Earth years in Earth years",
+      "property": "age",
+      "input": {
+        "planet": "Earth",
+        "seconds": 63115200
+      },
+      "expected": 2
```
ErikSchierboom:
```suggestion
      "expected": 2.0
```

This is just something to make some test generators a bit more happy, the ones that discern between ints and float.

bitfield:
I've literally never encountered a solution which hard-coded the results for each planet. If I did, I'd approve it. The tests are the tests; if the code passes the tests, it is correct. (That's not to say it couldn't be _improved_; that's what mentoring is for.)

NobbZ:
No, just passing the tests is not the barrier. The barrier is
understanding.

If the student just hard coded the return values for each testcase he
proved to not have understood the exercise. I'd try to explain it to them
then.

John Arundel <notifications@github.com> schrieb am Sa., 20. Apr. 2019,
13:13:

bitfield:
> If the student just hard coded the return values for each testcase he proved to not have understood the exercise.

This is a fundamental philosophical disagreement about what tests are for! (Which can't be resolved in a GitHub thread. I note it merely for the record, and to make it clear that intelligent people can disagree honourably about this issue.)

SleeplessByte:
True. But _because_ we have an agreement about not adding tests for specific implementation and we have an agreement about what exercism is and is for, in this case, I agree with @nobbz.

michaelglass:
I disagree, in so far as hardcoding the result is the minimal solution, and I think it's often easy to argue that the minimal solution is best, e.g. https://github.com/exercism/problem-specifications/pull/1512#issuecomment-485094254.

For me there are two questions:
1. What is it that exercism wants to teach?
2. Is it desirable to require mentors to intervene?

My opinions:
1. Personally, I think a more thorough test suite _teaches_ better practice. I also think the minimal implementation is often considered best. E.g. I find [this argument](https://github.com/exercism/problem-specifications/pull/1512#issuecomment-485094254) compelling.
2. The feedback loop of a test suite is much faster than a mentor. If I have to wait hours or days for a response, I've already lost flow. If a test fails, I know immediately. If I can document or codify my suggestions (e.g. through the readme, the test suite) then I'm providing students with more immediate feedback.

bitfield:
> we have an agreement about not adding tests for specific implementation

Then doesn't this PR fall foul of that agreement?

> we have an agreement about what exercism is and is for

I think this conversation demonstrates otherwise.

SleeplessByte:
> I think this conversation demonstrates otherwise.

You can find the [goal of exercism here](https://github.com/exercism/docs/blob/master/about/goal-of-exercism.md) and see [this document](https://github.com/exercism/automated-mentoring-support/blob/f2c7c647d92782481b84c46363766754bf803824/docs/guidelines.md) about what the automated mentoring project will approve and disapprove.

These two documents together are the agreement I am referring to :) 

> Then doesn't this PR fall foul of that agreement?

I think that's definitely an argument to be made. I don't see the point of adding these tests other than in the 1% of cases save a mentor from saying "but now it doesn't work for other input". For that reason I have _not_ ✔️ approved these changes myself.

michaelglass:
1. What is the _harm_ that this PR introduces?
- increased complexity?

2. How can I participate in a way that feels more light-weight.
- as a new member of the community I wish that
    - I had put in less work on this PR.
    - Others had put in less work on this PR. Three of this PR's six reviewers put in time editing something that will (likely) get thrown out.
    - I had clearer guidelines around how to propose a change like this. (It appears: a small issue describing the test case is the approach).
    - It was more clear to me who is responsible for making decisions.

Now, of course if this PR is going to make Exercism _worse_ than it's better that discarded then held dear because of the effort. But if it's discarded, I'd love to makee it harder to waste contributor's time.

To that end: I think issue & PR templates would be a nice next step.

SleeplessByte:
@michaelglass I think it's still very valuable that you made this PR. How we do things and why we do things aren't clear for everyone (including myself -- I had to look up the written docs) and discussions like these force all of us to up our game and look at the guideliness written in the various repo's. I apologise you put in so much work; but don't think this isn't fruitful. 

> I had clearer guidelines around how to propose a change like this. (It appears: a small issue describing the test case is the approach).

Did [CONTRIBUTING.md](https://github.com/exercism/problem-specifications/blob/master/CONTRIBUTING.md#problem-metadata) not suffice in this case? It says:
```md
We try to keep the generic descriptions generic--we avoid implementation-specific examples, and 
try not to be too prescriptive about suggesting how a problem might be solved.
```

> It was more clear to me who is responsible for making decisions.

Everyone here basically has "equal rights" as far as I can tell, but Jeremy (`iHiD`) and Katrina (kytrinyx) always have final and ultimate say. 

In this repository @petertseng takes care of a _lot_ of the important steps; e.g. I see him make sure all implementing tracks have a `sync` issue opened and interesting discussion around the merit of certain changes. Furthermore, some of the exercises are designed by people here such as the `resistor-xxx` series by @ErikSchierboom and @F3PiX, so we tend to ping them when a change is made to those exercises.

That said, the general consensus here is in `problem-specifications` that all us `track-maintainers` chime in when we want to and we merge if we get at least three ✔️ without strong ❌ .

> To that end: I think issue & PR templates would be a nice next steps.

I think this is a _great_ idea!

michaelglass:
for #1512, it's clear that I was pushing for a specific implementation strategy. For this case, I don't think I am (unless the specific implementation is, "actually do what the readme says").

I'll follow-up with some proposals for issue templates when I get the time. In the mean time, should I close this PR?

kytrinyx:
A few thoughts:

It sounds like here the key problem is that there was no discussion up front about what constraints the exercise has.

I think that it would be valuable to have a _short_ template, preferably within HTML comments, so that if the template pieces don't get deleted, we don't have to always look at the text. E.g. something like this.








In this case, I actually think that adding a few tests to ensure that the implementation is not hard-coded seems perfectly reasonable. That is generally considered a reasonable practice in TDD ("triangulation"). I didn't do it originally when I added the exercise, because it didn't occur to me that people would hard-code the values.

If we do add tests, I think that we should be careful to keep the values within the 32-bit range.

cmccandless:
> E.g. something like this. *[invisible comments below]*

```
<!-- NOTE: changing exercise specifications has a lot of complicated considerations. You may want to open an issue to discuss before opening a PR. -->

<!-- What problem did you observe? How do you propose to fix it? -->

<!-- Please look at earlier PRs and commits for this exercise. What relevant points have been discussed previously? -->
```

kytrinyx:
d'oh. Thank you @cmccandless !

SleeplessByte:
IMO, the only _correct_ way to ensure people don't hardcode the value is randomly _generate_ the input (and output). Adding a second set of values can still make some people hardcode the values, albeit probably less.

I'm assuming there was a reason the tests for `two-fer` are only <nothing>. `Alice` and `Bob`? We still see people hardcode these as well.

👍 * 💯 on the comment in the PR template!

sshine:
Perhaps we can document this policy against additional hardcoded unit tests to avoid hardcoded solutions so that @michaelglass'es efforts were not in vain, and so that others will not repeat the process of wasting time on similar PRs.

michaelglass:
The "only correct test are quickcheck" argument similar to "perfect is the enemy of good". Does the metadata support generative test?

The "What about the two-fer tests" is in the realm of [whataboutism](https://en.wikipedia.org/wiki/Whataboutism).

When is a test change to an existing test suite valid?

SleeplessByte:
@michaelglass my comment really isn't whataboutism. It's the first exercise in _most_ tracks and I am assuming here that the inputs were purposefully **not** generated. 

If we now are trending towards "let's make it harder for people to hard-code", I was going to suggest we actually make this change in _all_ exercises we see people hardcode, including `two-fer`.

ErikSchierboom:
> If we now are trending towards "let's make it harder for people to hard-code", I was going to suggest we actually make this change in all exercises we see people hardcode, including two-fer.

Personally, I see very, very little hardcoded exercises. I see so little of that in fact, that I'm against adding test cases for that specific purpose :)

SleeplessByte:
It's less than 1 in each hundred (going of the data I got for writing the analyzer)

SleeplessByte:
@michaelglass Do you think something like [this](https://github.com/exercism/problem-specifications/pull/1511#issuecomment-486400333) (my last comment) would have been useful in the docs?

michaelglass:
I think `Contributing` is already a huge doc to wrap your mind around, but I think it'd be useful for a PR template.

But, I think I would have considered this PR falling under this case

> completeness, e.g. a test is missing which allows for an incorrect implementation (for example tests were not covering a leading zero, or odd numbers).

SleeplessByte:
Yes, I understand what you're saying. Thank you for spending the time ❤️ . I don't think it was wasted effort but I am sorry this went the way it did.

PR #1513, opened 2019-04-19
Connect: remove obsolete sentence from description

### LINKED ISSUE #1435: connect: description is not accurate wrt spaces
> with spaces added for readability, which won't be in the representation passed to your code

In each language I've checked (Erlang, Python, Ruby) that is not accurate: the spaces are included in the test strings.

emengd:
The canonical test input is now similar to what is shown in the description, but the description says it isn't. Fixes #1435.

### DIFF
--- a/exercises/connect/description.md
+++ b/exercises/connect/description.md
@@ -13,8 +13,7 @@ Your goal is to build a program that given a simple representation of a board
 computes the winner (or lack thereof). Note that all games need not be "fair".
 (For example, players may have mismatched piece counts.)
 
-The boards look like this (with spaces added for readability, which won't be in
-the representation passed to your code):
+The boards look like this:
 
 ```text
 . O . X .


petertseng:
There is at least one language where the spaces are not present https://github.com/exercism/haskell/blob/master/exercises/connect/test/Tests.hs#L19

So for at least one language, the existing description is accurate and is not obsolete.

Maintainers of any such language are advised to be prepared when this PR is merged.

Therefore, my recommendation: Given there is at least one language track where spaces are not present and at least one language track where they are, it would be useful to examine the implementing tracks and get the number of tracks in each of those two categories.

This recommendation doesn't affect whether to merge this PR, so that decision of whether to do so may continue unimpeded.

petertseng:
EBWODP: consistently say language track, not language (we don't necessarily maintain a language)

emengd:
>Maintainers of any such language track are advised to be prepared when this PR is merged.

i just want to point out that any track where the sentence I remove is accurate is one where they do not use the canonical test data already.

petertseng:
> any track where the sentence I remove is accurate is one where they do not use the canonical test data already.

I wasn't sure if I understood that correctly. Can I use an example to illustrate?

Track https://github.com/exercism/haskell/blob/master/exercises/connect/package.yaml#L2 uses 1.1.0, which is same version as https://github.com/exercism/problem-specifications/blob/master/exercises/connect/canonical-data.json#L3. And all the cases in https://github.com/exercism/haskell/blob/master/exercises/connect/test/Tests.hs are the same. And as seen in https://github.com/exercism/haskell/blob/master/exercises/connect/test/Tests.hs#L19 , the removed sentence applies (the spaces are not in the input given to the function under test).

emengd:
You're right, it's more complicated than what I was saying. The Haskell track maintainers have taken it upon themselves to make the sentence accurate by removing the spaces from the input before forwarding it to the user. I was saying this under the assumption that no track maintainers would preprocess the test data, so if they had space-less data it meant that they had changed it.

I checked a couple of other tracks where they do not change the data and in the example solution they remove the spaces ([python](https://github.com/exercism/python/blob/222c02976c2c0e75d03440893f543c760ca39ba9/exercises/connect/example.py#L24)) or just adapt the neighbour-search logic ([javascript](https://github.com/exercism/javascript/blob/7e4ffe203d528e861d579fc1925fe6135457eec9/exercises/connect/example.js#L43)). I myself encountered this issue in the OCaml track, where the data is also fed to your function as-is.

So now I'm actually not sure what the intention of whoever wrote this sentence was. Perhaps they expected maintainers to preprocess the data as done in the Haskell track. Given that I found 3 tracks who failed to do so, and that I'm not sure how maintainers would reliably notice this, it still sounds correct to remove the sentence.

petertseng:
I agree that it is difficult to ensure that a person implementing an exercise for a given track will know to strip the spaces.

I think that if a track still wishes to strip the spaces, assuming this PR is merged, perhaps they add a short bit in their README saying that they have done so. On the other hand, if the PR is *not* merged, then the sentence is inaccurate for tracks that leave the input as-is, and there is not an easy mechanism to *remove* a sentence from the README. So merging this PR leaves an easier road forward for all tracks. Closing this PR would leave the road forward unclear.

In my eyes, the line of inquiry I mentioned (dealing with tracks that do strip spaces) can reasonably conclude here with the above conclusion. Note that a complete survey of the currently-implementing tracks is still possible, but it doesn't seem like the incentives are aligned for it to be required before merging.

PR #1514, opened 2019-04-24
Forth - Corrected minor typos in the error msgs

macta:
The error msg expectations don't match other similar error patterns and so it breaks auto generated tests

### DIFF
--- a/exercises/forth/canonical-data.json
+++ b/exercises/forth/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "forth",
-  "version": "1.7.0",
+  "version": "1.7.1",
   "comments": [
     "The cases are split into multiple sections, all with the same structure.",
     "In all cases, the `expected` key is the resulting stack",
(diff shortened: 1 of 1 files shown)


macta:
I'm curious why if other tracks generate their tests, that these errors didn't get spotted?

Possibly other tracks don't check the error message of the exception? e.g. I generate:

```
self
		should: [ result := forthCalculator evaluateInstructions: #('1 over' ) ]
		raise: Exception
		whoseDescriptionIncludes: 'only one value on the stack'
		description: 'Should get an error with the correct description'
```

NobbZ:
> I'm curious why if other tracks generate their tests, that these errors didn't get spotted?

Exactly because they are auto-generated and the generator does just use the current test case description to generate the test and therefore just compare the string ;)

I've not yet a forth generator on the Erlang track though...

macta:
@NobbZ - its true, unless you implement a solution to the generated exercise, you wouldn't notice the inconsistent error message (its only now that someone has implemented a solution that we noticed the failing tests and traced it back)

ErikSchierboom:
> Exactly because they are auto-generated and the generator does just use the current test case description to generate the test and therefore just compare the string ;)

This :)

ErikSchierboom:
I see three approvals, so I've merged. Thanks @macta!

Issue #1390, opened 2018-11-10
isogram: Add test case for duplicate following non-letter


bitfield:
I had a Go student submit a solution like this:

```go
	for _, char := range word {
		if !unicode.IsLetter(char) {
			break
		}
```

I'm sure they meant `continue` rather than `break`, but the test cases don't catch this, because there is no case where a non-letter character is followed by a duplicated character. 

An example test case which would catch this problem:

```json
        {
          "description": "hypothetical word with duplicated character after hyphen",
          "property": "isIsogram",
          "input": {
            "phrase": "thumbscrew-jappingly"
          },
          "expected": false
        },
```

rpottsoh:
@bitfield would you be willing to open a PR that makes the necessary changes to the [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/isogram/canonical-data.json) file?  Along with the new case you are proposing the `"version":` would also need to be bumped.  Refer to the README for this repo for direction on how to bump the version.  The PR should close/fix this issue.

bitfield:
Done! Thanks for the help.

PR #1517, opened 2019-05-05
Isogram: Add missing test case that distinguishes from continue and break for loops.

da-edra:
1. Why is this change neccesary?
Because the current test suite allows you to use a `break` a statement inside a
for loop and having all the test pass, this can lead to some confusion between
`break` and `continue` statements that are inside a for loop.

With the current test suite one could do something like this:

```
for i, letter := range word {
		letter = unicode.ToLower(letter)
		if !unicode.IsLetter(letter) {
			break
		}
		if strings.ContainsRune(word[i+1:], letter) {
			return false
		}
	}
```

And have all the test pass.

This could be problematic as the `break` statement gives a better benchmark but it's because of a coincidence with the test cases.

Screenshot_2019-05-05_13-36-18

2. How does it address the issue?
A test case was added where it won't pass if you use a `break` statement, you
need to use a `continue` statement for it to work.

Screenshot_2019-05-05_17-02-17

3. What side effects does this change have?
It clears out the confusion between `break` and `continue`.

### DIFF
--- a/exercises/isogram/canonical-data.json
+++ b/exercises/isogram/canonical-data.json
@@ -3,7 +3,7 @@
   "comments": [
     "An isogram is a word or phrase without a repeating letter."
   ],
-  "version": "1.7.0",
+  "version": "1.8.0",
   "cases": [
     {
       "description": "Check if the given string is an isogram",
@@ -112,6 +112,14 @@
           "property": "isIsogram",
           "input": {
             "phrase": "angola"
+          },
+          "expected": false
+        },
+        {
+          "description": "a non letter before a repeating pattern",
+          "property": "isIsogram",
+          "input": {
+            "phrase": "red bat cat"
           },
           "expected": false
         }


exercises/isogram/canonical-data.json
```diff
@@ -3,7 +3,7 @@
   "comments": [
     "An isogram is a word or phrase without a repeating letter."
   ],
-  "version": "1.7.0",
+  "version": "1.7.1",
```
SleeplessByte: I think this should be 1.8.0 as it's adding a new test case.

rpottsoh:
Are using break/continue idiomatic ways to solve this exercise for Go?  If it isn't then I would think a mentor would point that out to a student.  It is possible, in other languages, to solve this problem without using continue/break which probably makes this proposed additional exercise unnecessary.  

Don't interpret my comment as being for/against adding this case.  I am just trying to spark some discussion.

da-edra:
Hello @rpottsoh ,
Yes, using a `break/continue` is an idiomatic way to solve it as it avoid's nesting an `if` inside another `if` and makes the code easier to read, in fact, it's what a mentor suggested me.

exercises/isogram/canonical-data.json
da-edra: You're right, I've fixed it.

rpottsoh:
This may not matter, but the Go track is lacking this case from 1.7.0 of the canonical data:
```json
        {
          "description": "hypothetical word with duplicated character following hyphen",
          "property": "isIsogram",
          "input": {
            "phrase": "thumbscrew-jappingly"
          },
          "expected": false
        },
```
This case was added in #1390.  This brought the version up from 1.6.0 to 1.7.0.  Go appears to be following 1.6.0.  Can you see if the case I have shown above would accomplish the same goal as the case you are proposing?  If so then updating the test suite in Go may be all that is needed.  The test suite for Go isogram was last updated on November 10, '18, 2 days before the canonical data was updated by #1390.  The issue was known because of https://github.com/exercism/go/pull/1180 and according to https://github.com/exercism/go/pull/1180#issuecomment-438906814 it looks like a "fresh generate (go run .meta/gen.go) followed by a push."

da-edra:
Thank you very much for pointing that out!

I've updated the test cases and now that case is present and it fixes the problem, I'll point the PR with the updated test cases to [the Go repository](https://github.com/exercism/go) and close this one.

Issue #1496, opened 2019-04-02
Meta: adding new keys to the schema


SleeplessByte:
I've changed the title to be a meta discussion, so that we don't have the same discussion in multiple issues:

- [optional flags](https://github.com/exercism/problem-specifications/issues/1492) #1492
- [immutability flags](https://github.com/exercism/problem-specifications/issues/1493#issuecomment-478563670) #1493
- [other types of tests](https://github.com/exercism/problem-specifications/issues/1225#issue-314244907) #1225 
- multiline "type"

It seems there really is a lot of ask for new fields and there also seems to be a notion that we don't want to break the current generators.

Personally, I am currently parsing the comments for special behaviour and this means that my generators will all break when someone decides the comments should be a special format. I can derive no special meaning from the english in these comments. I don't think this is the right place to hold special cases, but it is for things not covered in the current scheme.

I know that you (@pgaspar @ErikSchierboom @coriolinus @sshine @SaschaMann and @NobbZ) are all involved with writing and maintaining the generators, or at least the discussions on GitHub:

- I think it would be wise to discuss first if we'll add new keys, if at all, because so far each issue has already raised the "yeah okay but we don't _really_ need it and it costs me work"-type comment
- ...and then which ones are candidates, so we can merge it at once and the burden on maintainers is much lower, as the raised point just mentioned _is valid_.
-----------

Original issue about multiline.

There is canonical data such as [`food-chain`](https://github.com/exercism/problem-specifications/blob/master/exercises/food-chain/canonical-data.json#L4) and [`twelve-days`](https://github.com/exercism/problem-specifications/pull/954/files#diff-dcda8ef37a4982f40c53174662e64ebdR5) that correctly comment:

> JSON doesn't allow for multi-line strings, so all verses are presented 
> here as arrays of strings. It's up to the test generator to join the
> lines together with line breaks.

Okay, that's great! But in this case I think it would be _great_ if we'd add a meta property (and keep a list of those, as there are more like these (such as a property preferably being a constant or at least immutable) so that we can automate generation of these exercises.

SleeplessByte:
This ties into my comment [here](https://github.com/exercism/problem-specifications/issues/1225#issuecomment-479002135):

> > Adding a new field breaks those generators.
>
> Yeah I don't think this is a good reason to _not_ add new fields. I actually think it's one of the worst reasons. In this case, we can reduce the workload for _all tracks_ significantly by adding new properties, and perhaps making the schema versioned. 
> 
[...]

Issue #1492, opened 2019-03-29
Proposal: Add flag to mark test cases as optional to canonical schema


SaschaMann:
**Proposal**: Add a way to mark test cases as optional in the canonical data.
For example with a flag `"optional": true` for optional cases, `null` or no flag for cases that (almost) every track should implement. Perhaps with an additional comment or reference to an issue/PR that explains when to implement it and when not to.

This allows automated test generators to handle these cases and leaves it up to the maintainers of the track if they want to add them or not.

**Motivation**: There are test cases (e.g. #1487 or [complex-numbers#L291-L300](https://github.com/exercism/problem-specifications/blob/master/exercises/complex-numbers/canonical-data.json#L291-L300)) that some tracks want to implement, but other tracks don't, for example because it would unnecessarily increase the complexity of the exercise. While every track can implement additional tests manually, once a certain amount of tracks would use such a case, it may be useful to add them to the canonical data. (full [context](https://exercism-team.slack.com/archives/CFWU94MJ6/p1553848677002800) that lead to this issue being created)

petertseng:
To help reviewers evaluate and decide on this proposal, I contribute three additional motivations as potential factors.

**Floating-point tests in triangle**:

https://github.com/exercism/problem-specifications/blob/bac9381d071caf437b68c1b4cd08ad9841b9ecf7/exercises/triangle/canonical-data.json#L186-L197

**Performance tests**, such as pythagorean-triplet (https://github.com/exercism/problem-specifications/issues/1347) and alphametics (https://github.com/exercism/problem-specifications/pull/1024)

**Invalid inputs**: recall that we've had discussions in https://github.com/exercism/problem-specifications/issues/902 . Currently, unless they are essential to the exercise, we will not put them in canonical-data.json , but acknowledge that tracks may choose to add their own. There may be situations where multiple tracks would choose to add such cases, in which cases it could be valuable to share them, and this could be a mechanism for doing so.

I'm interested to see how this proposal goes.

pgaspar:
A few questions:

1. This would require a change on the automatic test generators on every track, right? Because otherwise optional tests will be generated as usual.
2. What happens when an exercise has multiple tests that are optional for very different reasons? Is using "optional" as a category too broad? (Would we want to maintain a set of narrower categories?)

---

[...]

Issue #1493, opened 2019-03-30
High-Scores: add immutability test


rpottsoh:
I have a student that has written method `personalTopThree` and allowed it to mutate the original input data.  There are presently no tests to confirm that `latest` will return the same result before and after `personalTopThree` is called.  

This is the case I am proposing:
```json
    {
      "description": "latest not affected by personalTopThree",
      "property": "latest",
      "input": {
        "scores": [10, 30, 90, 30, 100, 20, 10, 0, 30, 40, 40, 70, 70, 25]
      },
      "expected": 25
    }
```

Is this a reasonable behavior to expect of `personalTopThree`?

NobbZ:
As an Erlang maintainer and elixir mentor (who didn't did this exercise yet) what is mutation? ;)

petertseng:
It is true that unintended or unexpected mutation is a cause of many bugs.

You might find it useful to compare notes with the submitter of https://github.com/exercism/problem-specifications/pull/1486. That pull request asks a similar question of "did the student implementation keep some sort of state, mutate this state when answering a query, and thus set itself up to give an incorrect answer to a subsequent query". The way I see it, you two want similar things and can operate under the principle of "two heads are better than one".

If I may ask a question, it seems that during the course of the test, you intend for *both* `personalTopThree` and `latest` messages to be sent? Do I have that right?
(I wouldn't mind you sketching out how you expect the test to be represented in an Exercism language of your choice or in pseudocode to illustrate)

If so, there will be the interesting question of how to represent in the JSON that this is what you intend to happen; currently that is done in the description I see.

NobbZ:
This seems to be very implementation specific.

This test assumes, that we have an object that is created from some initial input.

As someone who does about 75% FP style programming only, I'm not used (anymore) to that kind of style.

Instead I'm used to just have functions, which gets passed the same input any time and therefore aren't subject to mutation.

Also in the canonical data, I do only see a call to `latest`, but not `topThree`, so assuming isolated tests, that create a new object under test each time, I'm unsure how this will change anything.

rpottsoh:
@NobbZ :

> This seems to be very implementation specific.

Indeed it could very well be.

> This test assumes, that we have an object that is created from some initial input.

This would be a more typical way to go about a solution in an OOP language.  Certainly a more FP approach could be taken as well, which would make this issue moot I believe.

@petertseng :
[...]

PR #1518, opened 2019-05-10
[RFC] Add 'optional' key to canonical-schema

SaschaMann:
# WAIT FOR #1496 BEFORE MERGING

This adds the [most-liked](https://github.com/exercism/problem-specifications/issues/1492#issuecomment-481125708) implementation of an optional key. See #1492, #1496 for more info.

Ideally this should be a draft PR, but I wanted CI to run to spot potential issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ is easier to understand with an example:
 
 ```json
 { "exercise": "foobar"
-, "version" : "1.0.0"
+, "version" : "1.1.0"
 , "comments":
     [ " Comments are always optional and can be used almost anywhere.      "
     , "                                                                    "
@@ -92,6 +92,15 @@ is easier to understand with an example:
               }
             , "expected"   : null
             }
+            { "description": "Foo'ing a very big number returns nothing"
+            , "optional"   : "big-ints"
+            , "comments"   : [ "Making this test case pass requires using BigInts." ]
+            , "property"   : "foo"
+            , "input"      : {
+                "word"       : "28948022309329048855892746252171976962977213799489202546401021394546514198529"
+              }
+            , "expected"   : null
+            }
           , { "description": "Bar'ing a name with numbers gives an error"
             , "property"   : "bar"
             , "input"      : {
@@ -118,6 +127,7 @@ There are also some conventions that must be followed:
   - If an error is expected (because the input is invalid, or any other reason), the value at `"expected"` should be an object containing exactly one property, `"error"`, whose value is a string.
     - The string should explain why the error would occur.
     - A particular track's implementation of the exercise **need not** necessarily check that the error includes that exact string as the cause, depending on what is idiomatic in the language (it may not be idiomatic to check strings for error messages).
+  - Test cases that only some tracks should implement, for example because it would unnecessarily increase the complexity of the exercise in some but not all languages, mark it with an `optional`-key. Multiple cases related to the same reason for optionality should have the same key. The decision that a test case is optional will often be made in the PR discussion, so don't worry about it too much while creating a PR.
 
 ### Test Data Versioning
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,4 +7,5 @@ cache: yarn
 script:
   - bin/check_required_files_present
   - sh bin/check_versions
+  - bin/check_optional
   - yarn test
--- a/OPTIONAL-KEYS.txt
+++ b/OPTIONAL-KEYS.txt
@@ -1,0 +1,3 @@
+floating-point
+big-integer
+unicode
--- a/bin/check_optional
+++ b/bin/check_optional
@@ -1,0 +1,39 @@
+#!/bin/sh
+
+# This script checks that 'optional' fields in canonical-data.json are listed.
+
+optional_keys_file=OPTIONAL-KEYS.txt
+allowed_optional=$(jq -nR '[inputs]' $optional_keys_file)
+
+failed=0
+
+check_optional_for_file() {
+    json_file=$1
+    echo "Checking 'optional' fields in $json_file"
+
+    present_optional=$(jq '[ .cases[] | recurse(.cases[]?) | .optional | select(. != null) ]' $json_file)
+    invalid_optional=$(jq --null-input \
+                          --argjson allowed "$allowed_optional" \
+                          --argjson present "$present_optional" \
+                          --raw-output '$present - $allowed | join("\n")')
+
+    if [ ! -z "$invalid_optional" ]; then
+        echo "Invalid optional fields:"
+        echo "$invalid_optional" | perl -pe 's/^/ - /'
+        echo
+
+        failed=1
+    fi
+}
+
+for json_file in $(git diff --name-only master HEAD | grep '^exercises/.*/canonical-data\.json$'); do
+    check_optional_for_file $json_file
+done
+
+if [ $failed -gt 0 ]; then
+    echo "Allowed optional fields (see $optional_keys_file) are:"
+    cat $optional_keys_file | perl -pe 's/^/ - /'
+    exit 1
+fi
+
+exit 0
--- a/canonical-schema.json
+++ b/canonical-schema.json
@@ -27,7 +27,7 @@
    "self": { "vendor" : "io.exercism"
            , "name"   : "canonical-data"
            , "format" : "jsonschema"
-           , "version": "1-0-0"
+           , "version": "1-1-0"
            },
 
    "$ref": "#/definitions/canonicalData",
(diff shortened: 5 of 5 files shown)


README.md
```diff
@@ -92,6 +92,15 @@ is easier to understand with an example:
               }
             , "expected"   : null
             }
+            { "description": "Foo'ing a very big number returns nothing"
+            , "optional"   : "big-ints"
+            , "comments"   : [ "Making this test case pass requires using BigInts." ]
```
SaschaMann: The example in #1492 explains in the `comments` that "Your track may choose to skip this test". That applies to all cases with an `optional`-key, so it doesn't need to be added to every case, imo. The comment outlines **why** the test is optional.

README.md
```diff
@@ -118,6 +127,7 @@ There are also some conventions that must be followed:
   - If an error is expected (because the input is invalid, or any other reason), the value at `"expected"` should be an object containing exactly one property, `"error"`, whose value is a string.
     - The string should explain why the error would occur.
     - A particular track's implementation of the exercise **need not** necessarily check that the error includes that exact string as the cause, depending on what is idiomatic in the language (it may not be idiomatic to check strings for error messages).
+  - Test cases that only some tracks should implement, for example because it would unnecessarily increase the complexity of the exercise in some but not all languages, mark it with an `optional`-key. Multiple cases related to the same reason for optionality should have the same key. The decision that a test case is optional will often be made in the PR discussion, so don't worry about it too much while creating a PR.
```
SaschaMann: > The decision that a test case is optional will often be made in the PR discussion, so don't worry about it too much while creating a PR.

This isn't really a convention that must be followed, but I've added it to not discourage contributors since the decision if a case only works on certain tracks isn't easy and requires input from multiple maintainers. However, I don't mind dropping it either, if it doesn't fit this list.

ErikSchierboom:
@SaschaMann I think the WIP marker can be removed so we can merge this.

rpottsoh:
#1496 is still open.  Opening comment to this PR says to wait for #1496 before merging.  I assume that means this PR shouldn't be merged until that issue is closed.

sshine:
A practical note:

We should find a place to keep track of the values we use, since it would be a pity to have `int`, `ints`, `integers`, `float`, `floats`, `floating-points`, `bigints`, `big-int`, `big-ints`, `big-integer`, `big-integers`, and so on. Having the schema flexible is probably nice, but we want to document the keys being used.

Perhaps even make a CI test that fails if one is used but isn't listed in this place.

SleeplessByte:
I refer a similar solution to https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt, but then for these keys. It's easy to build CI around that.

sshine:
> WAIT FOR #1496 BEFORE MERGING

#1496 is a meta discussion mentioning 5 things:

- optional flags, which this PR addresses.
- immutability flags (#1493) which had stalled. I've proposed that we close it for reasons.
- other types of tests (#1225) which is a closed issue that is mentioned for inspiration.
- identity field (#1473) which is still in an early stage. I've proposed that we either grow on the idea or abandon it, but either way do not make this additional feature block the optional flags feature.
- multi-line strings (#1496) which has not reached a proposal yet.

I propose that we either:

1) Abandon #1493 (feel free to disagree, of course)
2) Wait with merging #1473 (early stage) and #1496 (early stage)
3) Go ahead and merge this PR and start adding optional keys.

Or:

1) Abandon #1493 (feel free to disagree, of course)
2) Set a time limit (e.g. 4-8 weeks) for growing the ideas of #1473 and #1496.
3) We wait with merging this PR and merge as many proposals as have reached accepted proposals in close succession.

As @SleeplessByte [says in #1496](https://github.com/exercism/problem-specifications/issues/1496#issuecomment-505883509):

> As long as we merge them in close succession (field proposals), we reduce the burden on generator maintainers.

Forgive my ignorance, I do not understand the cost of being a test generator maintainer.

If time is what we need, open source moves very slowly anyways.

But it creates an organizational problem that we put "Depends on #1496" on a PR, when #1496 branches five-ways into directions with hours worth of reading before you know if the particular proposals have stalled because of personal indifferences, design problems, design impossibilities, lack of activity, or something else.

So now I've went through all those and posted proposals to get the current PR's feature, because that's something I find to be missing on a weekly basis.

SaschaMann:
When I opened this PR and added

> WAIT FOR #1496 BEFORE MERGING

there was still an on-going discussion whether or not we should add any keys. Since then we found out that an extra key doesn't break anything, so I'm in favour of merging them independantly from each other. Maintainers who prefer handling all the changes at once can wait for the other issues to be resolved.

> multi-line strings (#1496) which has not reached a proposal yet.

There's a proposal in #1496 but no PR. 

---

> It's easy to build CI around that.

@SleeplessByte In that case, could you add it to this PR? :)

SleeplessByte:
> there was still an on-going discussion whether or not we should add any keys. Since then we found out that an extra key doesn't break anything, so I'm in favour of merging them independantly from each other. Maintainers who prefer handling all the changes at once can wait for the other issues to be resolved.

Yeah. This is my only concern.

I like your proposals @sshine!

bencoman:
Even if its useful to group addition of the other keys together, its also useful to concentrate on a single case for the first one.   

As the initiator of identity field (#1473), I'm having some progress with the alternative of slimming down the descriptions, but although the feedback is worthwhile its slow to work through.  So I'm not yet ready to let go of #1473, but it may end up not needed. So I don't think #1473 should hold up this PR.

> But it creates an organizational problem that we put "Depends on #1496" 

If #1496 was code in a PR, you'd wait for it to be closed.  As a discussion, I'd only expect a comment on #1496 that this PR will proceed as the first test case.

sshine:
In that case, can we focus on preventing divergence in values?

I do find it valuable that the schema does not need to be updated every time we add another category of optional keys. We could, for example, create CI hook that gathers all existing optional keys like so:

    jq '[ .cases[] | .optional | select(. != null) ]' exercises/rna-transcription/canonical-data.json
    jq '[ .cases[] | .cases[] | .optional | select(. != null) ]' exercises/robot-simulator/canonical-data.json

and fails unless "new optional key: <that>" is mentioned in the commit message?

Then we avoid typos, and we're forced to recognize when new optional keys are introduced.

ErikSchierboom:
I like the proposals made by @sshine. My preference would be to use option A: merge this PR, abandon #1493, and personally I would also abandon #1473.

SaschaMann:
> and fails unless "new optional key: " is mentioned in the commit message?

I prefer having a text file with all optional keys. If a new key is in it, CI passes. Otherwise it fails. This makes it easier to look up which keys exist already. This doesn't require an update to the schema, it just requires a file like TOPICS.txt. I have no experience with `jq` or other tools that deal with JSON in bash, so I don't know if this is easily possible, though.

sshine:
@SleeplessByte wrote:
> I prefer a similar solution to TOPICS.txt

@SaschaMann wrote:
> I'd prefer having a text file with all optional keys.

Then how about creating the file `optional-keys.json` which contains a list of strings?

```json
[
  "floating-point",
  "big-integer",
  "unicode"
]
```

Or alternatively a self-documenting file:

```json
{
  "description": "This file contains values for the 'optional' field in tests.",
  "optional": [
    "floating-point",
    "big-integer",
    "unicode"
  ]
}
```

sshine:
I'll try to provide a CI script shortly.

SaschaMann:
What's the benefit of a JSON file with just an array over a plain text file like TOPICS.txt?

sshine:
A common data format. It doesn't matter much to `jq`, though, as it has a `--raw-input` option.

For parsing purposes, TOPICS.txt isn't a good example, because it contains headers.

SaschaMann:
I don't think headers are necessary (for now anyway). 

> It doesn't matter much to jq, though, as it has a --raw-input option.

I'd argue against JSON for this list in that case because the lack of trailing commas will result in two lines shown as changed in the diff. This is a very minor thing obviously, but if there's no other benefit to JSON I feel like the added (tiny) annoyance isn't worth it, imo.

sshine:
@SaschaMann: I believe that since this PR is based on your `master`, I cannot push directly to it.

So I kindly advice you to `cherry-pick` https://github.com/sshine/exercism-problem-specifications/commit/e0194b5074026e65391d6f509c14df8f7e2eadee so that this contribution can become part of the present PR.

SaschaMann:
Strange, it's a branch on this repo, so you should be able to push to it. Anyway, I've cherry-picked the commit. Seems like a solid CI solution, thanks for writing it!

sshine:
Ah, shrug. There's a lot related to the GitHub workflows that I'm still getting used to.

Also, I took the liberty of naming the file `OPTIONAL-KEYS.txt` and pre-defining three optional keys: `floating-point`, `big-integer`, `unicode` in no particular order, since those are the ones I've been picking up on in the recent past. If anyone feels the file should have another name, or `floating-point-number` is more precise, or that `unicode` should be `𝕦𝕟𝕚𝕔𝕠𝕕𝕖` (OK, that's not possible), I have very few strong opinions.

SleeplessByte:
These changes look great @sshine . This is exactly like I wanted/preferred ❤️

PR #1519, opened 2019-05-19
reverse-string: Add a test with an even-sized word

siebenschlaefer:
Currently the tests have only odd sized strings. That allows an approach
to pass that does not work on even sized strings:

```cpp
std::string reverse_string::reverse_string(std::string str)
{
    if (str.size())
        for (auto a = str.begin(), b = str.end() - 1; a != b; ++a, --b)
            std::swap(*a, *b);
    return str;
}
```

The problem is `!=` instead of `<`.

Adding a test with an even sized string catches that invalid approach.

### DIFF
--- a/exercises/reverse-string/canonical-data.json
+++ b/exercises/reverse-string/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "reverse-string",
-  "version": "1.1.0",      
+  "version": "1.2.0",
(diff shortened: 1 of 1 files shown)


SleeplessByte:
That's 3! Let's do this 🚀

SleeplessByte:
Thank you @siebenschlaefer for this addition!

PR #1522, opened 2019-05-25
hamming: Remove empty strand errors

### LINKED ISSUE #1521: hamming: Empty strand errors don't make sense
canonical-data requires an error message stating `left/right strand must not be empty`, however this doesn't make sense due to the following:
1. There is an existing test case with both strands being empty as valid.
2. There is an existing test case for returning an error if the strands differ in length.

The error only applies if your input is already erroneous.

mienaikage:
Closes #1521

### DIFF
--- a/exercises/hamming/canonical-data.json
+++ b/exercises/hamming/canonical-data.json
@@ -1,6 +1,6 @@
 {
 "exercise": "hamming",
-"version": "2.3.0",
+"version": "2.4.0",
   "comments": [
     "Language implementations vary on the issue of unequal length strands.",
     "A language may elect to simplify this task by only presenting equal",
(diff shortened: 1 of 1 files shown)


sshine:
These test cases were added in #1450 after a discussion in exercism/haskell#796 and affects mainly functional programming languages that use pattern matching and recursion. In summary, they do catch errors that the existing ones didn't, and they do make the existing different-length tests somewhat redundant, but as it was also commented in #1450,

> I think the other two cases should remain, even though see seem redundant. They are not redundant to each other.

The other two cases are `"disallow first strand longer"` and `"disallow second strand longer"`.

Perhaps those two should be removed instead, or these two tests can be made optional once #1518 is resolved. Before that, any track is free to not implement them. I understand that with the Perl 5 track, your test generator fetches test cases directly from canonical data. I don't know if you have plans to handle the optional keyword in #1518 in any way.

SleeplessByte:
Since these test cases are not superfluous as per #1450, I will close this PR  but feel free to re-open if this is worth another look. I think redundancy is okay if it's in some languages -- don't even think the optional flag would be necessary here :)

petertseng:
Perhaps the sensible thing to do is to change the written `error`.

The error that "left strand must not be empty" wasn't the full picture. If it were taken to be the full picture it would be false because clearly the empty+empty case shows that left stand *is* allowed to be empty in some cases. So the true description is something like "left strand can only be empty if right strand is also empty" or "left strand must not be be empty if right strand is non-empty", but perhaps that was determined to be too hard to parse the conditionals and associated meaning.

Really these are all cases of "left and right strands must be of equal length". I submit that as my suggestion.

SleeplessByte:
> Really these are all cases of "left and right strands must be of equal length". I submit that as my suggestion.

Right, and the edge cases are left is empty and right is empty, but that's just because we can treat "empty" differently in many languages. So they three tests together are all testing the same thing, but different cases :)

PR #1501, opened 2019-04-05
Proposal: bracket-push: rename to matching-brackets

petertseng:
**Summary**: Here is my official proposal to rename bracket-push, which wants your comments. We have gained experience from the `retree`->`satellite` rename that renames are safe to do. Reviewers of this proposal should carefully consider whether the proposed change warrants the costs that will be incurred by the 31 tracks that have implemented this exercise.

We can consider discussion on what the new name should be, but if nobody has any other options, I defaulted to `matching-brackets`. Another possibility I thought of is `bracket-matching`.

I have another comment coming up with necessary follow-up steps that would need to be taken if this proposal is ratified.

Below is the original commit message, which contains rationale for the change.

---

I rehash my prior allegations of
https://github.com/exercism/problem-specifications/issues/693 below:

I contend that having the word `push` in the name unnecessarily biases
the solution space toward solutions that use a stack data structure (has
a *push* operation) or a *push*-down automaton, rather than other
solutions not using either of these two.

We have heard a principle that we want to name exercises by their story,
not by what they teach:
https://github.com/exercism/problem-specifications/issues/1451#issue-405853947

The story here is about matching brackets, so I posit that that serves
as the name we want.

I have changed the major version number, as was done in the
retree->satellite rename, even though I do not feel strongly about
whether that should be necessary:
https://github.com/exercism/problem-specifications/pull/1478

As we have gained experience in the rename of retree, we see that
problem-specifications is free to rename exercises at any point, without
waiting for all tracks to follow suit.
Of course, this operation should not be performed lightly since it
causes churn in the 31 (*thirty-one*) tracks implementing this exercise.

PR #1530, opened 2019-06-12
[Darts] Slim down generated method names

bencoman:
The purpose of the "description" field is to derive identifiers for naming generated test methods, but some descriptions have tended to verbosity ending up with long identifiers (up to 150 characters).  Such long method names are harder for students to work with and also are awkward for some IDE GUIs.

Issue #1473 proposed adding a separate "identifier" field, but the consensus there was to first try improving "descriptions" to better suit "identifier" generation.    This PR attempts to slim down the generated identifiers without losing substantive information about the test.

For the longest descriptions here, I couldn't recognize the significance of...
```
dart whose coordinates sum to > 1 but whose radius to origin is <= 1
```
While considering a shorter alternative, it helped consistency to also rearranged the shorter descriptions to be more explicit about each test condition, and add a couple of tests each side of boundary conditions.

### DIFF
--- a/exercises/darts/canonical-data.json
+++ b/exercises/darts/canonical-data.json
@@ -16,7 +16,7 @@
     },
     {
       "property": "score",
-      "description": "On outer border",
+      "description": "On outer circle",
       "input": {
         "x": 0,
         "y": 10
@@ -25,7 +25,7 @@
     },
     {
       "property": "score",
-      "description": "On middle-outer border",
+      "description": "On middle circle",
       "input": {
         "x": -5,
         "y": 0
@@ -34,7 +34,7 @@
     },
     {
       "property": "score",
-      "description": "On inner-middle border",
+      "description": "On inner circle",
       "input": {
         "x": 0,
         "y": -1
@@ -52,16 +52,16 @@
     },
     {
       "property": "score",
-      "description": "Just within inner-middle border",
+      "description": "Just within inner circle",
       "input": {
         "x": 0.7,
         "y": 0.7
       },
-      "expected": 1
+      "expected": 10
     },
     {
       "property": "score",
-      "description": "Just beyond inner-middle border",
+      "description": "Just beyond inner circle",
       "input": {
         "x": 0.8,
         "y": -0.8
@@ -70,7 +70,7 @@
     },
     {
       "property": "score",
-      "description": "Just within middle-outer border",
+      "description": "Just within middle circle",
       "input": {
         "x": -3.5,
         "y": 3.5
@@ -79,7 +79,7 @@
     },   
     {
       "property": "score",
-      "description": "Just beyond middle-outer border",
+      "description": "Just beyond middle circle",
       "input": {
         "x": -3.6,
         "y": -3.6
@@ -88,7 +88,7 @@
     },     
     {
       "property": "score",
-      "description": "Just within outer border",
+      "description": "Just within outer circle",
       "input": {
         "x": -7.0,
         "y": 7.0
@@ -97,7 +97,7 @@
     },     
     {
       "property": "score",
-      "description": "Just beyond outer border",
+      "description": "Just beyond outer circle",
       "input": {
         "x": 7.1,
         "y": -7.1
@@ -106,7 +106,7 @@
     },    
     {
       "property": "score",
-      "description": "Asymetric position within inner region",
+      "description": "Asymetric position within inner circle",
       "input": {
         "x": 0.4,
         "y": 0.8
@@ -115,7 +115,7 @@
     },
     {
       "property": "score",
-      "description": "Asymetric position within middle region",
+      "description": "Asymetric position between inner and middle circles",
       "input": {
         "x": -2,
         "y": 4
@@ -124,7 +124,7 @@
     },
     {
       "property": "score",
-      "description": "Asymmetric position within outer region",
+      "description": "Asymmetric position between middle and outer circles",
       "input": {
         "x": 4,
         "y": -8
--- a/exercises/darts/canonical-data.json
+++ b/exercises/darts/canonical-data.json
@@ -70,7 +70,7 @@
     },
     {
       "property": "score",
-      "description": "Middle-Outer border, dart just inside",
+      "description": "Middle-outer border, dart just inside",
       "input": {
         "x": -3.5,
         "y": 3.5
--- a/exercises/darts/canonical-data.json
+++ b/exercises/darts/canonical-data.json
@@ -2,12 +2,12 @@
   "exercise": "darts",
   "version": "2.0.0",
   "comments": [
-    "Return the correct amount earned by a dart landing in a given point in the target problem"
+    "Return the correct amount earned by a dart landing in a given point in the target."
   ],
   "cases": [
     {
       "property": "score",
-      "description": "A dart lands outside the target",
+      "description": "Dart missed target",
       "input": {
         "x": -9,
         "y": 9
@@ -16,7 +16,7 @@
     },
     {
       "property": "score",
-      "description": "A dart lands just in the border of the target",
+      "description": "Outer region, dart on border",
       "input": {
         "x": 0,
         "y": 10
@@ -25,34 +25,16 @@
     },
     {
       "property": "score",
-      "description": "A dart lands in the outer circle",
+      "description": "Middle region, dart on border",
       "input": {
-        "x": 4,
-        "y": 4
-      },
-      "expected": 1
-    },
-    {
-      "property": "score",
-      "description": "A dart lands right in the border between outer and middle circles",
-      "input": {
-        "x": 5,
+        "x": -5,
         "y": 0
       },
       "expected": 5
     },
     {
       "property": "score",
-      "description": "A dart lands in the middle circle",
-      "input": {
-        "x": 0.8,
-        "y": -0.8
-      },
-      "expected": 5
-    },
-    {
-      "property": "score",
-      "description": "A dart lands right in the border between middle and inner circles",
+      "description": "Inner region, dart on border",
       "input": {
         "x": 0,
         "y": -1
@@ -61,7 +43,7 @@
     },
     {
       "property": "score",
-      "description": "A dart lands in the inner circle",
+      "description": "Inner region, dart far inside",
       "input": {
         "x": -0.1,
         "y": -0.1
@@ -70,7 +52,61 @@
     },
     {
       "property": "score",
-      "description": "A dart whose coordinates sum to > 1 but whose radius to origin is <= 1 is scored in the inner circle",
+      "description": "Inner-Middle border, dart just inside",
+      "input": {
+        "x": 0.7,
+        "y": 0.7
+      },
+      "expected": 1
+    },
+    {
+      "property": "score",
+      "description": "Inner-Middle border, dart just outside",
+      "input": {
+        "x": 0.8,
+        "y": -0.8
+      },
+      "expected": 5
+    },
+    {
+      "property": "score",
+      "description": "Middle-Outer border, dart just inside",
+      "input": {
+        "x": -3.5,
+        "y": 3.5
+      },
+      "expected": 5
+    },   
+    {
+      "property": "score",
+      "description": "Middle-Outer border, dart just outside",
+      "input": {
+        "x": -3.6,
+        "y": -3.6
+      },
+      "expected": 1
+    },     
+    {
+      "property": "score",
+      "description": "Outer border, dart just inside",
+      "input": {
+        "x": -7.0,
+        "y": 7.0
+      },
+      "expected": 1
+    },     
+    {
+      "property": "score",
+      "description": "Outer border, dart just outside",
+      "input": {
+        "x": 7.1,
+        "y": -7.1
+      },
+      "expected": 0
+    },    
+    {
+      "property": "score",
+      "description": "Inner region dart, asymmetric position",
       "input": {
         "x": 0.4,
         "y": 0.8
@@ -79,19 +115,19 @@
     },
     {
       "property": "score",
-      "description": "A dart whose coordinates sum to > 5 but whose radius to origin is <= 5 is scored in the middle circle",
+      "description": "Middle region dart, asymmetric position",
       "input": {
-        "x": 2,
+        "x": -2,
         "y": 4
       },
       "expected": 5
     },
     {
       "property": "score",
-      "description": "A dart whose coordinates sum to > 10 but whose radius to origin is <= 10 is scored in the outer circle",
+      "description": "Outer region dart, asymmetric position",
       "input": {
         "x": 4,
-        "y": 8
+        "y": -8
       },
       "expected": 1
     }
--- a/exercises/darts/canonical-data.json
+++ b/exercises/darts/canonical-data.json
@@ -79,7 +79,7 @@
     },   
     {
       "property": "score",
-      "description": "Middle-Outer border, dart just outside",
+      "description": "Middle-outer border, dart just outside",
       "input": {
         "x": -3.6,
         "y": -3.6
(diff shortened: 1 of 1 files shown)


bencoman:
> I'd like to remove the word `dart` from all the descriptions

I agree entirely, just thought others might believe that a step too far. I'll wait for one more review before doing it.

exercises/darts/canonical-data.json
```diff
@@ -1,13 +1,13 @@
 {
   "exercise": "darts",
-  "version": "2.0.0",
+  "version": "2.1.0",
   "comments": [
-    "Return the correct amount earned by a dart landing in a given point in the target problem"
+    "Return the correct amount earned by a dart landing in a given point in the target."
```
ErikSchierboom: I thought maybe we could use darts-related terminology here?

```suggestion
    "Return the correct amount earned by a dart landing in a given point on the board."
```

exercises/darts/canonical-data.json
```diff
@@ -1,13 +1,13 @@
...
   ],
   "cases": [
     {
       "property": "score",
-      "description": "A dart lands outside the target",
+      "description": "Dart missed target",
```
ErikSchierboom:
```suggestion
      "description": "Dart missed board",
```

exercises/darts/canonical-data.json
```diff
@@ -25,25 +25,43 @@
...
+      "description": "Inner region, dart far inside",
+      "input": {
+        "x": -0.1,
+        "y": -0.1
+      },
+      "expected": 10
+    },
+    {
+      "property": "score",
+      "description": "Inner-Middle border, dart just inside",
```
ErikSchierboom:
```suggestion
      "description": "Inner-middle border, dart just inside",
```

exercises/darts/canonical-data.json
```diff
@@ -25,25 +25,43 @@
...
+      "input": {
+        "x": 0.7,
+        "y": 0.7
+      },
+      "expected": 1
     },
     {
       "property": "score",
-      "description": "A dart lands in the middle circle",
+      "description": "Inner-Middle border, dart just outside",
```
ErikSchierboom:
```suggestion
      "description": "Inner-middle border, dart just outside",
```

exercises/darts/canonical-data.json
```diff
@@ -52,25 +70,43 @@
     },
     {
       "property": "score",
-      "description": "A dart lands right in the border between middle and inner circles",
+      "description": "Middle-Outer border, dart just inside",
```
ErikSchierboom:
```suggestion
      "description": "Middle-outer border, dart just inside",
```

exercises/darts/canonical-data.json
```diff
@@ -52,25 +70,43 @@
...
+        "y": 3.5
       },
-      "expected": 10
-    },
+      "expected": 5
+    },   
     {
       "property": "score",
-      "description": "A dart lands in the inner circle",
+      "description": "Middle-Outer border, dart just outside",
```
ErikSchierboom:
```suggestion
      "description": "Middle-outer border, dart just outside",
```

exercises/darts/canonical-data.json
bencoman: Not necessarily against this (it occurred to me also), but first can you consider that its [description.md](https://github.com/exercism/problem-specifications/blob/master/exercises/darts/description.md) doesn't contain the word "board" and rather frequent use of "target".

ErikSchierboom: Right. Just ignore my comment, I'll create a separate PR later

bencoman:
I was finding the sub-word repetition of "inside inner-middle" & "outside middle-outer" slightly brain bending. Had an inspiration to substitute "within" & "beyond".  Not strongly tied to it, just sharing for your thoughts.

P.S. I just noticed that for integrating suggestions the "Files changed" tab has an extra button **"Add to batch"** that is missing from the "Conversations" tab. So I'll use that next time.

rpottsoh:
See if you can homogenize the descriptions.  What I mean is that some refer to a region and others a border.  Is kind of confusing.  Sometime there is hyphenation other times not.  However you decide best to "unbend your brain" oO works for me. 😄

exercises/darts/canonical-data.json
```diff
@@ -25,52 +25,88 @@
...
+      "description": "Just beyond outer circle",
+      "input": {
+        "x": 7.1,
+        "y": -7.1
+      },
+      "expected": 0
+    },    
+    {
+      "property": "score",
+      "description": "Asymetric position within inner circle",
```
ErikSchierboom:
```suggestion
      "description": "Asymmetric position within inner circle",
```

exercises/darts/canonical-data.json
```diff
@@ -79,19 +115,19 @@
     },
     {
       "property": "score",
-      "description": "A dart whose coordinates sum to > 5 but whose radius to origin is <= 5 is scored in the middle circle",
+      "description": "Asymetric position between inner and middle circles",
```
ErikSchierboom:
```suggestion
      "description": "Asymmetric position between inner and middle circles",
```

exercises/darts/canonical-data.json
ErikSchierboom: The "dart" bit has been removed from all the other cases, maybe also remove it here?

```suggestion
      "description": "Missed target",
```

bencoman:
Removed last commit. I had bumped the major version on the wrong PR.

exercises/darts/canonical-data.json
```diff
@@ -25,52 +25,88 @@
...
+      "description": "Just beyond outer circle",
+      "input": {
+        "x": 7.1,
+        "y": -7.1
+      },
+      "expected": 0
+    },    
+    {
+      "property": "score",
+      "description": "Asymmetric position within inner circle",
```
petertseng: that is interesting. I see that asymmetry (not being on the lines `x = y` or `x = -y`) is coincidentally a trait of the cases added in https://github.com/exercism/problem-specifications/pull/1488, but I don't think it is necessary for the thing being tested, with the following example: x = 3.5, y = 3.5

* is on `x = y`
* sums to > 5
* radius is <= 5

Therefore, asymmetry is not the actual thing being tested in these test cases, and writing their description as asymmetric loses information as to why this test case is here. Some possible choices I see here are:

* Decide that losing this information from the description is okay, with the justification being that it is possible to obtain it by reviewing commit history.
* Find some way to write this in the description that is still acceptable. Since I am raising this point it behooves me to give some suggestions... unfortunately some of my suggestions require some form of jargon
    * Within inner circle by Euclidean distance, not Manhattan
    * Within inner circle by Euclidean distance, not taxicab
    * Within inner circle by radius, not addition
    * Within inner circle but not its inscribed square
* Maybe this is a situation where nesting these three cases is appropriate, under a `cases` whose description describes what's in common between these three test cases.

bencoman: I liked all four suggestions. My preference would be the third as the least exotic, closely followed by the first as well defined terms.  

However, thanks @petertseng for the background on the purpose of those #1488 tests.  The existing description **"A dart whose coordinates sum to > 1 but whose radius to origin is <= 1 is scored in the inner circle"** didn't impart to me that it was to prevent **"an erroneous-but-passing solution where the sum of the absolute values of the coordinates of the dart are compared to the radius thresholds"**. This is a good example where the "what" didn't impart the "why" to me, and your third suggestion better describes "why".

Now considering the three test pairs I arranged [near either side of each boundary, e.g...](https://github.com/exercism/problem-specifications/blob/2f71f0332fe0f96008842fc00158b74d3cc333eb/exercises/darts/canonical-data.json#L72-L86)
```
    {
      "property": "score",
      "description": "Just within middle circle",
      "input": {
        "x": -3.5,
        "y": 3.5
      },
      "expected": 5
    },   
    {
      "property": "score",
      "description": "Just beyond middle circle",
      "input": {
        "x": -3.6,
        "y": -3.6
      },
      "expected": 1
    },     
```
simple addition won't be able to satisfy all three pairs.   

So it seems I inadvertently made those #1488 tests redundant.  
They might either be removed or re-purposed for asymmetry as I've done.

petertseng: I checked and you are right that the "just within" tests all will detect an implementation that tests absolute addition:

```
== anti-verify abs score ==
5. Just within inner circle: Got 5 instead of 10
7. Just within middle circle: Got 1 instead of 5
9. Just within outer circle: Got 0 instead of 1
11. Asymmetric position within inner circle: Got 5 instead of 10
12. Asymmetric position between inner and middle circles: Got 1 instead of 5
13. Asymmetric position between middle and outer circles: Got 0 instead of 1
```

So you are correct that these asymmetric points are all redundant.

I would say the answer for "should these tests be kept?" would depend on whether it is useful (would find any mistaken implementations) to have any points that aren't on `x = y` and `x = -y`. I wasn't sure if there are any, so I can be convinced either way. Though note that there are already some earlier test cases that show that.

I think it perhaps makes the most sense to remove these three tests then? Should the other three tests' descriptions be changed from "just within" to something else resembling the above suggestions? Or is "just within" enough information to figure it out?

I think I'm good trusting your judgment either way at this point

bencoman: I can't imagine anyone really doing it, but theoretically without an asymmetric test students might  
```
get away with... sqrt(2*x*x) 
rather than... sqrt(x*x + y*y) 
```
But one asymmetric test should do. Culled the other two.

bencoman:
So btw, we are now at commit 11.  Presuming some approval ticks aren't far away, when they arrive should I first squash commits in my branch before its merged into master?

rpottsoh:
This should be squashed as you said or by whomever does the merge.

SleeplessByte:
image

@bencoman this can just be squashed _on_ merge (like @rpottsoh) says. No need to do anything :)

ee7:
Yes, these commits should be squashed.

Reasons:
- Correctness. The commits are not standalone, and cannot be independently reverted. This can also interfere with `git bisect`.
- Brevity: it is easier for track maintainers to understand (and also eliminates an arguably needless merge commit).
- Usefulness: it automatically includes the PR number in the commit message, which makes `git log` more searchable.

See recent discussion at https://github.com/exercism/problem-specifications/pull/1501#issuecomment-482646134.

Here is some recent repo history from `git log --oneline`.

```
Merge pull request #1534 from exercism/bencoman-anagram-grammar
anagram: grammar
Merge pull request #1526 from exercism/bencoman-patch-1
(origin/bencoman-patch-1) also fix surplus square bracket
fix missing comma
and bump version
Correctly add to existing comment
Robot simulator descriptions too long (#1525)
Improve anagram problem statement (#1529)
expand description, response to @sleeplessbyte suggestion
ETL descriptions too long
```

Note that it's even worse because there are `robot-simulator` and `anagram` changes in the middle of `etl` commits that come from a single PR.

The above is much harder to understand than the following (which represents the same changes):

```
anagram: Fix grammar (#1534)
robot-simulator: Shorten descriptions (#1525)
anagram: Improve problem statement (#1529)
etl: Shorten descriptions (#1526)
```

Of course, `git log` has useful flags like `--no-merges` and `--graph` . But maintainers may browse via GitHub, or not use such flags.

Please also use the standard commit message style.

SleeplessByte:
@ee7 Thank you for this elaboration, and whilst appreciated, to squash or not to squash wasn't under debate. The question was if the OP needed to squash manually before merging. This PR isn't _really_ the place to start a discussion which type of merge is best for this repository. In general, the people who merge often usually squash, per the reasons you're giving (so we're all with you here!).
 
> standard commit message style

If you mean conventional commits: we don't use that in this repository, so it would be very out of place.

exercises/darts/canonical-data.json
ErikSchierboom: @petertseng Does the updated canonical data now match your suggestions?

ErikSchierboom:
@SleeplessByte Small bump asking for your review :)

exercises/darts/canonical-data.json
```diff
@@ -70,30 +52,66 @@
...
-        "y": 4
+        "x": 0.8,
+        "y": -0.8
       },
       "expected": 5
     },
     {
       "property": "score",
-      "description": "A dart whose coordinates sum to > 10 but whose radius to origin is <= 10 is scored in the outer circle",
+      "description": "Just within middle circle",
```
SleeplessByte:
```suggestion
      "description": "Just within the middle circle",
```

Perhaps "On the inner edge of the middle circle"?

exercises/darts/canonical-data.json
```diff
@@ -70,30 +52,66 @@
...
+      "input": {
+        "x": -3.5,
+        "y": 3.5
+      },
+      "expected": 5
+    },   
+    {
+      "property": "score",
+      "description": "Just beyond middle circle",
```
SleeplessByte:
```suggestion
      "description": "Just outside the middle circle",
```

```suggestion
      "description": "Just within the middle circle",
```

Perhaps "On the outer edge of the middle circle"?

exercises/darts/canonical-data.json
```diff
@@ -70,30 +52,66 @@
...
-        "x": 4,
-        "y": 8
+        "x": -3.6,
+        "y": -3.6
       },
       "expected": 1
+    },     
+    {
+      "property": "score",
+      "description": "Just within outer circle",
```
SleeplessByte:
```suggestion
      "description": "Just within the outer circle",
```

exercises/darts/canonical-data.json
```diff
@@ -70,30 +52,66 @@
...
+      "input": {
+        "x": -7.0,
+        "y": 7.0
+      },
+      "expected": 1
+    },     
+    {
+      "property": "score",
+      "description": "Just beyond outer circle",
```
SleeplessByte:
```suggestion
      "description": "Just outside the outer circle",
```

exercises/darts/canonical-data.json
```diff
@@ -70,30 +52,66 @@
...
+      "input": {
+        "x": 7.1,
+        "y": -7.1
+      },
+      "expected": 0
+    },    
+    {
+      "property": "score",
+      "description": "Asymmetric position between inner and middle circles",
```
SleeplessByte:
```suggestion
      "description": "Asymmetric position between the inner and middle circles",
```

exercises/darts/canonical-data.json
petertseng: I feel I have said what I wanted to say and that I do not object to the current situation. I remind others that I think it might still be possible to incorporate "by radius not by addition" (or whatever choice is chosen) if it will actually help a student understand what is wrong with a submitted implementation. if "just outside" (or whatever) is deemed to be sufficient, I accept the collective judgment.

ErikSchierboom:
@bencoman FYI. There is a merge conflict in this PR.

rpottsoh:
Master was merged into the branch.  Fixing.

bencoman:
@rpottsoh What was wrong above?  Clicking the <Resolve conflicts> button I used github's web editor to clean this up...
image

After your force-push the conflict still remains due to recent commit...
https://github.com/exercism/problem-specifications/commit/c5a1b8b75
How should this be handled?
(One thing I did forget was to re-bump the version after resolving the conflict)

rpottsoh:
My fault.  I was thinking that the PR should have been rebased when #1540 was merged.  Resolving the conflict via the GitHub interface appears to bring along the changes in master that have occurred since branch `bencoman-darts-slim-down-generated-method-names` was created and now appear as changes that have been made in this PR.  My intent was to eliminate all that so only the changes to darts/canonical-data.json appeared in the commit history.  I wasn't successful at accomplishing what I had set out to do and ended up with `master` merged into `bencoman-darts-slim-down-generated-method-names` to resolve the conflict.

bencoman:
I think I've cleared everything.   Anything further @petertseng @rpottsoh @SleeplessByte ?

SleeplessByte:
:rocket: great work @bencoman . Thanks @ErikSchierboom and @rpottsoh for the reviews.

PR #1532, opened 2019-06-13
Pangram: slim down test identifiers generated from descriptions

bencoman:
The purpose of the "description" field is to derive identifiers for naming generated test methods, but some descriptions have tended to verbosity ending up with long identifiers (up to 150 characters). Such long method names are harder for students to work with and also are awkward for some IDE GUIs.

Issue #1473 proposed adding a separate "identifier" field, but I was required to first try improving "descriptions" to better suit "identifier" generation. This PR aims to slim down the generated identifiers without losing substantive information about the test.

### DIFF
--- a/exercises/pangram/canonical-data.json
+++ b/exercises/pangram/canonical-data.json
@@ -4,7 +4,7 @@
     "A pangram is a sentence using every letter of the alphabet at least once.",
     "Output should be a boolean denoting if the string is a pangram or not."
   ],
-  "version": "1.4.1",
+  "version": "1.5.0",
   "cases": [
     {
       "description": "empty sentence",
--- a/exercises/pangram/canonical-data.json
+++ b/exercises/pangram/canonical-data.json
@@ -1,97 +1,90 @@
 {
   "exercise": "pangram",
   "comments": [
-    "A pangram is a sentence using every letter of the alphabet at least once."
+    "A pangram is a sentence using every letter of the alphabet at least once.",
+    "Output should be a boolean denoting if the string is a pangram or not."
   ],
   "version": "1.4.1",
   "cases": [
     {
-      "description": "Check if the given string is a pangram",
-      "comments": [
-        "Output should be a boolean denoting if the string is a pangram or not."
-      ],
-      "cases": [
-        {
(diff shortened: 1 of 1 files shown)


bencoman:
Having removed the nested level its too hard to see my other changes.
Going to redo this in a couple of commits to make review easier.

SleeplessByte:
FYI @bencoman , removing nesting is fine, but it's a _major_ semver increase as it's breaking. It would be 2.0.0 then.

bencoman:
@SleeplessByte, Happy to comply with that, but I don't understand. I'm only adding/deleting test cases,
changing inputs & outputs and changing test case ordering, which seems *minor* per...
https://github.com/exercism/problem-specifications/blob/master/README.md#major-version-changes

For my education, which of the *major* points have I crossed and what is breaking?

btw, restarted PR at #1533

SleeplessByte:
Ah, sorry, I thought you were _removing_ the levels (as in flattening).

```
[[a, b], [c,d]]
```

to

```
[a, b, c, d]
```

This does not constitute as removing 4 test cases (at level 1) and adding 4 (at level 0), but IIRC, this is removing 4 test cases and adding 1-4 new properties at level 0.

bencoman:
> I thought you were removing the levels (as in flattening).

yes, that is correct.  The way I was interpreting that semver info was that all Pangram test cases use a single property "isPangram".   That will be a function implemented by the solution such that CI testing will succeed regardless of which level a test case using moves to. Alternatively, the CI would fail if a new property "notPangram" was introduced since the solution would not have that function - thus would be a breaking change.  That is wrong?

SleeplessByte:
I understand your reasoning. I do not know the _correct_ answer, but given that multiple reviewers [suggested it here](https://github.com/exercism/problem-specifications/pull/1510) I think that even though conceptually you're right, generators and tracks might generate tests _based on the level_. 

As you know, not all tracks flatten/concatenate descriptions. For example, in JavaScript/TypeScript, nesting _does_ matter, because we can set-up state that is true for the nesting only.

bencoman:
ok. got it.

PR #1533, opened 2019-06-13
pangram: Slim down descriptions to generate better test identifiers 

bencoman:
The purpose of the "description" field is to derive identifiers for naming generated test methods, but some descriptions have tended to verbosity ending up with long identifiers (up to 150 characters). Such long method names are harder for students to work with and also are awkward for some IDE GUIs.

Issue #1473 proposed adding a separate "identifier" field, but I was required to first try improving "descriptions" to better suit "identifier" generation. This PR aims to slim down the generated identifiers without losing substantive information about the test.

To start with,  "Check if the given string is a pangram" is a rather banal redundant description unworthy of nesting.

### DIFF
--- a/exercises/pangram/canonical-data.json
+++ b/exercises/pangram/canonical-data.json
@@ -7,7 +7,7 @@
   "version": "1.5.0",
   "cases": [
     {
-      "description": "sentence empty",
+      "description": "empty sentence",
       "property": "isPangram",
       "input": {
         "sentence": ""
@@ -15,7 +15,7 @@
       "expected": false
     },
     {
-      "description": "recognizes a perfect lower case pangram",
+      "description": "perfect lower case",
       "property": "isPangram",
       "input": {
         "sentence": "abcdefghijklmnopqrstuvwxyz"
@@ -23,7 +23,7 @@
       "expected": true
     },
     {
-      "description": "pangram with only lower case",
+      "description": "only lower case",
       "property": "isPangram",
       "input": {
         "sentence": "the quick brown fox jumps over the lazy dog"
@@ -39,7 +39,7 @@
       "expected": false
     },
     {
-      "description": "another missing character, e.g. 'h'",
+      "description": "missing character 'h'",
       "property": "isPangram",
       "input": {
         "sentence": "five boxing wizards jump quickly at it"
@@ -47,7 +47,7 @@
       "expected": false
     },
     {
-      "description": "pangram with underscores",
+      "description": "with underscores",
       "property": "isPangram",
       "input": {
         "sentence": "the_quick_brown_fox_jumps_over_the_lazy_dog"
@@ -55,7 +55,7 @@
       "expected": true
     },
     {
-      "description": "pangram with numbers",
+      "description": "with numbers",
       "property": "isPangram",
       "input": {
         "sentence": "the 1 quick brown fox jumps over the 2 lazy dogs"
@@ -71,7 +71,7 @@
       "expected": false
     },
     {
-      "description": "pangram with mixed case and punctuation",
+      "description": "mixed case and punctuation",
       "property": "isPangram",
       "input": {
         "sentence": "\"Five quacking Zephyrs jolt my wax bed.\""
@@ -79,7 +79,7 @@
       "expected": true
     },
     {
-      "description": "upper and lower case versions of the same character should not be counted separately",
+      "description": "case insensitive",
       "property": "isPangram",
       "input": {
         "sentence": "the quick brown fox jumps over with lazy FX"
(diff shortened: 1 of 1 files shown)


bencoman:
I had split this into two commits to make it easier to separate review nesting and description changes.
Just discovered you get a nice overall view if you go...
   Files changed > Settings (gear icon) > Hide whitespace changes

exercises/pangram/canonical-data.json
```diff
@@ -1,97 +1,90 @@
...
+    {
+      "description": "only lower case",
+      "property": "isPangram",
+      "input": {
+        "sentence": "the quick brown fox jumps over the lazy dog"
+      },
+      "expected": true
+    },
+    {
+      "description": "missing character 'x'",
```
rpottsoh:
```suggestion
      "description": "missing an 'x'",
```
or perhaps
```suggestion
      "description": "lacks an 'x'",
```

Similar pattern could be repeated if others are on board.  Also, the whole file essentially appears to have been altered.  Some may find it more difficult to review.

bencoman:
btw, The strategy I use to create slim descriptions is imagine the word "Test" preceding the description (because including it would be redundant)

rpottsoh:
> Files changed > Settings (gear icon) > Hide whitespace changes

Nice find.  TIL something, and it is only the morning.

exercises/pangram/canonical-data.json
bencoman: I think "lacks" might be harder for non-native-english speakers to grok.

bencoman:
@rpottsoh I went to push a similar change to your suggestion into `missing character 'h'  `
but realised that camelCase identifier convention would generate `missing a 'h' `  into `missingAH` 
which is not nice.  I reverted your suggestion so it will generate `missingCharacterH` .
Let me know if you object.

exercises/pangram/canonical-data.json
kotp: Perhaps `"Missing the letter 'x'"`

rpottsoh: @bencoman `missingTheLetterX` or `missingAnX`?  I agree `lacksAnX` is the least desirable choice.

bencoman: In general I think scaffold words like "the" and "an" can often be left out without compromising meaning - but I can slide on it.  The majority of improvement from this PR comes in removing the nesting platitude description "Check if the given string is a pangram" and shrinking the description "upper and lower case versions of the same character should not be counted separately".  Then once I opened the box, other descriptions get some collateral cleaning (like removing redundant 'panagram' from descriptions.

As for length, my gut feel was 40-60 but I thought it would be good to put some science behind it by reviewing the stock Pharo 7 release...
```smalltalk
classes := Object allSubclasses select: [ :cc | cc isKindOf: TestCase class ].
methods := c flatCollect: [ :c | c allMethods  ].
tests := methods select: [ :m | m selector beginsWith: 'test' ].
lengths := tests collect: [ :m | m selector size ].
lengths asBag keysAndValuesDo: [ :len :count | Transcript crShow: len; show: ','; show: count  ]
```
and graphing in Excel...
image

So 40-60 looks like a reasonable target to aim for, but its not an absolute. I arbitrarily picked fixing descriptions longer than 100 chars.  When I get there I'll be able to better evaluate if more is needed. (and it looks like I should review of Pharo's longer ones also)

bencoman:
@kotp, @rpottsoh, I committed your suggestion. Anything further be giving this the tick?

exercises/pangram/canonical-data.json
```diff
@@ -1,97 +1,90 @@
...
-      ]
+      "description": "empty sentence",
+      "property": "isPangram",
+      "input": {
+        "sentence": ""
+      },
+      "expected": false
+    },
+    {
+      "description": "perfect lower case",
```
ErikSchierboom: I was wondering if we should maybe rephrase this test case or add something to the README to describe what a perfect pangram is.

kotp: That sounds like a good idea. (different PR though for the README.  `"perfect pangram in lower case"` might be a better description of the test

kotp: Context being the pangram, which is why I think that the wording ended up being the way it became.

SleeplessByte: I am in favour of adding to the readme as wel.

ErikSchierboom: @kotp @SleeplessByte If we do a separate PR for the readme, is this PR ok to be merged?

kotp: Yes, I think so.

SleeplessByte:
Another one merged @bencoman 🚀

bencoman:
thx all.

PR #1540, opened 2019-06-26
[Darts] Add additional edge case

Tyresius92:
Cover the edge case where a student might only think to check if `distance > 0` rather than `>= 0`, or similar.

### DIFF
--- a/exercises/darts/canonical-data.json
+++ b/exercises/darts/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "darts",
-  "version": "2.0.0",
+  "version": "2.1.0",
   "comments": [
     "Return the correct amount earned by a dart landing in a given point in the target problem"
   ],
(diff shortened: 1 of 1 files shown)


exercises/darts/canonical-data.json
```diff
@@ -77,6 +77,15 @@
       },
       "expected": 10
     },
+    {
+      "property": "score",
+      "description": "A dart lands right in the center of the dart board is scored in the inner circle",
```
ErikSchierboom: I like this test case. Maybe this should actually be the very first test case, as it requires the least amount of code to get right.

Tyresius92: Yeah, I thought about putting it first, but figured that the test case where you're returning 0 might be better? It seemed like the test suite is vaguely ordered to work from the outside inwards.

ErikSchierboom: Ah right. That makes sense.

SleeplessByte:
Thank you @Tyresius92

PR #1544, opened 2019-06-29
[WIP] Add schema key indicating multiline strings

SaschaMann:
This adds a key `multi-line-string-array: boolean` to the schema. If its value is `true`, arrays of strings in the test case should be joined together to a multiline string.

Previous discussion of this happened in #1496: 

Motivation:

> Original issue about multiline.
> 
> There is canonical data such as food-chain and twelve-days that correctly comment:
> 
>     JSON doesn't allow for multi-line strings, so all verses are presented
>     here as arrays of strings. It's up to the test generator to join the
>     lines together with line breaks.
> 
> Okay, that's great! But in this case I think it would be great if we'd add a meta property (and keep a list of those, as there are more like these (such as a property preferably being a constant or at least immutable) so that we can automate generation of these exercises.
> 
> _Originally posted by @SleeplessByte in https://github.com/exercism/problem-specifications/issues/1496#issue-428234215_

> I currently search `comments` in order to determine if something is a multiline string; because I then need to concatenate the values in the array using `\n` -- the same exercise also tests for single line strings. So there are not array + multiline string in one exercise, but I do need special handling.
> 
> If we don't want to add a property, that's fine :)
> 
> _Originally posted by @SleeplessByte in https://github.com/exercism/problem-specifications/issues/1496#issuecomment-492993261_

It affects the following exercises:

> 1. [`transpose`](https://github.com/exercism/problem-specifications/blob/5326cf2b32fd5df77cda3ade20d088d9a928b8b3/exercises/transpose/canonical-data.json#L4)
> 1. [`proverb`](https://github.com/exercism/problem-specifications/blob/5326cf2b32fd5df77cda3ade20d088d9a928b8b3/exercises/proverb/canonical-data.json#L4)
> 1. [`food-chain`](https://github.com/exercism/problem-specifications/blob/5326cf2b32fd5df77cda3ade20d088d9a928b8b3/exercises/food-chain/canonical-data.json#L4)
> 1. [`twelve-days`](https://github.com/exercism/problem-specifications/blob/84aa47979df7f526bdcb777cacc2c2cc9647dcaa/exercises/twelve-days/canonical-data.json#L4)
> 1. [`house`](https://github.com/exercism/problem-specifications/blob/5326cf2b32fd5df77cda3ade20d088d9a928b8b3/exercises/house/canonical-data.json#L4)
> 1. [`grep`](https://github.com/exercism/problem-specifications/blob/4f2efaa7a18fce99e91b01b3b11f13cac795735c/exercises/grep/canonical-data.json#L4)
> 
> And then there is [`beer-song`](https://github.com/exercism/problem-specifications/blob/df9c28dbeb2d0957fb7a32c4d588a697c47196e5/exercises/beer-song/canonical-data.json), which doesn't actually have the comment, but should...
> 
> _Originally posted by @SleeplessByte in https://github.com/exercism/problem-specifications/issues/1496#issuecomment-493167011_

> [`tournament`](https://github.com/exercism/problem-specifications/blob/ee01fe0c110b3c92fc3a52955a86427f18d3c219/exercises/tournament/canonical-data.json) should also be on that list.
> 
> _Originally posted by @cmccandless in https://github.com/exercism/problem-specifications/issues/1496#issuecomment-493478976_

---

Hiding whitespace changes when looking at the diff is helpful, because I had to adjust the indentation of various lines that are otherwise unchanged.

### DIFF
--- a/bin/check_multi-line-string-array_contents
+++ b/bin/check_multi-line-string-array_contents
@@ -10,7 +10,7 @@ check_keys() {
 
     present_keys=$(jq '[ .cases[] | ."multi-line-string-array" | select(. != null)[] ] | unique' $json_file)
 
-    allowed_keys=$(jq '[ .cases[] | recurse(.cases[]?) | .input | keys[] ] | unique + ["expected"]' $json_file)
+    allowed_keys=$(jq '[ .cases[] | recurse(.cases[]?) | .input | select(. != null) | keys[] ] + ["expected"] | unique' $json_file)
 
     invalid_keys=$(jq --null-input \
                       --argjson allowed "$allowed_keys" \
--- a/bin/check_multi-line-string-array_contents
+++ b/bin/check_multi-line-string-array_contents
@@ -18,7 +18,7 @@ check_keys() {
                       --raw-output '$present - $allowed | join("\n")')
 
     if [ ! -z "$invalid_keys" ]; then
-        echo "Invalid multi-line-string-array entries:"
+        echo "The following multi-line-string-array input parameters did not occur in any tests:"
         echo "$invalid_keys" | perl -pe 's/^/ - /'
         echo
 
--- a/bin/check_multi-line-string-array_contents
+++ b/bin/check_multi-line-string-array_contents
@@ -8,7 +8,7 @@ check_keys() {
     json_file=$1
     echo "Checking 'multi-line-string-array' contents in $json_file"
 
-    present_keys=$(jq '[ .cases[] | recurse(.cases[]?) | ."multi-line-string-array" | select(. != null)[] ] | unique ' $json_file)
+    present_keys=$(jq '[ .cases[] | ."multi-line-string-array" | select(. != null)[] ] | unique' $json_file)
 
     allowed_keys=$(jq '[ .cases[] | recurse(.cases[]?) | .input | keys[] ] | unique + ["expected"]' $json_file)
 
--- a/bin/check_multi-line-string-array_contents
+++ b/bin/check_multi-line-string-array_contents
@@ -1,0 +1,38 @@
+#!/bin/sh
+
+# This script checks that multi-line-string-array only contains valid keys.
+
+failed=0
+
+check_keys() {
+    json_file=$1
+    echo "Checking 'multi-line-string-array' contents in $json_file"
+
+    present_keys=$(jq '[ .cases[] | recurse(.cases[]?) | ."multi-line-string-array" | select(. != null)[] ] | unique ' $json_file)
+
+    allowed_keys=$(jq '[ .cases[] | recurse(.cases[]?) | .input | keys[] ] | unique + ["expected"]' $json_file)
+
+    invalid_keys=$(jq --null-input \
+                      --argjson allowed "$allowed_keys" \
+                      --argjson present "$present_keys" \
+                      --raw-output '$present - $allowed | join("\n")')
+
+    if [ ! -z "$invalid_keys" ]; then
+        echo "Invalid multi-line-string-array entries:"
+        echo "$invalid_keys" | perl -pe 's/^/ - /'
+        echo
+
+        failed=1
+    fi
+}
+
+for json_file in $(git diff --name-only master HEAD | grep '^exercises/.*/canonical-data\.json$'); do
+    check_keys $json_file
+done
+
+if [ $failed -gt 0 ]; then
+    echo "Only input keys are valid entries of multi-line-string-array"
+    exit 1
+fi
+
+exit 0
--- a/canonical-schema.json
+++ b/canonical-schema.json
@@ -27,7 +27,7 @@
    "self": { "vendor" : "io.exercism"
            , "name"   : "canonical-data"
            , "format" : "jsonschema"
-           , "version": "1-1-0"
+           , "version": "1-2-0"
            },
 
    "$ref": "#/definitions/canonicalData",
@@ -85,12 +85,13 @@
           , "type"       : "object"
           , "required"   : ["description", "property", "input", "expected"]
           , "properties" :
-                { "description": { "$ref": "#/definitions/description" }
-                , "optional"   : { "$ref": "#/definitions/optional"    }
-                , "comments"   : { "$ref": "#/definitions/comments"    }
-                , "property"   : { "$ref": "#/definitions/property"    }
-                , "input"      : { "$ref": "#/definitions/input"       }
-                , "expected"   : { "$ref": "#/definitions/expected"    }
+                { "description"             : { "$ref": "#/definitions/description"             }
+                , "optional"                : { "$ref": "#/definitions/optional"                }
+                , "multi-line-string-array" : { "$ref": "#/definitions/multi-line-string-array" }
+                , "comments"                : { "$ref": "#/definitions/comments"                }
+                , "property"                : { "$ref": "#/definitions/property"                }
+                , "input"                   : { "$ref": "#/definitions/input"                   }
+                , "expected"                : { "$ref": "#/definitions/expected"                }
                 }
           , "additionalProperties": false
           },
@@ -117,6 +118,11 @@
           { "description": "An identifier for similar optional test cases (kebab-case)"
           , "type"       : "string"
           , "pattern"    : "^[a-z]+(-[a-z]+)*$"
+          },
+    
+      "multi-line-string-array":
+          { "description": "Arrays of strings in the specified properties represent a multiline string"
+          , "type"       : "array"
           },
 
       "property":
--- a/canonical-schema.json
+++ b/canonical-schema.json
@@ -123,6 +123,9 @@
       "multi-line-string-array":
           { "description": "Arrays of strings in the specified properties represent a multiline string"
           , "type"       : "array"
+          , "items"      : {
+              "type": "string"
+            }
           },
 
       "property":
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,4 +8,5 @@ script:
   - bin/check_required_files_present
   - sh bin/check_versions
   - bin/check_optional
+  - bin/check_multi-line-string-array_contents
   - yarn test
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ is easier to understand with an example:
 
 ```json
 { "exercise": "foobar"
-, "version" : "1.1.0"
+, "version" : "1.2.0"
 , "comments":
     [ " Comments are always optional and can be used almost anywhere.      "
     , "                                                                    "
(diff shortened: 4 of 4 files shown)


canonical-schema.json
```diff
@@ -118,6 +119,11 @@
           , "type"       : "string"
           , "pattern"    : "^[a-z]+(-[a-z]+)*$"
           },
+    
+      "join-strings":
+          { "description": "If true, arrays of strings should be joined to a multiline string"
```
ErikSchierboom: I was wondering if we should add something like "arrays of strings in the input or expected properties", to make it clear that this property is only aimed at properties in the input and expected properties.

SleeplessByte:
> Or is that over-constraining it?

It's probably not cross-env correct.

Perhaps it should just be `multi-line-string-array` ?

ErikSchierboom:
> Perhaps it should just be multi-line-string-array

I like this option best, as it gives the test generators the freedom to do with it whatever they want (e.g. just keeping it an array).

SaschaMann:
`multi-line-string-array` is a much better name indeed.

canonical-schema.json
SaschaMann: Sounds like a good idea.

SleeplessByte:
> I like this option best, as it gives the test generators the freedom to do with it whatever they want (e.g. just keeping it an array).

Exactly! It doesn't prescribe that they HAVE to be joined, but only that that is the representation.

bencoman:
I like `multi-line-string-array` better.  
Now the description says... **Arrays of strings in `input` or `expected`**
So trying to understand better how it will apply by considering an [extract from `proverb`](https://github.com/exercism/problem-specifications/blob/5326cf2b32/exercises/proverb/canonical-data.json#L44-L59)
```
          {
            "description": "three pieces",
            "property": "recite",
            "input": {
              "strings": [
                "nail",
                "shoe",
                "horse"
              ]
            },
            "expected": [
                "For want of a nail the shoe was lost.",
                "For want of a shoe the horse was lost.",
                "And all for the want of a nail."
            ]
        },
```
with **description.md** saying... 
    `given the list ["nail", "shoe", "horse", "rider", "message", "battle", "kingdom"]`

I'm not clear if the `input` should be flattened as well as the `output` or how `multi-line-string-array` distinguishes between them.

SleeplessByte:
Perhaps: 

`multi-line-string-array: ['strings', 'expected']` ?

ErikSchierboom:
> multi-line-string-array: ['strings', 'expected'] 

This could work!

SaschaMann:
Is there a way to specify which entries are allowed in that array in the JSON schema directly?

SleeplessByte:
So I'm proposing keyof input and the literal 'expected'

NobbZ:
What if we had an input `expected`?

SleeplessByte:
We shouldn't and I suggest explicitly denying that from the input list. We also _shouldn't_ have any of the keywords of any of the 50 tracks be input and IIRC we try to satisfy that need.

bencoman:
> `multi-line-string-array: ['strings', 'expected'] `

Although in the case of **proverb** where the *description.md* specifies the input is a list of words, I'd expect  it to be `multi-line-string-array: ['expected'] - which further demonstrates the point.

bencoman:
[off topic]
> shouldn't have any of the keywords of any of the 50 tracks be input

Is there a list somewhere (or should there be?) used as a CI test?

ErikSchierboom:
> What if we had an input `expected`?

As @SleeplessByte said, I just think we should not ever do this.

sshine:
Since I'm not maintaining any test generators, I do not feel qualified to say that this will be beneficial.

SleeplessByte:
> Is there a list somewhere (or should there be?) used as a CI test?

Not that I know, butttt we should write this!

SaschaMann:
> So I'm proposing keyof input and the literal 'expected'

Yes, I was wondering if there's a way to represent this in the JSON schema itself, or if I just change the type to `array` and explain this behaviour in the description. I can't figure it out from the JSON schema spec and I'm not very familiar with it.

bencoman:
> > So I'm proposing keyof input and the literal 'expected'

> I was wondering if there's a way to represent this in the JSON schema itself

I'm also not very familiar with JSON schema, but ultimately the verification is done by the CI,
so just have a CI step that verifies it directly rather than indirectly through the schema.

> or if I just change the type to array and explain this behaviour in the description. 

That seems simplest and avoids the next person needing to get intimate with JSON Schema. 
The description can refer to a restricted keywords file tested by the CI.

SaschaMann:
Testing CI scripts is a bit annoying, #1547 demonstrates that the script correctly fails when `multi-line-string-array` contains an entry that's not used as input in this exercise but please double-check the CI script since this is the first time I've ever used `jq`. The script is primarily based on @sshine's `bin/check_optional` (thank you!).

SaschaMann:
The restricted keyword check should probably be moved to its own issue/PR because

> shouldn't have any of the keywords of any of the 50 tracks be input

requires an immense effort.

SaschaMann:
Thinking more about it: perhaps this should be a top-level flag since it's likely that the same input will always be a multi-line-string. What do you think?

bencoman:
>
> Thinking more about it: perhaps this should be a top-level flag since it's
> likely that the same input will always be a multi-line-string. What do you
> think?
>
Good insight - only I'm not sure about the impact on `expected` which could
vary.  Would it be too complicated having the key at multiple levels so
that `expected` could be per-test?

SaschaMann:
Good point, didn't think of that.

> Would it be too complicated having the key at multiple levels so that `expected` could be per-test?

Easy in the schema but likely annoying in the generators. So maybe it's not a good idea on the top-level.

sshine:
> shouldn't have any of the keywords of any of the 50 tracks be input

There's something very bad about such a constraint.

Much like an `if (x instanceof A) { ... } else if (x instanceof B) { ... } else ...` statement.

How about making it the responsibility of each test generator to deserialize JSON in such a way that it does not conflict with the syntax of the language? Surely no JSON parser should be so weak that they cannot deserialize a field that contains a string that overlaps with a keyword.

>> Would it be too complicated having the key at multiple levels so that expected could be per-test?
>
> Easy in the schema but likely annoying in the generators. So maybe it's not a good idea on the top-level.

Has someone building test generators tried and verified this yet?

Again, I'm not qualified to say.

bin/check_multi-line-string-array_contents
```diff
@@ -0,0 +1,38 @@
...
+
+# This script checks that multi-line-string-array only contains valid keys.
+
+failed=0
+
+check_keys() {
+    json_file=$1
+    echo "Checking 'multi-line-string-array' contents in $json_file"
+
+    present_keys=$(jq '[ .cases[] | recurse(.cases[]?) | ."multi-line-string-array" | select(. != null)[] ] | unique ' $json_file)
```
sshine: Since you have specified in canonical-schema.json that `multi-line-string-array` cannot be nested in `labeledTestGroup`, but only directly in `labeledTest` (which is unlike the `optional` parameter that is allowed to occur in both places), the use of `recurse(.cases[]?)` is not necessary and would incorrectly accept changes that break the schema.

It sounds like you were interested in not allowing nesting for the sake of simplicity.

Either way, the schema and the linter should conform.

Also, leading whitespace in `jq` string.

```suggestion
    present_keys=$(jq '[ .cases[] | ."multi-line-string-array" | select(. != null)[] ] | unique' $json_file)
```

bin/check_multi-line-string-array_contents
```diff
@@ -0,0 +1,38 @@
...
+
+    allowed_keys=$(jq '[ .cases[] | recurse(.cases[]?) | .input | keys[] ] | unique + ["expected"]' $json_file)
```
sshine: Here it *does* make sense to `recurse(.cases[]?)` if we assume that an outer `multi-line-string-array` key applies to nested cases in a `labeledTestGroup`.

I tried to run this on a canonical-data.json picked at random (high-scores), and it resulted in an error:

```
$ jq '[ .cases[] | recurse(.cases[]?) | .input | keys[] ] | unique + ["expected"]' exercises/high-scores/canonical-data.json
jq: error (at exercises/high-scores/canonical-data.json:83): null (null) has no keys
```

The problem here is that `labeledTestGroup` does not have direct `.input` fields, so this produces `null` values that, naturally, don't have any `keys`. The nested keys are still addressed by `recurse(.cases[]?)`.

One fix is to filter out those `null`s.

---

> What if we had an input `expected`?

> We shouldn't and I suggest explicitly denying that from the input list.

But we don't currently deny that, so *what if we had?*

Then `"expected"` might occur in the recursive search, potentially producing

```
[ "input1", "expected", "input3", "expected" ]
```

because `"expected"` is added to the allowed set *after* `unique`.

```suggestion
    allowed_keys=$(jq '[ .cases[] | recurse(.cases[]?) | .input | select(. != null) | keys[] ] + ["expected"] | unique' $json_file)
```

yawpitch:
> > shouldn't have any of the keywords of any of the 50 tracks be input
> 
> There's something very bad about such a constraint.

Agreed. If we try to avoid reserved keywords (nevermind the intersection of all language’s standard library names) we’ll run out of words quite quickly. Can anyone name a language that parses JSON data by evaluating it as native code?

bin/check_multi-line-string-array_contents
```diff
@@ -0,0 +1,38 @@
...
+
+    invalid_keys=$(jq --null-input \
+                      --argjson allowed "$allowed_keys" \
+                      --argjson present "$present_keys" \
+                      --raw-output '$present - $allowed | join("\n")')
+
+    if [ ! -z "$invalid_keys" ]; then
+        echo "Invalid multi-line-string-array entries:"
```
sshine: As far as I can determine, there can be two sources of errors:

1. There's a line `"multi-line-string-array": ["foo"]`, but there is no `input` parameter named `"foo"`.
2. There's a line `"multi-line-string-array": ["foo"]`, but the `input` parameter named `"foo"` is not an array of strings.

#1547 tests 1, and the spec changes enforce 2 partially: is an array, but not necessarily of strings.

I would like if the CI script would say something more informative like:

```suggestion
        echo "The following multi-line-string-array input parameters did not occur in any tests:"
```

sshine:
I'm marking my review as approved because I'm going on vacation for three weeks today, so I won't be able to re-review my proposed changes. I'm happy to look at it again in three weeks, but if you'd like to push forward before then, I suggest to be removed from the loop. :-)

SaschaMann:
@sshine Thanks for the fixes to CI!

> Can JSON Schema enforce that something is an array of only strings?
> 
> We would have to either encode that in the schema, accept that multi-line-string-array contains non-string values, or extend the CI check.

This is a good idea and can be done in the schema itself.

SaschaMann:
I'll add a test for

> 2. There's a line "multi-line-string-array": ["foo"], but the input parameter named "foo" is not an array of strings.

as well.

SaschaMann:
It wasn't, I kinda forgot that this existed after my last comment and then #1560 came up.

ErikSchierboom:
@SaschaMann Are you still interested in working on this PR?

SaschaMann:
No.

PR #1547, opened 2019-07-04
[WIP] Test that bin/check_multi-line-string-array_contents works (DO NOT MERGE!)

SaschaMann:
This tests the script added by #1544.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,4 +8,5 @@ script:
   - bin/check_required_files_present
   - sh bin/check_versions
   - bin/check_optional
+  - bin/check_multi-line-string-array_contents
   - yarn test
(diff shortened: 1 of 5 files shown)


SaschaMann:
This correctly fails due the presence of 

```json

    {
      "description": "change with Lilliputian Coins",
      "property": "findFewestCoins",
      "multi-line-string-array": ["target", "foo"],
      "input": {
        "coins": [1, 4, 15, 20, 50],
        "target": 23
      },
      "expected": [4, 4, 15]
    }
```

sshine:
I'm unsure how to interpret why this failed. Aren't there two reasons here?
1. `"foo"` does not occur as an input parameter name.
2. `"target"` is not a multi-line-string-array, but an integer.

SaschaMann:
1. is why it failed;
2. I've just added a check that the keys are valid keys, but not that they're string arrays. This _should_ be a reason why it fails, indeed.

PR #1548, opened 2019-07-05
Resistor Color: Additional info for problem spec

### LINKED ISSUE #1546: Resistor color: no task description
The explanation for how resistor colors (#1466) work is fine, but in terms of what the task actually is – there's not much there.

```Resistors have color coded bands, where each color maps to a number. The first 2 bands of a resistor have a simple encoding scheme: each color maps to a single number.

These colors are encoded as follows:

Black: 0
Brown: 1
Red: 2
Orange: 3
Yellow: 4
Green: 5
Blue: 6
Violet: 7
Grey: 8
White: 9
Mnemonics map the colors to the numbers, that, when stored as an array, happen to map to their index in the array: Better Be Right Or Your Great Big Values Go Wrong.

More information on the color encoding of resistors can be found in the Electronic color code Wikipedia article

```

 I propose the addition of a line that describes the expected input and output for this exercise. Maybe something like: 

`"Write a function that takes a band color as input and outputs the associated number."`

okeegan:
Resolves #1546.

### DIFF
--- a/exercises/resistor-color/description.md
+++ b/exercises/resistor-color/description.md
@@ -15,4 +15,6 @@ These colors are encoded as follows:
 
 Mnemonics map the colors to the numbers, that, when stored as an array, happen to map to their index in the array: Better Be Right Or Your Great Big Values Go Wrong.
 
-More information on the color encoding of resistors can be found in the [Electronic color code Wikipedia article](https://en.wikipedia.org/wiki/Electronic_color_code)
+More information on the color encoding of resistors can be found in the [Electronic color code Wikipedia article](https://en.wikipedia.org/wiki/Electronic_color_code).
+
+Create the means to look up the numerical value associated with a particular color band.


exercises/resistor-color/description.md
```diff
@@ -15,4 +15,6 @@ These colors are encoded as follows:
 
 Mnemonics map the colors to the numbers, that, when stored as an array, happen to map to their index in the array: Better Be Right Or Your Great Big Values Go Wrong.
 
-More information on the color encoding of resistors can be found in the [Electronic color code Wikipedia article](https://en.wikipedia.org/wiki/Electronic_color_code)
+More information on the color encoding of resistors can be found in the [Electronic color code Wikipedia article](https://en.wikipedia.org/wiki/Electronic_color_code).
+
+Create the means to look up the numerical value associated with a particular color band.
```
ErikSchierboom: Maybe move this line between the "Mnemonic" and "More information" lines? I feel like the "More information" line should probably be the last line.

yawpitch: I agree the "Create the means" line could be moved above the "More information" link, but then the prescriptive sense of the sentence is a little lost. Perhaps change it to "The goal of this exercise is to create" instead?

ErikSchierboom: I like that.

kotp: I like the move idea as well.

okeegan:
I question the usefulness of the wikipedia links in the task description. Surely if someone was interested they're only a google and a click away from that information?

ErikSchierboom:
> I question the usefulness of the wikipedia links in the task description. Surely if someone was interested they're only a google and a click away from that information?

I'm not sure I understand this argument to be honest. Surely the same reasoning could be applied to _any_ link?

okeegan:
Good point! I was really just thinking that very few other exercises have this kind of link to more information in the body, instead opting to give information about the source of the problem or concept in the 'Source' section.

ErikSchierboom:
Well, source should in general link to the page that served as the inspiration for the exercise. That could be a Wikipedia article, although it usually is another page (e.g. an issue in this repository).

PR #1550, opened 2019-07-13
Circular buffer - add test case for initial clear wrapping

marko213:
This change is intended to make the initial clear test case a little more strict - testing the overwrite and read fail capability after clearing as well.

The test case was also moved because it now uses overwrites and duplicating the test case didn't seem reasonable.

Original PR on the C++ track: exercism/cpp#296

This is an edge case that could happen when the clear alters the way the end of container is handled, as shown in the above PR. However, as also discussed, it is currently very specific and could be accidentally circumvented.

### DIFF
--- a/exercises/circular-buffer/canonical-data.json
+++ b/exercises/circular-buffer/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "circular-buffer",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "comments": [
     "In general, these circular buffers are expected to be stateful,",
     "and each language will operate on them differently.",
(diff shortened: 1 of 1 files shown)


petertseng:
The pull request description has a good description because it describes the reason for the change. The description should be included in the commit message as well; otherwise the commit message only says "modify test case" which is not helpful for understanding at all.

marko213:
To that I'd argue that two cases testing whether initially clearing changes something is repetitive. Besides, the case was moved for just this reason - the overwriting and failing reads would be tested in previous test cases, making this a complex, compound test case, kind of a "real world test" where many capabilities are tested, not just a few.

yawpitch:
> Even though I think it is possible to successfully argue for replacing the old test case instead of keeping it, we should prepare arguments against a reviewer who says the above quote

I'd be one of the reviewers who would say something like that quote.

> To that I'd argue that two cases testing whether initially clearing changes something is repetitive. 

There's nothing wrong with being repetitive in tests if the "something" under test is different: the original  test ensures you can use a cleared buffer, the second tests that clearing a buffer does not introduce a capacity issue in the buffer. They're discrete concerns under test so they merit discrete tests.

> Besides, the case was moved for just this reason - the overwriting and failing reads would be tested in previous test cases, making this a complex, compound test case - kind of a "real world" test where many capabilities are tested at once, not just a few at a time.

That isn't unit testing, though, is it? The more properties a test case exercises the harder it is to know _precisely_ what went wrong. More and simpler tests always trumps fewer and more complex ones.

marko213:
> That isn't unit testing, though, is it? The more properties a test case exercises the harder it is to know precisely what went wrong. More and simpler tests always trumps fewer and more complex ones.

The particular bug present in the linked solution only triggered if the buffer was cleared and then more writes than the size of the buffer occurred - I could not figure out a simpler test case which would trigger the bug.

petertseng:
anyone else who wishes to comment, you have about 24 hours to do so.

PR #1551, opened 2019-07-14
resistor-color-trio: add exercise

### LINKED ISSUE #1549: Proposal new exercise in the Resistor series: Resistor Color Trio
About the series https://github.com/exercism/problem-specifications/issues/1458
This is # 3 in the Resistor Series. It is a follow up for [Resistor Colors Duo](https://github.com/exercism/problem-specifications/blob/master/exercises/resistor-color-duo/description.md)  
This exercises introduces the third color band, that is needed to find the real value of a resistor.

### Description
If you want to build something using a Raspberry Pi, you'll probably use _resistors_. For this exercise, you need to know only three things about them:

- Each resistor has a resistance value.
- Resistors are small - so small in fact that if you printed the resistance value on them, it would be hard to read.
To get around this problem, manufacturers print color-coded bands onto the resistors to denote their resistance values. 
- Each band acts as a digit of a number. For example, if they printed a brown band (value 1) followed by a green band (value 5), it would translate to the number 15.
In this exercise, you are going to create a helpful program so that you don't have to remember the values of the bands. The program will take 3 colors as input, and outputs the correct value, in Ohms.
The color bands are encoded as follows:

- Black: 0
- Brown: 1
- Red: 2
- Orange: 3
- Yellow: 4
- Green: 5
- Blue: 6
- Violet: 7
- Grey: 8
- White: 9

In Resistor Colors Duo you de-coded the first two colors. For instance: orange-orange got the main value `33`
The third color stands for how many zeros need to be added to the main value. The main value plus the zeros gives us a value in Ohms.
For the exercise it doesn't matter what Ohms really are. 
For example:
* orange-orange-black would be 33 and no zeros, which becomes 33 Ohms
* orange-orange-red would be 33 and 2 zeros, which becomes 3300 Ohms
* orange-orange-orange would be 33 and 3 zeros, which becomes 33000 Ohms.

(If Math is your thing, you may want to think of the zeros as exponents of 10. If Math is not your thing, go with the zeros. It really is the same thing, just in plain English instead of Math lingo.)

This exercise is about translating the colors into a sentence:

> "The value of this resistor is … Ohms."

So an input of `“orange”, “orange”, “black”` should return 
> “The value of this resistor is 33 Ohms.”

The trick is that when we talk about Ohms, as soon as we get more than a thousand, we say “kilo”, in the same way that we say “kilograms” (kg) and “kilometers” (km). In the same way, we get “kilo-ohms” (kOhms).
So an input of `"orange", "orange", "orange"` should return
 > "The value of this resistor is 33 kOhms."

---

Question to track maintainers:
In the description, we reference the previous exercise in the series. That's because we think it makes sense to have the Duo and Trio exercises both in the track. Is our thinking correct? Or do you see reasons to add only Trio and not Duo? 

If maintainers choose to only implement the Trio exercise, we should adjust the description, to include how to compose the main value by using the first two color bands.

ErikSchierboom:
Closes #1549

### DIFF
--- a/exercises/resistor-color-trio/canonical-data.json
+++ b/exercises/resistor-color-trio/canonical-data.json
@@ -8,7 +8,7 @@
       "input": {
         "colors": ["orange", "orange", "black"]
       },
-      "expected": "33 Ohms"
+      "expected": "33 ohms"
     },
     {
       "description": "Blue and grey and brown",
@@ -16,7 +16,7 @@
       "input": {
         "colors": ["blue", "grey", "brown"]
       },
-      "expected": "680 Ohms"
+      "expected": "680 ohms"
     },
     {
       "description": "Red and black and red",
@@ -24,7 +24,7 @@
       "input": {
         "colors": ["red", "black", "red"]
       },
-      "expected": "2 kilo Ohms"
+      "expected": "2 kiloohms"
     },
     {
       "description": "Green and brown and orange",
@@ -32,7 +32,7 @@
       "input": {
         "colors": ["green", "brown", "orange"]
       },
-      "expected": "51 kilo Ohms"
+      "expected": "51 kiloohms"
     },
     {
       "description": "Yellow and violet and yellow",
@@ -40,7 +40,7 @@
       "input": {
         "colors": ["yellow", "violet", "yellow"]
       },
-      "expected": "470 kilo Ohms"
+      "expected": "470 kiloohms"
     }
   ]
 }
--- a/exercises/resistor-color-trio/canonical-data.json
+++ b/exercises/resistor-color-trio/canonical-data.json
@@ -8,7 +8,7 @@
       "input": {
         "colors": ["orange", "orange", "black"]
       },
-      "expected": "The value of this resistor is 33 Ohms."
+      "expected": "This is a 33 Ohms resistor."
     },
     {
       "description": "Blue and grey and brown",
@@ -16,7 +16,7 @@
       "input": {
         "colors": ["blue", "grey", "brown"]
       },
-      "expected": "The value of this resistor is 680 Ohms."
+      "expected": "This is a 680 Ohms resistor."
     },
     {
       "description": "Red and black and red",
@@ -24,7 +24,7 @@
       "input": {
         "colors": ["red", "black", "red"]
       },
-      "expected": "The value of this resistor is 2 kilo Ohms."
+      "expected": "This is a 2 kilo Ohms resistor."
     },
     {
       "description": "Green and brown and orange",
@@ -32,7 +32,7 @@
       "input": {
         "colors": ["green", "brown", "orange"]
       },
-      "expected": "The value of this resistor is 51 kilo Ohms."
+      "expected": "This is a 51 kilo Ohms resistor."
     },
     {
       "description": "Yellow and violet and yellow",
@@ -40,7 +40,7 @@
       "input": {
         "colors": ["yellow", "violet", "yellow"]
       },
-      "expected": "The value of this resistor is 470 kilo Ohms."
+      "expected": "This is a 470 kilo Ohms resistor."
     }
   ]
 }
--- a/exercises/resistor-color-trio/canonical-data.json
+++ b/exercises/resistor-color-trio/canonical-data.json
@@ -1,0 +1,46 @@
+{
+  "exercise": "resistor-color-trio",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "Orange and orange and black",
+      "property": "description",
+      "input": {
+        "colors": ["orange", "orange", "black"]
+      },
+      "expected": "The value of this resistor is 33 Ohms."
+    },
+    {
+      "description": "Blue and grey and brown",
+      "property": "description",
+      "input": {
+        "colors": ["blue", "grey", "brown"]
+      },
+      "expected": "The value of this resistor is 680 Ohms."
+    },
+    {
+      "description": "Red and black and red",
+      "property": "description",
+      "input": {
+        "colors": ["red", "black", "red"]
+      },
+      "expected": "The value of this resistor is 2 kOhms."
+    },
+    {
+      "description": "Green and brown and orange",
+      "property": "description",
+      "input": {
+        "colors": ["green", "brown", "orange"]
+      },
+      "expected": "The value of this resistor is 51 kOhms."
+    },
+    {
+      "description": "Yellow and violet and yellow",
+      "property": "description",
+      "input": {
+        "colors": ["yellow", "violet", "yellow"]
+      },
+      "expected": "The value of this resistor is 470 kOhms."
+    }
+  ]
+}
--- a/exercises/resistor-color-trio/canonical-data.json
+++ b/exercises/resistor-color-trio/canonical-data.json
@@ -4,51 +4,43 @@
   "cases": [
     {
       "description": "Orange and orange and black",
-      "property": "value",
+      "property": "label",
       "input": {
         "colors": ["orange", "orange", "black"]
       },
-      "expected": 33
+      "expected": "33 Ohms"
     },
     {
-      "description": "Green and brown and brown",
-      "property": "value",
+      "description": "Blue and grey and brown",
+      "property": "label",
       "input": {
-        "colors": ["green", "brown", "orange"]
+        "colors": ["blue", "grey", "brown"]
       },
-      "expected": 510
-    },
-    {
-      "description": "Blue and grey and yellow",
-      "property": "value",
-      "input": {
-        "colors": ["blue", "grey", "yellow"]
-      },
-      "expected": 680000
-    },
-    {
-      "description": "Brown and black and black",
-      "property": "description",
-      "input": {
-        "colors": ["brown", "black", "black"]
-      },
-      "expected": "This is a 1O Ohms resistor."
+      "expected": "680 Ohms"
     },
     {
       "description": "Red and black and red",
-      "property": "description",
+      "property": "label",
       "input": {
         "colors": ["red", "black", "red"]
       },
-      "expected": "This is a 2 kilo Ohms resistor."
+      "expected": "2 kilo Ohms"
+    },
+    {
+      "description": "Green and brown and orange",
+      "property": "label",
+      "input": {
+        "colors": ["green", "brown", "orange"]
+      },
+      "expected": "51 kilo Ohms"
     },
     {
       "description": "Yellow and violet and yellow",
-      "property": "description",
+      "property": "label",
       "input": {
         "colors": ["yellow", "violet", "yellow"]
       },
-      "expected": "This is a 470 kilo Ohms resistor."
+      "expected": "470 kilo Ohms"
     }
   ]
 }
--- a/exercises/resistor-color-trio/canonical-data.json
+++ b/exercises/resistor-color-trio/canonical-data.json
@@ -8,7 +8,10 @@
       "input": {
         "colors": ["orange", "orange", "black"]
       },
-      "expected": "33 ohms"
+      "expected": {
+        "value": 33,
+        "unit": "ohms"
+      }
     },
     {
       "description": "Blue and grey and brown",
@@ -16,7 +19,10 @@
       "input": {
         "colors": ["blue", "grey", "brown"]
       },
-      "expected": "680 ohms"
+      "expected": {
+        "value": 680,
+        "unit": "ohms"
+      }
     },
     {
       "description": "Red and black and red",
@@ -24,7 +30,10 @@
       "input": {
         "colors": ["red", "black", "red"]
       },
-      "expected": "2 kiloohms"
+      "expected": {
+        "value": 2,
+        "unit": "kiloohms"
+      }
     },
     {
       "description": "Green and brown and orange",
@@ -32,7 +41,10 @@
       "input": {
         "colors": ["green", "brown", "orange"]
       },
-      "expected": "51 kiloohms"
+      "expected": {
+        "value": 51,
+        "unit": "kiloohms"
+      }
     },
     {
       "description": "Yellow and violet and yellow",
@@ -40,7 +52,10 @@
       "input": {
         "colors": ["yellow", "violet", "yellow"]
       },
-      "expected": "470 kiloohms"
+      "expected": {
+        "value": 470,
+        "unit": "kiloohms"
+      }
     }
   ]
 }
--- a/exercises/resistor-color-trio/canonical-data.json
+++ b/exercises/resistor-color-trio/canonical-data.json
@@ -24,7 +24,7 @@
       "input": {
         "colors": ["red", "black", "red"]
       },
-      "expected": "The value of this resistor is 2 kOhms."
+      "expected": "The value of this resistor is 2 kilo Ohms."
     },
     {
       "description": "Green and brown and orange",
@@ -32,7 +32,7 @@
       "input": {
         "colors": ["green", "brown", "orange"]
       },
-      "expected": "The value of this resistor is 51 kOhms."
+      "expected": "The value of this resistor is 51 kilo Ohms."
     },
     {
       "description": "Yellow and violet and yellow",
@@ -40,7 +40,7 @@
       "input": {
         "colors": ["yellow", "violet", "yellow"]
       },
-      "expected": "The value of this resistor is 470 kOhms."
+      "expected": "The value of this resistor is 470 kilo Ohms."
     }
   ]
 }
--- a/exercises/resistor-color-trio/canonical-data.json
+++ b/exercises/resistor-color-trio/canonical-data.json
@@ -4,19 +4,35 @@
   "cases": [
     {
       "description": "Orange and orange and black",
-      "property": "description",
+      "property": "value",
       "input": {
         "colors": ["orange", "orange", "black"]
       },
-      "expected": "This is a 33 Ohms resistor."
+      "expected": 33
     },
     {
-      "description": "Blue and grey and brown",
+      "description": "Green and brown and brown",
+      "property": "value",
+      "input": {
+        "colors": ["green", "brown", "orange"]
+      },
+      "expected": 510
+    },
+    {
+      "description": "Blue and grey and yellow",
+      "property": "value",
+      "input": {
+        "colors": ["blue", "grey", "yellow"]
+      },
+      "expected": 680000
+    },
+    {
+      "description": "Brown and black and black",
       "property": "description",
       "input": {
-        "colors": ["blue", "grey", "brown"]
+        "colors": ["brown", "black", "black"]
       },
-      "expected": "This is a 680 Ohms resistor."
+      "expected": "This is a 1O Ohms resistor."
     },
     {
       "description": "Red and black and red",
@@ -25,14 +41,6 @@
         "colors": ["red", "black", "red"]
       },
       "expected": "This is a 2 kilo Ohms resistor."
-    },
-    {
-      "description": "Green and brown and orange",
-      "property": "description",
-      "input": {
-        "colors": ["green", "brown", "orange"]
-      },
-      "expected": "This is a 51 kilo Ohms resistor."
     },
     {
       "description": "Yellow and violet and yellow",
--- a/exercises/resistor-color-trio/description.md
+++ b/exercises/resistor-color-trio/description.md
@@ -33,9 +33,9 @@ This exercise is about translating the colors into a sentence:
 
 > "The value of this resistor is … Ohms."
 
-So an input of `“orange”, “orange”, “black”` should return
+So an input of `"orange”, "orange”, "black”` should return
 
-> “The value of this resistor is 33 Ohms.”
+> "The value of this resistor is 33 Ohms.”
 
 The trick is that when we talk about Ohms, as soon as we get more than a thousand, we say "kilo", in the same way that we say "kilograms" (kg) and "kilometers" (km). In the same way, we get "kilo-ohms" (kilo Ohms).
 So an input of `"orange", "orange", "orange"` should return
--- a/exercises/resistor-color-trio/description.md
+++ b/exercises/resistor-color-trio/description.md
@@ -37,7 +37,7 @@ So an input of `"orange", "orange", "black"` should return:
 
 > "The value of this resistor is 33 Ohms."
 
-The trick is that when we talk about Ohms, as soon as we get more than a thousand, we say "kilo", in the same way that we say "kilograms" (kg) and "kilometers" (km). In the same way, we get "kilo-ohms" (kilo Ohms).
+When we get more than a thousand Ohms, we say "kilo Ohms". That's similar to saying "kilometer" for 1000 meters, and "kilograms" for 1000 grams. 
 So an input of `"orange", "orange", "orange"` should return:
 
 > "The value of this resistor is 33 kOhms."
--- a/exercises/resistor-color-trio/description.md
+++ b/exercises/resistor-color-trio/description.md
@@ -37,7 +37,7 @@ So an input of `“orange”, “orange”, “black”` should return
 
 > “The value of this resistor is 33 Ohms.”
 
-The trick is that when we talk about Ohms, as soon as we get more than a thousand, we say “kilo”, in the same way that we say “kilograms” (kg) and “kilometers” (km). In the same way, we get “kilo-ohms” (kOhms).
+The trick is that when we talk about Ohms, as soon as we get more than a thousand, we say "kilo", in the same way that we say "kilograms" (kg) and "kilometers" (km). In the same way, we get "kilo-ohms" (kilo Ohms).
 So an input of `"orange", "orange", "orange"` should return
 
 > "The value of this resistor is 33 kOhms."
--- a/exercises/resistor-color-trio/description.md
+++ b/exercises/resistor-color-trio/description.md
@@ -1,0 +1,43 @@
+If you want to build something using a Raspberry Pi, you'll probably use _resistors_. For this exercise, you need to know only three things about them:
+
+- Each resistor has a resistance value.
+- Resistors are small - so small in fact that if you printed the resistance value on them, it would be hard to read.
+  To get around this problem, manufacturers print color-coded bands onto the resistors to denote their resistance values.
+- Each band acts as a digit of a number. For example, if they printed a brown band (value 1) followed by a green band (value 5), it would translate to the number 15.
+  In this exercise, you are going to create a helpful program so that you don't have to remember the values of the bands. The program will take 3 colors as input, and outputs the correct value, in Ohms.
+  The color bands are encoded as follows:
+
+* Black: 0
+* Brown: 1
+* Red: 2
+* Orange: 3
+* Yellow: 4
+* Green: 5
+* Blue: 6
+* Violet: 7
+* Grey: 8
+* White: 9
+
+In `resistor-color duo` you decoded the first two colors. For instance: orange-orange got the main value `33`.
+The third color stands for how many zeros need to be added to the main value. The main value plus the zeros gives us a value in Ohms.
+For the exercise it doesn't matter what Ohms really are.
+For example:
+
+- orange-orange-black would be 33 and no zeros, which becomes 33 Ohms.
+- orange-orange-red would be 33 and 2 zeros, which becomes 3300 Ohms.
+- orange-orange-orange would be 33 and 3 zeros, which becomes 33000 Ohms.
+
+(If Math is your thing, you may want to think of the zeros as exponents of 10. If Math is not your thing, go with the zeros. It really is the same thing, just in plain English instead of Math lingo.)
+
+This exercise is about translating the colors into a sentence:
+
+> "The value of this resistor is … Ohms."
+
+So an input of `“orange”, “orange”, “black”` should return
+
+> “The value of this resistor is 33 Ohms.”
+
+The trick is that when we talk about Ohms, as soon as we get more than a thousand, we say “kilo”, in the same way that we say “kilograms” (kg) and “kilometers” (km). In the same way, we get “kilo-ohms” (kOhms).
+So an input of `"orange", "orange", "orange"` should return
+
+> "The value of this resistor is 33 kOhms."
--- a/exercises/resistor-color-trio/description.md
+++ b/exercises/resistor-color-trio/description.md
@@ -33,11 +33,11 @@ This exercise is about translating the colors into a sentence:
 
 > "The value of this resistor is … Ohms."
 
-So an input of `"orange”, "orange”, "black”` should return
+So an input of `"orange", "orange", "black"` should return:
 
-> "The value of this resistor is 33 Ohms.”
+> "The value of this resistor is 33 Ohms."
 
 The trick is that when we talk about Ohms, as soon as we get more than a thousand, we say "kilo", in the same way that we say "kilograms" (kg) and "kilometers" (km). In the same way, we get "kilo-ohms" (kilo Ohms).
-So an input of `"orange", "orange", "orange"` should return
+So an input of `"orange", "orange", "orange"` should return:
 
 > "The value of this resistor is 33 kOhms."
--- a/exercises/resistor-color-trio/description.md
+++ b/exercises/resistor-color-trio/description.md
@@ -31,7 +31,7 @@ For example:
 
 This exercise is about translating the colors into a sentence:
 
-> "The value of this resistor is … Ohms."
+> "The value of this resistor is ... Ohms."
 
 So an input of `"orange", "orange", "black"` should return:
 
--- a/exercises/resistor-color-trio/description.md
+++ b/exercises/resistor-color-trio/description.md
@@ -4,7 +4,7 @@ If you want to build something using a Raspberry Pi, you'll probably use _resist
 - Resistors are small - so small in fact that if you printed the resistance value on them, it would be hard to read.
   To get around this problem, manufacturers print color-coded bands onto the resistors to denote their resistance values.
 - Each band acts as a digit of a number. For example, if they printed a brown band (value 1) followed by a green band (value 5), it would translate to the number 15.
-  In this exercise, you are going to create a helpful program so that you don't have to remember the values of the bands. The program will take 3 colors as input, and outputs the correct value, in Ohms.
+  In this exercise, you are going to create a helpful program so that you don't have to remember the values of the bands. The program will take 3 colors as input, and outputs the correct value, in ohms.
   The color bands are encoded as follows:
 
 * Black: 0
@@ -19,25 +19,25 @@ If you want to build something using a Raspberry Pi, you'll probably use _resist
 * White: 9
 
 In `resistor-color duo` you decoded the first two colors. For instance: orange-orange got the main value `33`.
-The third color stands for how many zeros need to be added to the main value. The main value plus the zeros gives us a value in Ohms.
-For the exercise it doesn't matter what Ohms really are.
+The third color stands for how many zeros need to be added to the main value. The main value plus the zeros gives us a value in ohms.
+For the exercise it doesn't matter what ohms really are.
 For example:
 
-- orange-orange-black would be 33 and no zeros, which becomes 33 Ohms.
-- orange-orange-red would be 33 and 2 zeros, which becomes 3300 Ohms.
-- orange-orange-orange would be 33 and 3 zeros, which becomes 33000 Ohms.
+- orange-orange-black would be 33 and no zeros, which becomes 33 ohms.
+- orange-orange-red would be 33 and 2 zeros, which becomes 3300 ohms.
+- orange-orange-orange would be 33 and 3 zeros, which becomes 33000 ohms.
 
 (If Math is your thing, you may want to think of the zeros as exponents of 10. If Math is not your thing, go with the zeros. It really is the same thing, just in plain English instead of Math lingo.)
 
-This exercise is about translating the colors into a sentence:
+This exercise is about translating the colors into a label:
 
-> "The value of this resistor is ... Ohms."
+> "... ohms"
 
 So an input of `"orange", "orange", "black"` should return:
 
-> "The value of this resistor is 33 Ohms."
+> "33 ohms"
 
-When we get more than a thousand Ohms, we say "kilo Ohms". That's similar to saying "kilometer" for 1000 meters, and "kilograms" for 1000 grams. 
+When we get more than a thousand ohms, we say "kiloohms". That's similar to saying "kilometer" for 1000 meters, and "kilograms" for 1000 grams.
 So an input of `"orange", "orange", "orange"` should return:
 
-> "The value of this resistor is 33 kOhms."
+> "33 kiloohms"
--- a/exercises/resistor-color-trio/metadata.yml
+++ b/exercises/resistor-color-trio/metadata.yml
@@ -1,0 +1,4 @@
+---
+blurb: "Convert color codes, as used on resistors, to a human-readable string."
+source: "Maud de Vries, Erik Schierboom"
+source_url: "https://github.com/exercism/problem-specifications/issues/1549"
--- a/exercises/resistor-color-trio/metadata.yml
+++ b/exercises/resistor-color-trio/metadata.yml
@@ -1,4 +1,4 @@
 ---
-blurb: "Convert color codes, as used on resistors, to a human-readable string."
+blurb: "Convert color codes, as used on resistors, to a human-readable label."
 source: "Maud de Vries, Erik Schierboom"
 source_url: "https://github.com/exercism/problem-specifications/issues/1549"


exercises/resistor-color-trio/description.md
```diff
@@ -0,0 +1,43 @@
...
+
+This exercise is about translating the colors into a sentence:
+
+> "The value of this resistor is … Ohms."
+
+So an input of `“orange”, “orange”, “black”` should return
+
+> “The value of this resistor is 33 Ohms.”
+
+The trick is that when we talk about Ohms, as soon as we get more than a thousand, we say “kilo”, in the same way that we say “kilograms” (kg) and “kilometers” (km). In the same way, we get “kilo-ohms” (kOhms).
```
SaschaMann: For consistency and readability I'd prefer either writing out the unit completely (`kilo Ohms`) or in short form (`kΩ`), but not mixed.

I think you're using the unicode quotation marks here `“”` instead of `""`, is that intended?

F3PiX: Thanks Sascha! 
> writing out the unit completely (`kilo Ohms`) or in short form (`kΩ`), but not mixed.

We want to avoid special characters in this one, so let's go with 'kilo Ohms'.

>Using kilograms as example is slightly misleading 
I understand your reasoning, but since it's just about explaining where the 'kilo'-part is coming from in a non-mathy way, and since the base unit is not really relevant for this particular exercise, I'd rather leave it as is. Can you live with that?

> Quotation marks: not intended.

F3PiX:
```suggestion
The trick is that when we talk about Ohms, as soon as we get more than a thousand, we say "kilo", in the same way that we say "kilograms" (kg) and "kilometers" (km). In the same way, we get "kilo-ohms" (kilo Ohms).
```

SaschaMann: > Can you live with that?

Yes, of course. All of my comments were meant as suggestions that you may or may not consider, not requests for change. I should've made that clearer perhaps. :)

exercises/resistor-color-trio/canonical-data.json
```diff
@@ -0,0 +1,46 @@
...
+  "exercise": "resistor-color-trio",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "Orange and orange and black",
+      "property": "description",
+      "input": {
+        "colors": ["orange", "orange", "black"]
+      },
+      "expected": "The value of this resistor is 33 Ohms."
```
rpottsoh: Would it sufficient to just expect `33 Ohms`?  The rest of the sentence seems unnecessary.

F3PiX: I like when programs talk in real sentences. So, just returning the value is too little imo. But I agree that the sentence is adding noise... 

I see 2 options: either a shorter sentence, like "This is a 33 Ohms resistor". Or "The value of orange-orange-black is 33 Ohms". 
That latter would introduce a second iteration on `colors`. I'm not sure if that adds something interesting.
@ErikSchierboom I don't mind going with the shortened sentence. Do you?

coriolinus: For the problem specification, I'd prefer if the expected value were the integer `33` instead of any kind of string.

String formatting is not integral to this exercise, and a trivial wrapper function makes a sentence out of an integer. The reverse operation is less simple. 

We should be teaching good practices. One good practice is to avoid conflating two different logical operations into one function. Computing a resistor's value from its colors and formatting a human-friendly string describing a resistor are two different logical operations.

yawpitch: > For the problem specification, I'd prefer if the expected value were the integer `33` instead of any kind of string.

I’d strongly second that point. The first two exercises are focused on calculating the integer value by parsing a string, and this adds on the complexity of that from the previous two. Adding string interpolation is unnecessary and it’s not like we’re lacking in exercises for that anyway. What we have been lacking is _simple_ exercises that gradually expand on integer maths.

The very debate about preference about what string to return displays another problem with having one function that addresses two concerns; the string you’d print would be culturally and locale sensitive in the real world, so you’d want one function doing the math and another doing the interpolation. If we don’t want to add a second function, then _should_ we make _another_ exercise that spends time on English-specific string formatting esoterica (ie 3000 > 3K)?

F3PiX: I like the concept of separating the two concerns, one calculating the values, and one for returning a string.
I do think we need the string part of this exercise. Not the interpolation, but for the part that is about Ohms vs kilo Ohms. 
That is also meant as a bridge to the next Resistor exercise(s), where the Ohms value is combined with other values in other units (like a percentage for accuracy).
My proposal would be: separate the concerns in 2 methods, one for calculation, one for the ‘report’/printing. Can we all live with that?

F3PiX: > should we make another exercise that spends time on English-specific string formatting esoterica (ie 3000 > 3K)
Agreed. And IIRC there’s an issue to create such an exercise.

coriolinus: Sure, a two-method solution works for me. One accepts as input a list of colors and returns an integer; the other accepts an integer and returns a string. We should test them independently.

ErikSchierboom: > I like the concept of separating the two concerns, one calculating the values, and one for returning a string.

Fine with me!  

> One accepts as input a list of colors and returns an integer; the other accepts an integer and returns a string.

What do you think @F3PiX?

F3PiX: I'm totally okay with it 👍

exercises/resistor-color-trio/description.md
SleeplessByte: I think reversing them here would already be helpful:

```suggestion
The trick is that when we talk about Ohms, as soon as we get more than a thousand, we say "kilo", in the same way that we say "kilometers" (km) and "kilograms" (kg). In the same way, we get "kilo-ohms" (kilo Ohms).
```

I'm not a fan of the double "in the same way".

Perhaps:

```suggestion
The trick is that when we talk about Ohms, as soon as we get more than a thousand, we say "kilo", which results in  "kilo-ohms" (kilo Ohms). In the same way when we have more than a thousand metres we say "kilometers" (km) and when we have more than a thousand grams, we say "kilograms" (kg).
```

F3PiX: I see your point, @SleeplessByte, but I'm not convinced by this solution. I'll think of something else.

yawpitch: Wouldn’t it be “kilohms” not “Kilo-ohms”? Wolfram Alpha seems think so, and that led me to NIST and https://www.nist.gov/physical-measurement-laboratory/nist-guide-si-chapter-6#623, which indicates prefixes are inseparable from units. Elsewhere it mentions lowercase presentation and that:

> there are three cases in which the final vowel of an SI prefix is commonly omitted: megohm (not megaohm), kilohm (not kiloohm), and hectare (not hectoare)

Although I’ll be honest the only one of those I’m totally sure of is hectare.

SleeplessByte: Yes, hectare is the one I'm also sure about. @F3PiX do as you see fit ❤️ !

ErikSchierboom:
I've tried to address the various comments made.

exercises/resistor-color-trio/description.md
```diff
@@ -0,0 +1,43 @@
...
+
+- orange-orange-black would be 33 and no zeros, which becomes 33 Ohms.
+- orange-orange-red would be 33 and 2 zeros, which becomes 3300 Ohms.
+- orange-orange-orange would be 33 and 3 zeros, which becomes 33000 Ohms.
+
+(If Math is your thing, you may want to think of the zeros as exponents of 10. If Math is not your thing, go with the zeros. It really is the same thing, just in plain English instead of Math lingo.)
+
+This exercise is about translating the colors into a sentence:
+
+> "The value of this resistor is … Ohms."
```
SaschaMann: There's another unicode character here, `…`

exercises/resistor-color-trio/description.md
```diff
@@ -0,0 +1,43 @@
...
+
+So an input of `"orange”, "orange”, "black”` should return
```
SaschaMann: Here too: `”`

exercises/resistor-color-trio/description.md
```diff
@@ -0,0 +1,43 @@
...
+
+> "The value of this resistor is 33 Ohms.”
```
SaschaMann: And here.

exercises/resistor-color-trio/description.md
SaschaMann: The dots are still there

exercises/resistor-color-trio/canonical-data.json
```diff
@@ -4,35 +4,43 @@
...
+      "expected": 33
     },
     {
-      "description": "Blue and grey and brown",
-      "property": "description",
+      "description": "Green and brown and brown",
+      "property": "value",
       "input": {
-        "colors": ["blue", "grey", "brown"]
+        "colors": ["green", "brown", "orange"]
```
F3PiX: s/orange/brown to get 510

exercises/resistor-color-trio/canonical-data.json
```diff
@@ -0,0 +1,54 @@
...
+      "input": {
+        "colors": ["blue", "grey", "yellow"]
+      },
+      "expected": 680000
+    },
+    {
+      "description": "Brown and black and black",
+      "property": "description",
+      "input": {
+        "colors": ["brown", "black", "black"]
```
coriolinus: The inputs for the `description` property should be integers, not color lists. The student has already written a three-color conversion function, at this point. No need to force them to call it; it is both obvious and natural to hook the output of the color converter into the input of the describer.

F3PiX:
On second thought, I'm not happy with the new tests, and I'd vote for going with the original set of tests (with the colors as input and the full description as output).

My thinking is: This exercise is like a mini-app, where a user inputs a list of colors they are reading from a resistor, and they get back the description with the value. To me it makes sense that that's the only thing we test. 
The rest of the design are 'implementation details', that we don't need to test. 
(This is similar to the Luhn tests: input is a string, output is valid true or false. https://github.com/exercism/problem-specifications/blob/master/exercises/luhn/canonical-data.json) 

I can imagine that a student puts all calculation in the `description` method, and that will make the tests pass. Now that we have green, it's up to the mentor to initiate a refactor, and point out that the code works but that it is not a good design. 

The 'testing the public interface' argument is the main reason. 
A second reason is the learning part of the exercise. If the tests force the student to follow an upfront design, they don't need to think about why we should seperate concerns. When they start with a suboptimal design, the mentor naming the problem and let them solve it is a better learning opportunity. 

TL;DR  
**My proposal is to revert to the previous set of tests.** Which is not questioning the best practice in the _implementation_ to separate calculation the `value` from composing the `desciption`.
I  briefly discussed this with @ErikSchierboom, so for anyone else: can you too live with reverting to the original tests, and if not, why?

yawpitch:
> My thinking is: This exercise is like a mini-app, where a user inputs a list of colors they are reading from a resistor, and they get back the description with the value. To me it makes sense that that's the only thing we test.

Isn't that fundamentally true of every exercise? In that every exercise _could_ be seen as _either_ a re-useable library function _or_ a discrete app, and that we could simply de-refactor every exercise to return a long descriptive string in the same manner?

My argument against that would partially be that it forefronts bad habits, but _mainly_ that it forces our users to be fluent in vernacular English idioms and to recreate a string that they may not understand and which is _mostly_ a matter of personal preference to begin with. In the Python track, at least, we strive not to test specific error text, for instance, to avoid putting up another barrier to students who are ESL ... I think that's a good thing to consider across all tracks.

Also it seems to me that the TDD approach we've taken treats every exercise as a library by default, since testing of app integration is usually not done with unit tests.

> (This is similar to the Luhn tests: input is a string, output is valid true or false. https://github.com/exercism/problem-specifications/blob/master/exercises/luhn/canonical-data.json)

Yes, but `luhn` is returning unambiguous and actionable data, as a good library function should in order to build applications with it. A string that encodes multiple pieces of data isn't an actionable piece of data. A response like "33 ohms" or "33 kilohms" (or kiloohms or kilo-ohms) would be _more_ actionable, but 33 or 3300 would be _most_ actionable. Wouldn't a separate exercise to handle "humanising" -- more accurately "English colloquialising" -- be _more_ instructive than, essentially, re-hashing the same problem in multiple smaller exercises?

> I can imagine that a student puts all calculation in the `description` method, and that will make the tests pass. Now that we have green, it's up to the mentor to initiate a refactor, and point out that the code works but that it is not a good design.

Will that not end up being pointed out on _most every_ submission? That just seems like overhead.

> The 'testing the public interface' argument is the main reason.

Sure ... but _why_ is the public interface as interpolated string _necessary_? What is that adding to the exercise across _all_ tracks that are likely to implement the exercise? Don't we already have a significant number of exercises whose entire _necessary_ interface is such a string? I'm open to the idea that it might be teaching something fundamentally new in relation to other exercises that do such string interpolation, I'm just not seeing what that is at the moment.

> A second reason is the learning part of the exercise. If the tests force the student to follow an upfront design, they don't need to think about why we should seperate concerns. When they start with a suboptimal design, the mentor naming the problem and let them solve it is a better learning opportunity.

I agree there's a learning opportunity there, but I'm not sure it's not served by having the tests de-normalised. It's also an argument that can also be applied to all the exercises ... why _not_ de-normalise others back to "bad" behaviour so students can be walked forward to "good"?

> TL;DR
> **My proposal is to revert to the previous set of tests.** Which is not questioning the best practice in the _implementation_ to separate calculation the `value` from composing the `desciption`.
> I briefly discussed this with @ErikSchierboom, so for anyone else: can you too live with reverting to the original tests, and if not, why?

The debate is mainly over what "value" is being returned. There seems to be a fair bit of controversy caused by that "value" being a sentence. That's certainly my issue, based on fielding a lot of issues with students who already struggle with English idioms and a _large_ number of exercises that essentially rely on those idioms. Can I live with it too? Sure ... but I want to better understand the need for `resistor-color-trio` to deviate from `resistor-color` and `resistor-color-duo`, and from other exercises, by de-normalising the test specification towards an _intentionally_ bad design. And I'd much rather we consider returning a value that the user can _easily_ parse into an actionable form -- ie `"VALUE UNITS"` or better yet `(VALUE, UNITS)` -- rather than testing them, again, on string formatting to incorporate a the actual data into an arbitrary series of words that may be semantically meaningless to them.

exercises/resistor-color-trio/canonical-data.json
SleeplessByte: Except that the integers don't _mean_ anything by themselves. 

"orange" "orange" "orange" would be [3, 3, 3].

```
(10 * colorValue('orange') + colorValue('orange')) * (10 ** colorValue('orange'))
```

Would be a readable solution. If we use integers, we are moving away from correct and optimal solutions that look like this:

```
(colorFirstValue('orange') + colorSecondValue('orange')) * colorMultiplier('orange')
```

coriolinus: I didn't mean to suggest a list of integers; I meant a single integer, the value in ohms.

The intent is to get people to write one `calculate_value` function, and one `format_description` function. It should be obvious and trivial to nest the calls: if what you have is a list of colors, and what you want is a string description, there's no reason not to write `format_description(calculate_value(color_list))`. 

However, it would teach bad factorization to include both operations in a single function.

coriolinus:
Thank you @yawpitch for expressing exactly my objections with greater eloquence than I have available at the moment. 

I think it is a very bad idea to intentionally denormalize the design: even if mentors will usually prompt students to normalize it, what fraction of the mentors will forget to do so? What fraction of students will simply move on to the next exercise, instead of continuing to refine an accepted solution, and thereby learn bad habits?

SaschaMann:
> Will that not end up being pointed out on most every submission? That just seems like overhead.

Indeed, it seems like this would end up being an automated comment on the tracks that have auto-mentoring for common problems anyway.

I agree with @yawpitch but also want to add another objection to having only strings in the shared test data.

> My thinking is: This exercise is like a mini-app, where a user inputs a list of colors they are reading from a resistor, and they get back the description with the value. To me it makes sense that that's the only thing we test.
The rest of the design are 'implementation details', that we don't need to test.

Testing a specific, arbitrary string representation of a value with a physical unit is also an implementation detail. You're forcing a specific way to implement string representation such values, that may go against the conventions in certain languages. E.g. in Julia, the canonical way to represent a resistance value would be `350Ω` (note that this is not a string) with its string representation `"350 Ω"`. Turning this into `"350 ohm"` would require reinventing the wheel and teach bad habits.

Of course, this is just how one language does it, but if this problem spec _only_ specifies strings, it can't be used to create/generate tests without a lot of string-parsing and -converting hassle. When both the final string and the "pure-value" are part of the tests, this will be much easier because the strings can be ignored and instead be generated from the "pure-value" test data.

exercises/resistor-color-trio/description.md
F3PiX:
```suggestion
> "The value of this resistor is ... Ohms."
```

exercises/resistor-color-trio/description.md
```diff
@@ -0,0 +1,43 @@
...
+
+This exercise is about translating the colors into a sentence:
+
+> "The value of this resistor is … Ohms."
+
+So an input of `"orange", "orange", "black"` should return:
+
+> "The value of this resistor is 33 Ohms."
+
+The trick is that when we talk about Ohms, as soon as we get more than a thousand, we say "kilo", in the same way that we say "kilograms" (kg) and "kilometers" (km). In the same way, we get "kilo-ohms" (kilo Ohms).
```
F3PiX:
```suggestion
When we get more than a thousand Ohms, we say "kilo Ohms". That's similar to saying "kilometer" for 1000 meters, and "kilograms" for 1000 grams. 
```

exercises/resistor-color-trio/canonical-data.json
SleeplessByte: Ah. I understand what you're (both, @coriolinus and @yawpitch) are saying. I think having `value` return 3300 is fine 👍 😄 .

I think having `label` return `"33 kilo Ohms"` is also something we should have.
I think having `description` is something a track can add based on preference on a sentence.

I don't agree with all the "this will lead to bad code" arguments. I _understand_ what you're saying, but it's up to the student to recognise the two concerns or for the mentor to point them out. Purposefully crafting problem descriptions so a student can't do it wrongly is kinda forcing a certain implementation, which is something we don't want.

SleeplessByte:
I've read all that is said above and whilst I mostly agree with the nits and content of the arguments by @yawpitch and @coriolinus , here are my two cents:

- Prefer a single property in the canonical data. I suggest `label` which returns a formatted string (33 kilo Ohms). This exercise is meant to do 2 things: the algorithm math and the branching of humanisation. Adding extra text is not preferred (re: description), because in a real world program there are various use cases for `label` and very little for `description`. I agree with @coriolinus (and the second explanation by @yawpitch ) that `description` is composable. 

- Optionally allow tracks to add a local track HINT to implement `value` which returns a number which is the raw number of `kilo` ohms. The reason I don't think this should be in the tests is because it **breaks** the abstraction on the fourth band. I think the arguments here are all very alike primitive obsession. FYI, the fourth band _may_ (depending on existence and colors of fifth / sixth) add an error margin e.g. 1%. `value` then giving 33000 is an _incorrect abstraction_. It should almost give a range, or a special object that has median 33000 and std of 0.01 * mu or whatever. 

- If languages / tracks choose to, add `description` which takes the `label` and turns it into a sentence.

exercises/resistor-color-trio/canonical-data.json
SleeplessByte: See: https://github.com/exercism/problem-specifications/pull/1551#issuecomment-512589784

`value` as number (33000) is still an incorrect abstraction.

SleeplessByte:
@F3PiX I suggest adding the actual TABLE of coloured bands + indices + meaning to the description, instead of only the "numbers". The information here is incomplete. I also suggest doing that to all the exercises.

| color | significant figure / digit / value | multiplier | tolerance | temp. coeff |
|-------|-----------------------------------|------------|-----------|--------------|
| - | - | - | ±20 % | |
| black | 0 | 10^0 | - | 250 ppm / K |
| brown  | 1 | 10^1 | ±1 % | 100 ppm/K |
| red 🍓 |  2 | 10^2 | ±2 % | 50 ppm/K |
| orange 🍊 | 3 | 10^3 | ±0.05 % | 15 ppm/K |
| yellow 🍋 | 4 | 10^4 | ±0.02 % | 25 ppm/K |
| green 🍏 | 5 | 10^5 | ±0.5 % | 20 ppm/K |
| blue | 6 | 10^6 | ± 0.25 % | 10 ppm/K |
| violet | 7 | 10^7 | ±0.1 % | 5 ppm/K |
| grey | 8 | 10^8  | ±0.01 % | 1 ppm/K |
| white | 9 | 10^9 | -
| gold | - | 10^-1 | ±5 %
| silver | - | 10^-2 | ±10 %
| pink | - | 10^-3

We probably don't want to _overwhelm_ the student with information, so cut form this table as you wish, but the notion that the _INDEX_ of the band is not important (from the description) is probably an oversight. 

🍓 does not mean 2
🍊 does not mean 3
🍋 does not mean 4
🍏 does not mean 5

🍏🍋🍊🍓 does not mean 5432 (but `54 kilo Ohms ±2%` if I'm correct), however

🍋 in the first position means `4 * 10^1` (4)
🍋 in the second position means `4 * 10^0` (40)
🍋 in the third position means multiply by `10 ^ 4` (`10 000`)
🍋 in the fourth position means a tolerance of ±0.02% (unless they are super old, then yellow is gold is ±5% (???)) 

**The position is significant**. The descriptions of `resistor-color`, `resistor-color-duo` and `resistor-color-trio` do not make that apparent right now.

ErikSchierboom:
I really like @SleeplessByte's suggestion to use a label property instead of a description, which admittedly had some problems associated with it. There were some objections against using a string value, so let me address most of them:

> [...] it forces our users to be fluent in vernacular English idioms and to recreate a string that they may not understand and which is mostly a matter of personal preference to begin with.

This was indeed a problem with the old description property, but I think the label property works around this by reducing the string to only contain essential information. Yes, the text is still English, but would non-native English speakers really have much of a problem reading "33 kilo Ohms?" Kilo is a fairly international word and Ohms is, well, the official unit.

> We should be teaching good practices. One good practice is to avoid conflating two different logical operations into one function. Computing a resistor's value from its colors and formatting a human-friendly string describing a resistor are two different logical operations.

We should definitely be teaching good practices. That computing a resistor's value should be separate from the formatting is not something I'd argue against, this absolutely makes sense. However, this does not mean that we need to enforce this structure in the canonical data. In fact, it can make for some great mentoring opportunities where we can point students about separating the operations into separate operations if they did not do so (and automated mentoring could automatically catch this). Furthermore, if we assume that this exercise will be unlocked by the first or second version of the resistor-color series, students will already be familiar and have implemented a color->value property.

> Testing a specific, arbitrary string representation of a value with a physical unit is also an implementation detail. You're forcing a specific way to implement string representation such values, that may go against the conventions in certain languages.

Yep, this is true. It is not the goal of the exercise though to teach people how to represent the physical unit best in their languages (shameless plug for F#' [units of measure](https://fsharpforfunandprofit.com/posts/units-of-measure/)) though. The goal of the exercise is to teach people how to combine basic string formatting with basic math. This exercise has been designed specifically to be a good fit into the early part of a track's anatomy. As such, you can see this as an ideal early exercise, one that combines the math of the resistor-color one or two exercise and the string formatting of the two-fer exercise. 

Hopefully this helps explain _why_ the exercise is designed as it is currently.

ErikSchierboom:
What do you, @F3PiX, think about the table suggestion made by @SleeplessByte?

coriolinus:
I think adding the table to the description could be useful, once the table is reasonably complete. Not sure what purpose the emoji serve in it.

> Optionally allow tracks to add a local track HINT to implement `value`... The reason I don't think this should be in the tests is because it breaks the abstraction on the fourth band.

I disagree with this assertion. If you were writing a program which depended on four-band properties of resistors, you would never move straight from the band pattern to a formatted string; you'd have a function which turned a list of colors into some kind of struct, and a function which stringified that struct.

The point isn't that we absolutely must use a primitive. It's that for the three-band problem in this exercise, an integer is both sufficient, and a better representation of the value than a string.  

> However, this does not mean that we need to enforce this structure in the canonical data. In fact, it can make for some great mentoring opportunities

Two objections there. One is that the last thing we need right now is to add work for the mentors! Rust's backlog is already large, and consistently growing on a weekly basis. Particularly for early exercises, we should design them such that students have the best chance to do it right in as few iterations as possible. 

The other objection is that for a level-1-difficulty exercise like this, I think it's unreasonable to expect students to figure out on their own that they need to refactor to improve the composition. Students at this level of difficulty should still be seeing examples of how to do things right. They don't have the experience yet to look at the problem and write two functions as their first iteration. 

> The goal of the exercise is to teach people how to combine basic string formatting with basic math.

That is a good and worthy goal. What is unreasonable about expanding it to "teach people how to combine basic string formatting with basic math, while demonstrating good program factorization"?

SleeplessByte:
The emoji's were just there so I could refer to them later without having to write the colour.

I don't understand this notion that if we don't have a property in the canonical data that we don't expect it in the program.

Just because the canonical data doesn't have a value property doesn't mean that you should now go and write a program that does a million things in one function.

F3PiX:
Re: the table. I like that the table shows that a band has 2 properties: a value and a position. 
Let's discuss it separately, because if we want to add it, it would make sense to me to add it to the whole series, so starting in Resistor Color 1.

yawpitch:
> > [...] it forces our users to be fluent in vernacular English idioms and to recreate a string that they may not understand and which is mostly a matter of personal preference to begin with.
> 
> This was indeed a problem with the old description property, but I think the label property works around this by reducing the string to only contain essential information. Yes, the text is still English, but would non-native English speakers really have much of a problem reading "33 kilo Ohms?" Kilo is a fairly international word and Ohms is, well, the official unit.

To be clear I'm fine with "33 ohms", "33 kiloohms", or "33 gigaohms" being emitted by a function that addresses the "label" property, as the number and metric unit is certainly international enough to be cognizable to all.

My minor objection to "33 kilo Ohms" or "33 kilo-Ohms" is that both are non-standard representations (at least according to [NIST](https://www.nist.gov/pml/nist-guide-si-chapter-9-rules-and-style-conventions-spelling-unit-names), SI only weighs in on the [ampere](https://www.bipm.org/en/measurement-units/), but note that it spells that unit in lowercase as well, despite it also being a unit derived from a proper name. Adding the hyphen or the space, or indeed title casing the name of the unit ... all of those are examples of English vernacular adaptations to the doubled vowel.

coriolinus:
> Just because the canonical data doesn't have a value property doesn't mean that you should now go and write a program that does a million things in one function.

Agree that students _shouldn't_ do that. However, I assert that if this is an early exercise in the track, it is very likely that students _will_.

I contend that if we write the tests to require only a function which accepts a list of color names and emits a label, then a large fraction of beginner programmers will write only a single function. They may or may not adapt one of the functions that they wrote in previous iterations of the exercise. They are unlikely to import the function they wrote in a previous iteration of the exercise. 

I keep hammering on the separation of concerns because at this level of experience, we should be guiding the students to properly factor their program by the design of the exercise, not delegating this responsibility to the mentors.

SleeplessByte:
This is not a Level 1 "first" exercise.

yawpitch:
> Yah I think either `kΩ` or `kiloohm` are better than `kilo Ohms`. I'm meh on the _no special symbols_ policy. This is not level one. We can provide it in the description. It should be copy-able?

I'd argue against special symbols just because Unicode is non-trivial in some of the languages we support.

F3PiX:
I think it's time to come to a conclusion. The first one being: It doesn't look we can get consensus here.  The question is how we act on that. Here's my proposal. 

In Ruby's Track Anatomy, we really need this exercise, including the string, to display the `kiloohms`. 
And, the Resistor Series is started as an experiment, to see how we can build up a series of exercises, including exploring: what is a meaningful difference from exercise to exercise. Also, related to the concept of progression from the Track Anatomy Project.  
- [x] Returning a 'label' string instead of the original proposal with the sentence is addressing some of the concerns raised, so that is a better solution. Erik made that change already. 
- [ ] It also makes sense for the label to use standard SI notation: I think that would be "1 kiloohm = 1000 ohms". But I'll do some more research to get this exactly right. 
    
Due to the Track Anatomy project, and due to the experiment with the series exercise, we need some room to experiment. All the experiments are being monitored, so if there's reason to revert this exercise, or if it needs to return values to make it useful to more tracks than just a few, we will. There is, just like any other exercise, no reason for a track to implement an exercise that doesn't fit the track.  And the series are not meant to be implemented as a whole. For tracks that don't want to deal with strings in the return value, we already have Resistor Color Duo. That exercise _is_ returning values.
I considered an alternative: implementing this exercise on track level only, but I find 3 reasons not to go there. First,  tracks that have implemented Resistor 2 may be interested in 3 too. Second, the series experiment, as mentioned above, deserves to be more public to gather more experiences from different tracks.  Third, this exercise has a specific role in the tracks that worked through the Track Anatomy beyond level 2, so I really want it to be available for other tracks that are working on the Track Anatomy.   
 
- [ ] If other tracks need a Ohms value returned _plus_ the conversion to kilo's , and since both are not part of Resistor Color 2, we could consider to implement Resistor Color 4 as an alternative to 3, instead of a progression. Meaning: we add the fourth band, make it return values, not strings. It will build upon Resistor Color 2, adding 2 bands with 2 different meanings (the multiplier to get the Ohms value, and the tolerance percentage). 
It was not the intention to use the 4th exercise like that, and I'll have to think about the consequences  for the rest of the series, but if that would solve some problems and concerns about this exercise, and make a useful alternative, it's worth exploring. Then tracks can choose which one fits best, regarding both the language specifics and its Track Anatomy. Let me know if there are tracks that would be interested in this alternative exercise.

There are other fundamental and important concerns and discussions that go beyond the scope of this one exercise. Such as the use of English and special symbols, the nature of the tests (if they should help design or not), the role of canonical data, the relation between mentoring and canonical data and such. I'd vote for discussing those separately. 

TL;DR
Because it doesn't look that we will reach consensus, I'd vote for "let's agree to disagree" and implement the exercise with the original test set up, but including the change to 'label' and SI notation.       

For that matter, I also propose to not merge with only 1 approval, as is the current guard in this repo, but a minimum of 2. 

I do understand this is not an ideal outcome, because it doesn't address, let alone solve all the concerns raised. It's also not meant to cut off the discussion, but rather a proposal to continue that discussion elsewhere.

exercises/resistor-color-trio/canonical-data.json
yawpitch: > I personally think that _forcing description to take "ohms"_ is also a language-specific implementation details. Having a class object with a description _property_ that doesn't take arguments is very reasonable in a lot of languages. Both your arguments are forcing a certain code style (that is functional).

I'm not sure that it does; it's up to the track implementer to decide on function vs class as the organizing structure, and for instance in Python I might reasonably decide that this should be the excercise in which we introduce subclassing of immutables and implement this as:

```python
class ResistorValue(int):
    def __new__(cls, color, *colors):
         pass

    @property 
    def description(self):
        pass
```

Now `description` is perfectly valid and doesn't take ohms as a functional parameter, because the class itself is inherently a value in ohms.

kytrinyx:
To my mind, the above discussion can be summarized as follows:

- the Ruby track needs an exercise that combines a math-y thing with a string-y thing
- this exercise has a whole host of conceptually problematic things about it
- it's worth using it in the Ruby track now as an experiment
- the experiment should be public

I think that because this discussion is public, that satisfies the final requirement.

My vote would be to put this exercise in the Ruby track as a track-level exercise, and consider alternatives for a level 1 exercise that addresses the need in the first point.

yawpitch:
I'd second @kytrinyx and vote for implementing as a track-level exercise, at least for now.

With an SI label string as the return value most of my concerns with student experience are minimized, and I'm glad for that change, but I've still got an issue with that being the expected value of a property in the canonical data. Canonical data _has_ historically been fairly strongly normalised, and this PR still de-normalises an exercise's canonical data intentionally. Merging as is seems premature, as it means pushing what is currently _known_ to be a track-specific need into the canonical description of a general problem. The discussion on what the canonical data is actually _for_ -- which hasn't really been addressed for quite a long time before Track Anatomy became a thing -- _needs_ to be had and a merge _does_ cut off that discussion. I'd prefer to table the merge while that discussion occurs, which a track-specific implementation would do.

Barring that perhaps a "semi-official" moratorium on other tracks implementing this exercise? At least ones not yet moving through Track Anatomy? Backwards incompatible changes to canonical data are painful for maintainers and can be disruptive for students, so the less heads on the hydra the better if that discussion leads to "always normalise canonical data".

yawpitch:
> For that matter, I also propose to not merge with only 1 approval, as is the current guard in this repo, but a minimum of 2.

Strongly second that point; I don't think a single voice on canonical data merges is sensible, given the automatic / semi-automatic generation of tests in some tracks.

ErikSchierboom:
I've updated to use the standardized outputs.

coriolinus:
Implementing as track-specific for Ruby, now, and then continuing to discuss the point of canonical data works for me.

coriolinus:
In my opinion the "host of conceptual problems" are largely tied into what @yawpitch describes as the discussion of what the canonical data is actually for. I don't think there's a problem with having a problem about parsing three-band resistors into their ohm values; I don't think there's a problem with a problem which requires two functions. I quite like the human-readable description that's been arrived at. I _do_ think there's a problem with introducing intentionally denormalized data into the canonical data for an exercise.

The arguments in favor of denormalization appear to me to be:

1. we can revert this if necessary
2. no track needs to implement this exercise
3. mentors can always pick up the slack and teach normalization
4. it is not the goal of the exercise though to teach people how to represent the physical unit best in their languages

The thing is, none of those are actually arguments in favor of denormalization. They would apply equally to the normalized form. 

The arguments in favor of normalization appear to me to be:

1. demonstrate good factorization for students too new to be expected to come up with it on their own
2. decouple the data structure from its human-readable representation
3. depending on mentors to correct the implementation is worse design than arranging the canonical data to support the correct design, even if the test suite is auto-generated without human intervention
4. mentors have quite enough to do already, so it's better for canonical data to be structured in such a way that minimizes their total work.
5. it hurts nothing, and helps the students' learning, to encourage them to represent the physical unit sell in this exercise

These appear to me to be both stronger and more specific arguments, but of course this is a matter of opinion.

---

> This is not a Level 1 "first" exercise.

Really? That is very surprising, to me. In my opinion, this exercise would be appropriate for the second or third week of an entry level programming course. What level would you assign to it, and why?

--- 

I guess I haven't mentioned it yet, but I very much agree that `label`, returning a minimally-formatted string, is a better test than `description`, returning a sentence. 

---

> https://github.com/exercism/problem-specifications/pull/1551#issuecomment-512589784
> 
> value as number (33000) is still an incorrect abstraction.

That's true, but I assert that it's a better abstraction than a string label. If we were to implement future exercises `resistor-color-quad`, `-quint`, `-sext`, we'd almost certainly be demonstrating building up a struct wrapper around an int, implementing math operators on that struct, doing math with error values, etc. In no case would we want to be wrapping a string as the fundamental data type and parsing it on demand. We'd want the struct to contain integer or real values. Therefore, it makes more sense _now_ to return an integer, because it builds more naturally into those potential future exercises. 

---

Let's talk about what the canonical data is for. In my understanding, good canonical data for an exercise should have these properties:

- broadly generalizable across a wide variety of languages
- doesn't constrain students to a single possible implementation
- constrains students enough that mentors know broadly what to expect from student code
- is structured to broadly avoid antipatterns, according to the level of difficulty<sup>1</sup>

1: This is likely to be a contentuous assertion, because it's fairly nebulous. What I mean is that for easy exercises such as this, we should be demonstrating good design by requiring good factorization from the structure of the tests. For difficult exercises, we can simplify the tests, because we can assume the students have better ideas about factorization. 

I would like to reach consensus on these goals in general, and then apply them to this exercise in particular.

yawpitch:
> In no case would we want to be wrapping a string as the fundamental data type and parsing it on demand. We'd want the struct to contain integer or real values. Therefore, it makes more sense _now_ to return an integer, because it builds more naturally into those potential future exercises.

Amen.

> Let's talk about what the canonical data is for. In my understanding, good canonical data for an exercise should have these properties:
> 
>     * broadly generalizable across a wide variety of languages
> 
>     * doesn't constrain students to a single possible implementation
> 
>     * constrains students enough that mentors know broadly what to expect from student code
> 
>     * is structured to broadly avoid antipatterns, according to the level of difficulty1
> 
> 
> 1: This is likely to be a contentuous assertion, because it's fairly nebulous. What I mean is that for easy exercises such as this, we should be demonstrating good design by requiring good factorization from the structure of the tests. For difficult exercises, we can simplify the tests, because we can assume the students have better ideas about factorization.

Those four items comport with my understanding of the purpose of the canonical data, though I would add:

* Doesn't constrain _maintainers_ to a single implementation: the exercise should be moldable to the needs of the track
* Doesn't hard-code the "tags" the exercise is "meant to teach": the track should be free to decide what the exercise brings to the track
* Defines _library_ functionality, not an _app_; ideally each "property" represents a discrete unit that has _one_ concern and _could_ be composed into an application. The actual composition, if any, should be left up to the tracks.

The first is to head off digressions like `33Ω` in Julia vs `"33 ohm"` in Ruby vs `Ohm(33)` in Python. The canonical representation common to all is `33`, and the maintainer can decide if that merits a unit type, a string, or a class.

The second is to expand on the first: a list in the canonical doesn't preclude an enumeration or a hashmap, but a property called "index" to a certain extent makes the exercise "about" lists. Similarly a string in the canonical that encodes ints is suddenly an exercise "about" string interpolation.
 
The third is meant to head off premature diversion into what is ultimately personal preference. If the problem _necessarily_ has multiple concerns then the canonical data, ideally, describes the testable interface of those discrete concerns, and little else, and that describes a library full of units. The moment we're trying to describe an _application_ of those concerns we're stepping into how you'd want to assemble those components using that language and make it something that human users can integrate with and enjoy... honestly that application is best left to the student's imagination.

SleeplessByte:
I'm backing out of this discussion.

I feel that most of the points made here are hiding some secondary agenda or issue with something either around Track Anatomy or this exercise (please don't take this as a personal attack of some sorts, it's not, just an observation and interpretation -- I can be completely wrong, which is why it's a _feeling_).

I don't think this PR is the place to debate what canonical data is and is not; I **do** think we as a team should have the debate, so thank you for starting it. I recommend extracting it from this PR.

Most of the argumentation above is making up scenario's that mighta coulda shoulda happen. There are many exercises with _far worse_ abstractions and canonical data. I don't understand why this is so difficult to accept. I don't understand the primitive obsession with this exercise. I don't understand why there is this meta discussion on what data-type something should be. Generators already decide if a property should be a function or an object or a class. I don't understand why this is any different, but most of all, I don't think the majority of discussion is helping _anyone_.

**Your track does not need to implement this exercise**

@kytrinyx said her p(i)e(a)ce and I think her suggestion makes sense. Let's try this out in Ruby, and go from there. We need to be able to make mistakes too. We need to be able to try these things out instead of coming up with "this might happen if someone does maybe this, in that specific case".

Personally, I think this exercise makes sense in a lot of the progression lines and I'm inclined to implement it _as is_ in JS and TS. The design of the series of exercises around the same story make sense to me. And I don't feel as strongly about what the "right expected output" is. I think our energy is better spent elsewhere. I don't think changes to this exercise are as expensive as some make it sound.

With that said, this back and forth about if a `label` string is better than a number or the other way around is not going anywhere. I personally think this should not have taken 4 days of discussion. Clearly we don't agree and you're bringing in your languages and potential implementation to fuel the argumentation. I don't think it's helpful.

----

Feel free to continue, I'm no longer commenting on any of this. Reach me on slack if you'd like to discuss further/more/something else.

yawpitch:
I am, and have been, working on creating an issue that summarizes the well-intentioned -- if long, and at times difficult to follow -- debate and highlights the key points where consensus obviously does not exist.

That issue will ignore -- in the interests of equanimity -- the last comment, in its entirety.

SaschaMann:
Perhaps making the `value` property `optional` could please "both sides"?

exercises/resistor-color-trio/canonical-data.json
```diff
@@ -0,0 +1,46 @@
...
+      },
+      "expected": "680 ohms"
+    },
+    {
+      "description": "Red and black and red",
+      "property": "label",
+      "input": {
+        "colors": ["red", "black", "red"]
+      },
+      "expected": "2 kiloohms"
```
petertseng: smartly dodged the question of:

> should 1200 ohms be formatted as `1200 ohms` or `1.2 kiloohms`?

nice! as long as there are no inputs that would raise this question, it can be avoided. otherwise the description may have to get a bit more specific about when to apply prefixes. keep this in mind if someone tries to suggest any other inputs.

ErikSchierboom: Good observation! I indeed avoided this situation by carefully picking the input values :)

ErikSchierboom:
> Perhaps making the value property optional could please "both sides"?

Not my side 😛 

Jokes aside, I'd rather not add the optional fields, as I think they wouldn't solve the issue, as the proposed inputs of the various properties are different between the two versions (one version (this PR) has label take an array of strings, whereas the other version would take a number).

SaschaMann:
> (one version (this PR) has label take an array of strings, whereas the other version would take a number)

I missed that. Then I agree it wouldn't work well.

ErikSchierboom:
I've thought about this a bit more, and the best argument against the current canonical data format is that it restricts languages in how they can implement this exercise. Let's see what we can do to try and remedy this.

One proposed solution was to separate the `value` and `label` properties:

```json
{
    "description": "Orange and orange and black value",
    "property": "value",
    "input": {
        "colors": ["orange", "orange", "black"]
    },
    "expected": 33
},
{
    "description": "Orange and orange and black label",
    "property": "label",
    "input": {
        "value": 33
    },
    "expected": "33 ohms"
}
```

While this approach works for languages that'd like to use a library-like approach, it doesn't work as well for the application-like approach. In the latter case, we probably don't want the `value` property at all and use the three strings as input to the `label` property. The `value` properties could be marked as optional and ignored. However, to automatically generate the correct `label` test, the integer `value` property would have to either be reverse engineered back into its original value, which I feel is complexity I don't want to force upon a test generator. Another option is to lookup the corresponding `value` property and find the case where the expected value matches the input value, and then find it input colors property. This is also a situation I am not comfortable with, as it tightly couples canonical test case properties. There are two major issues with this: it will be a breaking change to change of the `value` properties expected values, as well as the fact that for each `label` property there must be a corresponding `value` property working with the exact same data. This means either less variety in the canonical data if we want to keep the number of test cases small, or else having many test cases, both of which are not ideal I feel.

Taking inspiration from some suggested solutions in this thread, what if we encode both the label _and_ the value in the expected output? Something like this:

```json
{
    "description": "Orange and orange and black",
    "property": "label",
    "input": {
        "colors": ["orange", "orange", "black"]        
    },
    "expected": {
        "label": "33 ohms",
        "value": 33
    }
}
```

In this case, I think one could reasonably convert this canonical data to both a library-like approach that uses both a `value` and a `label` property and to an application-like approach that only uses a `label` property. It also opens up the possibility for a track to encode things in a completely different way, e.g. the F# track could convert it to a [unit of measure](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/units-of-measure). What do people think of this approach? 

Another alternative would be to just have the value and have the track be fully responsible for converting it to an appropriate test suite (as suggested earlier):

```json
{
    "description": "Orange and orange and black",
    "property": "value",
    "input": {
        "colors": ["orange", "orange", "black"]        
    },
    "expected": 33
}
```

With this approach, the canonical data is very generic and allows for any implementation. This is also its disadvantage, as it will mean that individual tracks will all have to do a bit of thinking and determine how to best implement the exercise. This is not bad _per se_, but it will make things a bit more difficult for track maintainers. Similarly, the canonical data will now not be strongly linked to a specific purpose. The current PR is designed for a specific purpose in mind, namely to be a relatively basic string concatenation/math combination exercise. While that is still possible with a bare-bones canonical data approach, it might be that in future iterations this won't be apparent anymore and there might be changes to the canonical data that would no longer it make a basic string concatenation/math combination exercise (there is precedence for this). 

One final thought: it might be that this exercise is just incidentally one in which the encoding of the canonical data is a bit problematic, so the actual issue present might not be something that is a sign of a larger issue at hand.

Well, I've now argued for and against my very own proposal and suggested some possible modifications. I'd be interested in hearing everybody's thoughts.

coriolinus:
My ranked preference given those options:

1. Separate the `value` and `label` properties entirely. Tracks which prefer an application-like approach can merge the cases manually.
2. Drop the `label` property, expect the value. However, this adds work for track maintainers who really prefer the application approach. 
3. Encode both `label` and `value` within the expected result. This implies a single tested function producing a tuple containing both of those, which feels like suboptimal factoring. The main point in favor of this approach is that it's the easiest to transform into one of the other forms using a single `jq` filter.

What do you guys think of this: use proposal 1, but include in the comments a `jq` filter which merges the `value` inputs into the `label` outputs, and gets rid of the extraneous `value` inputs? This minimizes the work required for maintainers who strongly prefer the application format.

yawpitch:
> Taking inspiration from some suggested solutions in this thread, what if we encode both the label _and_ the value in the expected output? Something like this:
> 
> ```json
> {
>     "description": "Orange and orange and black",
>     "property": "label",
>     "input": {
>         "colors": ["orange", "orange", "black"]        
>     },
>     "expected": {
>         "label": "33 ohms",
>         "value": 33
>     }
> }
> ```

Encoding the components of a composite return value is the closest thing to what I’d say is the “best compromise between extremes” after #1553, so I’d prefer something like that.  I would propose to tweak your suggestion slightly into something like this:

```
{
    "description": "Orange and orange and black",
    "property": "label",
    "input": {
        "colors": ["orange", "orange", "black"]        
    },
    "expected": {
        “value”: 33,
        "unit": "ohms"
    }
}
```

This way the implementing language can build whatever label that is most appropriate to both the language and where this language gets placed in the progression for that language. As it’s just the final state attributes of a generic representation a language can, relatively easily — and more easily than with an already encoded composite label — make the representation under test be a composite string, a struct or tuple, or whatever object type is most appropriate.

Having the “33 ohms” label in there as a third key would make this easier to implement for those who want this exercise to end in that string, but would make this less generic to the notion of composite return values, which are a generic problem.

ErikSchierboom:
> What do you guys think of this: use proposal 1, but include in the comments a jq filter which merges the value inputs into the label outputs, and gets rid of the extraneous value inputs

And by this you mean have separate properties, right? If so, what do you think about my objections to this approach (tightly-linked, duplicate value/label test case combinations required)? I'm especially hesitant to tightly-link the test cases.

@yawpitch I like the "ohms" unit suggestion:

```json
{
    "description": "Orange and orange and black",
    "property": "label",
    "input": {
        "colors": ["orange", "orange", "black"]        
    },
    "expected": {
        “value”: 33,
        "unit": "ohms"
    }
}
```

> End result is something not ideal for anyone, but which maximizes utility for all sides.

Yep, it is definitely a compromise, but if it's workable for all then maybe it is the best we can do. @coriolinus could you live with this compromise?

coriolinus:
Yes, that compromise works well for me, because the struct fields are orthogonal. We haven't decided what to do about choosing units for larger values. There are two options:

```json
    "expected": {
        "value": 33000,
        "unit": "ohms"
    }
```

```json
    "expected": {
        "value": 33,
        "unit": "kiloohms"
    }
```

The first is more general. The second better fits the original intent of this exercise. I mildly prefer the first, but I am willing to accept either.

coriolinus:
To actually answer the question asked: I agree that tightly linking the test cases isn't great, but it's necessary in order to be able to apply an automatic `jq` transform. I think the proposed compromise is a better idea.

ErikSchierboom:
>  We haven't decided what to do about choosing units for larger values. There are two options:

My preference would also be with the first option, as that is indeed the most general.

yawpitch:
> Yes, that compromise works well for me, because the struct fields are orthogonal. We haven't decided what to do about choosing units for larger values. There are two options:
> 
> ```json
>     "expected": {
>         "value": 33000,
>         "unit": "ohms"
>     }
> ```
> 
> ```json
>     "expected": {
>         "value": 33,
>         "unit": "kiloohms"
>     }
> ```
> 
> The first is more general. The second better fits the original intent of this exercise. I mildly prefer the first, but I am willing to accept either.

My instinct is to go for the more general approach, but in this case it strongly serves the special case that inspired the exercise well to have it be “kiloohms”, so I’d tend to say that approach would be best for this exercise’s needs. Mild tangent: if I read the description correctly three bands can encode up to 99,000,000,000 ohms... shouldn’t we have tests for megaohms and gigaohms as well? Even if only optionally ones? If optional should we consider dekaohms and hectoohms as well, for SI consistency?

ErikSchierboom:
@F3PiX @SleeplessByte what are your thoughts?

coriolinus:
> shouldn’t we have tests for megaohms and gigaohms as well? Even if only optional “bonus” ones? If optional should we consider dekaohms and hectoohms as well, for SI consistency?

That sounds like a good idea to me.

ErikSchierboom:
> That sounds like a good idea to me.

I think that is a good idea. I'll add some optional values tonight, and update the canonical data as well. My thinking now is that @yawpitch has it right, and it perhaps better to use "kiloohms' as the unit.

We're getting there! Thanks everybody for being so responsive and chipping in.

F3PiX:
I'm glad that there's a solution.
I do have 2 questions:
- Is this really easier for maintainers than the thing I proposed: splitting the requirements into 2 different exercises, where this Resistor III returns a label, and Resistor IV is going to return values only `[3300, 20]`? Maybe it's naïve thinking, but I'd think it's easier for maintainers to skip III and implement IV, or the other way around, than having to do the extra work.
- Is it really important that an exercise doesn't fit other tracks? That is news to me, because there are multiple exercises that don't fit the Ruby track. I would never expect that to be a reason not to implement that exercise, _if_ it's expected to be useful to at least 1 or 2 other tracks.

yawpitch:
> We're getting there! Thanks everybody for being so responsive and chipping in.

@ErikSchierboom I do believe we are getting there, but I think this fix might require an update to the canonical schema, since right now **expected** seems to only allow for either a single value or an **error**.

ErikSchierboom:
@yawpitch That won't be necessary, we already have exercises returning objects (e.g. [ETL](https://github.com/exercism/problem-specifications/blob/master/exercises/etl/canonical-data.json#L75)).

yawpitch:
@F3PiX 

>     * Is this really easier for maintainers than the thing I proposed: splitting the requirements into 2 different exercises, where this Resistor III returns a label, and Resistor IV is going to return values only `[3300, 20]`? Maybe it's naïve thinking, but I'd think it's easier for maintainers to skip III and implement IV, or the other way around, than having to do the extra work.

That is, as you say, naive ... sure, splitting the exercise here would eliminate the headache _for now_, but that is just a recipe for fracturing in the future when this comes up again, and again, and _again_. And it _will_. So you'd get an exponential growth in the number of exercises as the complexity of those exercises goes up, and most of the difference between them would be boilerplate. A general solution for encoding the most reduced form of a composite answer is going to reduce that growth and keep the core job of the maintainers (implementing new exercises for their track and maintaining existing exercises as canonical data changes) to a minimum.

>     * Is it really important that an exercise doesn't fit other tracks? That is news to me, because there are multiple exercises that don't fit the Ruby track. I would never expect that to be a reason not to implement that exercise, _if_ it's expected to be useful to at least 1 or 2 other tracks.

_Yes_, to some of us, it _is_ really that important ... this debate wouldn't be happening if that wasn't a strongly held value and priority of at least some of the maintainers, to say nothing of the mentors and students who don't have a voice here.

Broadly speaking _any_ exercise that Ruby can address can be equally addressed by _every_ other imperative (or multi-paradigm) language (which IIRC constitutes a majority, but at the least a very singificant block of, the languages we currently support) ... the only exercises, really, that _do not_ and _cannot_ apply to a language are those that are specific to one of the major paradigms (ie OOP, Functional, Declarative) that a given language _does not_ support. And for each of those (except for wherever Prolog fits) there's overlap with an entire family of languages. Consider for instance Julia to be a language in which _anything_ you want to to teach in Ruby has a necessary and valuable equivalent lesson to be taught, hence an equal amount to gain from the canonical data ... but only if the canonical data isn't hand-crafted to a track-specific outcome.

In fact I'd say that an exercise probably _should not_ have a canonical representation unless and until it's been normalised to a form that applies to its paradigmatic family of exercises ... which means that likely only Prolog can really _not_ think about this issue as important.

yawpitch:
BTW, one of the _key_ and _unique_ value propositions of Exercism, from my point of view as a would-be polyglot programmer, is the ability to attack the same problem in a dozen or more different languages, just to see how those languages approach the same issues. I don't believe that Exercism should become 50+ different CodeAcademy clones ... IMHO that would be best served by promoting to canonical those exercises that both _can_ and -- where reasonably applicable -- _should_ be implemented widely.

Admittedly that's a bias, but it's a bias based on what drew me to Exercism in the first place, _before_ the addition of mentoring. I just don't want to see that value unnecessarily degraded by what I see as a premature and unconsidered willingness to think of the canonical data as a 2 or 3 language specific data store.

F3PiX:
> shouldn’t we have tests for megaohms and gigaohms as well? 

The current test values all exist in real 3 band resistors. 
As far as I could find, gigaohms resistors exist, but with 4 bands, not 3.
megaohms do exist as a 3 band resistor.   

Two options
- We can save the gigaohms for the next exercise in the series, that will handle 4 bands
- We can add the extra cases if it's okay that maybe they don't exist as real resistors.

yawpitch:
> > shouldn’t we have tests for megaohms and gigaohms as well?
> 
> The current test values all exist in real 3 band resistors.
> As far as I could find, gigaohms resistors exist, but with 4 bands, not 3.
> megaohms do exist as a 3 band resistor.
> 
> Two options
> 
>     * We can save the gigaohms for the next exercise in the series, that will handle 4 bands
> 
>     * We can add the extra cases if it's okay that maybe they don't exist as real resistors.

I'd tend to think that, rather like `roman-numerals` and `leap`, we can assume that the result of the exercise need not refer back to a resistor that is actually manufactured. That said it'd be quite reasonable to say in the description.md that there's some upper limit beyond which you simply wouldn't practically use a 3 band resistor, even if you could actually encode values above that limit, and thus we won't test beyond it.

ErikSchierboom:
I've updated the PR to use the new, agreed-upon format for the `expected` value. Also, I've tried to add the optional test cases, but there is no precedent for having optional test cases, and the [existing keys](https://github.com/exercism/problem-specifications/blob/master/OPTIONAL-KEYS.txt) don't really make much sense. We can of course fix this, but I'd prefer doing that in a separate PR when the need arises.

Once again, thanks for the comments!

yawpitch:
> the [existing keys](https://github.com/exercism/problem-specifications/blob/master/OPTIONAL-KEYS.txt) don't really make much sense.

I don't mind this going out without those tests, just thought this was an ideal case for the optional flag. As I understood it that string is just arbitrary though, isn't it? So basically a comment and the only reason for OPTIONAL-KEYS.txt is to keep  those comments from -- potentially -- duplicating out of control? This is such a generic case that "humanized-outputs" or "extended-outputs" would probably cover it.

ErikSchierboom:
Merged! Thanks all. 🎉

ErikSchierboom:
> I don't mind this going out without those tests, just thought this was an ideal case for the optional flag. As I understood it that string is just arbitrary though, isn't it? So basically a comment and the only reason for OPTIONAL-KEYS.txt is to keep those comments from -- potentially -- duplicating out of control? This is such a generic case that "humanized-outputs" or "extended-outputs" would probably cover it.

Yep, it is a great candidate. "humanized-outputs" or "extended-outputs" sounds good. @sshine what do you think would be a good, generic value for this optional value?

sshine:
@ErikSchierboom: I didn't see this comment until now, sorry!

I understand that your question here is whether I think "humanized-outputs" or "extended-outputs" are good candidates for OPTIONAL-KEYS.txt. In general, I don't know. In this particular exercise, I don't think so personally:

If I were to port this exercise to one of the tracks I participate in currently:

- Haskell: Has no test generator, would create a `newtype Resistance = Resistance { ohms :: Int }` definition for the return value and ask the student to provide an `instance Show Resistance where show = ...` that produces something alike that in the canonical tests, and write the test suite manually based on the canonical tests, as is the custom on the track. So, same task, but split into parts.

  Come to think of it, I think this is actually an excellent way to introduce custom `Eq`, `Ord` and `Show` instances rather than derived ones. So a type class-centric exercise. I like it! Also, it will be fun to have an exercise called "-trio" when the two others aren't there. :-D

  This reminds me of a 1990s Danish comedy trio where two of the comedians didn't show up on stage until a while into one sketch, because they were out shopping. The show was called "Where are the two others?" -- I think it might've been a pun on the fact that they got asked that a lot in private.

- OCaml: Has a test generator, would create a data type definition for the return value, and an exercise-specific JSON deserializer for the canonical tests, so that there's a translation between the JSON object in the canonical test and the data type in the OCaml exercise test suite. I would probably accept that values of this data type does not have a canonical form (so testing that 33000 ohms and 33 kiloohms for equality cannot be done with `=` without first normalizing it), or try to achieve something alike the Haskell solution using module-level functors. If it weren't because F# has units of measure, this is probably similar to how I'd do it in F# -- do you think?

- Tcl: Has no test generator (yet), but regardless of this, I would probably return a list `{33 ohms}`. Tcl has coercion between strings and non-nested lists, so you can print or index that without problems. On the whole "but is 33000 ohms the same as 33 kiloohms?" debacle I just made up: I would say that I don't care, because if you cared about such things, you wouldn't use Tcl anyways. :-P

@yawpitch wrote:
> before anyone gets miffed at my picking out Prolog above, it's just the only language I can think of in our list of supported langauges that seems to exist in a paradigm on its own.

Yeah, I think you'd call it logic or constraint programming, and I've been taught to sub-classify it as declarative, just like FP. When I was taught Prolog at school, we were asked not to use the `!` cut operator, as [Wikipedia reiterates](https://en.wikipedia.org/wiki/Prolog): "*Prolog is not purely declarative: because of constructs like the cut operator".

I hope this is not miffed, but I can think of two better examples:

- The [Ballerina](https://ballerina.io) language, which mostly made up [its own exercises](https://github.com/exercism/ballerina/tree/master/exercises). I haven't made an effort to classify Ballerina's programming paradigms, and I don't know if they deviated from canonical exercises because they had to, or because they wanted to.
- I've wanted to make a `jq` track (to learn the damn thing). Since it's a programmable *query* language, I'd hardly expect any overlap with canonical exercises.

coriolinus:
It's tangential to this discussion, but attempting to write canonical data
for exercises query languages such as JQ and SQL could be a pretty
interesting problem.

ErikSchierboom:
Thanks for the detailed response @sshine! I think that the original question you responded to has been partially voided later on, as the structure of the returned data was modified after said question. It's great to hear that you see value in implementing this exercise in other languages.

PR #1552, opened 2019-07-14
Anagram - add test cases for differently-cased words

marko213:
Add two test cases to catch some weird solutions where all words are rejected if the matched words have different case.

Original PR on the cpp track: exercism/cpp#297

### DIFF
--- a/exercises/anagram/canonical-data.json
+++ b/exercises/anagram/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "anagram",
-  "version": "1.4.1",
+  "version": "1.5.0",
   "cases": [
     {
       "description": "no matches",
@@ -53,6 +53,15 @@
         ]
       },
       "expected": ["gallery", "regally", "largely"]
+    },
+    {
+      "description": "detects multiple anagrams with different case",
+      "property": "findAnagrams",
+      "input": {
+        "subject": "nose",
+        "candidates": ["Eons", "ONES"]
+      },
+      "expected": ["Eons", "ONES"]
     },
     {
       "description": "does not detect non-anagrams with identical checksum",
@@ -116,6 +125,15 @@
         "candidates": ["BANANA", "Banana", "banana"]
       },
       "expected": []
+    },
+    {
+      "description": "words other than themselves can be anagrams",
+      "property": "findAnagrams",
+      "input": {
+        "subject": "LISTEN",
+        "candidates": ["Listen", "Silent", "LISTEN"]
+      },
+      "expected": ["Silent"]
     }
   ]
 }


exercises/anagram/canonical-data.json
```diff
@@ -109,13 +118,13 @@
       "expected": []
     },
     {
-      "description": "words are not anagrams of themselves (case-insensitive)",
+      "description": "words are not anagrams of themselves (case-insensitive), mixed",
```
kotp: Since the test description has changed, perhaps this is a new test, not a replacement test.

marko213: There should be a test which excludes all possible capitalizations of the input string and matches at least one valid anagram at the same time - adding another test seemed a bit  repetetive, as the changed test already performed the first part.

kotp: But there shouldn't be a test that covers the candidates that are mixed case of the original word and expects no match?

marko213: The modified should perform the same function, shouldn't it? Or do you think that just the same word with different capitalizations could be an edge case in someone's code and should be tested for separately?

petertseng:
The pull request description has a good description because it describes the reason for the change. The description should be included in the commit message as well; otherwise the commit message only says "modify test cases" which is not helpful for understanding at all.

exercises/anagram/canonical-data.json
kotp: I am saying, basically, that this test was changed in regards to the expected as well, which indicates the the test that is being replaced was unnecessary, or testing the wrong thing.

I think adding the test and later deciding that the test concerning "bananas" is not needed is a better story line.

kotp: It feels like a new test, rather than a modification of the test.

marko213: Fair enough, I guess I'll create another test and only try checking for two case variants of the input.

Should I force-push this over the previous commit?

kotp: Feel free to force push on your branches for the pull requests, as it helps keep things clean.

SleeplessByte: It doesn't really matter. We'll squash merge. Do what you'd like; keeping the changes clean is helpful but in this case probably just fine anyway :)

kotp: @SleeplessByte I don't always squash merge, it depends on how the history presented in the commits.  I didn't now that it was unconditional that we would do that.

SleeplessByte: I also don't _always_ squash merge, but I think for about 95% of all PRs across most exercism repositories, squashing make sense ;)

In this case: adding a set of test cases in a single file usually doesn't really benefit from having more "history" than just a single commit, agreed?

kotp: Yeah, which I was thinking they could force push.  I thought that was why the force push might be asked about, otherwise it would probably be a normal push, and I encourage changing the history locally, so we don't have to remember to need to here remotely.

SleeplessByte: Got ya! I'm striving to have people squash merge, for the simple reason that it also adds a link back to this PR!

kotp: That's a nice feature.  I didn't know that happens.

PR #1557, opened 2019-07-22
Word count - add test case for substrings from the beginning of words

marko213:
This should catch cases when only up to either the first word's or second word's length of characters are compared, counting substrings from the beginning under the same word.

[Original thread in the track-c channel on Slack](https://exercism-team.slack.com/archives/CATCU7LSJ/p1563419221002800)


yawpitch:
Two approvals and no dissent, I'll merge.

PR #1559, opened 2019-07-23
Linked List: adding 'delete' to required methods

jocelo:
Adding 'delete' to the list of methods that the object should have; in the assertions the delete method is expected to be part of the object.

### DIFF
--- a/exercises/linked-list/description.md
+++ b/exercises/linked-list/description.md
@@ -18,6 +18,7 @@ offers an array-like interface for adding and removing items:
 * `pop` (*remove value at back*);
 * `shift` (*remove value at front*).
 * `unshift` (*insert value at front*);
+* `delete` (*remove any node by value*);
 
 To keep your implementation simple, the tests will not cover error
 conditions. Specifically: `pop` or `shift` will never be called on an


petertseng:
There are a few possible meanings for `delete`, so first let me list them out so we can get on same page

* delete: Given a way to refer to a node, delete that node from the list containing it, in O(1) time.
* delete: Given a value and a list, search the list for a node containing that value in O(n) time, then delete it (as above).
* delete: Given an index and a list, advance that many positions into the list in O(n) time, and delete it (as above)

When I read a description of "remove any node by value", it seems to me to imply the second possibility for `delete` I mentioned above. If that is as intended, then great. If not, let's decide which other possibility is intended and have a description that designates that particular possibility.

A note for tracks implementing this exercise (noting that we have no canonical-data.json) that if deletion is added, it will probably be wise to test deletion from the head of the list, from the tail of the list, and from the middle. Take this into account when deciding whether to add delete.

I will make my recommendation in a next comment.

petertseng:
Personally I'll recommend against adding any of the three above possibilites of `delete`.

For "delete, given a way to refer to a node", I recommend against with preference strength 3/10. Note that this is the first time we allow a client of this list to refer to a node. In all other existing operations, they only have to be able to refer to a list and can remain completely ignorant that any nodes exist at all. It's not inherently bad to give the client more knowledge, but it will require more work on the part of students + mentors to make sure the interface thus provided makes sense.

For "delete, given a value and a list" and "delete, given an index and a list" I will recommend against with preference strength 1/10 because it "feels bad" to add an O(n) operation to this list. There's nothing wrong with it since these are possibly things that someone would want to be able to do to a list, so I did not give it any higher strength. After all, if the client uses these variants of `delete`, knowing full well they are O(n) and that is what they want, then I can't deny them what they want.

kytrinyx:
Hello. Thank you for this PR - it's really appreciated.

This repo is currently on "lock-down" while we think through a few things. As such, I'm going to put this PR on hold for now, and we will come back to once we've worked through the issues with this repo that we need to address. You can read more [here](https://github.com/exercism/problem-specifications/issues/1560).

Sorry that you've hit us at this momentarily awkward juncture, and thank you again for the contribution.

jocelo:
Hi @petertseng thanks for your in-dept comment about this. From my short experience with this project, I have come to know that more "advanced" topics are not part of any discussion when mentoring students (advanced as in time/space complexity, recursion, best practices). Other mentors have always pointed out that we should maintain a non-strict posture when helping students. Your comments about algorithm complexity makes a lot of sense and I like the idea that this topics are in the discussion board, even if they are not part on the current state of the project. I'll be more than happy to continue contributing to this project as much as I can!

@kytrinyx don't worry about it and many thanks for letting me know about the lock down.

This is an awesome project and I enjoy the community that's forming around it.

ErikSchierboom:
@jocelo Are you still interested in working on this PR?

ErikSchierboom:
I agree. I'll close. Thanks for creating the PR!

jocelo:
Back at you!

Issue #1560, opened 2019-07-24
[NOTICE] This repo is temporarily only accepting bug-fixes.


iHiD:
Over the last few months there have been a number of passionate debates in this repository about new exercises and changes to the existing exercises. Maintainers often have very opposed views and cannot reach a consensus. This has lead to lots of frustration from people who all want to make Exercism better, but disagree on how that should be done.

This is a clear sign that the product team have not done enough to communicate guidelines around how exercises should be structured and how this repository should be structured. The reason for this is that the product team don't know how exercises should be structured or how the data in this repository should be structured.

This repository (and exercises on Exercism in general) have evolved organically without much structure or thought. For V1, this was generally ok, but since V2, where exercises are part of learning pathways and where everything is a bit more complicated in general, this has now become a real problem. We in the product team therefore need to go away and do some proper work around understanding how exercises should evolve in Exercism. Once we've got some clarity on that, we can then start to put together some clarity around how this repository should work, and everyone can move forward, hopefully without the uncomfortable conversations that have been ongoing recently.

Therefore, for the time being, we are closing this repository to any new exercises, or any changes-of-substance to the canonical data. We will still accept bugfixes (if tests are wrong) and improvements to the copy around exercises, as long as they do not alter the substance of an exercise. If you would like to add a new exercise to your track, please do it as a track-specific exercise for now, and we can discuss generalising it into this repository further down the line.

I want to finish with a personal apology that I haven't realised this a bit earlier and stepped in before people ended up clashing over things. I'm excited about getting this to a really good place and making both the Tracks the best they can be, and ensuring that maintainers lives are as easy as possible.

kytrinyx:
> If you would like to add a new exercise to your track, please do it as a track-specific exercise for now, and we can discuss generalising it into this repository further down the line.

I want to second this. We are absolutely open to the tracks evolving!

Once we have a better handle on the product side of this, we can start working on generalizing any of the experiments that end up in individual tracks.

Of course: if there's an exercise that goes into one track, there's no reason why maintainers can't collaborate to put the exercise in multiple repos, but still at a track-specific level.

❤️ Thanks @iHiD

sshine:
> for the time being, we are closing this repository to any new exercises, or any changes-of-substance to the canonical data. We will still accept bugfixes (if tests are wrong) and improvements to the copy around exercises, as long as they do not alter the substance of an exercise

[...]

PR #1563, opened 2019-07-31
Change test case order to be TDD orientated.

whatcoda:
I have rearranged the tests for this exercise to be more test driven development orientated as we would potentially want the easier test cases to come first so the user doing the exercise can write code to fail fast (eg. negative coins, 0 coins, etc) then build up to the difficult scenarios of using Lilliputian and Lower Elbonia Coins.

This is just a suggestion as I found the exercises as a great way of doing TDD - albeit the tests having already been wrote!

### DIFF
--- a/exercises/change/canonical-data.json
+++ b/exercises/change/canonical-data.json
@@ -8,6 +8,33 @@
   ],
   "cases": [
     {
+      "description": "no coins make 0 change",
+      "property": "findFewestCoins",
+      "input": {
+        "coins": [1, 5, 10, 21, 25],
+        "target": 0
+      },
+      "expected": []
+    },
+    {
+      "description": "cannot find negative change values",
+      "property": "findFewestCoins",
+      "input": {
+        "coins": [1, 2, 5],
+        "target": -5
+      },
+      "expected": {"error": "target can't be negative"}
+    },
+    {
+      "description": "error testing for change smaller than the smallest of coins",
+      "property": "findFewestCoins",
+      "input": {
+        "coins": [5, 10],
+        "target": 3
+      },
+      "expected": {"error": "can't make target with given coins"}
+    },
+    {
       "description": "single coin change",
       "property": "findFewestCoins",
       "input": {
@@ -24,24 +51,6 @@
         "target": 15
       },
       "expected": [5, 10]
-    },
-    {
-      "description": "change with Lilliputian Coins",
-      "property": "findFewestCoins",
-      "input": {
-        "coins": [1, 4, 15, 20, 50],
-        "target": 23
-      },
-      "expected": [4, 4, 15]
-    },
-    {
-      "description": "change with Lower Elbonia Coins",
-      "property": "findFewestCoins",
-      "input": {
-        "coins": [1, 5, 10, 21, 25],
-        "target": 63
-      },
-      "expected": [21, 21, 21]
     },
     {
       "description": "large target values",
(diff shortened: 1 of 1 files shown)


yawpitch:
I'm not exactly against this change, but I'm also not at all convinced that the new order is any _more_ TDD than the old order. Seems to me that it would be perfectly sensible to write the old order, based on the following user stories:

1. We need something that can tell Bob what coin to give back if the change can be given with a single coin.
2. We need something that can tell Bob what coins to give back if the change can only be given with multiple coins.
3. Management called, we're opening a store in Lilliput. Go figure.
4. Management called again, we're opening a store in Lower Elbonia... apparently Bob's from there.
5. Someone asked Bob a stupid question ... we need to tell him not to give any change back for no cash tendered.
6. Oh plotz! We never ran out of pennies before!
7. How do you make change for negative money? Bob ... look, Bob ... is this how we ran out of pennies?

Basically the three tests you've brought forward are edge cases for unlikely, unexpected, and effectively invalid inputs ... it's natural for those to be added in later cycles of TDD, as they're discovered. The tests you've moved to the end, though, test the essential functionality of the normal case(s).

whatcoda:
Thanks for the feedback. I do agree on the ordering in the view of the user stories you provided.

Just to justify my reasoning, I'm coming more from a difficulty perspective as these exercises are to help people learn/hone their skills in a programming language.

I was doing the exercise in Node and at first I though the exercise was easy by just keep using the highest denominator coin from the start. When I got to the Lilliputian and Lower Elbonia coins tests, the difficulty ramped up. After tacking those, the negative and invalid scenarios were trivial. That is why I moved the Lilliputian and Lower Elbonia coins tests to the end as where the real challenge lay.
Whilst looking at other submissions of solutions I noticed some fell through the entire code for the negative and zero scenarios. Which, while worked, simple conditionals at the beginning to fail fast on these scenarios would be much more efficient.

yawpitch:
> Thanks for the feedback. I do agree on the ordering in the view of the user stories you provided.
> 
> Just to justify my reasoning, I'm coming more from a difficulty perspective as these exercises are to help people learn/hone their skills in a programming language.
> 
> I was doing the exercise in Node and at first I though the exercise was easy by just keep using the highest denominator coin from the start. When I got to the Lilliputian and Lower Elbonia coins tests, the difficulty ramped up. After tacking those, the negative and invalid scenarios were trivial. That is why I moved the Lilliputian and Lower Elbonia coins tests to the end as where the real challenge lay.
> Whilst looking at other submissions of solutions I noticed some fell through the entire code for the negative and zero scenarios. Which, while worked, simple conditionals at the beginning to fail fast on these scenarios would be much more efficient.

The reasoning is understandable, but additions to the canonical tests happen all the time when edge cases are discovered. For many of these the new test would by necessity be a trivial fix, but we've been appending the new test to the end rather than trying to maintain some sort of canonical order ... on any new test it's going to end up being a judgement call (or, really, a coin toss) as to the "correct" new order, so it's seemed simpler to merely append.

Also _not_ assigning a fixed order is inline with the usual recommendation that tests _in no way_ depend on each other, which means that they should, really, be able to be run in any order at any time and still provide meaningful assurance that the code still works.

Lastly, the test runners in many languages simply won't respect the order we present here in the canonical data anyway. For example in Python they'll end up running in alphabetical order based on the test case name, while in other languages they may or may not appear in presentation order ... it's really up to the language's test runner. Thus trying to maintain a specific order is going to be problematic at best.

Personally I'm inclined to say we don't even try, though I'm certainly open to counter-arguments. At any rate the canonical data is essentially locked at the moment per #1560, and since there's no bugfix here I'm thinking we should probably close this one for now.

Is that okay with you? I am sorry for the work you put in; I don't disagree with your reasoning, I just think we can't easily apply it to all the tracks via the canonical data.

iHiD:
Hello. Thank you for this PR - it's really appreciated.

This repo is currently on "lock-down" while we think through a few things. As such, I'm going to put this PR on hold for now, and we will come back to once we've worked through the issues with this repo that we need to address. You can read more [here](https://github.com/exercism/problem-specifications/issues/1560).

Sorry that you've hit us at this momentarily awkward juncture, and thank you again for the contribution.

ErikSchierboom:
@whatcoda Are you still interested in working on this PR?

whatcoda:
No thank you, I'll close to clean up. Thank you everyone for the feedback!

PR #1565, opened 2019-08-05
raindrops: simplify description.md

yawpitch:
Currently the description.md implies a solution where the student generates _all_ factors of a number and then tests for membership within that set; that's a frustrating and unnecessary wild good chase for a beginner, and I've had multiple students trip up on it. This updates to be a little less ambiguous and to add some informational links.

### DIFF
--- a/exercises/raindrops/description.md
+++ b/exercises/raindrops/description.md
@@ -1,4 +1,4 @@
-In this variant of the children's game [Fizz Buzz](https://en.wikipedia.org/wiki/Fizz_buzz) -- which has often been used as a screening exercise in technical job interviews -- your task is to convert an number into a string that contains raindrop sounds corresponding to certain potential factors. A factor is a number that evenly divides into another number, leaving no remainder. The simplest way to test if a one number is a factor of another is to use the [modulo operation](https://en.wikipedia.org/wiki/Modulo_operation).
+Your task is to convert a number into a string that contains raindrop sounds corresponding to certain potential factors. A factor is a number that evenly divides into another number, leaving no remainder. The simplest way to test if a one number is a factor of another is to use the [modulo operation](https://en.wikipedia.org/wiki/Modulo_operation).
 
 The rules of `raindrops` are that if a given number:
 
--- a/exercises/raindrops/description.md
+++ b/exercises/raindrops/description.md
@@ -5,7 +5,7 @@ The rules of `raindrops` are quite simple; if a given number:
 - has 3 as a factor, add 'Pling' to the result.
 - has 5 as a factor, add 'Plang' to the result.
 - has 7 as a factor, add 'Plong' to the result.
-- _does not_ have any of 3, 5, or 7 as a factor, the result should be the digits of number.
+- _does not_ have any of 3, 5, or 7 as a factor, the result should be the digits of the number.
 
 ## Examples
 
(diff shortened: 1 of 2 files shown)


exercises/raindrops/description.md
```diff
@@ -1,16 +1,14 @@
...
-- If the number has 5 as a factor, output 'Plang'.
-- If the number has 7 as a factor, output 'Plong'.
-- If the number does not have 3, 5, or 7 as a factor,
-  just pass the number's digits straight through.
+The rules of `raindrops` are quite simple; if a given number:
+
+- has 3 as a factor, add 'Pling' to the result.
+- has 5 as a factor, add 'Plang' to the result.
+- has 7 as a factor, add 'Plong' to the result.
+- _does not_ have any of 3, 5, or 7 as a factor, the result should be the digits of number.
```
cmccandless:
```suggestion
- _does not_ have any of 3, 5, or 7 as a factor, the result should be the digits of the number.
```

exercises/raindrops/description.md
```diff
@@ -1,16 +1,14 @@
-Convert a number to a string, the contents of which depend on the number's factors.
+In this variant of the children's game [Fizz Buzz](https://en.wikipedia.org/wiki/Fizz_buzz) -- which has often been used as a screening exercise in technical job interviews -- your task is to convert an number into a string that contains raindrop sounds corresponding to certain potential factors. A factor is a number that evenly divides into another number, leaving no remainder. The simplest way to test if a one number is a factor of another is to use the [modulo operation](https://en.wikipedia.org/wiki/Modulo_operation).
```
SaschaMann: > The simplest way to test if a one number is a factor of another is to use the [modulo operation](https://en.wikipedia.org/wiki/Modulo_operation).

That pretty much gives away the solution.

yawpitch: It does give away how to determine if ie 3 is a factor of number, but it doesn't give away how to respond to that with "pling" ... in a sense forcing the student to discover _how_ to do modulo in their language is, I think, more important than figuring out _that_ they should use modulo. Given the position this is likely to take in a track it seems best to just tell them what operator / function to look for.

exercises/raindrops/description.md
```diff
@@ -1,16 +1,14 @@
...
 
-- If the number has 3 as a factor, output 'Pling'.
-- If the number has 5 as a factor, output 'Plang'.
-- If the number has 7 as a factor, output 'Plong'.
-- If the number does not have 3, 5, or 7 as a factor,
-  just pass the number's digits straight through.
+The rules of `raindrops` are quite simple; if a given number:
```
kytrinyx:
```suggestion
The rules of `raindrops` are that if a given number:
```

I hesitate to use "simple" or "obvious" in the description of any Exercism exercise, because if someone finds themself stumbling, it can make them feel stupid (which would be unintentional, and also a shame).

exercises/raindrops/description.md
kytrinyx:
```suggestion
Your task is to convert an number into a string that contains raindrop sounds corresponding to certain potential factors. A factor is a number that evenly divides into another number, leaving no remainder. The simplest way to test if a one number is a factor of another is to use the [modulo operation](https://en.wikipedia.org/wiki/Modulo_operation).
```

The FizzBuzz reference is accurate, but it's not the most important thing about the exercise, so I don't want this to be the opening sentence.

I love the more specific reference though. Can we add it to the YAML file instead as the source? That would bring it into the final README.

https://github.com/exercism/problem-specifications/blob/6823334bbbef8e6193c09483fb3e5865f5f87a2b/exercises/raindrops/metadata.yml#L3

exercises/raindrops/description.md
yawpitch: Fair enough, fixed.

exercises/raindrops/description.md
yawpitch: @kytrinyx happy to move it to the metadata.yml, but would that mean updating the source_url to the wikipedia link, or leaving in jumpstartlab.com? From the WP link FizzBuzz has been a children's game for teaching math since at least 2002, and its use as a programming interview question seems to have started in 2007 (https://imranontech.com/2007/01/24/using-fizzbuzz-to-find-developers-who-grok-coding/).

yawpitch: @kytrinyx pushed an update of the metadata.yml ... are those changes sufficient?

kytrinyx: Yeah, I like the new `source`! Much better.

Issue #1572, opened 2019-08-18
matrix: non-square matrices are not addressed


cmccandless:
The [exercise description](https://github.com/exercism/problem-specifications/blob/master/exercises/matrix/description.md) never states that only square matrices will be considered, yet the tests in the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/matrix/canonical-data.json) only cover 
square matrices.

One of the following solutions is required:
- Add a statement to the description that only square matrices will be considered (*Recommended; see below*)
- Add tests for non-square matrices

As this exercise tends to occur earlier in many track progressions, adding tests that complicate the exercise further is inadvisable. It would be preferable to simply clarify the problem statement to the student without making the exercise more difficult.

*This issue was brought to attention in exercism/python#1907*

sshine:
I don't think it's a principal problem that the exercise description doesn't mention rectangular matrices. There are, after all, [many ways](https://en.wikipedia.org/wiki/List_of_matrices) to characterize a matrix that we don't use.

I'm not sure why the exercise should be limited to only square matrices: Just because the tests do so now, will the exercise become any harder if there is a different number of rows than columns? And will it become any easier if we state that only square matrices are considered? We also make unwritten assumptions about what "numbers" are: Are they integers? positive integers? non-negative integers? Will the exercise get any easier if we state any assumptions here?

If you make the assumption that you know the number of rows because you know the number of columns, that assumption is unwarranted, and you would have to handle the error case when the assumption is wrong, rather than allow for an arbitrary amount of rows. So without having the assumption stated explicitly, the exercise is easier to solve without making it implicitly.

In general: You omit a third option: Leave the exercise under-specified as it is.

To deal with the origin of the problem, being that a particular kind of solution written with Python and numpy serializes the input into differently shaped data structures, I would suggest either:

* Decide if adding a non-square matrix to the test really does complicate the exercise. I may fail to see this. My impression is that as long as the matrices are not rugged (all rows having the same length), one would use the same assumptions and techniques scanning and representing square and rectangular matrices. If not, add another canonical test case. To tracks for which this matters, this is a benefit. To tracks for which this does not, the exercise has not become harder.

* Add a track-specific test, since the origin of the concern is rather implementation-specific.

yawpitch:
[...]

PR #1576, opened 2019-08-28
Matrix: Add test case

DagmarTimmreck:
In javascript the following wrong implementation passes all previous test cases but not the new one
```
export class Matrix {
  constructor(string) {
    this._rows = string
      .split('\n')
      .map(row => row.split(' ').map(Number))
    this._columns = this._rows
      .map((_, i) => this._rows.map(row => row[i]))
  }

  get rows() {
    return this._rows
  }

  get columns() {
    return this._columns
  }
}
```

### DIFF
--- a/exercises/matrix/canonical-data.json
+++ b/exercises/matrix/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "matrix",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "cases": [
     {
       "description": "extract row from one number matrix",
@@ -64,6 +64,15 @@
         "index": 3
       },
       "expected": [3, 6, 9, 6]
+    },
+    {
+      "description": "can extract column from non-square matrix with more columns than rows",
+      "property": "column",
+      "input": {
+        "string": "1 2 3\n4 5 6",
+        "index": 3
+      },
+      "expected": [3, 6]
     },
     {
       "description": "extract column where numbers have different widths",
--- a/exercises/matrix/canonical-data.json
+++ b/exercises/matrix/canonical-data.json
@@ -30,13 +30,13 @@
       "expected": [10, 20]
     },
     {
-      "description": "can extract row from non-square matrix",
+      "description": "can extract row from non-square matrix with no corresponding column",
       "property": "row",
       "input": {
         "string": "1 2 3\n4 5 6\n7 8 9\n8 7 6",
-        "index": 3
+        "index": 4
       },
-      "expected": [7, 8, 9]
+      "expected": [8, 7, 6]
     },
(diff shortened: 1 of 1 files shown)


DagmarTimmreck:
See also https://github.com/exercism/javascript/pull/718

kotp:
Related to #1572 I believe.

sshine:
Your current test has 3 columns, 2 rows, and asks to extract the 3rd column.

Would it make sense to make the transposed test also?

2 columns, 3 rows, and ask to extract the 3rd row.

yawpitch:
Existing tests are for a 3x4, why not just duplicate those tests but for a 4x3?

Also per #1560 the repo is locked to all but bug-fixes; does this constitute a bug, since we don’t appear to have ever specified the correct behavior for a non-square matrix?

sshine:
@yawpitch: If this PR should be exempted from the lockdown, I would argue that it's because of track-specific coverage issues: If someone presents a solution in a language that passes current tests but contains a bug, I'd consider that a bug. But perhaps @iHiD should decide, since he's the one who announced the lockdown and knows its boundaries best.

DagmarTimmreck:
I don't know if the proposed changes solve all issues with the wrong Python solutions.

But I think, that the symmetric test cases describe the desired outcome of the exercise. (Although the structure of the input makes it nearly impossible to accidentally write a solution that does not handle a row without corresponding column correctly.)

iHiD:
>  But perhaps @iHiD should decide, since he's the one who announced the lockdown and knows its boundaries best.

Does this extra test change the difficulty of the exercise? That would be the yardstick I would use. If it just increases coverage without making the exercise noticeably harder (and this can be presumed to be true across tracks) then I'm happy for it to be merged if the other reviewers approve it based on content.

sshine:
So far they address two tracks for which faulty solutions will no longer succeed.

@cmccandless proposed that testing non-square matrices might complicate the exercise in #1572, but I think this was a hypothetical concern coming from the fact that we don't actually know. I would counter-argue that because we already *have* tests that involve non-square matrices, adding more should not complicate the exercise.

DagmarTimmreck:
I think @cmccandless may have meant 'ragged' or not completely filled  matrices. These would be complicated to model in some languages, I supposed. But rectangular matrices with different numbers of rows and columns are already covered by the tests.

The alternative would be to remove all test cases for non-square matrices which makes the exercise easier.

DagmarTimmreck:
Is there an action required of me or @cmccandless?

kotp:
@DagmarTimmreck I believe it was just awaiting a third party to review and merge.

Thank you, Peeps!

PR #1577, opened 2019-09-02
Matrix: Fix input string for test case 'can extract column from non-s…

DagmarTimmreck:
…quare matrix with no corresponding row'

PR #1578, opened 2019-09-03
matrix: fix whitespace issue

yawpitch:
Fix whitespace issue introduced by #1576, closes #1577.

Not sure if the extra whitespace is breaking anything yet, but best to get it fixed as quickly as possible and no telling how soon @DagmarTimmreck will be back at desk.

### DIFF
--- a/exercises/matrix/canonical-data.json
+++ b/exercises/matrix/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "matrix",
-  "version": "1.2.0",
+  "version": "1.2.1",
   "cases": [
     {
       "description": "extract row from one number matrix",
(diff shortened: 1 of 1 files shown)


exercises/matrix/canonical-data.json
```diff
@@ -1,6 +1,6 @@
 {
   "exercise": "matrix",
-  "version": "1.2.0",
+  "version": "1.2.1",
```
cmccandless: @sshine was requesting 1.3.0 in #1577. Is there a strong argument for 1.2.1?

yawpitch: Bumped to 1.3.0; I don't think there's an argument for either, but the minor version bump is fine.

DagmarTimmreck:
Thanks for taking over @yawpitch

exercises/matrix/canonical-data.json
sshine: [Test data versioning](https://github.com/exercism/problem-specifications#test-data-versioning) suggests for MAJOR.MINOR.PATCH that MINOR should be used when "Changing the test cases inputs and/or outputs." Whereas with "PATCH changes would never break well-designed test generators, because the test data remains exactly the same."

So while you'd assume that any robust matrix parser will handle a little extra whitespace without a glitch, the test data still does not remain exactly the same.

PR #1582, opened 2019-09-08
Square-Root: Add new exercise

wolf99:
An exercise to calculate the square root of natural numbers.
The tests allow an implementation that is simplified because they do only require natural roots less than 256.

https://github.com/exercism/c/pull/408

### DIFF
--- a/exercises/square-root/description.md
+++ b/exercises/square-root/description.md
@@ -2,6 +2,8 @@ Given a natural radicand, return its square root.
 
 Check out the Wikipedia pages on [square root](https://en.wikipedia.org/wiki/Square_root) and [methods of computing square roots](https://en.wikipedia.org/wiki/Methods_of_computing_square_roots).
 
-Recall that natural numbers are positive real whole numbers (i.e. 1, 2, 3 and up).
+Note that the term "radicand" refers to the number for which the root is to be determined. That is, it is the number under the root symbol.
+
+Recall also that natural numbers are positive real whole numbers (i.e. 1, 2, 3 and up).
 
 When implementing this in a language with a built-in square root function, implement your own function.
--- a/exercises/square-root/description.md
+++ b/exercises/square-root/description.md
@@ -1,0 +1,7 @@
+Given a natural radicand, return its square root.
+
+Check out the Wikipedia pages on [square root](https://en.wikipedia.org/wiki/Square_root) and [methods of computing square roots](https://en.wikipedia.org/wiki/Methods_of_computing_square_roots).
+
+Recall that natural numbers are positive real whole numbers (i.e. 1, 2, 3 and up).
+
+When implementing this in a language with a built-in square root function, implement your own function.
--- a/exercises/square-root/description.md
+++ b/exercises/square-root/description.md
@@ -1,9 +1,7 @@
 Given a natural radicand, return its square root.
 
-Check out the Wikipedia pages on [square root](https://en.wikipedia.org/wiki/Square_root) and [methods of computing square roots](https://en.wikipedia.org/wiki/Methods_of_computing_square_roots).
-
 Note that the term "radicand" refers to the number for which the root is to be determined. That is, it is the number under the root symbol.
 
-Recall also that natural numbers are positive real whole numbers (i.e. 1, 2, 3 and up).
+Check out the Wikipedia pages on [square root](https://en.wikipedia.org/wiki/Square_root) and [methods of computing square roots](https://en.wikipedia.org/wiki/Methods_of_computing_square_roots).
 
-When implementing this in a language with a built-in square root function, implement your own function.
+Recall also that natural numbers are positive real whole numbers (i.e. 1, 2, 3 and up).
--- a/exercises/square-root/metadata.yml
+++ b/exercises/square-root/metadata.yml
@@ -1,0 +1,4 @@
+---
+blurb: "Given a natural radicand, return its square root."
+source: ""
+source_url: ""
--- a/exercises/square-root/metadata.yml
+++ b/exercises/square-root/metadata.yml
@@ -1,4 +1,4 @@
 ---
 blurb: "Given a natural radicand, return its square root."
-source: ""
-source_url: ""
+source: "wolf99"
+source_url: "https://github.com/wolf99"
--- a/exercises/square-root/metadata.yml
+++ b/exercises/square-root/metadata.yml
@@ -1,4 +1,4 @@
 ---
 blurb: "Given a natural radicand, return its square root."
 source: "wolf99"
-source_url: "https://github.com/wolf99"
+source_url: "https://github.com/exercism/problem-specifications/pull/1582"
--- a/exercises/square-root/canonical-data.json
+++ b/exercises/square-root/canonical-data.json
@@ -1,0 +1,54 @@
+{
+  "exercise": "square-root",
+  "version": "1.0.0",
+  "cases": [
+    {
+      "description": "root of 1",
+      "property": "squareRoot",
+      "input": {
+        "radicand": 1
+      },
+      "expected": 1
+    },
+    {
+      "description": "root of 4",
+      "property": "squareRoot",
+      "input": {
+        "radicand": 4
+      },
+      "expected": 2
+    },
+    {
+      "description": "root of 25",
+      "property": "squareRoot",
+      "input": {
+        "radicand": 25
+      },
+      "expected": 5
+    },
+    {
+      "description": "root of 81",
+      "property": "squareRoot",
+      "input": {
+        "radicand": 81
+      },
+      "expected": 9
+    },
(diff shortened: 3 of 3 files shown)


exercises/square-root/metadata.yml
```diff
@@ -0,0 +1,4 @@
+---
+blurb: "Given a natural radicand, return its square root."
+source: ""
```
rpottsoh: Cite yourself if it is your own original work.  Got a URL or blog etc?

wolf99: Original work, but I do not have any site. Will add my GitHub profile if acceptable?

sshine:
Isn't this repo still only accepting bugfixes and improvements?

exercises/square-root/description.md
```diff
@@ -0,0 +1,7 @@
+Given a natural radicand, return its square root.
```
sshine: Maybe the word "radicand" needs clarification in one way or another.

I didn't know this term and felt kinda stupid. :-)

Maybe we should simply write "natural number"?

After all, the text below implies that those terms can be interchanged.

```suggestion
Given a natural number, return its square root.
```

wolf99: Can add a description of the term to the `readme.md` instead. Removing radicand means removing it from the `canonical-data.json` also, where the alternatives are clumsy, not very descriptive or not very distinguishable from any other numerical method argument  (e.g. `number`, `naturalNumber`, `input`, etc)

wolf99:
Whoops, hadn't noticed that the repo is not accepting new exercises.
Can this PR be put on hold or something?

exercises/square-root/description.md
sshine: Just as fine by me.

sshine:
> Whoops, hadn't noticed that the repo is not accepting new exercises.
> Can this PR be put on hold or something?

Sure, why not! We'll add [WIP] and [ON HOLD] to remark this.

petertseng:
An interesting question: Is the intent that the canonical tests will only have perfect squares as inputs?

There could be a lot of follow-up questions to this one (such as "is it recommended that implementing tracks have the function return a value of integral type?", "should the behaviour on inputs that are not perfect squares be explicitly defined?"), but I also suspect a lot of those questions might be answered differently per track. So that might be the answer to a lot of the follow-up questions.

exercises/square-root/description.md
```diff
@@ -0,0 +1,9 @@
...
+
+Check out the Wikipedia pages on [square root](https://en.wikipedia.org/wiki/Square_root) and [methods of computing square roots](https://en.wikipedia.org/wiki/Methods_of_computing_square_roots).
+
+Note that the term "radicand" refers to the number for which the root is to be determined. That is, it is the number under the root symbol.
+
+Recall also that natural numbers are positive real whole numbers (i.e. 1, 2, 3 and up).
+
+When implementing this in a language with a built-in square root function, implement your own function.
```
petertseng: it looks like we are taking a different strategy vs https://github.com/exercism/problem-specifications/pull/1418 here. In that PR, we explicitly chose to remove a sentence that says "if your language already has this, instead implement it yourself". In this PR, we choose to leave such a sentence in. There is probably a unifying principle that ties the two decisions together. It may be good to state it.

This comment should not be taken as a recommendation for or against any particular decision.

wolf99: Hmmm. I understand the point being made on that PR. But disagree with it.
`sqrt()` is something that is present in libraries in a lot of languages. But so is a `list` object yet we have `list-ops`, so is base conversion yet we have `all-your-base`, so are BSTs, yet we have `binary-search-tree`, etc etc.

If we reason that exercises should allow using the language features that they are intended to teach then a very large amount of the exercises become one liners across a lot of tracks.
On the other hand removing such exercises results in a much reduced set. 

The one option left? Suggest that learners do not use those language features in such exercises. Seems reasonable if someone truly wants to learn. And there is nothing preventing them from using the features if they don't. In this way the line is more a confirmation of the intent of the exercise rather than some kind of rule.

wolf99:
@petertseng For this exercise yes, the intention is to have perfect squares so as to make the returned value a whole integer. For languages where this matters, it makes the implementation easier as bit shifting can be used (explained in the Wikipedia page linked in the readme).

I had the idea that more complex versions of the same exercise could form non-core exercises that would be unlocked by this one, for learners that are interested in the mathematical aspect of the implementation and the algorithms involved. These would cover non-perfect square inputs and perhaps progressing to one particular root-finding method, such as Newton-Raphson (though of the top of my head I don't know how this latter could be tested 🤷‍♂ ).

In relation to defining behaviour for non-perfect square inputs for this exercise, perhaps an addition to the readme to explain that these are not intended to be covered?

wolf99:
Rebased, removed version and added UUIDs, ready for repo reopening

exercises/square-root/description.md
```diff
@@ -0,0 +1,9 @@
+Given a natural radicand, return its square root.
+
+Check out the Wikipedia pages on [square root](https://en.wikipedia.org/wiki/Square_root) and [methods of computing square roots](https://en.wikipedia.org/wiki/Methods_of_computing_square_roots).
+
+Note that the term "radicand" refers to the number for which the root is to be determined. That is, it is the number under the root symbol.
```
ErikSchierboom, line 5: Maybe it would make sense to have this explanation be given just after the first line in which radicand was introduced?

exercises/square-root/description.md
ErikSchierboom: Am I correct in assuming that the discussion is still open on what we want to do here?

wolf99: Going with the idea in #1418, and the fact that that PR was merged, I have removed the line from the description here so as to follow suit.

exercises/square-root/description.md
wolf99: Done

iHiD:
(cc @exercism/reviewers)

neenjaw (approved):
This seems quite reasonable! Nice work! Patience pays off!

@ErikSchierboom Can you comment on the test json structure?

ErikSchierboom (approved):
Excellent!

rpottsoh (approved):

SleeplessByte:
This is great!

zimmah:
Maybe specify `x ** 0.5` and `x ** (1/2)` also count as built-in functions, because apparently that isn't obvious.

SleeplessByte:
Ah yes. ** Is the same as Math.pow. want to PR that @zimmah ?

zimmah:
If I compare the instructions on exercism

```
Given a natural radicand, return its square root.

Note that the term "radicand" refers to the number for which the root is to be determined. That is, it is the number under the root symbol.

Check out the Wikipedia pages on square root and methods of computing square roots.

Recall also that natural numbers are positive real whole numbers (i.e. 1, 2, 3 and up).

The idea is not to use the built-in javascript function Math.sqrt(x), it's to build your own.
```

to the [file on github](https://github.com/exercism/problem-specifications/blob/master/exercises/square-root/description.md)

```
Given a natural radicand, return its square root.

Note that the term "radicand" refers to the number for which the root is to be determined. That is, it is the number under the root symbol.

Check out the Wikipedia pages on square root and methods of computing square roots.

Recall also that natural numbers are positive real whole numbers (i.e. 1, 2, 3 and up).
```

they don't match, is that the correct file?

wolf99:
@zimmah Yes. Each language track is able to add to an exercise's readme to so as to be able to give track specific info.

zimmah:
I see, https://github.com/exercism/javascript/pull/929

SleeplessByte:
@zimmah I will point you to the appropriate places. Feel free to ping me on slack if needed.

- The file that causes the difference is [`.meta/hints.md`](https://github.com/exercism/javascript/blob/master/exercises/square-root/.meta/hints.md)
- It allow us to _generate_ a different [`README.md`](https://github.com/exercism/javascript/blob/master/exercises/square-root/README.md)
- How it's generated is explained in [`CONTRIBUTING.md#improving-the-readmemd`](https://github.com/exercism/javascript/blob/master/CONTRIBUTING.md#improving-the-readmemd)

In order to PR this for JavaScript:

- Edit `.meta/hints.md`
- Run [`bin/fetch_configlet`](https://github.com/exercism/javascript/blob/master/bin/fetch-configlet) or [`bin/fetch_configlet.ps1`](https://github.com/exercism/javascript/blob/master/bin/fetch-configlet.ps1) to get `configlet` (or `configlet.exe`)
- Run `bin/configlet generate -o square-root .` to regenerate the `README.md`

Your PR will have 2 changed files:

- `.meta/hints.md`
- `README.md`

I will also post this on your PR.

zimmah:
Oh, I see, thanks SleeplessByte

Issue #1553, opened 2019-07-18
Meta: purpose / goals of canonical data


yawpitch:
This is meant to continue a portion of the discussion initiated in #1551. That PR got rather contentious because of apparent differences in understanding / agenda surrounding the problem-specifications in general and the canonical-data.json in specific. Hopefully this jumping off point will lead to more fruitful discussion and resolve some of the contention.

### The Background
The `resistor-color-trio` exercise describes, broadly, a trivial transformation problem: a list of three strings `["orange", "orange", "orange"]` must ultimately become _some_ string that minimally includes the term "33 kiloohms". There was some healthy debate on what, precisely, that final string should be, but ultimately rough consensus was arrived at for, simply, "33 kiloohms", so we'll consider that point settled.

_Any_ solution to this problem _must_:

1. transform each color to its corresponding numerical value
2. transform the numerical values into a resistor value in ohms (ie 33000)
3. transform the resistor value in ohms to a humanized string  (ie "33 kiloohms")

### The Perspectives

There appear to be, broadly, two schools of thought about the intent of canonical data, though I'm all too happy to admit I might be summarizing one or both sides incorrectly, as the conversation was long and branching.

#### Test the Units

This perspective sees the canonical data as modelling a library and sees each of those three steps as a discrete unit and thus a potential _property_ to be included in the canonical data. In many (if not most) languages Point 1 is going to be provided by a language primitive, and thus is reasonably excluded as a property to test, but Point 2 and Point  3 are not. 

In this perspective then the canonical data would include two _properties_, one that converted the list of strings to an integer and the second that converted that integer to a label for storage or display.

Note that this perspective ignores two considerations:

1. The intended or assumed didactic value of this exercise -- this perspective leaves that up to the tracks... but this doesn't help when the exercise's goal is to fill a gap revealed by Track Anatomy.
2. The prior art in a series or track -- this perspective doesn't assume work done in previous exercises... could lead to recapitulation of essentially identical work.
3. "Bake-in" of an "ideal" solution -- may not let the student explore enough?

#### Test the Whole

This perspective sees the the canonical data as modelling an application: since Point 1 is a functional necessity of Point 2, and Point 2 is a functional necessity of Point 3, an integration test suite that adequately exercises Point 3 will sufficient to test a solution, therefore Point 3 is the only property the canonical data needs to describe.

Thus the canonical data would include just one property that converts a list of strings directly to a label for storage or display.

Note that this perspective ignores different considerations:

[...]

Issue #1545, opened 2019-07-01
Proposal: Add canonical data "nothingness" field


SleeplessByte:
I don't know _how_ to call this exactly.

`JSON` does not distinguish between `nothing` and `empty` (`undefined` and `null` in JavaScript). I propose a new field that is a string value which describes how `null` may be interpreted. This would for example apply to `twoFer`'s test which doesn't pass a value.

```
'empty' | 'absent'
```

- `'empty'`: interpret `null` as if the input is given but empty. Use whatever idiomatic value in the language represents empty. For example, even though `Kotlin` as `null`, it might be better, in the context of the exercise, to use a sealed data type with a `Data(val name: string)` and `NoData` implementation. In JavaScript this would be `''` or `null`.

- `'absent'`: interpret `null` as if there is no input. Use whatever idiomatic value in the language represents optional values or something that deals with absent data. For example, in C# there are optional parameters, same as in JavaScript and Kotlin. In JavaScript this could also interpret the value as `undefined`.

What I'm really trying to solve for is the fact that `null` doesn't describe the _intent_.

bencoman:
My knee-jerk reaction is "now we've cracked the seal on adding knew keys, lets **not** get carried away and quickly accumulate a large amount of them" - but that said, I'm willing to take the discussion on its merits.  

I wanted to understand the background better and the "JSON1" & "JSON2" cases were interesting here...
https://stackoverflow.com/questions/21120999/representing-null-in-json

So we might have...??
```
    {
      "description": "no name given",
      "property": "twoFer",
      "input": {
        "name": {}
      },
      "expected": "One for you, one for me."
    },
```

SleeplessByte:
> let's not get carried away and quickly accumulate a large amount of them

I agree, which is why this is a discussion 👍 

```
"input": {
  "name": {}
},
```

This, in my generator and understand would be to pass:

```javascript
twoFer({})
```

A valid solution, perhaps, could be:

```json
 {
    "description": "no name given",
    "property": "twoFer",
    "input": {},
    "expected": "One for you, one for me."
},
``` 

This, in my generator and understand would be to pass:

```javascript
twoFer()
```

I think in every way, this is much better than adding a new key, or using `null` to denote "absence".

cmccandless:
What about languages that support neither optional parameters nor overloads (I mention this without a specific example of such a language to offer)?

NobbZ:
Those languages are free to not implement `two-fer` at all.

[...]

PR #1586, opened 2019-09-18
say: Tests vulnerable to octal interpretation of chunks

thriqon:
The tests of `say` do not catch an erroneous interpretation of chunks as
octal (which might happen due to leading zeroes inside a chunk). For example, my code produced the following:

````bash
bash say.sh 1023
one thousand nineteen
````

This is due to `023 oct == 19 dec`. Unfortunately, none of the test
cases test for this (and I'm sure quite a lot of implementations in the
database are wrong because of this).

### CODE REVIEW

**Comment by yawpitch** [comment]
This repo is locked to anything but bug fixes (see #1560), and I’m not sure this qualifies. I’m also not sure how many languages _can_ actually be effected by this (for instance in Python casting “023” to an **int** will always give you `23` as to be interpreted as octal it would need to be in the `0o23` octal literal from), though obviously Bash is effected.

Perhaps this should be implemented as a track-specific test in those languages that see “023” as an octal literal?

**Comment by sshine** [comment]
I don't really know if it qualifies as a bug or as an improvement that does not alter the substance of the exercise, but I'd consider it passable.

@thriqon: Did you submit or review a solution that erroneously passed existing tests but fails on this one, and would you mind sharing it? We have two conflicting forces at work: these tests aim to address arbitrary kinds of mistakes in dozens of languages, and yet sadly, the format limits us to a degree where we would like the cases to be succint.

There are existing tests with inner zeroes, e.g. 1002345 -> "one million two thousand three hundred forty-five". Does this not evoke the same kind of error as the first test case in this PR? Is there a motivation to find a smaller counter-example such as 1023? Does "0023" not get interpreted as an octal by the same program?

The second test case in this PR, if somehow the inner part is interpreted as octal, would fail differently. I'm not sure the description "a number with higher digits" fully catches the intent; maybe go with the existing test naming of simply "one thousand nine", since a description of the octal corner case, I doubt, will make sense in a lot of languages.

**Comment by rpottsoh** [comment]
@iHiD repo `problem-specifications` has WIP on the brain.

**Comment by thriqon** [comment]
@sshine, unfortunately, I've since fixed my code and it is gone, I'd have to reintroduce the bug.

The testcase `1,002,345` contains inner zeroes, but not in the bug-triggering position. In order to make the bug visible a chunk has to be a) lower than 100 and b) greater than 7. I'd be happy to rewrite this PR changing `1,023,045`, as that should cover all the old intentions AND the octal case.

**Comment by petertseng** [comment]
I would approve either this PR as it stands, or 1023045. In both cases, please ensure that Travis CI succeeds. Please read the results of the current run at https://travis-ci.org/exercism/problem-specifications/builds/586724026 if unsure.

---

It seems we are uncertain what constitutes a bugfix. Here are some possible thoughts.

* "If and only if no types have changed, the PR is a bugfix". Correctly classifies https://github.com/exercism/problem-specifications/pull/1581 as a bugfix and https://github.com/exercism/problem-specifications/pull/1580 as not-a-bugfix. Classifies this PR as a bugfix.
* "If and only if the README is not changed, the PR is a bugfix". Correctly classifies https://github.com/exercism/problem-specifications/pull/1581 as a bugfix. Incorrectly classifies https://github.com/exercism/problem-specifications/pull/1580 as a bugfix. Classifies this PR as a bugfix.
* "If and only if we believe most submitted solutions already do this correctly, the PR is a bugfix". Relies on human guesswork. By my guesses, correctly classifies https://github.com/exercism/problem-specifications/pull/1581 as a bugfix and https://github.com/exercism/problem-specifications/pull/1580 as not-a-bugfix. By my guesses, classifies this PR as a bugfix. The only implementations affected are those that convert parts of string to an integer as an intermediate step. A reminder that the resulting integers then have to be converted back into a string.

That is the best I can do for how to determine whether it is a bugfix. Help is appreciated.

File: exercises/say/canonical-data.json [hunk 1]
```diff
@@ -127,4 +127,4 @@
       "expected": {"error": "input out of range"}
     }
   ]
-}
\ No newline at end of file
+}
```
sshine: Avoid this.

sshine:
*(The rest is unrelated to fixing this bug in particular.)*

@petertseng wrote:
> It seems we are uncertain what constitutes a bugfix.

Yes, we are. For the purpose of this PR, it should be enough to assert that it is *either* a bugfix *or* an improvement that does not alter the substance of the exercise. We could clarify what we mean by bugfix, since there is a distinction between adding a unit test that fixes a common bug in solutions on a given track, and fixing a bug in canonical data (#1319 is the only I can think of).

But I don't know if it's necessary, since the aim of categorizing something as a bugfix or an improvement that does not alter the substance of the exercise is to avoid the kind of tension that caused #1560.

exercises/say/canonical-data.json
thriqon: Removed it.

yawpitch:
I'm still not certain that this shouldn't be a track-specific test, since octal interpretation of the chunks is a language-implementation-specific issue, but I don't have a formal objection based on that.

What I do somewhat object to is modifying an existing test without also modifying description to incorporate the reasoning for the change ... without looking at the commit history it both numbers appear entirely arbitrary and a future commit could undo the "added" benefit of testing octal interpretation unknowingly.

Honestly I'd rather have a new test with a description that specifically mentioned that it would block an issue with octal interpretation of chunks in languages for which that's an issue.

sshine:
> I'm still not certain that this shouldn't be a track-specific test

I think that principal discussion goes into the work of @ihid's project group work later this year.

Making the distinction every time people add test cases is one of the most common roadblocks for public, iterative improvement of this repo. We tried to mitigate it with optional tests for common cases (#1492, #1518) and other proposals and discussions (e.g. #1545, your #1553), but "octal interpretation" would be a fringe case even in that model.

> What I do somewhat object to is modifying an existing test without also modifying description to incorporate the reasoning for the change ... without looking at the commit history it both numbers appear entirely arbitrary *[...]*

Yes; as I uncovered, #399 does not immediately justify its particular choice of 1002345, so we can only assert that it has some coverage, and modifying it to 1023045 changes that coverage arbitrarily.

> Honestly I'd rather have a new test with a description

Same. But I wasn't blocking either.

> that specifically mentioned that it would block an issue with octal interpretation of chunks in languages for which that's an issue.

It's a tradeoff. Some tracks will find that redundant as it's so fringey. The test descriptions in this exercise don't even mention what they're testing most of the time, anyways. Lots of loose ends wrt. having a standard for test descriptions and upholding that. If we're going to solve them in this PR, then it is both unreasonable and the scope of this PR would exceed the temporary lockdown.

As in: If this test case description is simply "one million twenty-three thousand forty-five", it's not broken a standard that the canonical data for this exercise currently upholds; and looking at the history for the test case, we'd at least find this PR rather than the initial commit that contains no coverage description.

I've reverted the test case to 1023, which was in @thriqon's original PR and later confused by my indecision.

exercises/say/canonical-data.json
```diff
@@ -92,7 +100,7 @@
       "input": {
         "number": 1002345
       },
-      "expected": "one million two thousand three hundred forty-five"
+      "expected": "one million twenty-three thousand forty-five"
```
petertseng: you would want to change this back, if leaving 1002345 untouched.

sshine: Whoops.

yawpitch:
> It's a tradeoff. Some tracks will find that redundant as it's so fringey. The test descriptions in this exercise don't even mention what they're testing most of the time, anyways. Lots of loose ends wrt. having a standard for test descriptions and upholding that. If we're going to solve them in this PR, then the scope change of this PR is both unreasonable and would exceed the temporary lockdown. _(Edit: Changed sentence structure.)_

That seems spurious. The existence of many loose ends doesn't mean there's any value in adding another loose end. Nor does _not_ adding another loose end here implicate that we need to fix all of the other tests in this exercise in this PR.

Moreover with the hold as described in #1560 a PR to update all the descriptions, including this one, _should_ be merged, but arguably this PR _should not_ as the existing tests are not broken. I'm confident that this is *not* a bug fix, but a new test that is intended to address a language-specific issue. Even before the the hold it likely should have been given an "optional" tag so that track maintainers would know if they had reason to implement it.

> As in: If this test case description is simply "one million twenty-three thousand forty-five", it's not broken a standard that the canonical data for this exercise currently upholds; and looking at the history for the test case, we'd at least find this PR rather than the initial commit that contains no coverage description.

You're right, this exercise currently upholds no standard whatsoever, including the minimal one implied by common meaning of "description" on that field, however recapitulating a previously existing error is just shooting yourself in the foot _again_.

I've had reservations about this PR from the start. There is a reason why this repo is on hold, which is why this:

> Making the distinction every time people add test cases is one of the most common roadblocks for public, iterative improvement of this repo.

Runs into the wall of my reservations ... until #1560 is closed **there are not supposed to be any public, iterative improvements to this repo** _except_ for clear and obvious bug fixes -- existing tests are _wrong_, not merely that the existing tests fail to address all possible issues -- and copy improvements. I'm all for copy improvements and I'd consider the description field to be copy ... but to bypass the intent of the lockdown while not even making a minimal effort at ensuring the new test has good copy seems unworkable to me.

sshine:
> this PR *should not [be merged]* as the existing tests are not broken

In that case, we'll put it on hold. I agree with your reasoning.

petertseng:
I see that I must have misunderstood, thinking https://github.com/exercism/problem-specifications/pull/1581 was a precedent. I now see the error of my ways. My bad!

The reason I now see my misunderstanding is that I now see that (in the *leap* exercise) dividing by a number other than 100 or 400 is an operation independent of a language (we can do it in our heads, we can do it on paper and pencil), whereas the knowledge that a string that starts with 0 will be interpreted as octal depends on the particular specification of how strings are interpreted as numbers in a given language.

It must be that I initially did not give this enough thought and therefore failed to see this distinction.

When it comes time to consider this PR again, please do consider a description that says why the case is added, not just echo the expected output. Thank you.

yawpitch:
> I see that I must have misunderstood, thinking #1581 was a precedent. I now see the error of my ways. My bad!

It's not an error, it's just a point of view on which rough consensus can be difficult to achieve; there's obviously some wiggle room on what constitutes a bug fix issue. Since #1581 was us making, essentially, a flawed statement about the _problem_, rather than about any given _solution_ to that problem, it leans a little more towards tipping to "bug fix" and therefore merge-able during the hold. This issue addresses a valid and interesting edge case, but one that will effect only _solutions_ to the problem implemented in certain languages... that leans a little away from "bug fix", even if it's something we _definitely_ want to address once the hold gets lifted.

ErikSchierboom:
@thriqon Are you still interested in working on this PR?

IsaacG:
@thriqon Is this PR still relevant or should it be closed out?

kytrinyx:
Given that it's been a year since the previous ping with no response, I think we are good to close it.

IsaacG:
Closing out for now. Feel free to reopen if you want to work on this further!

PR #1591, opened 2019-09-30
leap: Grammar tweak for canonical data

m-dango:
The grammar in the leap canonical data descriptions was flagged up in https://github.com/exercism/perl6/pull/347


yawpitch:
This repository is locked to impactful, non bug-fix changes (see #1560). As this is not a bug-fix and is impactful (it will trigger both automated and manual rebuilding of exercise tests in tracks), so should not be merged until the lock is lifted.

PR #1593, opened 2019-10-01
Fix broken link to website contributing document

sjwarner-bp:
I noticed a dead link in the Contributing guide.

This fixes it by pointing to the Website repo instead.


yawpitch:
Thanks for the fix!

Issue #1555, opened 2019-07-21
minesweeper tests do not contain border


BartMassey:
In the description of the `minesweeper` problem, the minefield is shown as being surrounded by border characters. The test cases have no borders. Borders make the problem a bit easier by obviating the need for clipping. Should fix either the tests or the description.

roxanaghenea:
I can work on this

yawpitch:
_How_ does the border make the problem easier? The function receives a transmitted representation of the play state of the game ... _why_ would you want to include the border in that play state? It's either unchanging, in which case you're just transmitting redundant information, or it's potentially malformed on its own, and now you've got to parse whether or not the border is correctly formed.

BartMassey:
Just means that you can avoid doing a bunch of bounds checks on array references without going off the edges of the array. It's not a big deal one way or the other, as long as it's consistent.

NobbZ:
But if you include the border, then usual arithmetic to map 2 coordinates in a single index for a flattened down array wouldn't work anymore properly.

In my opinion including the border would makes things harder rather than easier.

BartMassey:
The arithmetic still works fine, you just have to run loops etc from 1..n-1 rather than 0..n. Without sentinels for the borders you will still have to account for `*(a + i * n + j - 1)` not necessarily being the cell to the left of `*(a + i * n + j)` because of wraparound.

That said, I'm good either way. Just make the description agree with the API one way or the other.

NobbZ:
I'm not sure where you think there is a mismatch? The border in the example is to make it easier for the human to understand the example.

The JSON though is to be parsed by maintainers and test-generators, which again may choose to parse `["   ", " * ", "   "]` into `{3,3,"    *    "}` or even `{3,3,[empty, empty, empty, empty, mine, empty, empty, empty, empty]}`.

yawpitch:
> I'm not sure where you think there is a mismatch? The border in the example is to make it easier for the human to understand the example.

There is a mismatch in that the description.md _implies_ that borders will be in the data the student has to work with. Since the student never sees the JSON that's a reasonable inference.

This _could_ be resolved by incorporating some text in the description.md that clearly states that the borders are for display purposes only and will _not_ be in the test data.

NobbZ:
So if we left away the border how would you show the difference of the two examples?

```plain
*  
   
   
```

and

```plain
*    
    
    
```

Hin: The upper board is 3x3, the bottom one is 4x3.

[...]

PR #1594, opened 2019-10-02
updated minesweeper canonical data to contain borders

sprinkle24:


sshine:
Interesting addition. Is there a prior discussion of this?

I think this falls under the category of improvements that are held back by #1560.

roxanaghenea:
Ok, I found this issue: https://github.com/exercism/problem-specifications/issues/1555
 did not know it was blocked :).

exercises/minesweeper/canonical-data.json
```diff
@@ -20,81 +20,111 @@
...
       "description": "space surrounded by mines",
       "property": "annotate",
       "input": {
         "minefield": [
-          "***",
-          "* *",
-          "***"
+          "+---+",
+          "|***|",
+          "|***|",
```
petertseng: This change makes the test case no longer a space surrounded by mines, since there is no space anymore.

exercises/minesweeper/canonical-data.json
```diff
@@ -20,81 +20,111 @@
...
         ]
       },
       "expected": [
-        "111",
-        "1*1",
-        "111"
+        "+---+",
+        "|111|",
+        "|1*1|",
+        "|1*1|",
```
petertseng: This change makes the test case no longer a mine surrounded by spaces, since now there are two mines.

petertseng:
EBWODP: Clarify the third option to "No, and this PR should change the description to state that the border will always be valid." to make clear that selecting this option means the change should go in this PR.

exercises/minesweeper/canonical-data.json
roxanaghenea: Updated, added the missing space

exercises/minesweeper/canonical-data.json
roxanaghenea: Updated.

roxanaghenea:
> when merging, please amend the commit messages to:
> 
> * Link to the issue #1555
> * Summarise the reason for the change in this commit message. A short summary such as "So that the description and the tests are consistent" will be acceptable.
> 
> Please also see two comments on the tests

Done.

roxanaghenea:
Solved the two requested changes.

yawpitch:
I'm going to strongly suggest that this change is unnecessary and does nothing but increases the complexity of the solution without markedly increasing its value as a learning exercise.

As is the student need only consider a binary state for a given place on the board -- does it contain a mine, or no -- and the area outside the playing board needs to merely be ignored. With these changes the student needs to also consider three potential wall tiles, and also whether or not the border itself is properly formed.

As an aesthetic choice it might make sense to _display_ a minesweeper board with a border, but it doesn't make sense to _transmit_ one with that border.

So my vote would be for updating the description.md file to reflect the fact that the border in that file is for display purposes only and has no bearing on the problem.

PR #1596, opened 2019-10-02
topics: suggestion to add errors

tehsphinx:
This is a bit of a Go thing.

First of all there are no `exceptions` in Go. So we cannot use the tag `exception_handling` as that would be wrong. In Go we have `errors` (and `panics`).
Secondly, `errors` are values in Go. So they'd be a data type.

We could also add `error_handling` in `Basic Concepts` instead.

### DIFF
--- a/TOPICS.txt
+++ b/TOPICS.txt
@@ -6,6 +6,7 @@ conditionals
 events
 equality
 exception_handling
+error_handling
 generics
 inheritance
 interfaces
@@ -26,7 +27,6 @@ classes
 dates
 discriminated_unions
 enumerations
-errors
 floating_point_numbers
 graphs
 integers
--- a/TOPICS.txt
+++ b/TOPICS.txt
@@ -5,8 +5,7 @@ callbacks
 conditionals
 events
 equality
-exception_handling
-error_handling
+errors_and_exceptions
 generics
 inheritance
 interfaces
--- a/TOPICS.txt
+++ b/TOPICS.txt
@@ -5,7 +5,8 @@ callbacks
 conditionals
 events
 equality
-errors_and_exceptions
+error_handling
+exception_handling
 generics
 inheritance
 interfaces
--- a/TOPICS.txt
+++ b/TOPICS.txt
@@ -26,6 +26,7 @@ classes
 dates
 discriminated_unions
 enumerations
+errors
 floating_point_numbers
 graphs
 integers


tehsphinx:
@rpottsoh: not really. The standard error in Go is a string containing the error message. It does not have an error code.

rpottsoh:
Since an `error` is just a string then maybe `error_handling` would be more appropriate as you suggested.  Outside of Go, `error` doesn't mean anything.

NobbZ:
`error`s are not strings. Anything can implement the `error`-interface by providing a method `Error()` (IIRC) which returns a string representation of the error.

tehsphinx:
@NobbZ: I know. But it served the point of explaining how errors are different from exceptions in other languages.

The standard error implementation is:

```go
// errorString is a trivial implementation of error.
type errorString struct {
	s string
}

func (e *errorString) Error() string {
	return e.s
}
```

... basically a string wrapped in a struct with a function.

yawpitch:
Error handling is, fundamentally, a subset of exception handling, regardless of the semantics of how a particular language engages with the notion of exceptional circumstances. Adding a new topic seems like overkill, what about simply renaming the existing one to something like "exceptions & errors"?

tehsphinx:
@yawpitch: something like that?

`errors_and_exceptions`

We can also use `exception_handling` for the Go track... but any Go programmer will feel that it's wrong -- even if you could argue that it's a subset... which I could also argue against, that errors in Go have nothing 'exceptional' about them.

yawpitch:
> @yawpitch: something like that?
> 
> `errors_and_exceptions`
> 
> We can also use `exception_handling` for the Go track... but any Go programmer will feel that it's wrong -- even if you could argue that it's a subset... which I could also argue against, that errors in Go have nothing 'exceptional' about them.

To be perfectly honest any Go programmer that feels that particularly strongly may be being a bit too pedantic and borderline cultish for their own good ... exception handling, in the broadest sense, merely means handling circumstances which are the exception to the rule ... ie running into a situation that is not the normal course of business for the program. Now, in the course of a career you _might_ happen across a problem in which errors are the _normal_ thing to react to and a non-error state _is_ the exception, but that is -- not to put to fine a point on it -- an exceptional problem.

A language _may_ choose to react to such an exceptional circumstance by diverting control flow and having a -- possibly catch-able -- exception mechanism, or it can choose to react by having a return type that includes the notion of an error ... both are perfectly valid (as, indeed, is just outright panic crashing, though that's not particularly helpful), and both fall under the umbrella of handling exceptions.

That said I'd be fine with `exceptions_and_errors` if that reduces friction ... I'd still put exceptions first, because there are _many_ exceptional circumstances that are not actually errors (reaching the end of an iterable thing, timing out before backing off  and retrying, receiving OS signals, etc).

yawpitch:
Also @tehsphinx do notice that there are several items in that file that may not, do not, or for that matter _can_ not apply to any given language ... generics and Go, pointers and Python ... pattern matching, loops, threading, classes (abstract or otherwise) ... a programmer in any given language could see any number of items and feel that it's wrong _for them_, but this is meant to be a set of general topics that _may_ apply to a track, thus they need not be stated in such a way that they _do_ apply.

sshine:
You do know that you don't need to agree on a common topic, right?

For example, Haskell track uses the tags "maybe" and "either".

tehsphinx:
> > You do know that you don't need to agree on a common topic, right?
> > For example, Haskell track uses the tags "maybe" and "either".
> 
> Both also valid ways of handling exceptions, even when they’re not errors. ;-)

But still they wouldn't use `exception_handling` as the tag but instead they use `maybe` and `either`. That is exactly my point. I wanted another tag `errors` or `error_handling` because `exception_handling` is not really fitting.

yawpitch:
> > > You do know that you don't need to agree on a common topic, right?
> > > For example, Haskell track uses the tags "maybe" and "either".
> > 
> > 
> > Both also valid ways of handling exceptions, even when they’re not errors. ;-)
> 
> But still they wouldn't use `exception_handling` as the tag but instead they use `maybe` and `either`. That is exactly my point. I wanted another tag `errors` or `error_handling` because `exception_handling` is not really fitting.

These aren’t tags, they’re topic suggestions, they do not need to be expressed in a way that applies to every language equally. You’re quibbling over a semantic difference: exception handling is an umbrella term that includes Haskell’s notion of Maybe **and** `error` **and** `Either` and so on and so on. It also includes Go’s `error` underneath it’s umbrella, as well as `defer`, `panic`, and `recover`. There is no reason whatsoever for individual topics to exist in this list to be crafted towards the perhaps multiple ways in which a given language chooses to specifically approach a general topic, so by adding this topic you’re just duplicating a narrower grouping under a more general one.

Exception handling includes any concept of reacting to a situation in which there is both a “more normal” outcome and a “less normal” one. It’s only distinct from conditionals, really, because it represent a place in which a caller reacts to “less normal” in a callee, or where a callee can communicate “less normal” to a caller.

Error handling is a specific subset of that problem... it’s exception handling when the exception is also an error.

Removing any controversy is why I proposed `exceptions_and_errors` as the topic, because that incorporates both and also implies that the latter is a subset of the former.

sshine:
@yawpitch wrote:
> what about simply renaming the existing one to something like "exceptions & errors"?

@tehsphinx wrote:
> something like `errors_and_exceptions`?

It seems you're suggesting the same thing (assuming equivalence between "`_and_`" and "&", and commutativity of the two), and yet you do not agree? Although #1560 only addresses "the substance of an exercise", and this PR addresses "topics", this is an example of passionate debate that #1560 is currently guarding against. So while it may be the aim of one reviewer to achieve an ideal taxonomy, another may just want to get on with track maintenance.

As such, it should probably be put on hold.

@tehsphinx wrote:
>  I wanted another tag `errors` or `error_handling` because `exception_handling` is not really fitting.

My point with `maybe` and `either` is that they're not in TOPICS.txt. Is there something on the website that breaks if you go ahead and use a topic that's not mentioned in TOPICS.txt? In that case, exercism/exercism#4186 has an [outstanding task](https://github.com/exercism/exercism/issues/4186#issuecomment-533643622) to document the dependencies, including how a track might depend problem-specifications wrt. TOPICS.txt.

If not, the path of least resistence is to use the topics you like on the Go track. :-)

@yawpitch wrote:
> exception handling is an umbrella term

It depends on whether you think of exceptions as "exceptional states that any program can have", or "the language feature called exception in language *X*", or "the methods with which compiler makers implement exceptions", or any other specific view. Having a discussion over which of "exception" and "error" is the umbrella term seems to diverge this PR.

yawpitch:
> As such, it should probably be put on hold.

I’m fine with that.

> Is there something on the website that breaks if you go ahead and use a topic that's not mentioned in TOPICS.txt?

There’s a script in the Julia track that does some minimal parsing of the file, and a view on the website’s track maintenance dashboard that expects its existence, but AFAICT there’s never been anything that prevents a track from renaming or using whatever novel name the track prefers.

> @yawpitch wrote:
> 
> > exception handling is an umbrella term
> 
> It depends on whether you think of exceptions as "exceptional states that any program can have", or "the language feature called exception in language _X_", or "the methods with which compiler makers implement exceptions", or any other specific view. Having a discussion over which of "exception" and "error" is the umbrella term seems to diverge this PR.

Everything in the “Basic Topics” section is just that, a basic topic, and should be interpreted broadly by default. Yes, someone _can_ read “inheritance” and think that means the specific algorithm used in Python for deciding method lookup resolution issues, but that’s an unhelpful and ungenerous frame of reference to apply to other tracks, and arguably even too narrow for that language as well.

This PR diverges the TOPICS.txt file because it assumes that error handling is _not_ already incorporated by the term exception handling _entirely_ because from within the narrow frame of reference of Go an `error` Is a first class type while an exception is not. That’s looking through the binocular backwards... if Go had decided to name that same first class type an `exception` instead  — as is the case in Python, which features exception handling but lacks error handling — this PR would not exist at all. By either topic name there’s no need for both topic names... my suggestion for `exceptions_and_errors` was meant to remove any confusion that these two topics are not interrelated, while also maintaining some of the “all dolphins are whales but not all whales are dolphins” status quo, since — to perhaps over-flog the point — all errors are exceptional but not all exceptions are errorful.

tehsphinx:
The underlying questions we have to answer first is:

> What is `TOPICS.txt` for?

and

> Do we want to use the `TOPICS` in the future more extensively? (For example the in the new project @ErikSchierboom is spearheading.)

For now all I did was look at https://tracks.exercism.io/go/master/topics. Some of the red ones could just be renamed and then fit to the already existing topics, some can't.

One of them is `errors`. We cannot have a topic that contains `exceptions` on the Go track. Those flags are visible to the student and that would lead them into the wrong direction. I would be very careful with your broad definition of `exceptions`, @yawpitch. You are not there to explain to the student what you mean by that when they read it. IMO, we really should be more mindful of the students who we potentially confuse with our ideas/definitions.

That being said: We can add `errors` or `error_handling`, or not. That depends on what the purpose of the TOPICS.txt list and the purpose of the tags is. Is the list a superset or just a list of commonly used topics. Where are the topics used (apart from the exercise list the student sees) and do we need a list somewhere that contains a superset. That will largely depend on the current and future use case.

yawpitch:
>  For now all I did was look at https://tracks.exercism.io/go/master/topics. Some of the red ones could just be renamed and then fit to the already existing topics, some can't. 

I may be wrong, but I don’t believe that any of those topics come directly from TOPICS.txt. That copy is all from the track’s own [config.json](https://github.com/exercism/go/blob/master/config.json), which is where the rubber meets the road and the Go track gets to decide whatever name it likes for whatever topic it likes.

So, for instance, I believe the `exception_handling` topic displayed for the `matrix` exercise is the one the Go maintainers have chosen to display. The fact that it’s been copied verbatim from the TOPICS.txt isn’t something that’s been forced on the track.

> One of them is `errors`. The Go team very purposefully did **not** name Go's errors `exceptions`. We cannot have a topic that contains `exceptions` on the Go track. Since those flags are visible to the students, it would lead them into the wrong direction. 

Yes, which is why you have the power to give the general topic a specific name that’s appropriate to your track in the configuration for your track. That doesn’t _need_ to be telegraphed back to the TOPICS.txt, which at this point is really just talking points.

> I would be very careful with your broad definition of `exceptions`, @yawpitch. You are not there to explain to the student what you mean by that when they read it. IMO, we really should be more mindful of the students who we potentially confuse with our ideas/definitions. I don't mind your definition. But on exercism.io we are trying to teach. Teaching always requires an extra level of care.

I’ve been here for a while, I’m aware of who we’re trying to teach, thanks. The fact that these general topics in this file are not student-facing rather moderates your point, though.

> That being said: We can add `errors` or `error_handling`, or not. That depends on what the purpose of the TOPICS.txt list and the purpose of the tags is. Is the list a superset or just a list of commonly used topics?

Good question. From searching carefully all its used for is suggestions of things to consider as topics in a track-specific setting. There’s no obvious mechanism for enforcing their relationship to track topics.

> Where are the topics used (apart from the exercise list the student sees) and do we need a list somewhere that contains a superset?

They’re not, as far as I can see, shown to the student. These human readable topics aren’t the ones found in the machine readable config file actually displayed to the student.

If a superset is actually required then we should delete the TOPICS.txt and parse the current config.json files and roll them up into that superset.

iHiD:
> What is TOPICS.txt for?

Yeah, so I'm putting this discussion on hold until we have answered this question. This discussion is another example of the reason this repo is locked atm. I won't lock the issue (because then I have to remember to unlock it later), but please let's stop the conversation here.

workingjubilee:
`TOPICS.txt` actually has prior discussion documenting its intent, under which it was agreed on as an addition, which was not imbued inside the file. I have located this discussion and will be submitting a PR to clarify this matter.

ErikSchierboom:
With Exercism v3, we've dropped the support for topics in favor of concepts. Therefore, I'm closing this PR.

PR #1598, opened 2019-10-03
triangle: Add tests for invalid floating-point triangles

jalovatt:
For completeness' sake, and just in case someone comes up with a crazy solution that would pass everything else but break at this point, I've added a pair of test cases for invalid floating-point triangles.


ErikSchierboom:
@jalovatt Are you still interested in working on this PR?

jalovatt:
> @jalovatt Are you still interested in working on this PR?

No thanks, I've been away from Rust and Exercism for a couple of years at this point.

PR #1600, opened 2019-10-05
Update resource link

vjba:
exercism/typescript#284


yawpitch:
The original link is working for me, and while I'm content that the wikipedia link is far more exhaustive, it's also a lot less succinct.

PR #1602, opened 2019-10-07
minesweeper: remove border

### LINKED ISSUE #1555: minesweeper tests do not contain border

yawpitch:
Updates the board example to remove the need for a border that doesn't match the test data and clarifies a couple small points in the description.md.

If merged:
    Resolves #1555
    Closes #1594

### DIFF
--- a/exercises/minesweeper/description.md
+++ b/exercises/minesweeper/description.md
@@ -1,25 +1,35 @@
-Add the numbers to a minesweeper board.
+Add the score totals to a completed Minesweeper board.
 
 Minesweeper is a popular game where the user has to find the mines using
 numeric hints that indicate how many mines are directly adjacent
 (horizontally, vertically, diagonally) to a square.
 
 In this exercise you have to create some code that counts the number of
-mines adjacent to a square and transforms boards like this (where `*`
-indicates a mine):
+mines adjacent to a given empty square and replaces that square with the
+count.
 
-    +-----+
-    | * * |
-    |  *  |
-    |  *  |
-    |     |
-    +-----+
+The board is a rectangle composed of blank space (' ') characters. A mine 
+is represented by an asterix ('\*') character.
 
-into this:
+If a given space has no adjacent mines at all, leave that square blank.
 
-    +-----+
-    |1*3*1|
-    |13*31|
-    | 2*2 |
-    | 111 |
-    +-----+
+## Examples
+
+For example you may receive a 5 x 4 board like this (empty spaces are
+represented here with the '.' character for display on screen):
+
+```
+.*.*.
+..*..
+..*..
+.....
+```
+
+And your code will transform it into this:
+
+```
+1*3*1
+13*31
+.2*2.
+.111.
+```
--- a/exercises/minesweeper/description.md
+++ b/exercises/minesweeper/description.md
@@ -9,20 +9,20 @@ mines adjacent to a given empty square and replaces that square with the
 count.
 
 The board is a rectangle composed of blank space (' ') characters. A mine 
-is represented by an asterix ('\*') character.
+is represented by an asterisk ('\*') character.
 
 If a given space has no adjacent mines at all, leave that square blank.
 
 ## Examples
 
 For example you may receive a 5 x 4 board like this (empty spaces are
-represented here with the '.' character for display on screen):
+represented here with the '·' character for display on screen):
 
 ```
-.*.*.
-..*..
-..*..
-.....
+·*·*·
+··*··
+··*··
+·····
 ```
 
 And your code will transform it into this:
@@ -30,6 +30,6 @@ And your code will transform it into this:
 ```
 1*3*1
 13*31
-.2*2.
-.111.
+·2*2·
+·111·
 ```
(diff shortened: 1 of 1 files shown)


exercises/minesweeper/description.md
```diff
@@ -1,25 +1,35 @@
...
-    |1*3*1|
-    |13*31|
-    | 2*2 |
-    | 111 |
-    +-----+
+mines adjacent to a given empty square and replaces that square with the
+count.
+
+The board is a rectangle composed of blank space (' ') characters. A mine 
+is represented by an asterix ('\*') character.
```
petertseng: I usually see it spelled "asterisk", but I'm open to other spellings. But I couldn't find a resource online that told me that asterix was a valid spelling for the character, after I tried hard to do so. Open to being educated.

petertseng: I see that the character was escaped for the purposes of Markdown (you can see when viewing a rendered version of this file that the backslash isn't there). I have a reservation that this disadvantages those individuals who would view this file with a text editor instead of a Markdown renderer since it may otherwise take them some time to understand what the backslash is for.
(P.S. "those individuals" includes me) 

I hope there is a way to do this that is friendly both to readers using a text editor and to readers using a Markdown renderer.

If it turns out the former group does not make up a significant population and therefore their concerns are invalid, feel free to carry on merrily without regard to this comment.

exercises/minesweeper/description.md
```diff
@@ -1,25 +1,35 @@
-Add the numbers to a minesweeper board.
+Add the score totals to a completed Minesweeper board.
```
petertseng: I wasn't sure it was right to call them "score totals" since I haven't seen a case where the numbers on the board contribute to scoring; usually I saw score counted in time taken or number of steps taken.

Perhaps I could suggest "mine counts" if "numbers" is deemed unsuitable. I would consider "numbers" suitable.

exercises/minesweeper/description.md
```diff
@@ -1,25 +1,35 @@
...
-
-into this:
-
-    +-----+
-    |1*3*1|
-    |13*31|
-    | 2*2 |
-    | 111 |
-    +-----+
+mines adjacent to a given empty square and replaces that square with the
```
petertseng: will "a given" lead readers to think that a single square will be designated? Perhaps "each" instead?

exercises/minesweeper/description.md
yawpitch: No, you're correct ... apparently my computer knows the name of the French cartoon character and thought I was just being creative with case. I'll fix that.

exercises/minesweeper/description.md
yawpitch: I'm going to split the difference and go with "mine counts" ... "numbers" is ambiguous.

exercises/minesweeper/description.md
yawpitch: I'd say the example makes this self explanatory.

exercises/minesweeper/description.md
yawpitch: The file has a .md extension, hence we're going to have to assume that the reader understands markdown conventions (otherwise we can't use triple tildes for code blocks, etc). Unfortunately there's no easy way to be friendly to plain text rendering without treating the blank space differently than the asterisk.

yawpitch: Fixed.

exercises/minesweeper/description.md
yawpitch: Fixed

exercises/minesweeper/description.md
```diff
@@ -1,25 +1,35 @@
...
+count.
+
+The board is a rectangle composed of blank space (' ') characters. A mine 
+is represented by an asterisk ('\*') character.
```
petertseng: regarding https://github.com/exercism/problem-specifications/pull/1602#discussion_r332113861 - I'd like to broach the possibility of writing this using backticks. How would that look?

yawpitch: We could do backticks around quotes (ie `’ ‘` and `’*’`, but I’m uncertain that’s any form of improvement on the escaped form. It also implies that the quotes are meant to be included in the board strings, leading to the same problem as the border by another angle. The quotes are necessary to make the blank space clear, as ` ` is not going to render very distinctly in all possible readers. Personally I think this fix would cause more significant issues than the one it addresses.

At the end of the day these _are_  markdown files. We need not — and cannot — fight to retain their readability in plaintext as a priority over their readability in a markdown enabled viewer, since that’s their primary means of consumption in the current environment. We _could_ entertain putting a script in the toolchain somewhere that removes all markdown formatting from the generated README that the user downloads (ie README.md -> README.txt), but that would then have to be distinct from the one that’s rendered on the website.

petertseng: :eyes:

yawpitch:
As we've got two approvals I'm going to merge. @petertseng we can revisit the escaping issue if we can figure out a better way of rendering an empty string.

Issue #1603, opened 2019-10-12
rational-numbers: abs() is already builtin in some environments


wolf99:
https://github.com/exercism/problem-specifications/blob/dab442f119d7a7196e27840cf8b5c7dc5c6e57fa/exercises/rational-numbers/canonical-data.json#L205

"abs" is used as the property for taking the absolute value of a rational number.
`abs()`  is a builtin in some environments (e.g. C on clang) which prevents the tests and implementation from matching the problem specifications.

Per https://github.com/exercism/c/pull/417

Proposing a change all property values to full words instead of abbreviations.

Will submit a PR shortly for this.

sshine:
It's okay that a property overlaps with a built-in in some language.

It is basically an unreasonable constraint.

This was covered in some discussion a few months ago that I can't remember the issue for, but I remember @SleeplessByte was also in it. The TL;DR is that test generators can take this into account.

Also, please remember that #1560 effectively prevents updates like these.

yawpitch:
I'm going to close this one because @sshine is correct, this was discussed previously and conflicting with a builtin name is not, on its own, a sufficient reason to rename a property, as that would require us to validate every property name against every language. Instead the property name can be translated as required to remove the conflict in the track's implementation / generator.

sshine:
I've looked and looked for the discussion that covers this, and I simply cannot find the reference to the issue which details the argument, so I'll repeat it in short summary in case it needs to be referred to in the future:

It is impossible to avoid hitting some built-in in some language at some point in time, and it is not necessary to try and avoid it:

1. When test suites are generated manually, the track can pick a custom name that does not conflict with the built-in name, and the manual process of bumping the test suite can take that renaming into account without additional effort.
2. When test suites are generated automatically, the track can pick a custom name that does not conflict with the built-in name, and the test generator can specialize the generated test cases so that they refer to the specialized name.

Such renames are typical for when a track or a presumed culture around a language has a different casing or naming convention than in canonical data.

yawpitch:
> I've looked and looked for the discussion that covers this, and I simply cannot find the reference to the issue which details the argument, so I'll repeat it in short summary in case it needs to be referred to in the future:

@sshine I seem to remember at least part of that discussion happening on Slack, and also that it was quite some time ago ... unfortunately we seem to only have a history as far back as August 29th.

sshine:
Slack :crying_cat_face:

SleeplessByte:
[...]

PR #1604, opened 2019-10-12
Fix rational numbers config

wolf99:
Per #1603 this updates some properties to full words instead of abbreviations.

I have also updated the casing on multi-word properties from alllowercase to camelCase.

Additionally have reduced the verbosity of some test case names where possible as these can fall a foul of style linters that restrict line length where the test case description is usually used as such in the test case name in the code.

### DIFF
--- a/exercises/rational-numbers/canonical-data.json
+++ b/exercises/rational-numbers/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "rational-numbers",
-  "version": "1.1.0",
+  "version": "2.0.0",
   "comments": [
     " The canonical data assumes mathematically correct real   ",
     " numbers. The testsuites should consider rounding errors  ",
--- a/exercises/rational-numbers/canonical-data.json
+++ b/exercises/rational-numbers/canonical-data.json
@@ -247,7 +247,7 @@
       "cases": [
         {
           "description": "Raise a positive rational number to a positive integer power",
-          "property": "exprational",
+          "property": "expRational",
           "input": {
             "r": [1, 2],
             "n": 3
@@ -256,7 +256,7 @@
         },
         {
           "description": "Raise a negative rational number to a positive integer power",
-          "property": "exprational",
+          "property": "expRational",
           "input": {
             "r": [-1, 2],
             "n": 3
@@ -265,7 +265,7 @@
         },
         {
           "description": "Raise zero to an integer power",
-          "property": "exprational",
+          "property": "expRational",
           "input": {
             "r": [0, 1],
             "n": 5
@@ -274,7 +274,7 @@
         },
         {
           "description": "Raise one to an integer power",
-          "property": "exprational",
+          "property": "expRational",
           "input": {
             "r": [1, 1],
             "n": 4
@@ -283,7 +283,7 @@
         },
         {
           "description": "Raise a positive rational number to the power of zero",
-          "property": "exprational",
+          "property": "expRational",
           "input": {
             "r": [1, 2],
             "n": 0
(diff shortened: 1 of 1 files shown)


exercises/rational-numbers/canonical-data.json
```diff
@@ -26,7 +26,7 @@
               "expected": [7, 6]
             },
             {
-              "description": "Add a positive rational number and a negative rational number",
```
SaschaMann: This distinction is necessary because it's also possible to add a positive number, e.g. 6.348 or π, to rational numbers.

exercises/rational-numbers/canonical-data.json
```diff
@@ -26,7 +26,7 @@
...
+              "description": "Add a positive and a negative rational number",
```
SaschaMann: This distinction is necessary because it's also possible to add a positive number, e.g. 6.348 or π, to rational numbers.

wolf99:
It's more to try to fix the issue in implementing functions where things like 'abs()' are builtins in some languages.
As it was a breaking change to fix this I thought to add the other small changes also.

If it's expected to diverge from the specification in such cases then I can do that either

wolf99:
Just a small additional note: the lack of camel case quite confused me, googled what 'exprational' was expecting it to be a mathematical term, hopefully other implementers also figure it out (but we shouldn't be relying on that!)

SaschaMann:
Perhaps it should just be `exp` or `exponential`.

> If it's expected to diverge from the specification in such cases then I can do that either

Not sure what the general consensus is when property names clash with builtin names. Imo this causes far less work than a breaking change that will affect more than one track, so it'd be the preferred option.

wolf99:
The exponent related function does require some additional clarification as there are two of them.
One for exponentiation of rational numbers and one for real numbers. So it is understandable that it should have "rational" in the name, but it should be distinct, such as `expRational`, `expReal`.

(I would choose `powRational`, but that's beside the point 😄 )

exercises/rational-numbers/canonical-data.json
```diff
@@ -59,16 +59,16 @@
           "cases": [
             {
               "description": "Subtract two positive rational numbers",
-              "property": "sub",
+              "property": "subtract",
```
sshine: > Per #1603 this updates some properties to full words instead of abbreviations.

As the argument in #1603 is "`abs()` is a builtin in some environments", I should point out that `sub` is a reserved word in Perl, and `subtract` is a builtin in Haskell. As I pointed out in #1603, the premise to avoid keywords in all possible languages is neither reasonable nor necessary.

yawpitch:
> It's more to try to fix the issue in implementing functions where things like 'abs()' are builtins in some languages.

I've closed #1603 as insufficient justification for a breaking change: we do not need to avoid conflicts between property names and individual language reserved words / builtins in the canonical data; translating the property name to something appropriate for the language in question is a track-level implementation concern.

> As it was a breaking change to fix this I thought to add the other small changes also.

#1560 puts breaking changes like this on hold for the moment, so I've added the Hold label to this issue and we should not merge this issue until that one is resolved.

Both of those said I don't disagree with the changes you've made, just their justification; the property names need not have been abbreviated to begin with and the longer form is mildly easier to read (though at the cost of more characters). The various "exp" methods are indeed better in what I tend to call dromedaryCase, as that conforms with the style rule described in the schema. I would tend to agree with replacing "exp" with "pow", since "power" is already used extensively in the descriptions ... if we make that change I'd also replace "number" with "power" at the end of each of the `expReal` descriptions, for both consistency and clarity.

wolf99:
Given your comments @sshine, @SaschaMann I'll close this then.
For the comments on other non-breaking improvements including those mentioned by @yawpitch , I guess they would fall under #1560 .

PR #1610, opened 2019-10-17
robot-simulator: "Rotating (counter)clockwise" -> turning left/right

petertseng:
Was curious about the clockwise/counterclockwise concept. The README
only ever makes reference to "turn right" and "turn left", never
referring to clockwise/counterclockwise.
https://github.com/exercism/problem-specifications/blob/master/exercises/robot-simulator/description.md

While it may seem obvious to some that clockwise corresponds to turning
right (true if looking at robot from above), it does require
establishing the assumption that the point of view is from above.

It seems to make more sense to speak of turning right and left,
especially since the corresponding instructions are R and L.

Historical:
The first time these terms were introduced was with this canonical data
being added in c72f06a21d0233711f143bc5f72df51cfce4a393.

### DIFF
--- a/exercises/robot-simulator/canonical-data.json
+++ b/exercises/robot-simulator/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "robot-simulator",
-  "version": "3.2.0",
+  "version": "3.2.1",
   "comments": [
     "Some tests have two expectations: one for the position, one for the direction",
     "Optionally, you can also test",
@@ -51,7 +51,7 @@
       ]
     },
     {
-      "description": "Rotating clockwise",
+      "description": "Turning right",
       "cases": [
         {
           "description": "changes north to east",
@@ -132,7 +132,7 @@
       ]
     },
     {
-      "description": "Rotating counter-clockwise",
+      "description": "Turning left",
       "cases": [
         {
           "description": "changes north to west",


petertseng:
Next, someone will come to this PR and say "Aha, but calling this operation 'turn right' is flawed too! You didn't consider the case where the robot is standing ON ITS HEAD!!!"

... indeed, I must admit to that person (whoever you are) that I am indeed defeated by this point. I have no rebuttal. I guess we all make assumptions...

yawpitch:
And, to be honest, I'd rather close this than update the description, as even that will lead to some effect and effort for the tracks.

petertseng:
I wanted to confirm that the above messages are read! It is not yet wise to share my own opinion. I plan to share my own opinion when about 72 hours have passed since the submission of the PR. In the meantime, I thank both of you for the reviews and yawpitch for the review, opinion, and recommendation.

petertseng:
EBWODP: I am not sure what purpose GitHub intends for contributors to use the eyes emoji for, but I have started using as "I confirm I read this message"

yawpitch:
> I wanted to confirm that the above messages are read! (I am not sure what purpose GitHub intends for contributors to use the eyes emoji for, but I have started using as "I confirm I read this message") It is not yet wise to share my own opinion. I plan to share my own opinion when about 72 hours have passed since the submission of the PR. In the meantime, I thank both of you for the reviews and yawpitch for the review, opinion, and recommendation.

Right, well given the 72 hour wait and the fact that this would be blocked by #1560 as is I'm going to label it as Hold.

petertseng:
It's a good thing I had to wait 72 hours because I spent some of that time doing research and man oh man did that research reveal a fundamental mistake in my understanding. Correcting this mistake in turn revealed that we don't need this PR.
The rest of this comment just documents my mistake, so there is no need to read it if not interested in knowing how I went wrong.

As some may know, the concept of "clockwise" must be accompanied by a specification that we observe a rotation from above or below the plane in which it is made. Reading that stipulation was what made it click for me.

This PR was ultimately based on the reasoning: "A student could just as easily say the rotation should be observed from below instead of above; then clockwise would be left!" Reading the above statement about the definition of "clockwise" allowed me to see the flaw in the reasoning; do you see it?

For that same student to say "clockwise is left", that student has said that left is perceived from above the plane, but clockwise is perceived from below (or vice versa)! While theoretically a person *could* define that an exercise would follow these rules, it's inconsistent. In the absence of a specific edict demanding otherwise, it just seems like it should be common sense to perceive both clockwise vs anticlockwise and left vs right from the same point of view, not completely opposite ones.

Wow, that was a really unfortunate misunderstanding!
If only I had thought through it beforehand, I could have avoided wasting everyone's time.
I am sorry about that. Hopefully by documenting my mistake I prevent others from repeating it.

yawpitch:
@petertseng there's no requirement to view a clock from above or below the plane on which the clock face is sitting ...

The clock has a face, that face naturally points "up" (the up axis passes through the origin of the clock hands and runs perpendicular to them), and from any position from which the face can be viewed its hands will be rotating "right" from the viewer's POV. You need never consider the point of view of anyone on the other side of that plane.

This is because the precursor to all clocks is the sundial, which was most readily made by putting a vertical rod into the surface of the earth ... no one could _possibly_ view that clock from the opposite side. Anyone on the opposite side of the planet would need their own sundial, which would also be pointing up, and the "hands" (the shadow of the rod on the face) would also be rotating "right" to anyone able to view it ... though for this analogy to work you'd need a binary star system ... that's a problem inherent to sundials.

The shadow on a sundial turns "sunwise" (right) and were you to reverse it you'd have to move the shadow "widdershins" (left). These became "clockwise" and "counter-clockwise" (USA) or "anticlockwise" (UK) once you could put your artificial sundial on a wall or your wrist and have it work, but they still retain the same meaning ... it's the direction of rotation of the hands of the clock relative to the facing of the clock, and to anyone viewing in a position to view that facing the direction of rotation is always right / yawing to starboard / clockwise / sunwise or left / yawing to port / anticlockwise / counterclockwise / widdershins.

It also helps to remember that clocks aren't usually transparent, and that they also are usually accompanied by a whole bunch of machinery that blocks their view from anywhere but their relative "up".

petertseng:
:eyes:

Issue #1609, opened 2019-10-15
list-ops: append descriptions do not match inputs


wolf99:
In the `canonical-data.json` for the `list-ops` exercise test case 2 and 3 have descriptions that are conflicting, given the associated inputs.

In the `append entries to a list and return the new list`  cases:
- Test case 2: "empty list _to_ list" expects `list1` input to be appended to `list2` input
- Test case 3: "non-empty lists" expects `list2` input to be appended to `list1` input

(This is based on the description and the ordering of the values in the expected output..)

Any given implementation can only _either_ append or prepend, but not arbitrarily chose one of append or prepend. Given the name of the cases relates to appending, I understand that append is what was intended, but this is not what these test cases imply.

Additional notes on reading the spec by someone who doesn't know functional operations:
- In the foldl and foldr test cases it seems ambiguous as to which argument in the function is to be used for the initial value, i.e. is the value from the list `x` or `y`? This makes understanding the division operations harder than it should be.
- when concatenating a list of nested lists is it intended that the empty list is not removed completely as it was in the list of lists? I guess the implication is that it is a "shallow" operation?

wolf99:
Problem:

```json
        {
          "description": "empty list to list",
          "property": "append",
          "input": {
            "list1": [],
            "list2": [1, 2, 3, 4]
          },
          "expected": [1, 2, 3, 4]
        },
        {
          "description": "non-empty lists",
          "property": "append",
          "input": {
            "list1": [1, 2],
            "list2": [2, 3, 4, 5]
          },
          "expected": [1, 2, 2, 3, 4, 5]
        }
```

Solution proposal 1:

```json
        {
          "description": "list to empty list",  //<-- Change description
          "property": "append",
          "input": {
            "list1": [],
            "list2": [1, 2, 3, 4]
          },
          "expected": [1, 2, 3, 4]
        },
        {
          "description": "non-empty lists",
          "property": "append",
          "input": {
            "list1": [1, 2],
            "list2": [2, 3, 4, 5]
          },
          "expected": [1, 2, 2, 3, 4, 5]
        }
```

Solution proposal 2:

```json
        {
          "description": "empty list to list",
          "property": "append",
          "input": {
            "list1": [],
            "list2": [1, 2, 3, 4]
          },
          "expected": [1, 2, 3, 4]
        },
        {
          "description": "non-empty lists",
          "property": "append",
          "input": {
            "list1": [1, 2],
            "list2": [2, 3, 4, 5]
          },
[...]

PR #1612, opened 2019-10-19
list-ops: add append case

wolf99:
Subsequent to #1611 (which is itself pursuant to #1609), this will complete the permutation of the possible use cases for appending lists.
This is addition is from the suggestion of @petertseng in https://github.com/exercism/problem-specifications/issues/1609#issuecomment-542440691

Probably should be held until #1560 is resolved however, so adding the hold label already myself 😃

### DIFF
--- a/exercises/list-ops/canonical-data.json
+++ b/exercises/list-ops/canonical-data.json
@@ -30,6 +30,7 @@
           "expected": [1, 2, 3, 4]
         },
         {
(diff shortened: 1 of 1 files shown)


wolf99:
Rebased and test case UUID added, ready for repo reopening,

petertseng (approved):

SleeplessByte (approved):

SleeplessByte:
🥳 🥳 🥳

PR #1613, opened 2019-10-20
grains: add total test

smalley:
Update problem specification to provide a value for the "total" function. Current test case passes empty value to total() and expects the answer for total(64)

### DIFF
--- a/exercises/grains/canonical-data.json
+++ b/exercises/grains/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "grains",
-  "version": "1.2.0",
+  "version": "1.2.1",
   "comments": [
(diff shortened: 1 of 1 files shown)


NobbZ:
`total` is not bound to any square, but instead it is the sum of rice on all 64 squares.

As there will always be 64 squares on a chess board, there is no need to call this property with any input.

yawpitch:
@smalley I'm closing this because @NobbZ is correct, `total` needs no parametrization. The description.md already makes clear that it's meant to return the total number of grains on a chessboard and that a chessboard has 64 squares.

iHiD:
@smalley Thank you though :)

PR #1614, opened 2019-10-24
affine cipher: remove reference to an exit code

coriolinus:
While it is idiomatic for shell scripts to indicate an error status
via exit code, doing so is uncommon in other languages. For generality,
the problem specifications should not instruct the test suites to
return an exit code.

This includes a bump to the patch version as the test suite has
not changed at all.

### DIFF
--- a/exercises/affine-cipher/canonical-data.json
+++ b/exercises/affine-cipher/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "affine-cipher",
-  "version": "2.0.0",
+  "version": "2.0.1",
   "comments": [
     "The test are divided into two groups: ",
     "* Encoding from English to affine cipher",
--- a/exercises/affine-cipher/description.md
+++ b/exercises/affine-cipher/description.md
@@ -33,7 +33,7 @@ More information regarding how to find a Modular Multiplicative Inverse
 and what it means can be found [here.](https://en.wikipedia.org/wiki/Modular_multiplicative_inverse) 
 
 Because automatic decryption fails if `a` is not coprime to `m` your
-program should return status 1 and `"Error: a and m must be coprime."`
+program should error
 if they are not.  Otherwise it should encode or decode with the
 provided key.
  


coriolinus:
re: bumping the version: I quote from [README.md](https://github.com/exercism/problem-specifications/blob/master/README.md):

> **PATCH version changes**
> There are examples of changes requiring a PATCH version change:
> - Changing descriptions or comments.

While that policy remains in effect, I believe it requires this patch version bump.

I agree that "must error" is more general, and will make that change.

yawpitch:
To be more precise I believe you’d have to update the description or comments field within the test data to make a version able change to the test suites.

coriolinus:
I disagree: it seems straightforward that changing `description.md` counts as changing the description. 

We need other people to cast their votes.

yawpitch:
Fair enough, but it’s not been just my understanding. Lots of changes to the description.md and metadata.yml have occurred without corresponding version changes in the actual test data.

petertseng:
It seems that historically we have not considered `description.md` changes to prompt a change to the tests' version.

I wonder if there is a `git` command that is of the form "show me commits that modified description.md that did not modify canonical-data.json" . The best idea I have right now is a manual inspection of `git log --stat exercises/*/description.md exercises/*/canonical-data.json`. Since we don't like having to force people to do manual work, let's try a different tack.

Remember that handy script https://github.com/exercism/problem-specifications/blob/master/bin/check_versions that reminds us to change the version when it is necessary to change it? Its only condition that will cause it to say a version change is required is if a `json` file changed; `.md` files are not inconsideration. If we wanted `.md` files to cause a version change, I would guess that the script would have written in such a way.

Well! Those serve as hints of what the interpretation has historically been, but not what the interpretation should be in the future. For that, I am sure any well-reasoned decision will be acceptable, and will allow the interested parties to have a proper discussion about what that well-reasoned decision will be, without any interference from me.

I will note that the benefit of a version number for the JSON file is it helps tracks determine whether their tests are up to date, as tests are not generated in the same way for every language. In contrast, READMEs are generated in the same way for every language, so a working procedure for determining whether READMEs are up to date is "try to generate them all, and see if any of them changed". So it is not necessary to use the JSON's version number as a signaling mechanism for a README change, but surely there could be any other reason for wanting to change it.

yawpitch:
My argument against versioning the description.md and the metadata.yml is that changes in neither materially effect the task the student is performing... the canonical-data.json IS that task, so if a change to the copy implied anything more than a clarification or rewording of the task THEN it would necessarily mean updating the canonical-data.json and therefore trigger a version ratchet.

When/if we get around to considering i18n tokens in the README.md (as opposed to say a README.md.[lang]) then I’d consider the virtue of versioning it, though likely separately from the canonical-data.json.

iHiD:
I'm going to put this on hold for now. 

We're planning on making changes to how things are structured that consider exactly this thing. I'm hoping to get an issue written up asap, but other things keep getting in the way.

ErikSchierboom:
Looking at the [current description](https://github.com/exercism/problem-specifications/blob/main/exercises/affine-cipher/description.md), the proposed fix appears to have already been applied.

PR #1617, opened 2019-11-05
Imbue TOPICS.txt with guidance on usage

workingjubilee:
See exercism/discussions#159 for source of original intent.
And exercism/exercism#4130 for expansion on the "internal consistency, but embrace uniqueness" clause.

See exercism/problem-specifications#1596 for example of problems caused by current misreading.
See exercism/exercism/issues/4130#issuecomment-410118760 for someone predicting this result.
Thus, this PR.

Perhaps this PR is overly verbose? Regardless, it is preferable to the current state of things. Elsewhere in the Exercism project these ideas are expressed as well, so I do not actually think it too audacious to suggest this.

### DIFF
--- a/TOPICS.txt
+++ b/TOPICS.txt
@@ -1,3 +1,27 @@
+Using this List
+---------------
+This is a non-exhaustive list of topics for use in config.json and elsewhere.
+The primary purpose for these topics is enabling students to navigate a given
+track, both through the topic data on each problem and through UI elements
+that parse and display these topics. This means that between tracks, topics
+may vary freely without necessarily imposing requirements on the surrounding
+tracks.
+
+Within a given track, however, consistency is paramount. A given name for a
+hing should remain exactly consistent due to how the track data is parsed,
+i.e. you can technically use "string" but it must always be "string", not
+"strings", so for simplicity's sake a track should favor pluralization or
+innumerable words. The snake_case form should always be favored, also,
+because alternatives are liable to break the parsing and interpolation that
+the UI does.
+
+In spite of these two rules, the third governing direction for topic list
+is that between tracks it is desirable to provide associative context, so
+that someone from another programming language can use concepts they
+already have learned, where applicable, to understand the language. This is
+mentioned very last because it should be last:  Exercism is dedicated
+first and foremost to embracing the uniqueness of a given language.
+
 Basic Concepts
 --------------
 abstract_classes
--- a/TOPICS.txt
+++ b/TOPICS.txt
@@ -8,7 +8,7 @@ may vary freely without necessarily imposing requirements on the surrounding
 tracks.
 
 Within a given track, however, consistency is paramount. A given name for a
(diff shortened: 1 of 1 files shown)


TOPICS.txt
```diff
@@ -1,3 +1,27 @@
...
+---------------
+This is a non-exhaustive list of topics for use in config.json and elsewhere.
+The primary purpose for these topics is enabling students to navigate a given
+track, both through the topic data on each problem and through UI elements
+that parse and display these topics. This means that between tracks, topics
+may vary freely without necessarily imposing requirements on the surrounding
+tracks.
+
+Within a given track, however, consistency is paramount. A given name for a
+hing should remain exactly consistent due to how the track data is parsed,
```
coriolinus:
```suggestion
thing should remain exactly consistent due to how the track data is parsed,
```

SaschaMann:
I would prefer if this went to a separate file (perhaps the README?). It documents topics.txt but doesn't need to be inside the file itself.

As it is, it would break tooling around new exercise generation so if you decide against moving it elsewhere, I'd appreciate if there will be a waiting period before merging to update tools that rely on the current format.

workingjubilee:
Previously similar words were in separate files (hidden in other repos, no
less!). The primary reason TOPICS.txt hasn't started
fistfights, with people differing on how to read this, seems to honestly
just be because it's hard to punch people over TCP/IP.

Accepting it of course can be delayed as long as necessary, and I
understand if an alternative solution is alighted upon also;
I proposed one in #1618. This is more of an... RFC. But I don't believe it
being in the root, with a LOUD_CAPS_FILENAME,
and not having a guidance at the top of it is long-term viable.

>

sshine:
@workingjubilee: Prior discussion on intent sounds like a good thing to have written down.

As for placing documentation: Local is better so contributors learn about this as they contribute, and central is better because we have a single point of reference (so we know where to look). One place is better than two, so things don't desynchronize.

It appears that exercism/discussions#159 is 200 pages long, so it is not possible for me to quickly assess what part of its conclusion reflects the content of this PR. However, there is nothing controversial in the formulations made.

I don't know if people parse TOPICS.txt, and if adding a text section to it will make this harder.

Do you know, @workingjubilee?

SaschaMann:
I believe @SleeplessByte also parses it for the track maintenance tool.

> if adding a text section to it will make this harder.

In my case it would be a trivial change but it's a breaking change nonetheless

SleeplessByte:
It will break my tool but I can fix it once it's broken.

yawpitch:
Given that it’s a change that would cause additional work and since this file is going to be mooted by work ongoing for V3 anyway (see exercism/exercism#5088), I’m going to put this on hold.

ErikSchierboom:
With Exercism v3, we've dropped the support for topics in favor of concepts. There is some guidance in this PR that can be useful, but we should add that information to the docs page. Therefore, I'm closing this issue.

PR #1627, opened 2019-12-22
fix: Remove trailing spaces

ffflorian:
This PR will remove trailing spaces in all files. Best viewed [without whitespace changes](https://github.com/exercism/problem-specifications/pull/1627/files?w=1).


exercises/ocr-numbers/description.md
```diff
@@ -60,18 +60,18 @@ Is converted to "1234567890"
...
+|_||_ |_
   | _||_|
-         
- _  _  _ 
+
+ _  _  _
   ||_||_|
   ||_| _|
-         
+
```
NobbZ: The whitespace here is significant and not allowed to be removed

rpottsoh:
I have updated the initial comment.  These changes may now be merged without the requirement to bump versions.

rpottsoh:
@ffflorian please include the statement `I have confirmed that no JSON version change is needed` as part of the comment in your next commit.  I am not sure that where I have inserted it will make any difference.  The override message will work if it is in a commit message.

exercises/ocr-numbers/description.md
ffflorian: What about changing it to this?

```text
    _  _  #
  | _| _| #
  ||_  _| #
          # (blank fourth row)
```

ffflorian:
@rpottsoh 
> @ffflorian please include the statement I have confirmed that no JSON version change is needed

Done, it works :) Just waiting for an answer from @NobbZ.

exercises/ocr-numbers/description.md
NobbZ: With literal double crosses?

I think that would lead to similar discussion we had on another exercise where a border was drawn in the examples for visual clarification but not in the test suits for easier programmatic access... I think it was the minesweeper exercise.

Just let the whitespace where it belongs and don't add arbitrary characters which could be misinterpreted

ffflorian: Alright. I reverted the changes to this file.

rpottsoh:
Happy New Year @NobbZ!  I think @ffflorian has addressed both of the issues we had.  What do you think?

PR #1635, opened 2020-02-08
luhn: check a number with an even remainder

pranasziaukas:
I saw a student accidentally pass the test suite just by checking the parity (divisibility by 2) of the final sum instead of the divisibility by 10. A lot of algorithms involve checking of parity so I can see such mistakes slipping in from time to time.

The only two proper numbers that are not valid in the test suite sum to 41 and 57.
Initially, I thought about testing all the possible remainders using the simplest inputs of `01` through to `09` but maybe that's beyond the purpose to provide a sane yet fairly straightforward test coverage.

Anyway, I think it's a good idea to test at least one proper number that is not valid and has an even remainder.

### DIFF
--- a/exercises/luhn/canonical-data.json
+++ b/exercises/luhn/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "luhn",
-  "version": "1.6.2",
+  "version": "1.7.0",
   "cases": [
     {
       "description": "single digit strings can not be valid",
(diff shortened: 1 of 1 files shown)


pranasziaukas:
Adding a new test case should've been a minor version bump instead of a patch.

SleeplessByte:
This seems like a fine "bug fix" to me. @ErikSchierboom see if you agree.

PR #1643, opened 2020-03-01
scale-generator: fix name of augmented interval

### LINKED ISSUE #1642: Scale generator: s/augmented first/augmented second/
See https://en.wikipedia.org/wiki/Augmented_second

r4f4:
The interval described is an augmented *second*, not an augmented first.
Added also its composition in terms of steps for completeness.

Fixes #1642

Signed-off-by: Rafael Fonseca <r4f4rfs@gmail.com>


yawpitch:
I agree, this is a doc improvement and doesn't change the test data, so it's fine to merge under #1560

PR #1624, opened 2019-12-07
#1623 `grade-school` canonical data does not correspond to the exercise description

eparovyshnaya:
*Whole roster query* - case description is altered according to the discussion #1623 results.

Closes #1623

PR #1644, opened 2020-03-03
[two-bucket] Make valid moves clearer

iHiD:
There has been confusion over the reuse of the word "one" in both the descriptions of the bucket ("Bucket one") and the moves ("Move from one bucket"). This removes that confusion, which will help reduce misunderstanding, especially amongst non-native English speakers.

With reference to https://github.com/exercism/exercism/issues/5205#event-3090216965

### DIFF
--- a/exercises/two-bucket/description.md
+++ b/exercises/two-bucket/description.md
@@ -21,8 +21,8 @@ Example:
 Bucket one can hold up to 7 liters, and bucket two can hold up to 11 liters. Let's say bucket one, at a given step, is holding 7 liters, and bucket two is holding 8 liters (7,8). If you empty bucket one and make no change to bucket two, leaving you with 0 liters and 8 liters respectively (0,8), that counts as one "move". Instead, if you had poured from bucket one into bucket two until bucket two was full, leaving you with 4 liters in bucket one and 11 liters in bucket two (4,11), that would count as only one "move" as well.
 
 To conclude, the only valid moves are:
-- pouring from one bucket to another
-- emptying one bucket and doing nothing to the other
-- filling one bucket and doing nothing to the other
+- pouring from either bucket to another
+- emptying either bucket and doing nothing to the other
+- filling either bucket and doing nothing to the other
 
 Written with <3 at [Fullstack Academy](http://www.fullstackacademy.com/) by Lindsay Levine.


coriolinus:
Actually, this should probably bump the patch field of the test version.

yawpitch:
@coriolinus argument as to why? Nothing’s been changed in the canonical-data.json, and my reading of the patch bump rule is that it would apply only to changes in the “description” or “comment” fields of that file.

coriolinus:
I think we've argued this out before. We haven't changed the description field within `canonical-data.json`, but we've changed the (arguably more important in this case) human-readable description of the problem. 

IIRC, last time we started arguing about this, the end result was the lockdown in this repo. I don't want to take it that far again, for fear that the result would be shutting down the problem specifications entirely.

yawpitch:
Me either, but the problem is IF making the copy more clear _outside_ the canonical data implies a version bump _inside_ the canonical data than we’re in a natural deadlock and this PR should not be merged.

My argument: the version number is in in the machine readable file because the version number is important to downstream automation, not to the student. If it were important to the student we’d display it in the description.md as well for them to consume.

Thus the version bump exists to signal a material change in the test data to scripts that generate tests, or that indicate to humans that they need to generate tests... a change solely  to copy in the description.md doesn’t qualify as such a material change.

I also don’t want to lock up this repo any further, but to my view If the versioning rules in the repo root README.md apply to copy-only changes like this (and therefore not only the test data file) then we’re already in a position where this repo **is** hard-locked until those rules are clarified further. So maybe this is a point on which @kytrinyx and/or @iHiD could make a decision and update the README.md to reflect that?

petertseng:
I am linking past discussions because I don't want them to be lost, I already looked for them, and I don't want my effort to go to waste.

* https://github.com/exercism/problem-specifications/pull/1614
* https://github.com/exercism/problem-specifications/pull/1624

Here is what I have to add today:

* If the decision is made to require a version change when a description.md is changed, contributors cannot reliably be expected to adhere to this decision unless https://github.com/exercism/problem-specifications/blob/master/bin/check_versions is modified to enforce it. So if you make this decision, I hope the decision maker will please change that script.
* To require a version change for description.md change holds zero value for me as a maintainer. If I as a maintainer want to know whether READMEs have changed, the procedure is simple: Attempt to generate them, and see if any of them have changed. No version number is needed for this procedure. So, if the decision is made to require it, I hope it is for a different reason than that.

petertseng:
EBWODP: That approval message was too presumptive and tone-ambiguous. I will modify it to clarify neutral tone, as follows:

This approval is an approval made in advance for the case where the decision is made that a change to description.md does not require a change to the canonical-data.json version. It neither advocates for nor against such a requirement.

iHiD:
I'm going to merge this, not because I agree with one side of the argument or the other (I'm not a maintainer and don't really understand the nuances), but because I want to get it merged and it seems that changing the version violates my own rules with regards to this being locked down, and I don't want to do that.

PR #1645, opened 2020-03-08
Resistor Color: Add requirements of the exercise

### LINKED ISSUE #1546: Resistor color: no task description

zuzia-kru:
Resolves #1546.

I know there's already a PR but it seems it hasn't been updated since July last year and I was doing this exercise today in the Ruby track and it got me really confused.

### DIFF
--- a/exercises/resistor-color/description.md
+++ b/exercises/resistor-color/description.md
@@ -13,6 +13,10 @@ These colors are encoded as follows:
 - Grey: 8
 - White: 9
 
+The goal of this exercise is to create a way:
+- to look up the numerical value associated with a particular color band
+- to list the different band colors
+
 Mnemonics map the colors to the numbers, that, when stored as an array, happen to map to their index in the array: Better Be Right Or Your Great Big Values Go Wrong.
 
 More information on the color encoding of resistors can be found in the [Electronic color code Wikipedia article](https://en.wikipedia.org/wiki/Electronic_color_code)


SleeplessByte:
It should _perhaps_ increase the version number (using the PATCH component) so that automated systems receive pings that this exercise has been updated.

zuzia-kru:
> It should _perhaps_ increase the version number (using the PATCH component) so that automated systems receive pings that this exercise has been updated.

Sure, let me know if you think it's needed, I wasn't sure how versions work and if they are there only for changes in the test code.

yawpitch:
Most recent precedent (#1644 and #1643) is no version bump required when the canonical-data.json does not change. I’m going to merge as is based on that.

PR #1650, opened 2020-04-22
Fix apophenia error slowing student progress

workingjubilee:
apophenia - noun
"the tendency to perceive a connection or meaningful pattern between unrelated or random things"
compare: pareidolia - noun
"to perceive a specific, often meaningful image in a random or ambiguous visual pattern"
- Merriam Webster's

### DIFF
--- a/exercises/luhn/description.md
+++ b/exercises/luhn/description.md
@@ -17,27 +17,27 @@ are disallowed.
 ## Example 1: valid credit card number
 
 ```text
-4539 1488 0343 6467
+4539 3195 0343 6467
 ```
 
 The first step of the Luhn algorithm is to double every second digit,
 starting from the right. We will be doubling
 
 ```text
-4_3_ 1_8_ 0_4_ 6_6_
+4_3_ 6_9_ 0_4_ 6_6_
 ```
 
 If doubling the number results in a number greater than 9 then subtract 9
 from the product. The results of our doubling:
 
 ```text
-8569 2478 0383 3437
+8569 6195 0383 3437
 ```
 
 Then sum all of the digits:
 
 ```text
-8+5+6+9+2+4+7+8+0+3+8+3+3+4+3+7 = 80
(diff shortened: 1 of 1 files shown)


yawpitch:
Thanks for the contribution, however can you provide some explanation of how apophenia applies here? As someone who apparently does not have this tendency I see no meaningful value in the change.

Also, your second change appears wrong... on line 27 you perform the doubling on the second digits, when for all other digits that line only acts as a mask to identify which digits to double.

yawpitch:
Note, if this is in reference to one of the most ignorant phrases in human history I'm fine with changing the string, I'd just fix that doubling issue.

workingjubilee:
It is _definitely_ what you are thinking of, yes. I did not wish to give them further resonance by echoing them, as they are well-recorded elsewhere, in appropriate context, and this is not a historical project. Thus the somewhat vague commit remark.

Fixed the issue you noted.

rpottsoh:
@workingjubilee please modify your initial comment to explain what purpose is being served by this pull request.  I am not seeing any logical reason for these changes you have proposed.

coriolinus:
@rpottsoh There exist people who use certain digit strings as a code. Unfortunately, the codes that become well-known are the ones which belong to hate groups, because they tend not to be subtle. 

The removed digit strings include such a code. Nobody is suggesting that its inclusion was anything other than coincidental, but the intent for removing it, in the original commit message, is to keep people--both supporters and victims--from seeing things that aren't really there.

sshine:
I'll spell it out a bit more: 88 means "heil hitler" (8 = h) in some circles.

Having once infiltrated neo-nazi systems, I have an eye for those numbers, too.

I can imagine that someone who was done something similar and whose mental health suffers, it may be somewhat of a fixation. Since we're dealing with minorities, this may be an exaggeration. Then again, I've witnessed three panic attacks in malls over queue distancing in the last month.

I would like to think of this as an act of kindness rather than any attempt at political correctness.

rpottsoh:
Thanks for the explanation.  It is clear to me now.

PR #1652, opened 2020-04-28
resistor-color-trio: Add missing test case from description

### LINKED ISSUE #1651: Add test to resistors-colors-trio
Hi!
The story for the `resistors-colros-trio` includes the following line:
```
- orange-orange-red would be 33 and 2 zeros, which becomes 3300 ohms.
```
However, my code in Julia had the output `3 kiloohms` for that input and passed the tests.

I will be PRing the corresponding json so that the tests corresponds to the specifications.

### LINKED ISSUE #1989: Resistor Color Trio: add test case >1000 but not divisible by 1000
Currently some solutions check if resistance value is >= 1000 and use integer division to switch to kilo-ohms, which causes loss of precision for values that are larger than 1000, but are not divisible by 1000.

This testcase will ensure divisibility by 1000 before transitioning to kilo-ohms.
```
{
  "description": "Red and red and red",
  "property": "label",
  "input": {
    "colors": ["red", "red", "red"]
  },
  "expected": {
    "value": 2200,
    "unit": "ohms"
  }
},
```

miguelraz:
resolves #1651 

- brought to y'all by the Julia Gang

### DIFF
--- a/exercises/resistor-color-trio/canonical-data.json
+++ b/exercises/resistor-color-trio/canonical-data.json
@@ -55,7 +55,8 @@
       "expected": {
         "value": 470,
         "unit": "kiloohms"
-      },
+      }
+    },
      {
       "description": "Orange and orange and red",
       "property": "label",
--- a/exercises/resistor-color-trio/canonical-data.json
+++ b/exercises/resistor-color-trio/canonical-data.json
@@ -1,6 +1,6 @@
 {
   "exercise": "resistor-color-trio",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "cases": [
     {
       "description": "Orange and orange and black",
--- a/exercises/resistor-color-trio/canonical-data.json
+++ b/exercises/resistor-color-trio/canonical-data.json
@@ -55,6 +55,16 @@
       "expected": {
         "value": 470,
         "unit": "kiloohms"
+      },
+     {
+      "description": "Orange and orange and red",
+      "property": "label",
+      "input": {
+        "colors": ["orange", "orange", "red"]
+      },
+      "expected": {
+        "value": 3300,
+        "unit": "ohms"
       }
     }
   ]
--- a/exercises/resistor-color-trio/canonical-data.json
+++ b/exercises/resistor-color-trio/canonical-data.json
@@ -57,7 +57,7 @@
         "unit": "kiloohms"
       }
     },
-     {
+    {
       "description": "Orange and orange and red",
       "property": "label",
       "input": {


SaschaMann:
This test case is already given as example in the description, but not covered by tests. I'd argue it's a bug fix and not covered by #1560.

SaschaMann (approved):

exercises/resistor-color-trio/canonical-data.json
```diff
@@ -56,6 +56,17 @@
         "value": 470,
         "unit": "kiloohms"
       }
+    },
+     {
```
rpottsoh, line 60:
```suggestion
    {
```
Remove single extra leading space.

rpottsoh (changes requested):

yawpitch:
I'd  argue that it is _not_ a bug and that this PR is blocked by #1560. As stated by @iHiD in the body of that Issue a test would have to be _wrong_, not merely missing, to be considered a bug:

> We will still accept bugfixes (**if tests are wrong**) and improvements to the copy around exercises, as long as they do not alter the substance of an exercise

Emphasis added.

Though it would take some effort to survey them all, I believe there are a number of description.md files mention examples that do not literally exist in the tests. Certainly there are many descriptions that make assumptions / implications not actually addressed in tests. If this addition isn't actually testing anything _new_ and otherwise uncovered in the tests -- which it isn't _obviously_ doing, as all three colors appear in the same orientation in other tests -- then I'd argue it's not actually necessary, on top of not really being a bug fix.

rpottsoh (approved):

rpottsoh:
@miguelraz thanks for the making the change I requested.  I have to agree with @yawpitch though.  I recommend following the suggestion outlined in #1560 that suggests that changes be added at the local (track) level.  This test case could be added to the track where you came across this issue in the canonical data.  @exercism/julia was it?

Personally I think the suggested change is a good one for the canonical data.   I am going to close this PR for now.  

This PR should be reconsidered whenever the restrictions of #1560 are reversed.

miguelraz:
Thanks, I think @SaschaMann can help with the Julia local changes.

iHiD:
I've reopening with the `hold` tag for us to consider after #1560 is reversed.

SaschaMann (changes requested):
(just to block the merge button while it's on hold)

iHiD:
@SaschaMann Nice idea. I believe y'all can also now convert PRs to drafts (I've just done that here)

miguelraz:
I believe #1560 is now reversed and this can be merged @SaschaMann .

SaschaMann:
It needs to be rebased, I think. The version can be dropped and the test case itself needs a UUID now.

angelikatyborska (changes requested):
I'm not sure if `3300 ohms` is the correct return value for this case. The description says:

> When we get more than a thousand ohms, we say "kiloohms".

3300 is more than a thousand, so my interpretation of the description would be that the expected value is `3,3 kiloohms`.

SaschaMann:
> The return values are described as objects with value and unit keys, not strings, so it's just about using a float for the value.

Some implementations of this exercise use strings as expected value, e.g. C#'s.

SleeplessByte:
https://neurophysics.ucsd.edu/courses/physics_120/resistorcharts.pdf switches at exactly 1000 ohms. I find that compelling enough to fix the description to `3.3 k`

IsaacG:
We've gone back and forth on 3300 vs 3.3k. I think the conclusion was that it's simpler to just not add it and leave it unspecified.

PR #1653, opened 2020-04-28
simple-cipher: Update description.md

chivalry:
The test suite is testing against the key having only lowercase letters, not alphanumeric.

### DIFF
--- a/exercises/simple-cipher/description.md
+++ b/exercises/simple-cipher/description.md
@@ -59,7 +59,7 @@ substitution cipher a little more fault tolerant by providing a source
 of randomness and ensuring that the key contains only lowercase letters.
 
 If someone doesn't submit a key at all, generate a truly random key of
-at least 100 alphanumeric characters in length.
+at least 100 lowercase characters in length.
 
 ## Extensions
 


exercises/simple-cipher/description.md
```diff
@@ -59,7 +59,7 @@ substitution cipher a little more fault tolerant by providing a source
 of randomness and ensuring that the key contains only lowercase letters.
 
 If someone doesn't submit a key at all, generate a truly random key of
-at least 100 alphanumeric characters in length.
+at least 100 lowercase characters in length.
```
rpottsoh, line 62: Others are welcome to weigh in, but I don't think `lowercase` is implied here.  A truly random key would not be limited to lower case letters.  Individual tracks are welcome to add test cases of their own and should not be limited to lowercase letters.  `alphanumeric` was added during #1346.

chivalry: The [`canonical-data.json`](https://github.com/exercism/problem-specifications/blob/master/exercises/simple-cipher/canonical-data.json) specifically has a test for the key being only lowercase letters on line [40](https://github.com/exercism/problem-specifications/blob/master/exercises/simple-cipher/canonical-data.json#L40), so it isn't track-specific. This actually tripped me up, because I drew my random key from `strings.ascii_lowercase + strings.digits` in the Python track because the instructions said "alphanumeric."

I don't really care which one it is, but if the instructions don't change, then the test in `connonical-data.json` should be changed or removed.

SleeplessByte: Agreed that this is a bug. I'm in favour of changing the test to alphanum, instead of changing the description back.

yawpitch: Changing the test will be rejected due to #1560, this is clearly not a bug.

Think of it like this: the canonical-data.json is, by definition, the _canon_. The description.md is a _story_ about the canon. From the moment #1560 dropped no change, omission, or imperfection in the story can override the canon... the canon can only be (un)fixed where the canon is internally inconsistent or fundamentally wrong and the inputs given in a test _should not_ produce the outputs described in that test.

There is no fundamental, non-personal-preference, reason why uppercase characters **must** be included — we have artificially constrained many exercises in artificial ways — but if a test exists that enforces that constraint then the canon has spoken and the description is _wrong_.

iHiD: 1. Change the description to lowercase and this can be merged. It's the "wrong" fix but it can move forward under #1560.
2. Change the test and this can't be merged (#1560) but it's probably the right fix.

I'd vote for (1) for now. Then do (2) once #1560 is released (which will probably be once v3 is launched and all exercises in this repo become practice exercises again)

yawpitch: I think it's likely obvious from what I said before, but to be explicit I'd also vote for 1 and for an Issue to be created, on hold, for making this case-insensitive in the future.

I wish there was a way for us to simply prevent a PR from being submitted without an Issue already existing and being referenced. I hate the idea of people doing work that will end up stuck.

iHiD: >  I hate the idea of people doing work that will end up stuck.

Me too! But it will get merged eventually.

PR #1655, opened 2020-05-04
rational-numbers: Make formulas more readable

SaschaMann:
This makes the given formulas easier to read, by using `a₁` instead of `a1` etc.

I don't know if this is covered by #1560, please turn it into a draft if it is.

### DIFF
--- a/exercises/rational-numbers/description.md
+++ b/exercises/rational-numbers/description.md
@@ -10,11 +10,11 @@ The product (multiplication) of two rational numbers `r₁ = a₁/b₁` and `r�
 
 Dividing a rational number `r₁ = a₁/b₁` by another `r₂ = a₂/b₂` is `r₁ / r₂ = (a₁ * b₂) / (a₂ * b₁)` if `a₂ * b₁` is not zero.
 
-Exponentiation of a rational number `r = a/b` to a non-negative integer power `n` is `r^n = a^n/b^n`.
+Exponentiation of a rational number `r = a/b` to a non-negative integer power `n` is `r^n = (a^n)/(b^n)`.
 
-Exponentiation of a rational number `r = a/b` to a negative integer power `n` is `r^n = b^m/a^m`, where `m = |n|`.
+Exponentiation of a rational number `r = a/b` to a negative integer power `n` is `r^n = (b^m)/(a^m)`, where `m = |n|`.
 
-Exponentiation of a rational number `r = a/b` to a real (floating-point) number `x` is the quotient `a^x/b^x`, which is a real number.
+Exponentiation of a rational number `r = a/b` to a real (floating-point) number `x` is the quotient `(a^x)/(b^x)`, which is a real number.
 
 Exponentiation of a real number `x` to a rational number `r = a/b` is `x^(a/b) = root(x^a, b)`, where `root(p, q)` is the `q`th root of `p`.
 
--- a/exercises/rational-numbers/description.md
+++ b/exercises/rational-numbers/description.md
@@ -2,21 +2,21 @@ A rational number is defined as the quotient of two integers `a` and `b`, called
 
 The absolute value `|r|` of the rational number `r = a/b` is equal to `|a|/|b|`.
 
-The sum of two rational numbers `r1 = a1/b1` and `r2 = a2/b2` is `r1 + r2 = a1/b1 + a2/b2 = (a1 * b2 + a2 * b1) / (b1 * b2)`.
+The sum of two rational numbers `r₁ = a₁/b₁` and `r₂ = a₂/b₂` is `r₁ + r₂ = a₁/b₁ + a₂/b₂ = (a₁ * b₂ + a₂ * b₁) / (b₁ * b₂)`.
 
-The difference of two rational numbers `r1 = a1/b1` and `r2 = a2/b2` is `r1 - r2 = a1/b1 - a2/b2 = (a1 * b2 - a2 * b1) / (b1 * b2)`.
+The difference of two rational numbers `r₁ = a₁/b₁` and `r₂ = a₂/b₂` is `r₁ - r₂ = a₁/b₁ - a₂/b₂ = (a₁ * b₂ - a₂ * b₁) / (b₁ * b₂)`.
 
-The product (multiplication) of two rational numbers `r1 = a1/b1` and `r2 = a2/b2` is `r1 * r2 = (a1 * a2) / (b1 * b2)`.
+The product (multiplication) of two rational numbers `r₁ = a₁/b₁` and `r₂ = a₂/b₂` is `r₁ * r₂ = (a₁ * a₂) / (b₁ * b₂)`.
 
-Dividing a rational number `r1 = a1/b1` by another `r2 = a2/b2` is `r1 / r2 = (a1 * b2) / (a2 * b1)` if `a2 * b1` is not zero.
+Dividing a rational number `r₁ = a₁/b₁` by another `r₂ = a₂/b₂` is `r₁ / r₂ = (a₁ * b₂) / (a₂ * b₁)` if `a₂ * b₁` is not zero.
 
-Exponentiation of a rational number `r = a/b` to a non-negative integer power `n` is `r^n = (a^n)/(b^n)`.
+Exponentiation of a rational number `r = a/b` to a non-negative integer power `n` is `rⁿ = aⁿ/bⁿ`.
 
-Exponentiation of a rational number `r = a/b` to a negative integer power `n` is `r^n = (b^m)/(a^m)`, where `m = |n|`.
+Exponentiation of a rational number `r = a/b` to a negative integer power `n` is `rⁿ = bᵐ/aᵐ`, where `m = |n|`.
 
-Exponentiation of a rational number `r = a/b` to a real (floating-point) number `x` is the quotient `(a^x)/(b^x)`, which is a real number.
+Exponentiation of a rational number `r = a/b` to a real (floating-point) number `x` is the quotient `aˣ/bˣ`, which is a real number.
 
-Exponentiation of a real number `x` to a rational number `r = a/b` is `x^(a/b) = root(x^a, b)`, where `root(p, q)` is the `q`th root of `p`.
+Exponentiation of a real number `x` to a rational number `r = a/b` is `x^(a/b) = root(xᵃ, b)`, where `root(p, q)` is the `q`th root of `p`.
 
 Implement the following operations:
  - addition, subtraction, multiplication and division of two rational numbers,
--- a/exercises/rational-numbers/description.md
+++ b/exercises/rational-numbers/description.md
@@ -10,13 +10,13 @@ The product (multiplication) of two rational numbers `r₁ = a₁/b₁` and `r�
 
 Dividing a rational number `r₁ = a₁/b₁` by another `r₂ = a₂/b₂` is `r₁ / r₂ = (a₁ * b₂) / (a₂ * b₁)` if `a₂ * b₁` is not zero.
 
-Exponentiation of a rational number `r = a/b` to a non-negative integer power `n` is `rⁿ = aⁿ/bⁿ`.
+Exponentiation of a rational number `r = a/b` to a non-negative integer power `n` is `r^n = a^n/b^n`.
 
-Exponentiation of a rational number `r = a/b` to a negative integer power `n` is `rⁿ = bᵐ/aᵐ`, where `m = |n|`.
+Exponentiation of a rational number `r = a/b` to a negative integer power `n` is `r^n = b^m/a^m`, where `m = |n|`.
 
-Exponentiation of a rational number `r = a/b` to a real (floating-point) number `x` is the quotient `aˣ/bˣ`, which is a real number.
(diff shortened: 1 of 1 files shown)


coriolinus:
Tests I have performed, checked if the numbers display correctly

- [x] Firefox on Windows
- [x] Notepad on Windows
- [x] `cat` on WSL terminal on Windows
- [x] `bat` on WSL terminal on Windows
- [ ] `type` on Windows 10 Command Prompt

Tests that I have not performed, which would be of interest

- OSX terminal
- Ubuntu terminal
- Android 
- Iphone / Ipad

I assume that most browsers and coding editors will properly support unicode characters; I think the testing priority should be terminals. Still, tests of all kinds are of interest.

The only failure so far has been using `type` on the command prompt in Windows. I imagine that this represents a small enough portion of the user population that we can accept that failure, particularly as notepad succeeds and is built into that OS. Data would be helpful, but may be unavailable.

I'm a bit concerned about readability, particularly for the superscript characters. Perhaps I am abnormal, but I keep my terminal font sized as small as is comfortable to read for an extended time. What this means is that `b^m` is comfortable for me to read, but `bᵐ` is not. Yes, it's easy to just expand the text in most environments these days, but I'd still prefer to avoid the necessity.

My personal inclination is against this PR, with regret. It's nice to have real mathematical notation, but it's even nicer to be able to read the text without squinting; the readability concern--which, as previously stated, may exist only for me--tips my vote.

SaschaMann:
Thanks a lot for testing those!

> The only failure so far has been using type on the command prompt in Windows. I imagine that this represents a small enough portion of the user population that we can accept that failure, particularly as notepad succeeds and is built into that OS. Data would be helpful, but may be unavailable.

The Windows command prompt is being replaced with a terminal that has much better unicode support, so this will mostly affect legacy Windows systems in the not so distant future.

> I'm a bit concerned about readability, particularly for the superscript characters. [...] the readability concern--which, as previously stated, may exist only for me--tips my vote.

Would you be fine if only the subscript characters were introduced? I mostly care about `b1` being replaced with `b₁`, `^m` is less of an issue imo and I'd be more than happy to drop those changes if that means the rest would be accepted.

rpottsoh:
I messed around with the file in Windows and the best end result I ended up with is all the subscripts reverted back to pre-PR state.
- If I list the raw file in Chrome browser window and copy and paste file contents to Notepad the contents of the file in Notepad appear correct.  However when I save the file Notepad puts up a prompt that unless I change the encoding to UTF8 the file will be stored as ANSI; I don't see an obvious way to make this suggested setting change in Notepad.
- I next tried pasting the file into Notepad++, none of the subscripts appeared correctly.  With more tinkering maybe this could be made to work.
- My final test was to copy the console to a file.  At the command prompt I executed `copy con filename.md` and then proceeded to paste from the web browser the raw contents of the file.  In the cmd window the contents appeared correctly.  Upon closing the file and I then executed `type filename.md` the results were ANSI :(   

It may be possible to have the changes appear correctly in Windows but it is cut and dry.....

Personally speaking I am OK with the changes.

kotp:
Notepad on Windows seems to do OK, from reports.

coriolinus:
> Would you be fine with the PR if only the subscript characters were introduced?

Yes, I would.

SaschaMann:
> 
> 
> > Would you be fine with the PR if only the subscript characters were introduced?
> 
> Yes, I would.

I've removed them.

kotp:
@coriolinux and @SaschaMann Do you think we should convert this to draft as well?

rpottsoh:
I might have to reverse my positive opinion on this change because #441.  It speaks to test cases but I think is applicable here.

SaschaMann:
> I would, however, object to the removal of the parentheses on lines 13, 15, and 17, especially while similar use of parentheses remains on 5,7,9, and 11 ... while I can tell at a glance that exponentiation binds before division, that's simply not obvious to all our students, many of whom already struggle with the more "mathy" exercises.

I forgot to undo that when removing the superscripts. Thanks for pointing that out.

exercises/rational-numbers/description.md
```diff
@@ -2,19 +2,19 @@ A rational number is defined as the quotient of two integers `a` and `b`, called
...
+The difference of two rational numbers `r₁ = a₁/b₁` and `r₂ = a₂/b₂` is `r₁ - r₂ = a₁/b₁ - a₂/b₂ = (a₁ * b₂ - a₂ * b₁) / (b₁ * b₂)`.
 
-The product (multiplication) of two rational numbers `r1 = a1/b1` and `r2 = a2/b2` is `r1 * r2 = (a1 * a2) / (b1 * b2)`.
+The product (multiplication) of two rational numbers `r₁ = a₁/b₁` and `r₂ = a₂/b₂` is `r₁ * r₂ = (a₁ * a₂) / (b₁ * b₂)`.
 
-Dividing a rational number `r1 = a1/b1` by another `r2 = a2/b2` is `r1 / r2 = (a1 * b2) / (a2 * b1)` if `a2 * b1` is not zero.
+Dividing a rational number `r₁ = a₁/b₁` by another `r₂ = a₂/b₂` is `r₁ / r₂ = (a₁ * b₂) / (a₂ * b₁)` if `a₂ * b₁` is not zero.
 
-Exponentiation of a rational number `r = a/b` to a non-negative integer power `n` is `r^n = (a^n)/(b^n)`.
+Exponentiation of a rational number `r = a/b` to a non-negative integer power `n` is `r^n = a^n/b^n`.
```
SaschaMann, line 13:
```suggestion
Exponentiation of a rational number `r = a/b` to a non-negative integer power `n` is `r^n = (a^n)/(b^n)`.
```

exercises/rational-numbers/description.md
```diff
@@ -2,19 +2,19 @@ A rational number is defined as the quotient of two integers `a` and `b`, called
...
 
-Exponentiation of a rational number `r = a/b` to a negative integer power `n` is `r^n = (b^m)/(a^m)`, where `m = |n|`.
+Exponentiation of a rational number `r = a/b` to a negative integer power `n` is `r^n = b^m/a^m`, where `m = |n|`.
```
SaschaMann, line 15:
```suggestion
Exponentiation of a rational number `r = a/b` to a negative integer power `n` is `r^n = (b^m)/(a^m)`, where `m = |n|`.
```

exercises/rational-numbers/description.md
```diff
@@ -2,19 +2,19 @@ A rational number is defined as the quotient of two integers `a` and `b`, called
...
 
-Exponentiation of a rational number `r = a/b` to a real (floating-point) number `x` is the quotient `(a^x)/(b^x)`, which is a real number.
+Exponentiation of a rational number `r = a/b` to a real (floating-point) number `x` is the quotient `a^x/b^x`, which is a real number.
```
SaschaMann, line 17:
```suggestion
Exponentiation of a rational number `r = a/b` to a real (floating-point) number `x` is the quotient `(a^x)/(b^x)`, which is a real number.
```

exercises/rational-numbers/description.md
```diff
@@ -2,13 +2,13 @@ A rational number is defined as the quotient of two integers `a` and `b`, called
...
+The sum of two rational numbers `r₁ = a₁/b₁` and `r₂ = a₂/b₂` is `r₁ + r₂ = a₁/b₁ + a₂/b₂ = (a₁ * b₂ + a₂ * b₁) / (b₁ * b₂)`.
 
-The difference of two rational numbers `r1 = a1/b1` and `r2 = a2/b2` is `r1 - r2 = a1/b1 - a2/b2 = (a1 * b2 - a2 * b1) / (b1 * b2)`.
+The difference of two rational numbers `r₁ = a₁/b₁` and `r₂ = a₂/b₂` is `r₁ - r₂ = a₁/b₁ - a₂/b₂ = (a₁ * b₂ - a₂ * b₁) / (b₁ * b₂)`.
 
-The product (multiplication) of two rational numbers `r1 = a1/b1` and `r2 = a2/b2` is `r1 * r2 = (a1 * a2) / (b1 * b2)`.
+The product (multiplication) of two rational numbers `r₁ = a₁/b₁` and `r₂ = a₂/b₂` is `r₁ * r₂ = (a₁ * a₂) / (b₁ * b₂)`.
 
-Dividing a rational number `r1 = a1/b1` by another `r2 = a2/b2` is `r1 / r2 = (a1 * b2) / (a2 * b1)` if `a2 * b1` is not zero.
+Dividing a rational number `r₁ = a₁/b₁` by another `r₂ = a₂/b₂` is `r₁ / r₂ = (a₁ * b₂) / (a₂ * b₁)` if `a₂ * b₁` is not zero.
```
SaschaMann, line 11:
```suggestion
Dividing a rational number `r₁ = a₁/b₁` by another `r₂ = a₂/b₂` is `r₁ / r₂ = (a₁ * b₂) / (a₂ * b₁)` if `a₂` is not zero.
```

This might belong in a different PR, so I'll just leave it as a suggestion. If `b₁ = 0`, then `r₁` is not a rational number anyway.

iHiD:
Thanks for the tag, @yawpitch. So, I'm happy for this to be merged in terms of #1560. But I have zero knowledge of the impact of this across systems. I'm happy that the website itself can render it, but can people read it when downloaded across systems (what @yawpitch said basically)? So, who here knows about this stuff and how it might be problematic? And then, after highlighting ways it might go wrong, how can we test the impact of this in READMEs across different systems? Get a Docker of Windows 3.1 and roll the dice? :) I'll need one of you to give me more guidance here, but I **would** like to get this merged if we can prove it's non-damaging.

coriolinus:
I've seen several suggestions for targets which should be tested before this is merged. Collecting them:

- OSX terminal
- Ubuntu terminal
- Android
- Iphone / Ipad
- screen readers (windows/mac/linux default ones?)
- Old Windows (via Docker or vmware, probably)

I think it is a good idea to test all of these platforms. However, I don't think any single failure other than the screen readers should automatically veto this PR; what matters isn't so much the number of systems for which this doesn't work, as the number of people who can't get a legible representation of the file.

I'm not convinced by the argument that people using low-level or old languages will be unwilling or unable to use modern tooling: if I were to attempt to pick up cobol or fortran, my first step after installing the compiler would be to install the appropriate plugin into VS Code, which is known to display these characters properly. 

> how can we test the impact of this in READMEs across different systems

This is the crux of the problem. I can see what amounts to two approaches possible:

- Perform a study: create an exercise called "Exercism Survey" with a README containing unicode, and a single test, whose name is "any character in the README render improperly for you", and which accepts any boolean. Inject it into all tracks, unlocked immediately. Collect the results, tabulate them once a few hundred have come in. This inconveniences no students, and someone could probably publish a legitimate paper on "Prevalence of Unicode-Compatible Editors In Programming Students" from the data. It would take non-trivial work from all maintainers, though.

- YOLO it: test the matrix of viewers that we can think of. If sufficient of them pass, merge this PR, wait for bug reports. If we see, say, half a percent of students filing bug reports about this, we can revert the PR.

My personal preference leans toward the second option, because I suspect that the number of people truly inconvenienced by this PR will be exceedingly low.

iHiD:
Yeah - I'm all up for YOLO here and agree with your post in general, @coriolinus. As long as we get some breadth of types, I'm good to go.

So everyone, let's call [this list](https://github.com/exercism/problem-specifications/pull/1655#issuecomment-623762059) canonical. If there are things missing from there, please say and @coriolinus can add. If you can test one (positively or negatively) please say and @coriolinus can update.

SaschaMann:
> Android
> Iphone / Ipad

The CLI doesn't support these systems, so presumably they'd have to go through a browser to download them?

They are displayed without issues in Firefox on Android for me, however in mobile Chrome, only the <sub>2</sub> renders, the <sub>1</sub> is a block (wat?). I'm not sure how relevant that is to the overall distribution. @iHiD since mobile Chrome doesn't support adblockers, Google Analytics probably has an accurate picture of how many people that would affect (if any).

iHiD:
> They are displayed without issues in Firefox on Android for me, however in mobile Chrome, only the 2 renders, the 1 is a block (wat?). I'm not sure how relevant that is to the overall distribution. 

So. This is probably/maybe/possible because each different OS renders its native font for v2. In v3, this will probably change.

> @iHiD since mobile Chrome doesn't support adblockers, Google Analytics probably has an accurate picture of how many people that would affect (if any).

11% of our traffic is mobile Chrome.

yawpitch:
We _might_ want to consider Linux distros outside Ubuntu ... I run Arch and X and the subscripts render fine (if **extremely** small) in my virtual terminal in vi, nano, etc ... but if I drop out of X and use the system console they all render as unknown blocks.

Now, admittedly, _most_ people I know use X or Wayland, but I do know people who boot to the console and only kick of an X session when they're browsing or otherwise forces into a GUI.

SaschaMann:
@iHiD is this covered by #1686's "All PRs that were on hold can be merged (many will probably need updating in line with immutability)."?

iHiD:
Yes. They still need the there green ticks though.

@rpottsoh - you reversed your approval. I'm personally ok with this because I think the reason that #441 made those rules was for the languages that have implement the test cases, not the READMEs themselves. Do you still feel this is inappropraite?

SaschaMann:
@iHiD it's been more than 14 days, can we merge this?

iHiD:
I'm going to say yes, and merge. But reserve the right to revert if we decide it was a bad idea in the future :)

PR #1656, opened 2020-05-08
Scale-Generator: Update description.md

aimorris:


exercises/scale-generator/description.md
```diff
@@ -16,28 +16,28 @@ A, Bb, B, C, Db, D, Eb, E, F, Gb, G, Ab
 The major and minor scale and modes are subsets of this twelve-pitch
 collection. They have seven pitches, and are called diatonic scales.
 The collection of notes in these scales is written with either sharps or
-flats, depending on the tonic. Here is a list of which are which:
+flats, depending on the tonic. Here is a list of which use sharps, flats, or neither:
```
iHiD, line 19:
```suggestion
flats, depending on the tonic:
```

aimorris:
Ahh so that's why they were in lower-case. It was completely unclear to me and looked just like typing errors.

aimorris:
Perhaps the description should mention this?

yawpitch:
> Perhaps the description should mention this?

@aimorris I'd be all for that ... since the rest of the changes in this PR are quite sensible, why don't you add some language to that effect? It just needs to make it clear that the case is relevant and that `F#` indicates that the student should return a major scale while `f#` indicates they should return a minor scale.

yawpitch:
BTW, we should probably reject exercism/javascript#860 because of the case-sensitivity issue.

SleeplessByte:
I purposefully did the do-not-merge because of that :)

ErikSchierboom:
Looking at the [current description](https://github.com/dzegel/exercism-problem-specifications/blob/main/exercises/scale-generator/description.md), I think we now have this information presented in a way that this PR can be closed.

Let me know if someone disagrees, and we can re-open again.

PR #1657, opened 2020-05-08
Add testcases to matching-brackets

ironhaven:
These are up-streamed test cases from exercism/rust#946 to catch bugs discovered in this https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=1661c960aaf13f061dbeb747a587cece matching-brackets rust implementation that where not caught with the preexisting testcases.

As explained in the [other pull request](https://github.com/exercism/rust/pull/946) these test cases handle bugs related to implementations not quitting when there is a error in situations such as isolated mismatched brackets or closing brackets when the stack is empty.

### DIFF
--- a/exercises/matching-brackets/canonical-data.json
+++ b/exercises/matching-brackets/canonical-data.json
@@ -137,6 +137,24 @@
       "expected": false
     },
     {
+      "uuid": "a345a753-d889-4b7e-99ae-34ac85910d1a",
+      "description": "early unexpected brackets",
+      "property": "isPaired",
+      "input": {
+        "value": ")()"
+      },
+      "expected": false
+    },
+    {
+      "uuid": "21f81d61-1608-465a-b850-baa44c5def83",
+      "description": "early mismatched brackets",
+      "property": "isPaired",
+      "input": {
+        "value": "{)()"
+      },
+      "expected": false
+    },
+    {
       "uuid": "99255f93-261b-4435-a352-02bdecc9bdf2",
       "description": "math expression",
       "property": "isPaired",


ErikSchierboom:
Looking at the [latest canonical data](https://github.com/exercism/problem-specifications/blob/main/exercises/matching-brackets/canonical-data.json), it looks like there already are test cases that verify the behavior this PR adds:

1. Start with closing bracket: https://github.com/exercism/problem-specifications/blob/main/exercises/matching-brackets/canonical-data.json#L36
2. Wrong closing bracket: https://github.com/exercism/problem-specifications/blob/main/exercises/matching-brackets/canonical-data.json#L42

Therefore, I think this PR can be closed. @ironhaven thanks for wanting to help and sorry that we've not looked at this PR for a while. If you feel like my above reasons are not valid, do let me know.

petertseng:
I can verify that there exists a **possible** implementation that fails only these two new test cases, getting the expected result on all other test cases in current canonical data. The implementation is a little rare, so I will leave it to y'all to judge whether it bears adding these two new test cases. The implementation has all of these behaviours:

* Push opening brackets onto a stack.
* A closing bracket always pops a character from that stack
* If the final character is a closing bracket, the final result is based solely on whether that final closing bracket matched the character it popped, **without regard to whether the stack is empty at that time**
* If the final character is any other character, the final result is based solely on whether the stack is empty at that time, **without regard to any errors that may or may not have happened prior**

An additional case that such an implementation would call well-balanced is `"[} 2"`.

Here is code for such an implementation

```ruby
      stack = []
      i['value'].each_char.reduce(true) { |_, c|
        if (o = BracketMatcher::CLOSES[c])
          stack.pop == o
        else
          stack << c if BracketMatcher::PAIRS.has_key?(c)
          stack.empty?
        end 
      }
```

petertseng (approved):
I support adding, but the implementation being tested for is really quite rare. Most common incorrect implementations are already handled by other tests. But I don't think extra cases are much of a burden(most implementations will just pass them without having to do any extra work), so I figure might as well add.

ErikSchierboom (approved):
@petertseng Thanks for digging into this. Having read your thoughts, I'm fine with adding these test cases.

wolf99 (approved):

PR #1658, opened 2020-05-25
Pangram: check case insensitive behavior

Wikunia:
I received one solution where it was only checked whether there are exactly 26 letters (case sensitive)
This would result in that `a...m A...M` is a pangram for example.
Also fails if there are more than 26 letters when counting i.e o and O as two different ones.

### DIFF
--- a/exercises/pangram/canonical-data.json
+++ b/exercises/pangram/canonical-data.json
@@ -4,7 +4,7 @@
     "A pangram is a sentence using every letter of the alphabet at least once.",
     "Output should be a boolean denoting if the string is a pangram or not."
   ],
-  "version": "2.0.0",
+  "version": "2.1.0",
   "cases": [
     {
       "description": "empty sentence",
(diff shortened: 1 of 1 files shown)


Wikunia:
One could also change the existing test for it:
```
the quick brown fox jumped over the lazy FOX
```
to 
```
the quick brown fox jumps over the lazy FOX
```

iHiD:
@cmcaine FYI, this should not have been merged. It's blocked under #1560. 

As a general rule, for anything in this repo, it's worth getting a sign off from at least two maintainers, as changes affect all tracks on the site.

@exercism/track-maintainers Anyone want this reverting or is everyone ok with this as an honest error? I'll revert it based on a single objection.

cmcaine:
Woops! Sorry

yawpitch:
I’m not officially here, so ignore this if anyone wants to counterpoint me if you’d like, but if I were here I’d object on general principle; we’ve held off on merging equally (if not more) valuable additions because of #1560. These are missing tests, not bugs.

Also I don’t believe the second test adds anything (it’s another iteration on the “the quick brown fox[...]” known pangram with surplus characters) while also implying that a case sensitive solution is in any way correct.

The first test is obviously valuable because it heads off a student misunderstanding the exercise spec, but one way or another a case sensitive simple count should be caught by a mentor even if the tests pass. Our test suites typically represent minimal functionality without any real attempt to head off all possible edge cases, so it’s always going to be possible to write a test-passing but still insufficient solution, at the very least by a simple conditional that responds to every test input with its appropriate output (which I’ve seen at least a hundred times for, among others, `bob`).

But, again, not really here. Tried to turn off all notifications, apparently can’t stop group-wide ones. Pay no attention to the ghost. Woo.

iHiD:
Reverted and reopened as #1660

SaschaMann:
> As a general rule, for anything in this repo, it's worth getting a sign off from at least two maintainers, as changes affect all tracks on the site.

Maybe we could bump the branch protection to 3 required approvals to make this an explicit rule rather than a convetion?

iHiD:
@SaschaMann Done.

@Wikunia Sorry you got bounced around a bit here. Let's continue the discussion here: https://github.com/exercism/problem-specifications/pull/1660

@cmcaine No problem. These things happen :) I appreciate your desire to be proactive! :)

PR #1660, opened 2020-05-25
Pangram: check case insensitive behavior

iHiD:
This is a re-opening of https://github.com/exercism/problem-specifications/pull/1658

### DIFF
--- a/exercises/pangram/canonical-data.json
+++ b/exercises/pangram/canonical-data.json
@@ -97,6 +97,7 @@
     },
     {
       "uuid": "7138e389-83e4-4c6e-8413-1e40a0076951",
+      "reimplements": "2577bf54-83c8-402d-a64b-a2c0f7bb213a",
       "description": "a-m and A-M are 26 different letters but not a pangram",
       "property": "isPangram",
       "input": {
--- a/exercises/pangram/canonical-data.json
+++ b/exercises/pangram/canonical-data.json
@@ -98,7 +98,7 @@
     {
       "uuid": "7138e389-83e4-4c6e-8413-1e40a0076951",
       "reimplements": "2577bf54-83c8-402d-a64b-a2c0f7bb213a",
-      "description": "a-m and A-M are 26 different letters but not a pangram",
+      "description": "a-m and A-M are 26 different characters but not a pangram",
       "property": "isPangram",
       "input": {
         "sentence": "abcdefghijklm ABCDEFGHIJKLM"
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
I've rebased this on the latest main and made one change, inspired by https://github.com/exercism/problem-specifications/pull/1658#issuecomment-633702171: removing the second test case.

CC @exercism/reviewers

petertseng (approved):
I'm not huge on the necessity of this case, because supposedly case 2577bf54-83c8-402d-a64b-a2c0f7bb213a "case insensitive" already handles it.

```
$ ruby -e 'p ("the quick brown fox jumps over with lazy FX".chars - [" "]).uniq.size' 
26
```

But, this is certainly a simpler and easier-to-understand way of expressing the same thing. Maybe it could even go before case 2577bf54-83c8-402d-a64b-a2c0f7bb213a "case insensitive" ...

exercises/pangram/canonical-data.json
```diff
@@ -94,6 +94,15 @@
         "sentence": "the quick brown fox jumps over with lazy FX"
       },
       "expected": false
+    },
+    {
+      "uuid": "7138e389-83e4-4c6e-8413-1e40a0076951",
+      "description": "a-m and A-M are 26 different letters but not a pangram",
```
ee7, line 100:
```suggestion
      "uuid": "7138e389-83e4-4c6e-8413-1e40a0076951",
      "reimplements": "2577bf54-83c8-402d-a64b-a2c0f7bb213a",
      "description": "a-m and A-M are 26 different letters but not a pangram",
```

Would this make sense?

The sentence for that test case is pretty weird:
> "the quick brown fox jumps over with lazy FX"

and makes the reader wonder about "with" and "FX".

So if we think they test the same thing, we could just replace it.

ErikSchierboom: I like that.

petertseng (approved):
I think they test the same thing. Replacing seems a fine idea.

SleeplessByte (approved):

exercises/pangram/canonical-data.json
```diff
@@ -94,6 +94,16 @@
         "sentence": "the quick brown fox jumps over with lazy FX"
       },
       "expected": false
+    },
+    {
+      "uuid": "7138e389-83e4-4c6e-8413-1e40a0076951",
+      "reimplements": "2577bf54-83c8-402d-a64b-a2c0f7bb213a",
+      "description": "a-m and A-M are 26 different letters but not a pangram",
```
ee7, line 101:
```suggestion
      "description": "a-m and A-M are 26 different characters but not a pangram",
```

Let's avoid "letters" here? 'a' and 'A' are the same letter.

ee7 (approved):

SleeplessByte:
Thanks all!

PR #1662, opened 2020-06-02
Palindrome Products: refine the problem

DmitrySamoylov:
For me it took a while to understand that we calculate product of exactly two numbers. BTW in the [original problem](https://projecteuler.net/problem=4) it is stated clearly.


iHiD:
@DmitrySamoylov Thanks :)

PR #1663, opened 2020-06-11
robot-name: remove dead link

cmcaine:
gSchool.it is now a domain parking spot

### DIFF
--- a/exercises/robot-name/metadata.yml
+++ b/exercises/robot-name/metadata.yml
@@ -1,4 +1,3 @@
 ---
 blurb: "Manage robot factory settings."
 source: "A debugging session with Paul Blackwell at gSchool."
-source_url: "http://gschool.it"


rpottsoh:
@iHiD, it is strange that GitHub is still expecting for the check to be completed but when I check Jenkins CI I see that the build was green.  In a situation like this usually something went wrong with the build and I can just restart it and everything works itself out.  This time I am not seeing a way to restart the build (assuming doing so will "fix" this).

exercises/robot-name/metadata.yml
```diff
@@ -1,4 +1,3 @@
 ---
 blurb: "Manage robot factory settings."
 source: "A debugging session with Paul Blackwell at gSchool."
-source_url: "http://gschool.it"
```
rpottsoh, line 4: This line maybe should be
```
source_url: ""
```
But I am not positive.

iHiD:
Closed and reopened the PR to restart the build.

iHiD:
@rpottsoh Looks like this has worked :)

rpottsoh:
@ihid, indeed.  I was looking for some documentation on the metadata.yml file and I am not finding anything.  @cmcaine's change may be fine as is, I just am not sure....

In any event I don't think this PR should be blocked.

iHiD:
I'm unsure.

SaschaMann:
There are various exercises without it (GH search only finds 98 files with `source_url` in this repo), so it should be fine, I think. E.g. [all-your-base](https://github.com/exercism/problem-specifications/blob/master/exercises/all-your-base/metadata.yml).

ErikSchierboom:
I'm almost certain the `source_url` can be left out.

PR #1665, opened 2020-08-02
linked-list: Clarified that existing functions should not be used

DanielNoord:
Added a line that tells the student not to use existing functions, similar to the problem specification in: 
https://github.com/exercism/problem-specifications/blob/master/exercises/list-ops/description.md
This was unclear for me at the time of making the exercise initially.

This is my first pull request so please do tell if I should have done this differently or opened an issue first!

### DIFF
--- a/exercises/linked-list/description.md
+++ b/exercises/linked-list/description.md
@@ -9,10 +9,11 @@ A linked list is a collection of data elements called *nodes*. In a
 In a *doubly linked list* each node also holds a link to the previous
 node.
 
-You will write an implementation of a doubly linked list. Implement a
-Node to hold a value and pointers to the next and previous nodes. Then
-implement a List which holds references to the first and last node and
-offers an array-like interface for adding and removing items:
+You will write an implementation of a doubly linked list without using 
+existing functions. Implement a Node to hold a value and pointers to 
+the next and previous nodes. Then implement a List which holds references 
+to the first and last node and offers an array-like interface for 
+adding and removing items:
 
 * `push` (*insert value at back*);
 * `pop` (*remove value at back*);


coriolinus:
In functional languages, it is very common for operations like `foldl`, `map`, `length`, and the like to be pre-implemented and part of the language prelude. It makes sense to tell the student not to simply delegate to the built-in implementations.

It's less clear that this applies for doubly-linked lists. I could agree with a statement like "If your language includes a doubly-linked list implementation, do not use that; write your own." However, it feels overly broad to prohibit use of all existing functions.

DanielNoord:
I understand your point. Is it possible to modify track-specific specifications? Or is that frowned upon?

My only experience is with the Python track and there I understood what was expected of me only after looking at other people's solutions. That seems counter-intuitive as the problem specification should make this clear from the start.

coriolinus:
It is both possible and recommended to edit track-specific things first, then bring them to the wider community if they seem more broadly applicable. In this case, you probably want https://github.com/exercism/python.

cmccandless:
The implicit instruction in _every_ exercise is to write your own solution, not to use an existing class. Similarly to @coriolinus' point, you must be allowed to use at least a basic subset of the language features to accomplish this, but the idea is to write it yourself, not `import solution`.

iHiD:
Hi @DanielNoord. Thank you for the suggestion. Opening a PR is great :) 

As the others have said, this sounds like something worth doing on a track-specific level, or actually probably a wider problem with Exercism not making clear the "rules" around submitting things (as @cmccandless alludes to). I think my take on this is that it's worth us having a "guidelines" doc for people using Exercism which explains these sorts of things, which we point people to, rather than adding the line in every exercise. Thoughts on that?

cmccandless:
Sounds like a great idea!

Issue #1668, opened 2020-08-22
Error texts for "largest series product" is misleading


wederbrand:
The problem description, README.md, is missing information on that zero length spans should return a product of 1, yet there is an error that is expected that says "span must be greater than zero" when in reallity spans can't be negative.

PR #1669, opened 2020-08-22
Largest-Series-Product: Change error text for test case where span is negative.

wederbrand:
Solves #1668

### DIFF
--- a/exercises/largest-series-product/canonical-data.json
+++ b/exercises/largest-series-product/canonical-data.json
@@ -178,6 +178,18 @@
         "span": -1
       },
       "expected": {"error": "span must be greater than zero"}
+    },
+    {
+      "uuid": "c859f34a-9bfe-4897-9c2f-6d7f8598e7f0",
+      "description": "rejects negative span",
+      "reimplements": "5fe3c0e5-a945-49f2-b584-f0814b4dd1ef",
+      "comments": ["Previous error message was incorrect"],
+      "property": "largestProduct",
+      "input": {
+        "digits": "12345",
+        "span": -1
+      },
+      "expected": {"error": "span must not be negative"}
     }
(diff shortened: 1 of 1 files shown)


exercises/largest-series-product/canonical-data.json
```diff
@@ -1,6 +1,6 @@
 {
   "exercise": "largest-series-product",
-  "version": "1.2.0",
+  "version": "1.2.1",
```
cmccandless, line 3: Would appreciate if another maintainer can chime in here and corroborate (or contradict) me here, but since some tracks actually use the error text here, I believe this should be a minor version bump, not a patch.
```suggestion
  "version": "1.3.0",
```

rpottsoh: I re-read the [README](https://github.com/exercism/problem-specifications#test-data-versioning) and I agree, the version should be `"1.3.0"`.

ErikSchierboom:
@wederbrand Are you still interested in working on this PR?

petertseng (dismissed):
be nice to get this in since the existing error is clearly incorrect. Can always correct this at the track level but better to fix the problem at the source.

BethanyG (approved):

junedev (approved):

junedev:
To clarify why the CI is failing: Instead of changing the text in the existing test case, a new one with the needs to be created with the "reimplements" key.

petertseng:
I even considered ignoring that rule and merging it anyway, but actually reimplementing it will have a greater chance of encouraging tracks to change the message, so I think it is better to reimplement.

kotp (approved):

kotp:
Thanks for the work @petertseng and the pause for review.

Issue #1674, opened 2020-09-24
Reopening the Problem Specifications repo


ErikSchierboom:
**TL;DR;** @exercism/track-maintainers We are planning on reopening problem specs in the next couple of weeks. The key change we are making is that problem-specs should be thought of as set of optional test cases. Individual tracks should choose which test cases to implement on a per-test case basis. We are adding a simple, per-exercise configuration file that keeps track of which test cases the exercise implements. We'll provide a simple mechanism for tracks to keep this file up-to-date with the test cases in the exercise's canonical data. The only change required for track maintainers is to update their test-generators to generate tests for just the test cases enabled in the configuration file. All test cases will be immutable; changes can only be introduced by adding new test cases, which can be flagged as re-implementing an existing test case.

---

A year ago we temporarily put the Problem Specifications repo in bug-fixes only mode (see this issue). 
This was done as there were systematic problems with the Problem Specifications repo that were causing conflict between maintainers, which we did not have the resources to fix in the immediate term.
Now that we're making good progress on Exercism v3, we have been able to dedicate time to designing a solution to the problems in the Problem Specifications repo. 

In this issue I will outline the problems that we're trying to solve and the solution we have come up with. 
We are very excited to re-open the Problem Specifications repo, which has always been one of the most active parts of Exercism and has brought a great deal of joy to many, and how these changes will allow us to do that. 
Let's get to it! 🎉

## Intro to Problem Specifications

The Problem Specifications repo describes a set of language-agnostic exercises, which maintainers can use to build exercises for their tracks. Over time, exercises have grown to not only contain shared titles and descriptions, but also canonical data that describes a set of test cases for the exercise. This setup has been very helpful for maintainers for tasks such as adding new exercises to tracks, updating existing existing exercises as well as bootstrapping entirely new tracks.

As the canonical data is defined as JSON data, some tracks have built tooling to automatically generate test suites from them. This has allowed these tracks to quickly scaffold test suites for new exercises, as well as fix bugs in existing (generated) test suites.

Unfortunately, as different tracks (and different maintainers) have different desires for how their tests are structured and written, this has caused tension. In this document, we'll describe what has caused this tension and how we aim to resolve this tension.

## Issues with Problem Specifications

Issues with the Problem Specifications repo occur when there is a 1-to-1 correspondence between an exercise defined in Problem Specifications and a track implementation of that exercise. In other words, if changes to the exercise data in Problem Specifications result in changes to a track's implementation, there can be trouble.

### README

[...]

PR #1676, opened 2020-09-30
Add uuid field to all test cases

ErikSchierboom:
This is the first step towards re-opening the problem specifications repo, as discussed in [this issue](https://github.com/exercism/problem-specifications/issues/1674). The PR does the following:

- Adds a (randomly generated) `uuid` field to all test cases
- Updates the JSON schema
- Updated the documentation

FYI @iHiD


README.md
```diff
@@ -128,6 +133,7 @@ There are also some conventions that must be followed:
     - The string should explain why the error would occur.
     - A particular track's implementation of the exercise **need not** necessarily check that the error includes that exact string as the cause, depending on what is idiomatic in the language (it may not be idiomatic to check strings for error messages).
   - Test cases that only some tracks should implement, for example because it would unnecessarily increase the complexity of the exercise in some but not all languages, mark it with an `optional`-key. Multiple cases related to the same reason for optionality should have the same key. The decision that a test case is optional will often be made in the PR discussion, so don't worry about it too much while creating a PR.
+  - Each test case must have a unique UUID specified in its `"uuid"` key. A UUID can be randomly generated using the [UUID Generator](https://www.uuidgenerator.net/version4).
```
SaschaMann, line 136: Why not `configlet uuid` instead?

canonical-schema.json
```diff
@@ -113,6 +114,12 @@
           , "type"       : "string"
           },
 
+      "uuid":
+          { "description": "A unique identifier"
```
SaschaMann, line 118:
```suggestion
          { "description": "A universally unique identifier in the v4 format"
```

`1` would also be a unique identifier.

README.md
ErikSchierboom: My reasoning was that `configlet` is harder to get setup locally. One has to download first, whereas using a website is easier. I can add a second note about using `configlet`?

SaschaMann: But the other doc says that we will require `configlet` for formatting anyway, so that setup cost is already there.

ErikSchierboom: > But the other doc says that we will require `configlet` for formatting anyway, so that setup cost is already there.

True, although we won't get there initially. `configlet` will be updated sometime later, so I think I'll leave the current comment in and then update it later once we've updated `configlet`.

SaschaMann: Presumably you added the UUIDs via a script, right? Could you maybe share it so that one can run `./add-uuids` to just deal with it as another alternative until it's properly automated?

ErikSchierboom: I added it with an F# script, so that's probably not ideal. That said, I _am_ working on tooling to automatically add missing uuids. I hope to finish this somewhere this week or the next. I'll then update the instructions.

iHiD (approved):
Not going to claim to have checked each uuid but it looks good! 🙂

iHiD:
Merging this to move us forward.

ErikSchierboom:
If you have some comments, do feel free to post them here.

PR #1677, opened 2020-09-30
Replace optional key with scenarios

ErikSchierboom:
This PR does not yet document the new `tests.toml` functionality and its tooling that we'll introduce to allow tracks to keep track of which tests are included/excluded.


README.md
```diff
@@ -96,8 +96,8 @@ is easier to understand with an example:
             , "expected"   : null
             }
           , { "uuid"       : "8790a635-e8a8-4343-a29f-7da2929b9378"
-            , "description": "Foo'ing a very big number returns nothing"
-            , "optional"   : "big-ints"
+            , "description": "Foo'ing a very big number returns nothing"
```
iHiD, line 99:
```suggestion
            , "description": "Foo'ing a very big number returns nothing"
```

canonical-schema.json
```diff
@@ -102,7 +102,7 @@
           , "required"   : ["description", "cases"]
           , "properties" :
                 { "description": { "$ref": "#/definitions/description" }
-                , "optional"   : { "$ref": "#/definitions/optional"    }
+                , "scenarios"  : { "$ref": "#/definitions/scenarios"   }
```
ErikSchierboom, line 105: For now, I've allowed the possiblity of also having scenarios at the test group level.

canonical-schema.json
```diff
@@ -120,10 +120,17 @@
...
           },
 
-      "optional":
-          { "description": "An identifier for similar optional test cases (kebab-case)"
-          , "type"       : "string"
-          , "pattern"    : "^[a-z]+(-[a-z]+)*$"
+      "scenario":
+          { "description": "An identifier for a specific scenario (kebab-case)"          
+          , "type": "string"
+          , "enum": ["big-integer", "floating-point", "library-test", "unicode"]
```
ErikSchierboom, line 126: I've taken the liberty of adding the actual values here, which makes the JSON schema validation a lot nicer.

SaschaMann: This seems easy to miss when adding a new scenario. Could we somehow merge both lists?

SaschaMann (approved):

canonical-schema.json
ErikSchierboom: I thought of maybe having a CI check to verify that the two match? The benefit of the file is that is easier to read. This has the benefit of actually verifying the actual scenario values.

README.md
```diff
@@ -4,11 +4,11 @@ Shared metadata for Exercism exercises.
...
 ## Contributing Guide
 
-Please see the [contributing guide](https://github.com/exercism/problem-specifications/blob/master/CONTRIBUTING.md)
+Please see the [contributing guide](./CONTRIBUTING.md).
 
-## Problem metadata
+## Exercise Metadata
 
-Each problem's data lives in a directory under `exercises/`
+Each exercise's metadata lives in a directory under `exercises/` .
```
kotp, line 11:
```suggestion
Each exercise's metadata lives in a directory under `exercises/`.
```

canonical-schema.json
SaschaMann: But a check works too, this isn't blocking

ErikSchierboom: > Can we regenerate the schema when the file changes? 

We could. Good idea.

kotp (approved):

iHiD (approved):

PR #1688, opened 2020-10-13
diffie-hellman: Reword 'in range 1 .. p'

ee7:
There are some languages where the syntax `1 .. p` includes the endpoints.

For example, the Nim code:
```Nim
for i in 1 .. 3:
  echo i
```

shows the output:
```
1
2
3
```

Therefore "in range 1 .. p" would have been especially confusing, given that it was intended to **exclude** the endpoints.

So let's use the wording from the `description.md`:
https://github.com/exercism/problem-specifications/blob/126568d7bddcb0696a3a4202030b240866a41889/exercises/diffie-hellman/description.md#L13-L14

### DIFF
--- a/exercises/diffie-hellman/canonical-data.json
+++ b/exercises/diffie-hellman/canonical-data.json
@@ -17,7 +17,7 @@
   "cases": [
     {
       "uuid": "1b97bf38-4307-418e-bfd2-446ffc77588d",
-      "description": "private key is in range 1 .. p",
+      "description": "private key is greater than 1 and less than p",
       "property": "privateKeyIsInRange",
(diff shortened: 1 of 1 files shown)


iHiD (approved):
Solid approve from me. I wouldn't have know what `range 1 .. p` meant. Same comment as https://github.com/exercism/problem-specifications/pull/1687#issuecomment-708358142 applies though - let's make this a standard somewhere.

SaschaMann (approved):

ErikSchierboom (approved):

ErikSchierboom:
Thanks @ee7! 🎉

PR #1691, opened 2020-10-13
perfect-numbers: 'natural number' -> 'positive integer'

ee7:
There is no convention whether zero is a natural number, so let's avoid
the issue.

For example, see:
- https://mathworld.wolfram.com/NaturalNumber.html
- https://en.wikipedia.org/wiki/Natural_number
- https://math.stackexchange.com/questions/283/is-0-a-natural-number

The previous wording of
```
  "description": "Zero is rejected (not a natural number)",
```
could be especially jarring for some tracks. For example, Nim defines
`Natural` and `Positive` types, and `Natural` includes 0. See:
- https://nim-lang.github.io/Nim/system.html#Natural

Note, the other exercises that mention "natural" are:

```
difference-of-squares/description.md
1:Find the difference between the square of the sum and the sum of the squares of the first N natural numbers.
3:The square of the sum of the first ten natural numbers is
6:The sum of the squares of the first ten natural numbers is
10:ten natural numbers and the sum of the squares of the first ten
11:natural numbers is 3025 - 385 = 2640.

difference-of-squares/metadata.yml
2:blurb: "Find the difference between the square of the sum and the sum of the squares of the first N natural numbers."

prime-factors/description.md
1:Compute the prime factors of a given natural number.

prime-factors/metadata.yml
2:blurb: "Compute the prime factors of a given natural number."

pythagorean-triplet/description.md
1:A Pythagorean triplet is a set of three natural numbers, {a, b, c}, for

spiral-matrix/description.md
3:The matrix should be filled with natural numbers, starting from 1

sum-of-multiples/description.md
4:If we list all the natural numbers below 20 that are multiples of 3 or 5,
```

### DIFF
--- a/exercises/perfect-numbers/canonical-data.json
+++ b/exercises/perfect-numbers/canonical-data.json
@@ -120,24 +120,24 @@
             "cases": [
                 {
                     "uuid": "72445cee-660c-4d75-8506-6c40089dc302",
-                    "description": "Zero is rejected (not a natural number)",
+                    "description": "Zero is rejected (as it is not a positive integer)",
                     "property": "classify",
                     "input": {
                         "number": 0
                     },
                     "expected": {
-                        "error": "Classification is only possible for natural numbers."
+                        "error": "Classification is only possible for positive integers."
                     }
                 },
                 {
                     "uuid": "2d72ce2c-6802-49ac-8ece-c790ba3dae13",
-                    "description": "Negative integer is rejected (not a natural number)",
+                    "description": "Negative integer is rejected (as it is not a positive integer)",
                     "property": "classify",
                     "input": {
                         "number": -1
                     },
                     "expected": {
-                        "error": "Classification is only possible for natural numbers."
+                        "error": "Classification is only possible for positive integers."
                     }
                 }
             ]
--- a/exercises/perfect-numbers/description.md
+++ b/exercises/perfect-numbers/description.md
@@ -1,7 +1,7 @@
 Determine if a number is perfect, abundant, or deficient based on
-Nicomachus' (60 - 120 CE) classification scheme for natural numbers.
+Nicomachus' (60 - 120 CE) classification scheme for positive integers.
 
-The Greek mathematician [Nicomachus](https://en.wikipedia.org/wiki/Nicomachus) devised a classification scheme for natural numbers, identifying each as belonging uniquely to the categories of **perfect**, **abundant**, or **deficient** based on their [aliquot sum](https://en.wikipedia.org/wiki/Aliquot_sum). The aliquot sum is defined as the sum of the factors of a number not including the number itself. For example, the aliquot sum of 15 is (1 + 3 + 5) = 9
+The Greek mathematician [Nicomachus](https://en.wikipedia.org/wiki/Nicomachus) devised a classification scheme for positive integers, identifying each as belonging uniquely to the categories of **perfect**, **abundant**, or **deficient** based on their [aliquot sum](https://en.wikipedia.org/wiki/Aliquot_sum). The aliquot sum is defined as the sum of the factors of a number not including the number itself. For example, the aliquot sum of 15 is (1 + 3 + 5) = 9
 
 - **Perfect**: aliquot sum = number
   - 6 is a perfect number because (1 + 2 + 3) = 6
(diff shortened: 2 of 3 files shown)


exercises/perfect-numbers/canonical-data.json
```diff
@@ -129,6 +129,19 @@
                         "error": "Classification is only possible for natural numbers."
                     }
                 },
+                {
+                    "uuid": "16a621fe-bbd8-4fdd-b861-d5332cf95c02",
+                    "description": "Zero is rejected (not a positive number)",
```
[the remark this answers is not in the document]
ee7: I agree. Thanks.

SaschaMann (approved):

exercises/perfect-numbers/canonical-data.json
```diff
@@ -129,6 +129,19 @@
                         "error": "Classification is only possible for natural numbers."
                     }
                 },
+                {
+                    "uuid": "16a621fe-bbd8-4fdd-b861-d5332cf95c02",
+                    "description": "Zero is rejected (not a positive integer)",
+                    "comments": ["'natural' -> 'positive'"],
```
ee7, line 135:
```suggestion
                    "comments": ["'natural number' -> 'positive integer'"],
```

exercises/perfect-numbers/canonical-data.json
```diff
@@ -139,6 +152,19 @@
                     "expected": {
                         "error": "Classification is only possible for natural numbers."
                     }
+                },
+                {
+                    "uuid": "008de030-ce51-4e0d-8b70-2ea980f17953",
+                    "description": "Negative integer is rejected (not a positive integer)",
+                    "comments": ["'natural' -> 'positive'"],
```
ee7, line 159:
```suggestion
                    "comments": ["'natural number' -> 'positive integer'"],
```

ee7 (commented):
Update the `comments` too.

exercises/perfect-numbers/canonical-data.json
```diff
@@ -120,24 +120,24 @@
             "cases": [
                 {
                     "uuid": "72445cee-660c-4d75-8506-6c40089dc302",
-                    "description": "Zero is rejected (not a natural number)",
+                    "description": "Zero is rejected (as it is not a positive integer)",
```
ee7, line 123: Here I thought about adding "as it is" or similar.

Otherwise it's possible to misread it as "zero is rejected (instead of a natural number)"

iHiD (approved):
Again, solid approve from me on making this require less maths knowledge. Let's again add this to wherever we put the standards document.

ErikSchierboom:
Interesting. This PR is not only updating the descriptions, but also the expected value. Technically this should result in new test cases. I think in this case it is unlikely to break anyone, but I do think we should use new test cases anyway.

SaschaMann:
> This PR is not only updating the descriptions, but also the expected value.

Where does it do that?

ErikSchierboom:
> https://github.com/exercism/problem-specifications/pull/1691/files#diff-30d81c46b2a3d984020efb84025fc4478e23882e1ee05411e74a62820df986bdR129

It changes the error's value. Or do we add an exception for that?

ee7:
> Technically this should result in new test cases. I think in this case it is unlikely to break anyone, but I do think we should use new test cases anyway.

Okay.

> Where does it do that?

@SaschaMann here:
```diff
                    "expected": {
-                        "error": "Classification is only possible for natural numbers."
+                        "error": "Classification is only possible for positive integers."
                    }
```

ErikSchierboom:
On a side note: I considered having a separate key for errors. 

Instead of:

```json
"input": INPUT,
"expected": {
  "error": "ERROR DESCRIPTION"
}
```

it would become:

```json
"input": INPUT,
"error": "ERROR DESCRIPTION"
```

ErikSchierboom:
One way to look at this is that the error text is actually a _description_ of the error, not the actual error.

SaschaMann:
> It changes the error's value. Or do we add an exception for that?

Ah, I missed that.

ErikSchierboom (approved):
As discussed in https://github.com/exercism/problem-specifications/issues/1709#issuecomment-709311827, we'll allow updates to the error values depending on certain conditions being met. This PR meets those conditions.

PR #1698, opened 2020-10-13
CI: Add /rebase command

SaschaMann:
(stolen from v3 repo)

### DIFF
--- a/.github/workflows/rebase.yml
+++ b/.github/workflows/rebase.yml
@@ -1,0 +1,21 @@
+name: Rebase command
+
+on: 
+  issue_comment:
+    types: [created]
+
+jobs:
+  rebase:
+    name: Rebase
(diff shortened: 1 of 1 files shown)


iHiD:
This won't run on forks, right?

SaschaMann:
Nope, as usual. But it'll still help people with commit access (e.g. when required checks change ;)

ErikSchierboom (approved):
We should probably also document that this is possible somewhere. This can be done later.

SleeplessByte (approved):

iHiD (approved):

ynfle:
There is also [autorebase](https://github.com/tibdex/autorebase) & [autosquash](https://github.com/marketplace/actions/autosquash) which may be useful

SaschaMann:
Thanks!

PR #1700, opened 2020-10-13
CI: Check that UUIDs are unique

SaschaMann:

Issue #1693, opened 2020-10-13
CI fails with the new schema


ee7:
See for example: https://github.com/exercism/problem-specifications/pull/1689

```
exercises/grade-school/canonical-data.json invalid
[
  {
    keyword: 'additionalProperties',
    dataPath: '.cases[2]',
    schemaPath: '#/additionalProperties',
    params: { additionalProperty: 'reimplements' },
    message: 'should NOT have additional properties'
  },
  {
    keyword: 'additionalProperties',
    dataPath: '.cases[2]',
    schemaPath: '#/additionalProperties',
    params: { additionalProperty: 'uuid' },
    message: 'should NOT have additional properties'
  },
  {
    keyword: 'oneOf',
    dataPath: '.cases[2]',
    schemaPath: '#/oneOf',
    params: {},
    message: 'should match exactly one schema in oneOf'
  }
]
```

PR #1702, opened 2020-10-13
CI: Add check that reimplemented-values are valid UUIDs

SaschaMann:
Note that this will pass if a test case reimplements itself. I don't think that's a problem or not allowed by the spec though.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,7 +69,7 @@ jobs:
               # Abort early if the exercise doesn't have reimplemented test cases
               [[ -e $reimplements ]] && continue
               
-              uuids=$(grep '"uuid":' "$f" | sed -E 's/.*"([^"]+)".*/\1/')
+              uuids=$(jq -r '.. |."uuid"? | select(. != null)' "$f")
               for reimplemented_uuid in $reimplements; do
                   if [[ $uuids != *"$reimplemented_uuid"* ]]; then
                       echo "$f: reimplemented value refers to a test case with UUID $reimplemented_uuid, but the file does not contain any test cases with that UUID."
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,7 +64,7 @@ jobs:
 
           for f in exercises/*/canonical-data.json; do
               # Search for reimplemented test cases
-              reimplements=$(grep '"reimplements":' "$f" | sed -E 's/.*"([^"]+)".*/\1/')
+              reimplements=$(jq -r '.. |."reimplements"? | select(. != null)' "$f")
           
               # Abort early if the exercise doesn't have reimplemented test cases
               [[ -e $reimplements ]] && continue
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,3 +57,25 @@ jobs:
               echo "$duplicate_uuids"
               exit 1
           fi
+
+      - name: Verify that reimplements-values refer to existing UUID
+        run: |
+          fail=0
+
+          for f in exercises/*/canonical-data.json; do
+              # Search for reimplemented test cases
+              reimplements=$(grep '"reimplements":' "$f" | sed -E 's/.*"([^"]+)".*/\1/')
+          
+              # Abort early if the exercise doesn't have reimplemented test cases
+              [[ -e $reimplements ]] && continue
+              
+              uuids=$(grep '"uuid":' "$f" | sed -E 's/.*"([^"]+)".*/\1/')
+              for reimplemented_uuid in $reimplements; do
+                  if [[ $uuids != *"$reimplemented_uuid"* ]]; then
+                      echo "$f: reimplemented value refers to a test case with UUID $reimplemented_uuid, but the file does not contain any test cases with that UUID."
+                      fail=1
+                  fi
+              done
+          done
+          
+          exit "$fail"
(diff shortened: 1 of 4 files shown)


SaschaMann:
Wait for #1693 to be resolved. The schema doesn't know about `reimplements` yet.

iHiD:
Closed and reopened to see if that triggered the codeowners correctly. It didn't.

SaschaMann:
It kinda does, it just doesn't show it in the reviewer list:

grafik

iHiD:
Right, ok. So it probably needed the team when the PR was opened.

SaschaMann:
It works now 🤷

.github/workflows/ci.yml
```diff
@@ -45,3 +45,25 @@ jobs:
...
       - name: Verify that canonical-data.json files adhere to canonical-schema.json
         run: yarn test
+
+      - name: Verify that reimplements-values refer to existing UUID
+        run: |
+          fail=0
+
+          for f in exercises/*/canonical-data.json; do
+              # Search for reimplemented test cases
+              reimplements=$(grep '"reimplements":' "$f" | sed -E 's/.*"([^"]+)".*/\1/')
```
ErikSchierboom, line 55: You could possibly use jq here too, see https://github.com/exercism/problem-specifications/pull/1700/files#r504241547.

.github/workflows/ci.yml
```diff
@@ -45,3 +45,25 @@ jobs:
...
+          
+              # Abort early if the exercise doesn't have reimplemented test cases
+              [[ -e $reimplements ]] && continue
+              
+              uuids=$(grep '"uuid":' "$f" | sed -E 's/.*"([^"]+)".*/\1/')
+              for reimplemented_uuid in $reimplements; do
+                  if [[ $uuids != *"$reimplemented_uuid"* ]]; then
+                      echo "$f: reimplemented value refers to a test case with UUID $reimplemented_uuid, but the file does not contain any test cases with that UUID."
+                      fail=1
```
ErikSchierboom, line 64: Good idea to not immediately bail out.

ErikSchierboom (approved):

.github/workflows/ci.yml
SaschaMann: I don't fully understand it and if seems to be slower so I'd prefer the naive grep/sed approach

coriolinus: The downside of the grep/sed approach is that it depends on formatting: if for any reason the reimplemented UUID drops down a line, sed won't see it anymore. jq is slower (because it's parsing the JSON), but it's immune to formatting (because it's parsing the JSON). I have never yet encountered a case where jq is sufficiently slower than sed that it makes a difference in human terms.

SaschaMann: Yeah, I assumed they were formatted correctly because that's something that (eventually) will be checked by CI as well.

ee7 (approved):
I approve on the basis that I tested that it works, and `shellcheck` doesn't complain.

However, I think I'm in the "use `jq` instead" camp. Then it wouldn't silently fail if some JSON was somehow in non-pretty format.

I think it's a case of comparing the maintenance burdens of
1. People being less familiar with `jq` and
2. Ensuring that the JSON formatting stays working, and ensuring that nobody moves this check to happen before the formatting, potentially years in the future.

The `jq` option just seems more robust. But sure, we hope that the failure modes for the `grep/sed` version shouldn't happen.

SaschaMann:
Yea that's a good summary. I personally don't think that duplicating checks is necessary but I'm fine with switching to jq if that's what people prefer. I'll change this and the other workflow tomorrow :)

> I approve on the basis that I tested that it works, and shellcheck doesn't complain.

I check them with shellcheck as well. I wonder if it'd be possible to have shellcheck check relevant parts of the workflow files without copying them out.

.github/workflows/ci.yml
SaschaMann:
```suggestion
              reimplements=$(jq -r '.. |."reimplements"? | select(. != null)' "$f")
```

.github/workflows/ci.yml
```diff
@@ -57,3 +57,25 @@ jobs:
...
+          fail=0
+
+          for f in exercises/*/canonical-data.json; do
+              # Search for reimplemented test cases
+              reimplements=$(jq -r '.. |."reimplements"? | select(. != null)' "$f")
+          
+              # Abort early if the exercise doesn't have reimplemented test cases
+              [[ -e $reimplements ]] && continue
+              
+              uuids=$(grep '"uuid":' "$f" | sed -E 's/.*"([^"]+)".*/\1/')
```
SaschaMann, line 72:
```suggestion
              uuids=$(jq -r '.. |."uuid"? | select(. != null)' "$f")
```

ErikSchierboom (approved):
Nice!

coriolinus (approved):
My shellcheck is misbehaving on this; it's not reporting any errors, and I think it probably should. In particular, I suspect that it would normally complain about [word splitting](https://github.com/koalaman/shellcheck/wiki/SC2046) on lines 67 and 72.

That said, if it works, it works; I'm not going to be the Shellcheck Enforcer.

SaschaMann:
Thanks for the reviews!

@coriolinus It never shows that error to me when assigning the output of a command to a variable, maybe it's not needed? 🤷

PR #1703, opened 2020-10-13
Add reimplements to schema

SaschaMann:
(view the commits separately for a more readable diff)

### DIFF
--- a/canonical-schema.json
+++ b/canonical-schema.json
@@ -78,14 +78,14 @@
           , "type"       : "object"
           , "required"   : ["uuid", "description", "property", "input", "expected"]
           , "properties" :
-                { "uuid"       : { "$ref": "#/definitions/uuid" }
+                { "uuid"        : { "$ref": "#/definitions/uuid" }
                 , "reimplements": { "$ref": "#/definitions/uuid" }
-                , "description": { "$ref": "#/definitions/description" }
-                , "scenarios"  : { "$ref": "#/definitions/scenarios"   }
-                , "comments"   : { "$ref": "#/definitions/comments"    }
-                , "property"   : { "$ref": "#/definitions/property"    }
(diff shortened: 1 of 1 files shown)


SaschaMann:
See #1704 for an example that this fixes the build failures in #1691.

petertseng (approved):

SaschaMann:
Thanks for the really quick reviews @petertseng :)

ee7 (approved):
I approve on the basis that CI now passes for a `canonical-data.json` that contains `reimplements`.

But two questions:
1) Do we have an official ordering that the keys are supposed to be in?
2) If so, should we enforce the key order during CI? Maybe it helps with readability when adding tests. The ordering seemed pretty consistent before, but  there will be nearly double the number of keys now.


I notice that the `reimplements` in this PR is after `uuid`, but when I made my recent PRs I instead ordered the keys as I saw them in https://github.com/exercism/problem-specifications/issues/1674, that is:
```json
  {
    "uuid": "82d32c2e-07b5-42d9-9b1c-19af72bae860",
    "description": "two times one is two",
    "comments": ["Expected value is changed to 2"],
    "reimplements": "e46c542b-31fc-4506-bcae-6b62b3268537",
    "property": "twice",
    "input": {
      "number": 1
    },
    "expected": 2
  }
```

iHiD:
We should do, and we should enforce it, yes. Can whoever merges this (or anyone before), please create an issue about deciding order, and another about CI enforcing?

ErikSchierboom:
> We should do, and we should enforce it, yes. Can whoever merges this (or anyone before), please create an issue about deciding order, and another about CI enforcing?

This is also planned for when we update configlet with the sync functionality BTW.

ErikSchierboom:
Thanks for the quick fix @SaschaMann!

Issue #1709, opened 2020-10-15
Should the value of `error` be mutable?


ee7:
### Example
In `perfect-numbers` we have this test case:
https://github.com/exercism/problem-specifications/blob/8df6cb2ed8e77aa3ac0c02744cd9ddb83920f5f2/exercises/perfect-numbers/canonical-data.json#L121-L130

And we notice (see https://github.com/exercism/problem-specifications/pull/1691) that we would like to make this change:
```diff
 { 
     "uuid": "72445cee-660c-4d75-8506-6c40089dc302", 
-     "description": "Zero is rejected (not a natural number)", 
+     "description": "Zero is rejected (not a positive integer)", 
     "property": "classify", 
     "input": { 
         "number": 0 
     }, 
     "expected": { 
-         "error": "Classification is only possible for natural numbers." 
+         "error": "Classification is only possible for positive integers." 
     } 
```

We seem to have decided that `description` is mutable, and therefore we do not add a new test case that `reimplements` for a `description`-only change.

The question is: should the contents of an `error` also be considered mutable, or should it require adding a new test case?

### Discussion
My initial opinion is something like:
1. We said that `description` should be mutable if we don't change the "meaning" of the test.
1. There are some useful error-message changes that don't change the "meaning" of the test either, and the value of `error` is similar to the "description" of the error. 
1. Therefore, directly mutating the `error` value should be allowed.

Drawbacks:
- Is there a track that asserts something about the contents of the error message for this kind of test? Such a track would have to update that assertion, even though their `tests.toml` didn't change.
- The CI check for immutability would need to check that `expected` doesn't change, but make a special case for an `error`-only change.

coriolinus:
I'd vote: it's mutable, except if the mutation would change the total set of errors in an exercise. Using your example, I'd expect a Rust test generator to come up with

```rust
enum Error {
  ZeroIsRejectedNotANaturalNumber,
  ClassificationIsOnlyPossibleForNaturalNumbers,
}
```

If a PR updates the text on all such errors, we now have

```rust
enum Error {
  ZeroIsRejectedNotAPositiveInteger,
  ClassificationIsOnlyPossibleForPositiveIntegers,
}
```

The set of errors, and the semantics, hasn't changed; only the enum's label. I think this is fine.

If, on the other hand, we rewrite things to `enum Error` had some number of variants other than 2, then that's a breaking change; that should be added as a new test, because there are now new semantics in the exercise.

Make sense?

[...]

PR #1712, opened 2020-10-15
CI: Ensure immutability of test cases

SaschaMann:
This is gonna be a bit tricky to test completely but maybe it's sufficient to test it for PR events


SaschaMann:
18dbfa2 demonstrates that it works in a PR.

SaschaMann:
@iHiD Once this is merged, this should also become a required check :)

.github/workflows/ci.yml
```diff
@@ -15,6 +15,37 @@ jobs:
...
+        with:
+          ref: '${{ github.base_ref }}'
+          path: 'old'
+      
+      # Pushes
+      - name: Checkout the previous commit (Pushes)
+        uses: actions/checkout@v2
+        if: github.event_name == 'push' || github.event_name == 'workflow_dispatch'
+        with:
+          ref: '${{ github.sha }}~'
```
SaschaMann, line 40: This is only really gonna work reliably on the main branch fwiw

bin/check-immutability.py
```diff
@@ -0,0 +1,34 @@
...
+new = {c['uuid']: c for c in new}
+
+fails = set()
+
+# Iterate through old cases as only those could potentially be mutated
+for case in old:
+    for k in immutable_keys:
+        if case[k] != new[case['uuid']][k]:
+            fails.add(case['uuid'])
+            break
```
ErikSchierboom, line 26: Just to be sure: does this fail when the new file has removed one of the old test cases?

ErikSchierboom (approved):
Nice work. I think there are two checks that we need to add (can be done in follow-up PRs):

1. The `error` property's text value can be modified, provided it is modified for _all_ occurences (see [this issue]( https://github.com/exercism/problem-specifications/issues/1709#issuecomment-709311827))
1. The `scenarios` key should only be updated additively (see [this section](The scenarios field can be mutated additively, by adding new scenarios. Existing scenarios must not be changed or removed. Adding new scenarios does therefore does not mean adding a new test case.)).

bin/check-immutability.py
SaschaMann: It should because the dict key won't exist, but I'll test to be sure.

ErikSchierboom: Great!

SaschaMann: It already worked because it threw an error, but b010ec0 adds better handling. Sample output for a file that has been modified and deleted:

```
$ python3 bin/check-immutability.py test/canonical-data.json test/canonical-data-mutated-and-deleted.json

The following tests contain illegal mutations:
 - 9f98e133-eb7f-45b0-9676-cce001cd6f7a (Real part of a purely real number)
 - 07988e20-f287-4bb7-90cf-b32c4bffe0f3 (Real part of a purely imaginary number)
The following tests have been deleted illegally:
 - 657c55e1-b14b-4ba7-bd5c-19db22b7d659
```

SaschaMann:
> The scenarios key should only be updated additively (see [this section](The scenarios field can be mutated additively, by adding new scenarios. Existing scenarios must not be changed or removed. Adding new scenarios does therefore does not mean adding a new test case.)).

Fixed by af3969e

SaschaMann:
Should I upload the test changes to `.github/tests` or so? Might be useful to have them already when modifying this script in the future.

bin/check-immutability.py
```diff
@@ -0,0 +1,54 @@
...
+
+    # Check if the case has been deleted
+    if uuid not in new:
+        deleted.add(uuid)
+        continue
+
+    # Check that scenarios are only updated additively
+    if 'scenarios' in case and not set(case['scenarios']).issubset(set(new[uuid]['scenarios'])):
+        fails.add(uuid)
+
```
SaschaMann, line 34:
```suggestion
        fails.add(uuid)
        continue

```

bin/check-immutability.py
```diff
@@ -0,0 +1,54 @@
...
+newf = sys.argv[2]
+
+immutable_keys = ('property', 'input', 'expected')
+
+# Use jq to flatten the test data, and parse it
+old = json.loads(subprocess.run([f"jq -r '[.. | objects | select(.uuid != null)]' {oldf}"], stdout=subprocess.PIPE, shell=True).stdout.decode('utf-8'))
+new = json.loads(subprocess.run([f"jq -r '[.. | objects | select(.uuid != null)]' {newf}"], stdout=subprocess.PIPE, shell=True).stdout.decode('utf-8'))
+
+# Convert new to dict uuid => case
+new = {c['uuid']: c for c in new}
```
BethanyG, line 17:
```suggestion
new = {case['uuid']: case for case in new}
```

bin/check-immutability.py
```diff
@@ -0,0 +1,54 @@
...
+        continue
+
+    # Check that scenarios are only updated additively
+    if 'scenarios' in case and not set(case['scenarios']).issubset(set(new[uuid]['scenarios'])):
+        fails.add(uuid)
+        continue
+    # Check for changes to immutable keys
+    for k in immutable_keys:
+        if case[k] != new[uuid][k]:
+            fails.add(uuid)
```
BethanyG, line 38:
```suggestion
    for key in immutable_keys:
        if case[key] != new[uuid][key]:
            fails.add(uuid)
```

bin/check-immutability.py
```diff
@@ -0,0 +1,54 @@
...
+
+if len(fails) > 0:
+    print('The following tests contain illegal mutations:')
+    for f in fails:
+        print(f" - {f} ({new[f]['description']})")
+
+if len(deleted) > 0:
+    print('The following tests have been deleted illegally:')
+    for d in deleted:
+        print(f" - {d}")
```
BethanyG, line 52:
```suggestion
    for failure in fails:
        print(f" - {failure} ({new[failure]['description']})")

if len(deleted) > 0:
    print('The following tests have been deleted illegally:')
    for deletion in deleted:
        print(f" - {deletion}")
```

bin/check-immutability.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python 
+
+import json
+import subprocess
```
cmccandless, line 4:
```suggestion
import shlex
import subprocess
```

bin/check-immutability.py
```diff
@@ -0,0 +1,54 @@
...
+import sys
+
+oldf = sys.argv[1]
+newf = sys.argv[2]
+
+immutable_keys = ('property', 'input', 'expected')
+
+# Use jq to flatten the test data, and parse it
+old = json.loads(subprocess.run([f"jq -r '[.. | objects | select(.uuid != null)]' {oldf}"], stdout=subprocess.PIPE, shell=True).stdout.decode('utf-8'))
+new = json.loads(subprocess.run([f"jq -r '[.. | objects | select(.uuid != null)]' {newf}"], stdout=subprocess.PIPE, shell=True).stdout.decode('utf-8'))
```
cmccandless, line 14: Requires `import shlex` (suggestion above).

If using stdout with `subprocess.run`, `capture_output=True` is recommended.

```suggestion
# Use jq to flatten the test data, and parse it
def get_test_data(filename):
	cmd = shlex.split(f"jq -r '[.. | objects | select(.uuid != null)]' {filename}")
    return json.loads(subprocess.run(cmd, capture_output=True, shell=True).stdout


old = get_test_data(oldf)
new = get_test_data(newf)
```

bin/check-immutability.py
```diff
@@ -0,0 +1,54 @@
...
+        fails.add(uuid)
+        continue
+    # Check for changes to immutable keys
+    for k in immutable_keys:
+        if case[k] != new[uuid][k]:
+            fails.add(uuid)
+            break
+
+if len(fails) == 0 and len(deleted) == 0:
+    exit(0)
```
cmccandless, line 42: See https://stackoverflow.com/a/6501134/3229611

```suggestion
    sys.exit(0)
```

bin/check-immutability.py
```diff
@@ -0,0 +1,54 @@
...
+    print('The following tests contain illegal mutations:')
+    for f in fails:
+        print(f" - {f} ({new[f]['description']})")
+
+if len(deleted) > 0:
+    print('The following tests have been deleted illegally:')
+    for d in deleted:
+        print(f" - {d}")
+
+exit(1)
```
cmccandless, line 54:
```suggestion
sys.exit(1)
```

cmccandless (changes requested):

bin/check-immutability.py
SaschaMann: Could you explain to me why `shlex.split` should be used here? The docs for it describe what it does but it's not clear to me why this should be used instead of calling the command directly.

coriolinus: Really, there's no need to use `jq` to flatten the cases at all; it's prefectly possible (and even easier) in standard python:

```python3
def walk_cases(data):
    if "cases" in data:
        for case in data["cases"]:
            yield from walk_cases(case)
            if "uuid" in case and "description" in case:
               # if desired, yield case["uuid"]
                yield case
               
```

coriolinus: @SaschaMann Depending on your input string, `subprocess.run` may interpret `jq -r ...` as one big command with whitespace in the filename, and then fail when it determines that no executable with that name exists in your path. Using `shlex.split` means that it gets parsed properly into an input list ahead of time.

SaschaMann: Thanks for the explanation. It didn't do that here, so I think shlex isn't needed in this particular case because the format of the inputs is well-known, while making it harder to follow/read for non-Python users. If it fails, that's a sign that the input is somehow wrong which means the commit also contains wrong data.

While I like that Python-only implementation, I think I'd rather go with the jq approach since we already use jq elsewhere in the repo and the Python-only solution introduces more complexity without, imo, much to gain. I'd rather keep a script like this as "dumb" as possible and stick to basic constructs whenever possible. The `subprocess.run` call is a mess but it's enough to understand that it runs the command and saves the output without having to understand the details.

I appreciate the suggestions though

ErikSchierboom (approved):
LGTM. Testing seems to be the main thing we should be looking at. Did you manage to test the functionality on a fork @SaschaMann?

SaschaMann:
@ErikSchierboom I tested it for PR events and locally. I can upload the test cases if you want? Might make things easier in the future.

ErikSchierboom:
> I tested it for PR events and locally. I can upload the test cases if you want? Might make things easier in the future.

You could do that, whatever you prefer. I just wanted to verify that you checked if this worked (I assumed you did of course).

BethanyG (approved):

cmccandless (approved):

ErikSchierboom:
I'm gonna merge this. Thanks @SaschaMann!

It would be good to check this works in upcoming PRs.

SaschaMann:
@iHiD Could you make `CI / No test has been mutated` a required check?

PR #1713, opened 2020-10-15
Add Style Guide

iHiD:
As discussed in a few PRs in the last few days, I'd like to add a style guide that we can use as a guide for grammatical/language changes in descriptions etc. These should avoid debates on language, as all discussions can defer to this document, and in cases where this document does not document a style and there is not consensus, an executive decision can be made, added to this document, and then adhered to moving forward.

To get the ball rolling, I propose having a `STYLE-GUIDE.md` file, with a guidance and example for each point. I've put this together as a rough PR to get discussion started.

Please consider it both from the general case (how the style guide should work) and the specific points in this PR.

### DIFF
--- a/STYLE-GUIDE.md
+++ b/STYLE-GUIDE.md
@@ -1,0 +1,22 @@
+# Exercism's Style Guide
+
+This document acts as a Style Guide for the language and wording used in exercises.
+
+## General principle
+
+All content should be written in US english. 
+In the future other translations may occur, but the "official" Exercism natural langauge is US English.
+
+## Explain or substitude mathmatical and esotoric terms.
+
+In any place that mathematical terms are used they should be explained or substituted out for terms that require less domain knowledge.
+
+Examples:
+- Rather than using "natural numbers", we should use "positive integers" or "zero and positive integers". 
+- If we want to use the phrase "rational numbers", it must be explained in the introduction to the exercise.
+- Rather than using the word range (which can have different meanings in different contexts) use "greater than x and less than y".
+
+## Use consistency within an exercise.
+
+There are some terms that have multiple valid spellings (e.g. "lower case" vs "lowercase"). 
+Where a consistent style has not been agreed within this document, these must be consistent within an exercise.
--- a/STYLE-GUIDE.md
+++ b/STYLE-GUIDE.md
@@ -15,7 +15,7 @@ In any place that mathematical terms are used they should be explained or substi
 Examples:
 - Rather than using "natural numbers", we should use "positive whole numbers".
 - If we want to use the phrase "rational numbers", it must be explained in the introduction to the exercise.
-- Rather than using the word range (which can have different meanings in different contexts) use "greater than x and less than y".
+- Rather than using the word range (which can have different meanings in different contexts) use "x < ? < y (greater than x and less than y)".
 
 ## Use consistency within an exercise.
 
--- a/STYLE-GUIDE.md
+++ b/STYLE-GUIDE.md
@@ -7,7 +7,7 @@ This document acts as a Style Guide for the language and wording used in exercis
 All content should be written in US english. 
 In the future other translations may occur, but the "official" Exercism language is US English.
 
-## Explain or substitude mathematical and esotoric terms.
+## Explain or substitute mathematical and esoteric terms.
 
 In any place that mathematical terms are used they should be explained or substituted out for terms that require less domain knowledge.
 
--- a/STYLE-GUIDE.md
+++ b/STYLE-GUIDE.md
@@ -4,7 +4,8 @@ This document acts as a Style Guide for the language and wording used in exercis
 
 ## General principle
 
-All content should be written in US english. 
+All content should be written in US English.
+All units of measurement must be [SI](https://en.wikipedia.org/wiki/International_System_of_Units) or [SI-derived](https://en.wikipedia.org/wiki/SI_derived_unit) units.
 In the future other translations may occur, but the "official" Exercism language is US English.
 
 ## Explain or substitute mathematical and esoteric terms.
--- a/STYLE-GUIDE.md
+++ b/STYLE-GUIDE.md
@@ -5,7 +5,7 @@ This document acts as a Style Guide for the language and wording used in exercis
 ## General principle
 
 All content should be written in US english. 
-In the future other translations may occur, but the "official" Exercism natural langauge is US English.
+In the future other translations may occur, but the "official" Exercism language is US English.
 
 ## Explain or substitude mathematical and esotoric terms.
 
--- a/STYLE-GUIDE.md
+++ b/STYLE-GUIDE.md
@@ -7,7 +7,7 @@ This document acts as a Style Guide for the language and wording used in exercis
 All content should be written in US english. 
 In the future other translations may occur, but the "official" Exercism natural langauge is US English.
 
-## Explain or substitude mathmatical and esotoric terms.
+## Explain or substitude mathematical and esotoric terms.
 
 In any place that mathematical terms are used they should be explained or substituted out for terms that require less domain knowledge.
 
--- a/STYLE-GUIDE.md
+++ b/STYLE-GUIDE.md
@@ -13,7 +13,7 @@ In the future other translations may occur, but the "official" Exercism language
 In any place that mathematical terms are used they should be explained or substituted out for terms that require less domain knowledge.
 
 Examples:
-- Rather than using "natural numbers", we should use "positive integers" or "zero and positive integers". 
+- Rather than using "natural numbers", we should use "positive whole numbers".
 - If we want to use the phrase "rational numbers", it must be explained in the introduction to the exercise.
 - Rather than using the word range (which can have different meanings in different contexts) use "greater than x and less than y".
 


STYLE-GUIDE.md
```diff
@@ -0,0 +1,22 @@
+# Exercism's Style Guide
+
+This document acts as a Style Guide for the language and wording used in exercises.
+
+## General principle
+
+All content should be written in US english.
```
SaschaMann, line 7:
```suggestion
All content should be written in US English. 
```

STYLE-GUIDE.md
```diff
@@ -0,0 +1,22 @@
+# Exercism's Style Guide
+
+This document acts as a Style Guide for the language and wording used in exercises.
+
+## General principle
+
+All content should be written in US english. 
+In the future other translations may occur, but the "official" Exercism natural langauge is US English.
```
SaschaMann, line 8:
```suggestion
In the future other translations may occur, but the "official" Exercism language is US English.
```

STYLE-GUIDE.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Explain or substitude mathmatical and esotoric terms.
```
SaschaMann, line 10:
```suggestion
## Explain or substitude mathematical and esotoric terms.
```

STYLE-GUIDE.md
```diff
@@ -0,0 +1,22 @@
...
+
+In any place that mathematical terms are used they should be explained or substituted out for terms that require less domain knowledge.
+
+Examples:
+- Rather than using "natural numbers", we should use "positive integers" or "zero and positive integers".
```
SaschaMann, line 15: `zero`, `positive` and `integers` are mathematical terms.

STYLE-GUIDE.md
```diff
@@ -0,0 +1,22 @@
...
+All content should be written in US english. 
+In the future other translations may occur, but the "official" Exercism natural langauge is US English.
+
+## Explain or substitude mathmatical and esotoric terms.
+
+In any place that mathematical terms are used they should be explained or substituted out for terms that require less domain knowledge.
+
+Examples:
+- Rather than using "natural numbers", we should use "positive integers" or "zero and positive integers". 
+- If we want to use the phrase "rational numbers", it must be explained in the introduction to the exercise.
```
SaschaMann, line 16: Is a link to an explanation sufficient?

STYLE-GUIDE.md
```diff
@@ -0,0 +1,22 @@
...
+- Rather than using the word range (which can have different meanings in different contexts) use "greater than x and less than y".
```
SaschaMann, line 17: Depending on the context `x < _ < y` may be better (as in more readable), imo.

STYLE-GUIDE.md
SaschaMann, line 7: Can we add a rule that all units must be SI units or SI-compatible, though?

STYLE-GUIDE.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Use consistency within an exercise.
+
+There are some terms that have multiple valid spellings (e.g. "lower case" vs "lowercase").
```
SaschaMann, line 21: Perhaps this could be the first example of an agreed spelling?

STYLE-GUIDE.md
SaschaMann, line 10: Could you perhaps add an example of a non-maths esoteric term? The content of this section only mentions maths terms.

STYLE-GUIDE.md
neenjaw: But they are more plainly understood than `natural numbers` which appears to be the intent

SaschaMann: I'm not convinced that "integers" is more plainly understood than "natural numbers".

neenjaw: ¯\\\_(ツ)_/¯

STYLE-GUIDE.md
SleeplessByte: Maybe:

```
Regrettably, all content should be written in Simple English, colloquially known as US English.
```

Just kidding of course. But we might want to consider linking to a document that has the most prominent differences between English and US English, as they are **not** apparent to everyone.

STYLE-GUIDE.md
SleeplessByte: I prefer both.

```
x < ? < y (greater than x and less than y)
```

STYLE-GUIDE.md
SleeplessByte: Yes: https://en.wikipedia.org/wiki/Non-SI_units_mentioned_in_the_SI + SI please.

SaschaMann: I remember there being discussions on Slack about the use of `pants` vs `trousers` where @iHiD veto'd the use of US English. Perhaps that should be added as an exception to this rule? 🙃

ceddlyburge (approved):
The freecodecamp editorial guide is good and has some stuff about simple language. One thing it mentions is not to use Latin things like e.g, and instead use English things, like for example.

SaschaMann:
> general case (how the style guide should work)

I think in general it may be a good idea to have a rule that exercise-specific style decisions overrule the general style guide, when they are necessary or reasonable in the context of the exercise.

STYLE-GUIDE.md
iHiD: I'm happy to have "no pants" as a documented rule!

STYLE-GUIDE.md
iHiD: > I'm not convinced that "integers" is more plainly understood than "natural numbers".

I'm pretty convinced it is as `Integer` is a commonly used class / type in most languages, whereas `natural` isn't a programming "concept" I've ever used.

But I'd be happy to say "positive whole numbers" too.

STYLE-GUIDE.md
iHiD: Sure. But I also want a rule that explains what happens when multiple valid spellings are used without a documented decision in this guide (ie, they should be exercise-consistent)

iHiD:
@ceddlyburge Nice!

> I think in general it may be a good idea to have a rule that exercise-specific style decisions overrule the general style guide if they are necessary or reasonable in the context of the exercise.

I'm ok with this in principle. However, I'd probably want that signing off by other maintainers as a valid exception, rather than being something that maintainers can just arbitarily override. Else we risk a "rogue" maintainer who doesn't want to adhere to the rules just making "exceptions" for all their exercises (I'm thinking track-specific overrides). ie, we should all adhere to these rules Exercism-wide unless there's a specific situation where a cohort of maintainers agree there's a valid exception.

I'd also like to roll this guide out to mentoring and stuff further down the line too.

That said, I'm not sure of a situation where the style-guide **should** be overriden. Do you have an example?

STYLE-GUIDE.md
iHiD: Yep. Sounds good. (This is actually a potential example where the "exception" thing mentioned below fits in - if an example is about converting from one unit to another and we deliberately choose not to use `si` units).

@SaschaMann Either add a suggestion to this conversation or follow up with a seperate PR (the later is probably preferable so we can get this one over the line quicker)

STYLE-GUIDE.md
```diff
@@ -0,0 +1,22 @@
+# Exercism's Style Guide
+
+This document acts as a Style Guide for the language and wording used in exercises.
+
+## General principle
+
+All content should be written in US english. 
+In the future other translations may occur, but the "official" Exercism language is US English.
+
+## Explain or substitude mathematical and esotoric terms.
```
rpottsoh, line 10:
```suggestion
## Explain or substitute mathematical and esoteric terms.
```
Spelling....

STYLE-GUIDE.md
rpottsoh: lowercase seems to be more popular:
image

https://english.stackexchange.com/questions/59409/lowercase-lower-case-or-lower-case

STYLE-GUIDE.md
SaschaMann: Yea that's precisely the thing. I hear the term integers mostly in a programming context, while natural numbers are used everywhere and more likely to be known even by programming novices because it's a term they (probably) heard in school.

As you know, I'm really not a fan of singling out maths as some scary voodoo thing that people must be afraid of. I'd be happier if it was an example where, instead of a mathematical term, an everyday term would be used. Maybe something like `Rather than using "bijective map", we should use "1-to-1 map"`? (Not a great example either.)

> But I'd be happy to say "positive whole numbers" too.

That's worse imo because positive whole numbers are just natural numbers :D

STYLE-GUIDE.md
SaschaMann:
```suggestion
All content should be written in US English.
All units of measurement must be [SI](https://en.wikipedia.org/wiki/International_System_of_Units) or [SI-derived](https://en.wikipedia.org/wiki/SI_derived_unit) units.
```

SaschaMann:
Somewhat off-topic: Perhaps we should add a spelling linter like https://github.com/marketplace/actions/orthograph-err?

STYLE-GUIDE.md
SleeplessByte: Good enough for me 👍

STYLE-GUIDE.md
iHiD: I'm happy to use ~zero~ no esoteric terms (where integer might be considered esoteric here). Especially as it's kinder to non-native speakers.

But, we do actively presume someone has a programming background, and we equally actively presume someone doesn't have a maths background. So we should be more comfortable using programming words than maths words.

> That's worse imo because positive whole numbers are just natural numbers :D

Sometimes natural numbers include zero, sometimes they don't :P

All concepts have words that label them. But requiring someone to know more words is a less kind thing than using two words to explain something when they might not.

Saying 
> positive whole numbers (know as "natural numbers")

Then using `natural numbers` throughout the rest of the description is fine.

SaschaMann:
> However, I'd probably want that signing off by other maintainers as a valid exception, rather than being something that maintainers can just arbitarily override. 

I agree. I also think it should explicitely be documented in the exercise itself, not just in a git commit/PR.

> That said, I'm not sure of a situation where the style-guide should be overriden. Do you have an example?

I can't think of a specific example right now.

STYLE-GUIDE.md
SleeplessByte: Isn't it good for understanding to do "whole numbers" (integers)" and "whole positive numbers (unsigned integers)"? If we're trying to be as _accessible_ as possible, this might be preferable.

My problem with _integer_ is that it's NOT the same thing across programming languages.

SaschaMann: > But, we do actively presume someone has a programming background, and we equally actively presume someone doesn't have a maths background. So we should be more comfortable using programming words than maths words.

I'd argue that high school words are an even lower requirement than programming words, therefore we should use those when possible.

That said, this is kinda pointless, maybe we can come up with a differenti example instead of arguing whether integers or natural numbers are a more commonly known term?

neenjaw (approved):

STYLE-GUIDE.md
SaschaMann:
```suggestion
- Rather than using "bijective map", we should use "1-to-1 map".
```

STYLE-GUIDE.md
kotp: Dogs have pants, people have trousers.

STYLE-GUIDE.md
wolf99: More popular != correct 😜

STYLE-GUIDE.md
```diff
@@ -0,0 +1,23 @@
...
+
+This document acts as a Style Guide for the language and wording used in exercises.
+
+## General principle
+
+All content should be written in US English.
+All units of measurement must be [SI](https://en.wikipedia.org/wiki/International_System_of_Units) or [SI-derived](https://en.wikipedia.org/wiki/SI_derived_unit) units.
+In the future other translations may occur, but the "official" Exercism language is US English.
+
+## Explain or substitute mathematical and esoteric terms.
```
wolf99, line 11: Should there be a similar rule for abbreviations, acronyms and initialisms ?

For example, a rule that is often used in academic writing is that an initialism should have its expansion in brackets immediately following its first use and thereafter can be used freely.

cmcaine: I think we should have a no jargon rule. And don't use initialisms etc at all unless they are better known by the initialism (e.g. HTTP, HTML, radar) or if the text would be ridiculously verbose otherwise.

wolf99: In favour of no jargon. 
However, outside of the specific language related initialism, should we assume that all users will have knowledge of what I or you might consider better known ones, even for non-native US English speakers?

Yet we would still need to use initialisms in some cases. If we, hypothetically, agree that RADAR is jargon to non-US English speakers, then we would not have to be forced to use the long form each and every time it is mentioned, especially if it might form the name of test cases or some other oft repeated part of an exercise.

thus we need some kind of middle solution (hence my suggestion 😉).

I can think of a host of these that might be used depending on the exercise and track: FIFO, NASA, ABS, GIF, DIY, SONAR, SUV, UFO, CRUD, ANSI, API, ASCII, ICANN, CMYK, HTTP, FTP, DPI, DSL, JPEG, ... these are not especially unusual in programming or jargon-y, yet I can easily say that not everyone around the world would be familiar with all them

STYLE-GUIDE.md
iHiD: https://github.com/exercism/problem-specifications/issues/1715

STYLE-GUIDE.md
iHiD: I (genuinely) don't understand "1 to 1 map".

Wikipedia says:

> In mathematics, a bijection, bijective function, one-to-one correspondence, or invertible function, is a function between the elements of two sets, where each element of one set is paired with exactly one element of the other set, and each element of the other set is paired with exactly one element of the first set.

If we needed to use the word bijection, we would need to be using some level of information that covers this. Maybe "where each element in one set has a paired element in another set"? (I don't know if that's correct - but that's how I'm interpretting that ^^)

```suggestion
- Rather than using "natural numbers", we should use "positive whole numbers".
```

STYLE-GUIDE.md
iHiD:
```suggestion
- Rather than using the word range (which can have different meanings in different contexts) use "x < ? < y (greater than x and less than y)".
```

STYLE-GUIDE.md
iHiD: Let's have a seperate PR and standalone discussion for that. I've created an issue 🙂 

https://github.com/exercism/problem-specifications/issues/1716

STYLE-GUIDE.md
iHiD: I can't think of one off the top of my head. But I'll try and come up with one and PR it. Feel free to do the same if you can think of one.

iHiD:
I've "resolved" all the discussions that are blocked, so I'm going to merge so we've got a document on master that people can add to. Let's continue any discussions on individual PRs, so this doesn't become a huge PR (we're already at 2x comments to LOCs :wink:).

Thanks everyone 🎉

STYLE-GUIDE.md
iHiD: I think at the very least, I'd want an example of what is and is not a rational number, and any information that is needed to solve the exercise. For rational numbers, I'd probably have:

> This exercise is about Rational Numbers: Numbers that can be expressed as a fraction of two whole numbers. e.g. 1/4 or 12/6. You can learn more about rational numbers at [maths.com](...).

If I needed to know about signs, I'd add whether rational numbers can be negative or not. If I didn't need to know for the exercise, I wouldn't mention it.

SaschaMann: Thanks!

iHiD: Feel free to PR that clarification to the guide, etc 🙂

STYLE-GUIDE.md
kotp: I would guess that if I were talking about "lower case" it would be the beer that is lower than the other case of beer.  While saying "lowercase" is talking about letters.

STYLE-GUIDE.md
iHiD: I'd agree with lowercase, but I think this falls into the "it doesn't matter as long as its consistent within an exercise" bucket. Unless anyone feels strongly about this particular word, and wants to encode `lowercase` as a rule and find me another example for this bit :)

STYLE-GUIDE.md
cmcaine: I think if a bijection were required there would be some easy contextual way of describing the property (maybe we need things to be reversible, or maybe we just need a unique name for each robot and for there to be no names that don't correspond to robots). So this could be covered under "no jargon".

In general, I think "a one-to-one mapping" or "one-to-one correspondence" is more clear. If someone starts talking about surjections and bijections then I almost always just have to look it up. And btw, your definition is slightly wrong in that it doesn't ensure that the second set only contains elements paired with the first, nor that each element in the first set is paired with a unique element in the second.

This table is a nice way of explaining it for me:

image

Issue #1716, opened 2020-10-15
Add a rule to the style guide to explain abbreviations, acronyms and initialisms


iHiD:
_Originally posted by @wolf99 in https://github.com/exercism/problem-specifications/pull/1713#discussion_r505751243_

Please read the discussion there.

cmcaine:
@wolf99 said:

> Should there be a similar rule to explain abbreviations, acronyms and initialisms ?

> For example, a rule that is often used in academic writing is that an initialism should have its expansion in brackets immediately following its first use and thereafter can be used freely.


I said:

> I think we should have a no jargon rule. And don't use initialisms etc at all unless they are better known by the initialism (e.g. HTTP, HTML, radar) or if the text would be ridiculously verbose otherwise.

@wolf99 said:

> In favour of no jargon.

>However, outside of the specific language related initialism, should we assume that all users will have knowledge of what I or you might consider better known ones, even for non-native US English speakers?

>Yet we would still need to use initialisms in some cases. If we, hypothetically, agree that RADAR is jargon to non-US English speakers, then we would not have to be forced to use the long form each and every time it is mentioned, especially if it might form the name of test cases or some other oft repeated part of an exercise.

>thus we need some kind of middle solution (hence my suggestion 😉).

>I can think of a host of these that might be used depending on the exercise and track: FIFO, NASA, ABS, GIF, DIY, SONAR, SUV, UFO, CRUD, ANSI, API, ASCII, ICANN, CMYK, HTTP, FTP, DPI, DSL, JPEG, ... these are not especially unusual in programming or jargon-y, yet I can easily say that not everyone around the world would be familiar with all them

I think most of those terms are better known by their initialisms / acronyms, so there is no benefit from spelling them out. For example, if you're talking about NASA I won't really benefit from you saying the "The National Aeronautics and Space Administration (NASA)", instead you should say "The American space agency, NASA".

The only exceptions here are: ABS (could be one of several things); CRUD (jargon: avoid if possible); DPI (could be dots or deep packet inspection (jargon), just say the whole thing once or use a better term (pixel density or packet inspection)); DSL (jargon, spell it out); FIFO (jargon: prefer queue and stack to FIFO, FILO).

If there is some genuine reason to say CRUD or DSL loads (what?), then you can use the initialism with explanation under the second part of the rule I gave.

iHiD:
(cc @kotp who has opinions on acronyms)

kotp:
[...]

PR #1718, opened 2020-10-16
Add abbreviations and restructure slightly

iHiD:
This:
- Adds a block on abbreviations to resolve https://github.com/exercism/problem-specifications/issues/1716
- Restructures the document into sections.

### DIFF
--- a/STYLE-GUIDE.md
+++ b/STYLE-GUIDE.md
@@ -47,6 +47,16 @@ An some examples of good usage:
 - "NASA, the United States' space agency, launched the Mariner 2 space probe in..." _(NASA is not expanded because the "National Aerospace and Space Administration" is much better known by its acronym than by its expanded name)_
 - "The Department of Motor Vehicles (DMV) is filled with sloths. That's why everything takes forever at the DMV" _(Define DMV the first time it is used)_
 
+#### Exceptions
+
+Some abbreviations are considered common, useful, and non-technical enough that we have decided to permit them:
+- `e.g.` or `eg`
+- `i.e.` or `ie`
+- `etc.` or `etc`
+- `docs`
+
+Contractions (e.g. "won't", "I'm", "that's") should be used sparingly if at all in exercise descriptions, but are not restricted in other language around the site (e.g. website copy, mentoring).
+
 ## Choice of words
 
 ### Mathematical and jargon terms.


kotp (approved):

STYLE-GUIDE.md
```diff
@@ -2,22 +2,41 @@
...
+### Measurements
 All units of measurement must be [SI](https://en.wikipedia.org/wiki/International_System_of_Units) or [SI-derived](https://en.wikipedia.org/wiki/SI_derived_unit) units.
-In the future other translations may occur, but the "official" Exercism language is US English.
 
-## Explain or substitute mathematical and esoteric terms.
+### Abbreviations, acronyms and initialisms
+> All abbreviations, acronyms and initialisms should be defined. 
+> - If the unabbreviated term is clear, always include the full wording at first use (e.g. "Domain-specific language (DSL)");
+> - If the term needs more explaination to be clear to someone unfamiliar, explain it: (e.g. "The four basic types of SQL commands are Create, Read, Update, Delete. The initials of these come together to form the word CRUD.");
+> - If explaining the acryonm adds no tangible value to the student, simply introduce the acryonm as if it were a "proper noun" (e.g. "People's inherit DNA, a set of chemical instructions that influence how their bodies are constructed")
```
SleeplessByte, line 31:
```suggestion
> - If explaining the acryonm adds no tangible value to the student, simply introduce the acryonm as if it were a "proper noun" (e.g. "People's inherit DNA, a set of chemical instructions that influence how their bodies are constructed")
```

This literally leads to people worrying about that they miss something vital about the exercise and quitting because of it, complaining about it or struggling with it

STYLE-GUIDE.md
```diff
@@ -2,22 +2,41 @@
...
+All content should be written in US English. In the future other translations may occur, but the "official" Exercism language is US English.
+
+### Measurements
 All units of measurement must be [SI](https://en.wikipedia.org/wiki/International_System_of_Units) or [SI-derived](https://en.wikipedia.org/wiki/SI_derived_unit) units.
-In the future other translations may occur, but the "official" Exercism language is US English.
 
-## Explain or substitute mathematical and esoteric terms.
+### Abbreviations, acronyms and initialisms
+> All abbreviations, acronyms and initialisms should be defined. 
+> - If the unabbreviated term is clear, always include the full wording at first use (e.g. "Domain-specific language (DSL)");
```
SleeplessByte, line 29: What does "If the unabbreviated term is clear" mean? The term "Domain" in this case is jargon *and* ambiguous in the context of programming languages. It's term also doesn't _explain_ what it is. So what is the rule exactly here?

STYLE-GUIDE.md
```diff
@@ -2,22 +2,41 @@
...
+> - If the term needs more explaination to be clear to someone unfamiliar, explain it: (e.g. "The four basic types of SQL commands are Create, Read, Update, Delete. The initials of these come together to form the word CRUD.");
```
SleeplessByte, line 30: Assume _all_ acronyms are unfamiliar, because the reality is, most of them are.

SleeplessByte (changes requested):

SleeplessByte (changes requested):
I've left my actual commentary here: https://github.com/exercism/problem-specifications/issues/1716#issuecomment-709676906

STYLE-GUIDE.md
iHiD: I don't get your comment at all :) You think we should have to write out "deoxyribonucleic acid" rather than just using DNA? I think that just trying to pronounce that in my head would put me off.

cmcaine, line 31:
```suggestion
> - If explaining the abbreviation adds no tangible value to the student, simply introduce it as if it were a "proper noun" (e.g. "People inherit DNA, a set of chemical instructions that influence how their bodies are constructed")
```

Did you mean to put this whole section in a blockquote too?

I think there should be an additional rule about avoiding jargon or abbreviations where another term would be fine (so don't talk about FIFOs if you could instead talk about queues; don't talk about things being RESTful if you can instead talk about what specific properties are good; don't talk about CRUD if you don't need to; don't use a jargon-heavy theme unless it really adds to the exercise; etc.)

SleeplessByte: (see slack for response, then we can come to a clear message and put it here)

SleeplessByte: It should be made clear that _understanding_ what DNA is or what it stands for, isn't important for the exercise or completing it. That should be the rule when something doesn't add tangible value.

It's fine for the story to have "things", but we must make clear that they're not important to understand. Students _will_ spend time looking things up, which is, as you mentioned on slack, a waste.

STYLE-GUIDE.md
```diff
@@ -2,22 +2,41 @@
...
+Where a consistent style has not been agreed within this document, these must be consistent within an exercise.
+
+### Exceptions
+
+Exercises may vary from these rules if there is general consensus across maintainers that this is reasonable. 
+For example, an exercise about converting between different units might choose not to use SI measurements.
+
+## Language
+
+All content should be written in US English. In the future other translations may occur, but the "official" Exercism language is US English.
```
ErikSchierboom, line 22: I was wondering if there is a resource we could link to that lists some of the differences between US English and UK English? That way people have a convenient resource to check their text against.

STYLE-GUIDE.md
ErikSchierboom: I had the same question :) The example seems to me an application of the rule above it ("All abbreviations, acronyms and initialisms should be defined.")

STYLE-GUIDE.md
ErikSchierboom: ☝️ I like this

ErikSchierboom (approved):

STYLE-GUIDE.md
```diff
@@ -2,22 +2,41 @@
...
+
+### Measurements
 All units of measurement must be [SI](https://en.wikipedia.org/wiki/International_System_of_Units) or [SI-derived](https://en.wikipedia.org/wiki/SI_derived_unit) units.
-In the future other translations may occur, but the "official" Exercism language is US English.
 
-## Explain or substitute mathematical and esoteric terms.
+### Abbreviations, acronyms and initialisms
+> All abbreviations, acronyms and initialisms should be defined.
```
SaschaMann, line 28: Could we use `<abbr>` (and perhaps `<dfn>`) for this? That way people who know what it means can ignore it, but those that don't can look up the meaning. It avoids cluttering up the text and given that it's part of HTML, it'll be accessible too.

E.g. `<dfn id="html"><abbr title="HyperText Markup Language">HTML</abbr>
</dfn> is a markup language used to create the semantics and structure
of a web page.` (from https://developer.mozilla.org/de/docs/Web/HTML/Element/abbr#Eine_Abk%C3%BCrzung_definieren)

cmcaine: Personally, I don't really like the hover-to-show definition user experience and prefer just spelling it out in the prose.

iHiD:
Rewritten based on @cmcaine and @SleeplessByte's wording.

SaschaMann (approved):
LGTM! TY, @iHiD et al.!

SleeplessByte (approved):
This is great now. Thanks!

STYLE-GUIDE.md
```diff
@@ -2,13 +2,54 @@
...
+- Prefer "queue" to "FIFO" and "stack" to "FILO"
+- Instead of describing an interface as "RESTful", describe it's specific properties
+- If you must use CRUD, explain that it stands for Create, Read, Update, Delete and that these are the basic SQL actions
+
+An some examples of good usage:
+- "HyperText Markup Language (HTML) is the language used to describe document structure and content on the web" _(expanded and explained)_
+- "DNA, a set of chemical instructions that influence how our bodies are constructed" _(DNA is not expanded because "deoxyribonucleic acid" is unlikely to help explain what DNA is to our audience)_
+- "NASA, the United States' space agency, launched the Mariner 2 space probe in..." _(NASA is not expanded because the "National Aerospace and Space Administration" is much better known by its acronym than by its expanded name)_
+- "The Department of Motor Vehicles (DMV) is filled with sloths. That's why everything takes forever at the DMV" _(Define DMV the first time it is used)_
+
```
cmcaine, line 49: The current rules forbid uses of "e.g.", "docs", "etc." that don't also define those terms. That's okay if we're sure it's what we want.

STYLE-GUIDE.md
```diff
@@ -2,13 +2,54 @@
...
+- The text will be excessively verbose without abbreviation
+
+When using abbreviations always explain what the term means on its first use. This will often, but not always, include expanding the abbreviation. It will rarely be sufficient to only expand the abbreviation.
+
+Here are some example rules:
+- Prefer "if I recall correctly" to "IIRC"
+- Prefer "as far as I know" to "AFAIK"
+- Prefer "queue" to "FIFO" and "stack" to "FILO"
+- Instead of describing an interface as "RESTful", describe it's specific properties
+- If you must use CRUD, explain that it stands for Create, Read, Update, Delete and that these are the basic SQL actions
```
cmcaine, line 42:
```suggestion
- If you must use "CRUD", explain that it stands for Create, Read, Update, Delete and that these are the basic actions in standard databases
```

Under the current rules, the author would also have to define what SQL means.

STYLE-GUIDE.md
iHiD: Yeah - can you put a seperate issue together for someone to do this pls?

STYLE-GUIDE.md
iHiD: Good spot.

cmcaine (approved):
This part is missing, but maybe that's deliberate or still being hashed out. Otherwise this seems good :+1: 

> When the theme of an exercise includes a potentially complex or confusing topic that doesn't matter to complete an exercise (e.g. DNA and nucleotides in the nucleotide count and Hamming distance exercises) include a disclaimer that it doesn't need to be understood to complete the exercise.

STYLE-GUIDE.md
SleeplessByte: Right. `etc.` `e.g.` and `ie.` id' like to have as _exceptions_.

kotp: What about contractions?

- I'd vs I would
- won't vs would not
- couldn't vs could not

cmcaine: won't is short for "will not" :)

I think contractions are fine and don't fall under this rule, but we could add a rule that they're fine?

SleeplessByte: Yeah, contractions in général are okay I think.

kotp: But sometimes hard to get right.  Regionally (to some ill-conceived idea of "region"), "I'd" might be "I could" or "I would" or "I should" but how do you know which is which as a non-native speaker?  And of course the "would" or "will" problem with "won't".  It is just as easy to write carefully and avoid the contraction for the reader.  (Burden on the writer).  The same way we are careful in how we write e-mail, and don't top post, because it is easier for the writer, but harder on the reader.  (Our readers are kind, we should be kind as well.)

cmcaine: I don't know of any regional dialect that interprets won't as would not or I'd as anything but "I would" 🤷 

I don't want to have to write out don't and won't without contractions all the time, that would feel and (imo) read weird; as if I was putting on an accent.

STYLE-GUIDE.md
SleeplessByte: (could be a user setting, expand by default)

STYLE-GUIDE.md
SleeplessByte: Don't won't can't shouldn't couldn't wouldn't are probably all fine and more well known.

Shan't mayn't mightn't mustn't oughtn't needn't and daren't probably are not.

kotp: Some of those are pretty common where I am, and some are even more common where I have been.  For writing, I avoid them, because I know I write normally to a more "global" audience.

SleeplessByte: What I meant is that the top list, together with the following:

- that's 
- there's
- what's 
- we're 
- there's
- they're
- that's 
- he'll 
- she'll 
- you'll 
- I'm, 
- I've, 
- let's 
- isn't 
- aren't

...are taught as common contraction, according to my recollection, and the bottom ones aren't. In some of these cases, they might even be more well known than the un-contracted form, but I have no data on that. Just a feeling.

I really don't want to disallow _contractions_, but IF we will disallow them, I would favour a list that's only the uncommon contractions.

kotp: I try to avoid all of them, but yes, the common ones, when I write, I generally write in my "voice" and so they do at times sneak in.  Maybe we should state "disfavor" contractions, but not outright disallow 'em.

wolf99 (approved):

STYLE-GUIDE.md
```diff
@@ -2,13 +2,54 @@
...
+- Prefer "as far as I know" to "AFAIK"
+- Prefer "queue" to "FIFO" and "stack" to "FILO"
+- Instead of describing an interface as "RESTful", describe it's specific properties
+- If you must use "CRUD", explain that it stands for Create, Read, Update, Delete and that these are the basic actions in standard databases
+
+An some examples of good usage:
+- "HyperText Markup Language (HTML) is the language used to describe document structure and content on the web" _(expanded and explained)_
+- "DNA, a set of chemical instructions that influence how our bodies are constructed" _(DNA is not expanded because "deoxyribonucleic acid" is unlikely to help explain what DNA is to our audience)_
+- "NASA, the United States' space agency, launched the Mariner 2 space probe in..." _(NASA is not expanded because the "National Aerospace and Space Administration" is much better known by its acronym than by its expanded name)_
+- "The Department of Motor Vehicles (DMV) is filled with sloths. That's why everything takes forever at the DMV" _(Define DMV the first time it is used)_
```
iHiD, line 48:
```suggestion
- "The Department of Motor Vehicles (DMV) is filled with sloths. That's why everything takes forever at the DMV" _(Define DMV the first time it is used)_

#### Exceptions

Some abbreviations are considered common, useful, and non-technical enough that we have decided that should be allowed:
- `e.g.` or `eg`
- `i.e.` or `ie`
- `etc.` or `etc`
- `docs`

Contractions (e.g. "won't", "I'm", "that's") should be used sparingly if at all in exercise descriptions, but are not restricted in other language around the site (e.g. website copy, mentoring).
```

STYLE-GUIDE.md
iHiD: Good discussion - thanks. 

I've added/merged a suggestion above - pls check it over. I don't think we need contractions generally in exercises, but I'm not going to restrict them in mentoring as I want people to be able to use their voice.

cmcaine (approved):

STYLE-GUIDE.md
```diff
@@ -47,6 +47,16 @@ An some examples of good usage:
...
 
+#### Exceptions
+
+Some abbreviations are considered common, useful, and non-technical enough that we have decided to permit them:
+- `e.g.` or `eg`
+- `i.e.` or `ie`
+- `etc.` or `etc`
+- `docs`
+
+Contractions (e.g. "won't", "I'm", "that's") should be used sparingly if at all in exercise descriptions, but are not restricted in other language around the site (e.g. website copy, mentoring).
```
SaschaMann, line 58: Doesn't this belong in "Choice of words"?

iHiD: Maybe, but as the discussion came up here I figured it might fit. As we flesh this whole thing out more we'll probably find a better structure in general.

iHiD:
Merging. Thanks everyone. Feel free to PR improvements 🙂

PR #1720, opened 2020-10-17
Style: Fix whitespace issues

ee7:
This PR:
- Removes trailing whitespace (ignoring `ocr-numbers`, where it is apparently significant).
- Adds final newlines where they were missing

It fixes those whitespace issues in `configlet`-generated `README.md` files for:
- `darts`
- `resistor-color`
- `resistor-color-duo`
- `rest-api`

Note that, because the `darts` and `rest-api` description files were missing a final newline, `configlet generate` can produce a `README.md` that looks like:
```
The outer circle has a radius of 10 units (This is equivalent to the total radius for the entire target), the middle circle a radius of 5 units, and the inner circle a radius of 1. Of course, they are all centered to the same point (That is, the circles are [concentric](http://mathworld.wolfram.com/ConcentricCircles.html)) defined by the coordinates (0, 0).

Write a function that given a point in the target (defined by its `real` cartesian coordinates `x` and `y`), returns the correct amount earned by a dart landing in that point.
## Running the tests
```

That is, missing a newline before the `## Running the tests`.

It would be nice to have a CI check for whitespace issues.


exercises/ocr-numbers/description.md
```diff
@@ -38,10 +38,10 @@ Update your program to recognize multi-character binary strings, replacing garbl
 Update your program to recognize all numbers 0 through 9, both individually and as part of a larger string.
 
 ```text
- _ 
+ _
  _|
-|_ 
-   
+|_
+
```
SaschaMann, line 44: Are you sure that it's not intended here?

ee7: I believe that it's not intended.

And if I run `configlet generate` using a `problem-specifications` repo that's checked out to this PR then I get:
[before](https://github.com/ee7/python/blob/2d846bd69b3145e012c19e5e989d2a1516357c9b/exercises/ocr-numbers/README.md) vs [after](https://github.com/ee7/python/blob/7d1c76a267846cba6f886ee80d21050a6ac21593/exercises/ocr-numbers/README.md)

petertseng: Although we declared that the whitespace should be kept in https://github.com/exercism/problem-specifications/pull/1627#pullrequestreview-335650064, any decision is acceptable to me here, just wanted to make sure the previous information is considered.

ee7: @petertseng Thanks for linking the previous discussion.

In my opinion: if this PR's whitespace changes to `ocr-numbers` are not significant then now is a good opportunity to make this change (now that there's no version string we have to bump).

But this PR includes a whitespace change that was said to be significant in that past discussion. Why is it significant? Pinging @NobbZ, who wrote:  
> The whitespace in this code block is significant and not allowed to be removed

I just want to be able to:
1. Run `configlet generate`.
1. Commit the `README.md` changes without seeing an error (I configure `git` to complain about trailing whitespace).

SleeplessByte: It's significant in that exercise. It _must_ output a newline.

Add `<!-- prettier-ignore -->`, or better `<!-- prettier-ignore-start -->` `<!-- prettier-ignore-end-->` around the codeblock to solve this.

ee7: Okay, thanks. I've reverted the change to `ocr-numbers`.

My motivation for this PR was fixing the whitespace in `resistor-color` and `resistor-color-duo`, rather than poking things elsewhere.

exercises/rest-api/description.md
```diff
@@ -35,4 +35,4 @@ Your task is to implement a simple [RESTful API](https://en.wikipedia.org/wiki/R
 - https://restfulapi.net/
 - Example RESTful APIs
   - [GitHub](https://developer.github.com/v3/)
-  - [Reddit](https://www.reddit.com/dev/api/)
```
SaschaMann, line 38: Why's this highlighted as change? I can't see any change here?

SaschaMann (approved):

iHiD (approved):

kotp (approved):

PR #1725, opened 2020-10-24
Workflows: Avoid duplicate builds

SaschaMann:
This will prevent duplicate builds in PRs where the PR branch is part of the repo and not a fork.

### DIFF
--- a/workflows/configlet.yml
+++ b/workflows/configlet.yml
@@ -2,7 +2,13 @@
 
 name: configlet
 
-on: [push, pull_request, workflow_dispatch]
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - master
+  pull_request:
 
 jobs:
   lint:


SleeplessByte:
Just so I can learn some more, can you explain what goes wrong currently and how does this fix it?

SaschaMann:
Of course. Right now, if you, as a maintainer with write access to the repo, create a branch and push a commit to it, it will run the build (-> the `push` event). If you open a PR that merges that branch into the main branch, this will trigger a second build (-> the `pull_request` event). It will have two identical builds for the same commit on those PR-branches. For an example, look at the checks here: https://github.com/exercism/julia/pull/260

This change makes it so that pushes only trigger a build if they're on one of the listed branches, i.e. `master` (and for future-proofing/copy-pastability `main`).

This means that in the situation above, it would only run the `pull_request` builds, not the `push` builds.

The downside is that if you're working on something before opening a PR, it will not run builds on that branch. That can be worked around by either triggering a build manually or using Draft PRs though.

SleeplessByte:
So if I look at that PR, you mean that I see:

- Check Notebooks / check_notebooks (pull_request)
- Check Notebooks / check_notebooks (push)

Is that what you mean with "twice with the same commit"?

SaschaMann:
Yes, that's what I meant. 

> but pull_request actually runs on a detached head after merging.

I wasn't aware of that difference[1] but for our case, that just strengthens the case to restrict push to the main branch. Whether or not the branch on its own works is irrelevant, we want to know if it would break when merged into master, don't we?

***

[1] The docs make it pretty hard to find out:

> The workflow runs on your repository's code at the same commit SHA and Git ref that triggered the event. When a workflow runs, GitHub sets the GITHUB_SHA (commit SHA) and GITHUB_REF (Git ref) environment variables in the runner environment. 

The UI shows that the commit that triggered the event is the same:

grafik

SleeplessByte:
The docs are horrible on this.

I found out by remebering that someone staff member said the events do something different, and then go through the actual log.

SleeplessByte:
I agree with your thinking. We want to know if it works after push.

However, you lose information (and time) if you can't tell if your branch' is in an invalid state, or the merge thereafter will be. For this reason, I think it's fine as a suggestion in the readme.md, but not as a default.

In my opinion, I'd consider having them both -- I wish we could rename them be clear "after merge" and "commit".

PR #1729, opened 2020-11-01
proverb: add one test case on inputs leading with vowels

qudongfang:


exercises/proverb/canonical-data.json
```diff
@@ -103,6 +103,21 @@
                 "For want of a soldier the battle was lost.",
                 "And all for the want of a pin."
             ]
+        },
+        {
+          "uuid": "4da2d184-a631-4f31-b770-52cccca54dd1",
+          "description": "five pieces modernized",
```
petertseng, line 109: the reason this test case is important is because there is an element beginning with a vowel; the number of pieces is unimportant. Therefore, the description should reflect this and say "item beginning with vowel" or similar.

qudongfang: Yeah, def
I'll update the comment.

glennj:
This could very well be its own exercise, since the rules are not quite so straightforward for "an" vs "a". For example:
* "an umbrella" but "a university"
* "an hour" but "a horse"
* "an FBI badge" but "a CIA badge"
https://www.grammar.com/a-vs-an-when-to-use/

qudongfang:
> https://www.grammar.com/a-vs-an-when-to-use/

Good call! I'll close this PR then.

PR #1735, opened 2020-12-11
Grade school exercise: A student cannot simultaneously be in two grad…

pedrorolo:
In the grade school exercise names are meant to be unique and thus a user should only belong in a grade. This PR adds canonical data to verify this condition. And it has been discussed and implemented for the typescript class:

- [Former Issue](https://github.com/exercism/typescript/issues/361)
- [Pull Request on the Typescript track](https://github.com/exercism/typescript/pull/362)

### DIFF
--- a/exercises/grade-school/canonical-data.json
+++ b/exercises/grade-school/canonical-data.json
@@ -15,6 +15,19 @@
             "expected": ["Aimee"]
         },
         {
+          "uuid": "dece43c8-3ba5-11eb-8fdf-7f8daeaeb5f2",
+          "description": "A student can't be in two different grades",
+          "property": "roster",
+          "input": {
+            "students": [
+              ["Aimee", 2],
+              ["Aimee", 1]
+            ],
+            "desiredGrade": 2
+          },
+          "expected": []
+        },
+        {
             "uuid": "233be705-dd58-4968-889d-fb3c7954c9cc",
             "description": "Adding more students adds them to the sorted roster",
             "property": "roster",


iHiD (approved):
Looks good to me. Thanks!

SaschaMann (approved):

SleeplessByte (approved):
Thank you for doing this @pedrorolo

exercises/grade-school/canonical-data.json
```diff
@@ -14,6 +14,19 @@
             },
             "expected": ["Aimee"]
         },
+        {
+          "uuid": "dece43c8-3ba5-11eb-8fdf-7f8daeaeb5f2",
```
ee7, line 18: Note that this isn't a valid version 4 UUID (because the 14th character is not 4).

It looks like all the other UUIDs in `problem-specifications` are valid version 4 UUIDs.

SleeplessByte: I don't think we have the limitation of it being a v4. 😬

ErikSchierboom: I think for v3 we wanted to use v4 I think, but I don't know why anymore 🤷 @iHiD might know

SleeplessByte: Someone can PR a change for this one. I don't necessarily think there's much value in forcing it to be totally random (instead of also using date-time and MAC address). The purpose is to be unique, and that it still is 🥇

ee7: > I don't think we have the limitation of it being a v4.

Indeed, it's not currently in the `problem-specifications` CI, or the v2 `configlet lint`. But I think it should be. As I understand it, we'll fix the (not-so-unique) `uuid`s that are duplicated between tracks. So I think we might as well enforce a consistent format for `uuid`s on track repos at the same time. Currently, about 2% of `uuid`s on Exercism are not valid version 4 UUIDs that are also lowercase. 

It's true that there isn't a meaningful difference in the collision probability, but UUID version 4 is the right version for a randomly generated UUID. It's also what `configlet uuid` generates (in both in the v2 and v3 versions of `configlet`), and every other `problem-specifications` `uuid` is version 4.

> Someone can PR a change for this one.

I'll do that, but we'd also have to PR to any track that adds the old UUID to a `tests.toml` file. So far, that's only on the `javascript` track [here](https://github.com/exercism/javascript/blob/316084ddfe56b5e6950615de166ecc405806767f/exercises/grade-school/.meta/tests.toml#L7). I'll PR there too.

> The purpose is to be unique, and that it still is

True. But we shouldn't go too far with that argument - otherwise it implies that we should merge a `uuid` that looks like [this](https://github.com/exercism/bash/blob/59adf298d54d20a2b498a813abe2c2e4a6cd0385/config.json#L932). That *is* the most unique `uuid`, though :)

SleeplessByte: > Furthermore, a reader that understands UUIDs might see a version 1 UUID and assume that the timestamp component is useful, causing them to wonder why the others are version 4.

Yeah, that's a good point!

> 'll do that, but we'd also have to PR to any track that adds the old UUID to a tests.toml file. So far, that's only on the javascript track here. I'll PR there too.

Sweet thanks! I'll ✅ the PR if you ping me, and lets have reviewers ✅ it to, as to reduce the amount of tracks that will need to do this.

> True. But we shouldn't go too far with that argument - otherwise it implies that we should merge a uuid that looks like this. That is the most unique uuid, though :)

😏

iHiD: I'd quite like to have `${PREFIX}-${v4 UUID}` everywhere. Where the prefix signifies the context (e.g. this would be `PSTC` (Problem specs test case) or similar)

PR #1736, opened 2020-12-12
rational-numbers: Add ordering to spec

SaschaMann:
This is a fun one, because I can't think of any sensible way to represent this in JSON. Therefore the test data is described in a comment. Testing only specific order operations would require adding dozens, possibly hundreds, of tests to cover all cases, so I don't think that's a feasible way of adding it to the canonical data.

An example implementation can be found [on the Julia track](https://github.com/exercism/julia/blob/005c0042b25e861240fc18c7fd83eb0a4aeb0dc9/exercises/rational-numbers/runtests.jl#L103-L129).

---

This also raises the question: How are new optional test cases handled if they require an addition to the description?

### DIFF
--- a/exercises/rational-numbers/canonical-data.json
+++ b/exercises/rational-numbers/canonical-data.json
@@ -422,6 +422,30 @@
           "expected": [1, 1]
         }
       ]
+    },
+    {
+      "uuid": "bd1eab16-7603-4a77-ae4e-23c930fe39cd",
+      "description": "Ordering",
+      "property": "<",
+      "comments": [
+        "Depending on your language's features, you may have to implement the following operators:",
+        "=, !=, <=, <, >=, >",
+        "",
+        "To test the ordering, it's recommended to iterate through a = -5..5, b = -5..5, c = -5..5, d = -5..5,",
+        "define rational numbers r = a / b, s = c / d,",
+        "and compare the result of r $op c,",
+        "where $op = {=, !=, <=, <, >=, >}",
+        "to the result of the corresponding floating point comparisons",
+        "a / b $op c. Analogously, compare r $op s with a / b $op c / d.",
+        "Skip the iteration step when a == b == 0 or c == d == 0, since 0/0 is undefined.",
+        "",
+        "Depending on your implementation of the exercise, you may need to consider",
+        "division by zero differently in the iteration.",
+        "",
+        "This testing method is based on Julia Base tests[1], released under the MIT licence[2].",
+        "[1]: https://github.com/JuliaLang/julia/blob/52bafeb981bac548afd2264edb518d8d86944dca/test/rational.jl",
+        "[2]: https://github.com/JuliaLang/julia/blob/52bafeb981bac548afd2264edb518d8d86944dca/LICENSE.md"
+      ]
     }
   ]
 }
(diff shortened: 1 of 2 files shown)


exercises/rational-numbers/canonical-data.json
```diff
@@ -422,6 +422,30 @@
           "expected": [1, 1]
         }
       ]
+    },
+    {
+      "uuid": "bd1eab16-7603-4a77-ae4e-23c930fe39cd",
+      "description": "Ordering",
```
ErikSchierboom, line 428: I don't have a better alternative, but this description is fairly minimal. Maybe something like `Rational numbers can be ordered` or something like that?

exercises/rational-numbers/canonical-data.json
```diff
@@ -422,6 +422,30 @@
...
+      "property": "<",
```
ErikSchierboom, line 429: In the [diffie-hellman exercise](https://github.com/exercism/problem-specifications/blob/master/exercises/diffie-hellman/canonical-data.json#L31), we experimented with using a textual description of the property. This might be useful here too.

ErikSchierboom (approved):
> This also raises the question: How are new optional test cases handled if they require an addition to the description?

I don't understand this. Addition to what description?

exercises/rational-numbers/canonical-data.json
SaschaMann: The name is in the style of the other names, like "absolute value" or "multiplication". Do you think it should be longer anyway?

SaschaMann (commented):
(the app won't let me add a single comment)

exercises/rational-numbers/canonical-data.json
ErikSchierboom: There is some inconsistency in the test naming. I'm fine with keeping it as is therefore, but let's see what others think.

SaschaMann:
> I don't understand this. Addition to what description?

To the description.md

exercises/rational-numbers/canonical-data.json
SaschaMann: Interesting, didn't know about that exercise. Do you think that format would be better here? I feel like it'd need special handling in generators either way, so I'm not sure if a description like in diffie-hellman has an advantage? It would likely need the comment anyway for extra explanations.

ErikSchierboom:
No we don't yet have a way to elegantly handle this.

exercises/rational-numbers/canonical-data.json
ErikSchierboom: It would definitely need the comment. I do think a textual property might be somewhat easier to understand. What do you think @SleeplessByte?

SleeplessByte: Special handling is required either way (see linked-list); I think written out properties are more consistent with the other exercises.

Wouldn't care here 🤷🏽‍♂️🤗

SaschaMann: Okay, I'll write it out (that'll also make the schema check happy)

ErikSchierboom:
@SaschaMann Are you still interested in working on this PR?

SaschaMann:
No but if anyone wants to pick it up, feel free.

kotp:
Should we have a "Available for pickup" label created for issues like these? @ErikSchierboom @iHiD

PR #1739, opened 2020-12-17
change: add simplest test case

SleeplessByte:
In JavaScript we've had this testcase to kick-off the exercise, input and output, without needing to write an actual algorithm. Adding it here for completeness.

### DIFF
--- a/exercises/change/canonical-data.json
+++ b/exercises/change/canonical-data.json
@@ -7,83 +7,187 @@
   ],
   "cases": [
     {
+      "uuid": "d0ebd0e1-9d27-4609-a654-df5c0ba1d83a",
+      "description": "change for 1 cent",
+      "property": "findFewestCoins",
+      "input": {
+        "coins": [
+          1,
+          5,
+          10,
+          25
+        ],
+        "target": 1
+      },
+      "expected": [
+        1
+      ]
+    },
+    {
       "uuid": "36887bea-7f92-4a9c-b0cc-c0e886b3ecc8",
       "description": "single coin change",
       "property": "findFewestCoins",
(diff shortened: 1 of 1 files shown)


SaschaMann (approved):
LGTM. For better readability of the commit history, I'd prefer if you could split the test addition and the formatting to two commits and then rebase them.

SleeplessByte:
I'm unwilling to turn of editor formatting. It really breaks up the workflow where I am editing the track(s) in one window and the tests in another. What I could do is "just save it first", commit that, and then make my addition in a separate commit. Would it suffice?

If yes, I'll do that going forward.

SaschaMann:
> What I could do is "just save it first", commit that, and then make my addition in a separate commit. Would it suffice?

Yeah that's fine. Order doesn't matter but if for whatever reason we need to find something in the commit history, mashing formatting with content changes makes it harder.

(Perhaps we should setup the everloved prettier actions here too :D )

SleeplessByte:
Understood! Fixed behaviour in new PRs :)

ErikSchierboom:
> (Perhaps we should setup the everloved prettier actions here too :D )

I think that for this repo we could even go as far as automatically applying prettier :)

ErikSchierboom (approved):
> LGTM. For better readability of the commit history, I'd prefer if you could split the test addition and the formatting to two commits and then rebase them.

I agree with this. It is slightly inconvenient to figure out what exactly was added and what was reformatted. 

Good addition!

valentin-p (approved):

ErikSchierboom:
Thanks @SleeplessByte!

PR #1740, opened 2020-12-17
diffie-hellman: add immutability / purity test case

SleeplessByte:
JavaScript has a test case to test if the algorithm is side-effect free, or has the appearance that it is. 

- Tracks that have a functional implementation may test if running the same function twice in succession generates the same result (not this test case)
- Tracks that have an object-oriented approach may test if running the same method with different input generates the correct result (which it arguably wouldn't if p or g were mutated in the process)

### DIFF
--- a/exercises/diffie-hellman/canonical-data.json
+++ b/exercises/diffie-hellman/canonical-data.json
@@ -1,18 +1,20 @@
 {
   "exercise": "diffie-hellman",
   "comments": [
-    "Optional tests to consider:",
-    "* Validation of parameters:",
-    "  Validate that `p`, `g` are valid.",
-    "  Validate that keys given as inputs are valid.",
-    "  Resources that show what happens if parameters are not validated:",
-    "  http://cryptopals.com/sets/5/challenges/34",
-    "  http://cryptopals.com/sets/5/challenges/35",
-    "* Large numbers:",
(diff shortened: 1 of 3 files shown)


ErikSchierboom (approved):

exercises/diffie-hellman/canonical-data.json
```diff
@@ -45,6 +47,26 @@
...
+        "public key from different values, for example from the test with     ",
+        "uuid (b4161d8e-53a1-4241-ae8f-48cc86527f22).                         "
+      ],
+      "property": "publicKey",
+      "input": {
+        "p": 23,
+        "g": 5,
+        "privateKey": 15
+      },
+      "expected": 19,
```
petertseng, line 65: this value is correct because `5 ** 15 % 23 == 19`.

petertseng (approved):
the case added is correct. please do not squash commits when merging. the schema change would ideally go into its own commit. I express no views on other issues such as immutability.

SleeplessByte:

ErikSchierboom:
@SleeplessByte There is a merge conflict.

SleeplessByte:
@exercism/reviewers it's ready!

neenjaw (approved):
🎉

SleeplessByte:
Ha. I swear I didn't press squash 🤷🏽

PR #1741, opened 2020-12-17
flatten-array: add intermediate TDD-supporting tests

SleeplessByte:
JavaScript has these additional tests that do two things:
- provide better stepping stones (TDD)
- test null values are not handled differently when there is nothing to flatten

### DIFF
--- a/exercises/flatten-array/canonical-data.json
+++ b/exercises/flatten-array/canonical-data.json
@@ -2,56 +2,236 @@
   "exercise": "flatten-array",
   "cases": [
     {
+      "uuid": "8c71dabd-da60-422d-a290-4a571471fb14",
+      "description": "empty",
+      "property": "flatten",
+      "input": {
+        "array": []
+      },
+      "expected": []
+    },
+    {
       "uuid": "d268b919-963c-442d-9f07-82b93f1b518c",
       "description": "no nesting",
       "property": "flatten",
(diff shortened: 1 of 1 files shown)


exercises/flatten-array/canonical-data.json
```diff
@@ -1,57 +1,237 @@
...
+          null,
+          [
+            [
+              null,
+              null
+            ],
+            null
+          ],
+          null
+        ]
```
ErikSchierboom, line 234: I think this a great example of where the formatting helps me visualize the structure better 👍

ErikSchierboom (approved):
LGTM

petertseng (approved):
the new test cases added should be pretty obviously correct, but I confirm they have checked programmatically just in case.. formatting not in its own commit is regrettable, but track maintainers can take comfort in the fact that immutable tests means that it's guaranteed that the test content has not changed, only the formatting. If that comment looks familiar, yes it's the same comment as was made in another PR, but it is just as applicable.

SleeplessByte:
I have moved to formatting separately in future PRs to make that even better DX-wise.

SleeplessByte:
For the person merging: Squash!

ErikSchierboom:
CC @exercism/reviewers

wolf99 (approved):

PR #1742, opened 2020-12-17
forth: add negative numbers and local-scope test

SleeplessByte:
@SaschaMann here I've done format first as separate commit.

### DIFF
--- a/exercises/forth/canonical-data.json
+++ b/exercises/forth/canonical-data.json
@@ -610,6 +610,36 @@
           "expected": {
             "error": "undefined operation"
           }
+        },
+        {
+          "uuid": "3c8bfef3-edbb-49c1-9993-21d4030043cb",
+          "description": "only defines locally",
+          "comments": [
+            "This test is meant to determine if instructions that define      ",
+            "words are executed in the state of the program or locally. For   ",
+            "example, in OOP-based approaches, the words should be defined in ",
+            "instance the instructions were evaluated only.                   "
+          ],
+          "scenarios": [
+            "local-scope"
+          ],
+          "input": {
+            "instructionsFirst": [
+              ": + - ;",
+              "1 1 +"
+            ],
+            "instructionsSecond": [
+              "1 1 +"
+            ]
+          },
+          "expected": {
+            "first": [
+              0
+            ],
+            "second": [
+              2
+            ]
+          }
         }
       ]
     },
--- a/exercises/forth/canonical-data.json
+++ b/exercises/forth/canonical-data.json
@@ -25,6 +25,23 @@
             4,
             5
           ]
+        },
+        {
+          "uuid": "fd7a8da2-6818-4203-a866-fed0714e7aa0",
+          "description": "pushes negative numbers onto the stack",
+          "property": "evaluate",
+          "input": {
+            "instructions": [
+              "-1 -2 -3 -4 -5"
+            ]
+          },
+          "expected": [
+            -1,
+            -2,
+            -3,
+            -4,
+            -5
+          ]
         }
       ]
     },
--- a/exercises/forth/canonical-data.json
+++ b/exercises/forth/canonical-data.json
@@ -14,9 +14,17 @@
           "description": "numbers just get pushed onto the stack",
           "property": "evaluate",
           "input": {
-            "instructions": ["1 2 3 4 5"]
+            "instructions": [
+              "1 2 3 4 5"
+            ]
           },
-          "expected": [1, 2, 3, 4, 5]
+          "expected": [
+            1,
+            2,
+            3,
+            4,
+            5
+          ]
         }
       ]
     },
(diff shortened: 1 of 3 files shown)


exercises/forth/canonical-data.json
```diff
@@ -424,25 +598,66 @@
...
+          }
+        },
+        {
+          "uuid": "3c8bfef3-edbb-49c1-9993-21d4030043cb",
+          "description": "only defines locally",
+          "comments": [
+            "This test is meant to determine if instructions that define      ",
+            "words are executed in the state of the program or locally. For   ",
+            "example, in OOP-based approaches, the words should be defined in ",
+            "instance the instructions were evaluated only.                   "
```
ErikSchierboom, line 638: This reads a bit weird to me.

ErikSchierboom (approved):
Small comment.

exercises/forth/canonical-data.json
SleeplessByte: Oh it is a bit weird.

SleeplessByte, line 638:
```suggestion
            "example, in OOP-based approaches, the words should be defined in ",
            "the same scope, or local to the instance the instructions were   ",
            "evaluated in.                                                    "
```

ErikSchierboom: New version is 👍

exercises/forth/canonical-data.json
```diff
@@ -424,25 +598,67 @@
...
+            "example, in OOP-based approaches, the words should be defined in ",
+            "the same scope, or local to the instance the instructions were   ",
+            "evaluated in.                                                    "
+          ],
+          "scenarios": [
+            "local-scope"
+          ],
+          "property": "evaluate",
+          "input": {
+            "instructionsFirst": [
```
petertseng, line 646: I do not believe we have any other instances of canonical data where two test cases with the same `property` have different keys in their `input`. Before, it the case that same `property` implies that `input` contains the same keys for all instances of that `property`. However, I have looked for any documentation of this fact, and found none in README.md or any other place I thought to look.

This would have been a bigger problem before we had an `input` object, but it may not be an issue anymore, since generators now can know to look in a well-defined place (the `input` object) for the presence or absence of any keys that they might be expecting.

Therefore, you'll not get an objection from me if you proceed with this, but you would have gotten one before we had an `input` object. You'll get to be the one breaking new ground.

SleeplessByte: I didn't know how to do this in any other way, but based it on "similar" test cases where we have thingA and thingB. 

Do you think this is an _okay enough_ aproach?

SleeplessByte:

exercises/forth/canonical-data.json
SleeplessByte: Oh, maybe we do: `evaluateBothAreEqual` ?

ErikSchierboom: I like the separate property 👍

SleeplessByte: I think I do too. Not overloading any meaning

SleeplessByte:
I re-formatted this on a different machine, with much more sensical results 🙈

ErikSchierboom (dismissed):
Changes look good!

angelikatyborska (approved):

ErikSchierboom:
@SleeplessByte Needs a rebase.

ErikSchierboom:
CC @exercism/reviewers

petertseng (approved):
only way I was willing to approve this is if the formatting changes are removed. Thought it was better to have separate commands on separate lines:

```diff
-            "instructions": [": foo dup ;", "1 FOO Foo foo"]
+            "instructions": [
+              ": foo dup ;",
+              "1 FOO Foo foo"
+            ]
```

The `+` version looks more readable to me.

So I've kept just the new tests. What do you think?

petertseng:
Since I've made unilateral changes that have not been reviewed yet, I think it would be a violation of process if I merged this. I'm going to ask for a review of what I did.

petertseng:
Just so that I am not carelessly discarding the formatting work that was previously done, I will extract them to their own PRs. In this way, the changes will be split up for individual review, rather than some of them being lost.

petertseng:
I'd like to apologise for my error in judgment - it was a mistake to make unilateral changes to this PR.

The thought process that led to my error was "This PR has been here for a long time. I like this part but not that part. Let's approve the part of it that I like and remove the rest. Don't want to delay the PR any more than it needs to be".

That was a mistake since I should have discussed it and voiced my concerns, and gotten agreement on a good course of action before proceeding. We don't want to get into an environment where we all feel the need to unilaterally override work that others have put their valuable time into, without any discussion. That drives us apart instead of having us work together.

I apologise for this error. I can rectify it by bringing back the original changes in their entirety all in this PR, or we can resolve it via https://github.com/exercism/problem-specifications/pull/1912 and https://github.com/exercism/problem-specifications/pull/1913. I will not repeat this mistake; I will discuss instead of making unauthorised changes.

exercises/forth/canonical-data.json
```diff
@@ -452,6 +461,24 @@
...
+        },
+        {
+          "uuid": "3c8bfef3-edbb-49c1-9993-21d4030043cb",
+          "description": "only defines locally",
+          "comments": [
+            "This test is meant to determine if instructions that define      ",
+            "words are executed in the state of the program or locally. For   ",
+            "example, in OOP-based approaches, the words should be defined in ",
+            "the same scope, or local to the instance the instructions were   ",
+            "evaluated in.                                                    "
```
kotp, line 473: To me, the padding could be cause for noisy diffs in the future, while sentence per line could make sense in this context as well.

petertseng:
Just had to rebase because #1913 got merged

petertseng:
Did the comment reflow (instead of wrapping at set width) since it's been requested without objection.

PR #1744, opened 2020-12-18
kindergarten-garden: Completionist

SleeplessByte:


exercises/kindergarten-garden/canonical-data.json
```diff
@@ -66,43 +91,183 @@
       "cases": [
         {
           "uuid": "149b4290-58e1-40f2-8ae4-8b87c46e765b",
-          "description": "first student's garden",
+          "description": "for Alice, first student's garden",
```
ErikSchierboom, line 94: I think we said that changing the description was a breaking change. I'm not totally sure though.

ErikSchierboom (commented):
I like the completionist additions!

exercises/kindergarten-garden/canonical-data.json
iHiD: I think description is mutable as long as it clarifies or fixes typos. Which this does.

ErikSchierboom: Agreed. I _think_ we forgot to document this :D

ErikSchierboom (approved):

exercises/kindergarten-garden/canonical-data.json
SleeplessByte: We forgot to document this, but I remembered 🥇

valentin-p (approved):

petertseng (approved):
all new test cases added are correct. formatting not in its own commit is regrettable, but track maintainers can take comfort in the fact that immutable tests means that it's guaranteed that the test content has not changed, only the formatting

PR #1745, opened 2020-12-18
linked-list: add canonical data

SleeplessByte:


exercises/linked-list/canonical-data.json
```diff
@@ -0,0 +1,532 @@
...
+            "operation": "push",
+            "value": 7
+          },
+          {
+            "operation": "pop",
+            "expected": 7
+          }
+        ]
+      },
+      "expected": []
```
ErikSchierboom, line 35: We don't yet have a fixed rule for this, but I recall two exercises which canonical-data was also in a form where the input contains expected values and the expected value should be ignored:

- [`react`](https://github.com/exercism/problem-specifications/blob/master/exercises/react/canonical-data.json#L66)
- [`circular-buffer`](https://github.com/exercism/problem-specifications/blob/master/exercises/circular-buffer/canonical-data.json#L40)

Both these canonical data use an empty object for the expected value. Granted, it doesn't really matter much, but it _may_ be useful to have some consistency.

exercises/linked-list/canonical-data.json
```diff
@@ -0,0 +1,532 @@
+{
+  "exercise": "linked-list",
+  "comments": [
+    "In order to keep the interfae for the exercise close or equal to the     ",
```
ErikSchierboom, line 4:
```suggestion
    "In order to keep the interface for the exercise close or equal to the     ",
```

exercises/linked-list/canonical-data.json
```diff
@@ -0,0 +1,532 @@
...
+    "description.md and also satisfying the canonical-schema.json, the only   ",
+    "properties are pop, push, shift, unshift. Some tracks also implement     ",
+    "delete and count, via a Track-Inserted hints.                            ",
```
ErikSchierboom, line 7: Did you mean

```suggestion
    "delete and count, via Track-Inserted hints.                              ",
```

or

```suggestion
    "delete and count, via a Track-Inserted hints.md file.                    ",
```

exercises/linked-list/canonical-data.json
```diff
@@ -0,0 +1,532 @@
...
+    "- If operation is push, unshift, delete, then value is given             ",
+    "- If operation is pop, shift, count, the expected value can be given     ",
+    "                                                                         ",
+    "Encoding tests and operations using the same field is necessary to have  ",
+    "tests that don't just initialize the list, and then have one operation.  "
+  ],
+  "cases": [
+    {
+      "uuid": "7f7e3987-b954-41b8-8084-99beca08752c",
+      "description": "pop gets element from the list",
```
ErikSchierboom, line 21: Maybe 

```suggestion
      "description": "pop gets element from the singleton list",
```

or 

```suggestion
      "description": "pop gets last element from singleton list",
```

My thinking here being that we might want to specify that pop gets the last element.

exercises/linked-list/canonical-data.json
```diff
@@ -0,0 +1,532 @@
...
+            "operation": "pop",
+            "expected": 11
+          }
+        ]
+      },
+      "expected": []
+    },
+    {
+      "uuid": "00ea24ce-4f5c-4432-abb4-cc6e85462657",
+      "description": "shift gets element from the list",
```
ErikSchierboom, line 65:
```suggestion
      "description": "shift gets first element from the list with only one element",
```

exercises/linked-list/canonical-data.json
```diff
@@ -0,0 +1,532 @@
...
+            "operation": "count",
+            "expected": 0
+          }
+        ]
+      },
+      "expected": []
+    },
+    {
+      "uuid": "fd65e422-51f3-45c0-9fd0-c33da638f89b",
+      "description": "deletes the element with the specified value from the list      ",
```
ErikSchierboom, line 367:
```suggestion
      "description": "deletes the element with the specified value from the list",
```

ErikSchierboom (approved):
Excellent! Left some minor nits.

exercises/linked-list/canonical-data.json
SleeplessByte: I'm fine with that!

exercises/linked-list/canonical-data.json
SleeplessByte: No, it's purposefully testing you get "a" element. The next test tests that you get the correct element.

exercises/linked-list/canonical-data.json
SleeplessByte: No, it's purposefully testing you get "a" element. The next test tests that you get the correct element.

exercises/linked-list/canonical-data.json
ErikSchierboom: Okay!

SleeplessByte: I should probably add this as comments?

exercises/linked-list/canonical-data.json
SleeplessByte: I should probably add this as comments?

ErikSchierboom: Might be useful. But perhaps I'm the only one being tripped up by this 🤷

exercises/linked-list/canonical-data.json
```diff
@@ -0,0 +1,543 @@
...
+          },
+          {
+            "operation": "push",
+            "value": 59
+          },
+          {
+            "operation": "pop"
+          },
+          {
+            "operation": "pop"
```
petertseng, line 336: these are `pop`s, but the description says "shifting to empty doesn't break the list" - do I understand correctly to mean that these should be `shift` instead, or have I misunderstood?

exercises/linked-list/canonical-data.json
```diff
@@ -0,0 +1,543 @@
...
+          {
+            "operation": "count",
+            "expected": 2
+          },
+          {
+            "operation": "pop",
+            "expected": 79
+          },
+          {
+            "operation": "shift",
```
petertseng, line 407: although I think having both of these be `pop` or both `shift` would catch more mistakes (forgetting to link the elements on the two sides of the deleted element together), it is only a guess and I have no ground to demand changes based on just a guess.

exercises/linked-list/canonical-data.json
```diff
@@ -0,0 +1,543 @@
...
+            "operation": "push",
+            "value": 73
+          },
+          {
+            "operation": "push",
+            "value": 9
+          },
+          {
+            "operation": "push",
+            "value": 9
```
petertseng, line 512: it strikes me that because these two 9 are next to each other, it's not defined nor tested whether the element closer to the head or to the tail should be deleted. but perhaps that is intentional.

exercises/linked-list/canonical-data.json
SleeplessByte: You have not. I think I broke it when I fixed something else. Let me fix this! Thanks for the fresh 👁️

exercises/linked-list/canonical-data.json
SleeplessByte: That was intentional; it's not in the description and wanted to "leave it in the middle". Do yo think it's better to _enforce_ it? Or maybe I should add a comment that it's left to the implementor?

exercises/linked-list/canonical-data.json
SleeplessByte: I don't know either! I can change it 🤷🏽 😄 .

We can also have _both_ test cases. What do you think?

petertseng: I'm the sort of person who likes to make my tests try to catch as many bugs as possible. So now that you've mentioned having both... I think I like that! Now I'm only guessing at what the possible bugs could be that a student might write, and I wouldn't know for sure since I'm not a mentor. But this is my best guess, and my suggestion based on that guess.

SleeplessByte:
For the person merging: Squash!

exercises/linked-list/canonical-data.json
petertseng: Right now I don't really think it increases educational value to enforce one way or the other. Probably a comment that it's left up to dealer's choice would pre-emptively answer anyone else who asks a similar question as I did.

exercises/linked-list/canonical-data.json
SleeplessByte: I think your guess is good here. I'll add the extra test case.

I did try to add as many tests as possible that catch edge cases I've seen in the wild, and the one you've mentioned is one I didn't see/think of, but it's good to check!

exercises/linked-list/canonical-data.json
SleeplessByte: I'll add a helpful comment :)

exercises/linked-list/canonical-data.json
SleeplessByte: Hmm. It was actually the `pop` on line 347 that was incorrect, but fixed them all!

exercises/linked-list/canonical-data.json
```diff
@@ -0,0 +1,636 @@
...
+            "operation": "push",
+            "value": 47
+          },
+          {
+            "operation": "count",
+            "expected": 1
+          },
+          {
+            "operation": "pop",
+            "value": 47
```
petertseng, line 306:
```suggestion
            "expected": 47
```

exercises/linked-list/canonical-data.json
```diff
@@ -0,0 +1,636 @@
...
+            "operation": "push",
+            "value": 47
+          },
+          {
+            "operation": "count",
+            "expected": 1
+          },
+          {
+            "operation": "shift",
+            "value": 47
```
petertseng, line 348:
```suggestion
            "expected": 47
```

petertseng (commented):
Okay, I finished writing the verifier. It only found two things that don't match my understanding of the specification, and I am ready to approve once either I'm told that my understanding is wrong and it's correct as-is or they're changed

(I don't mind posting the verifier but it's not really the subject of discussion of this PR, so I'll post it and mark it as off-topic)

petertseng:
verifier (to be marked as off-topic)

```ruby
require 'json'

def run_ops(list, ops)
  ops.each_with_index { |op, j|
    expected_keys = case op['operation']
    when 'push'
      list << op['value']
      {value: :required}
    when 'unshift'
      list.unshift(op['value'])
      {value: :required}
    when 'pop'
      got = list.pop
      if (exp = op['expected'])
        raise TestFailure, "pop #{j} expected #{exp} but got #{got}" if got != exp
      end
      {expected: :optional}
    when 'shift'
      got = list.shift
      if (exp = op['expected'])
        raise TestFailure, "shift #{j} expected #{exp} but got #{got}" if got != exp
      end
      {expected: :optional}
    when 'count'
      got = list.size
      exp = op['expected']
      raise TestFailure, "count #{j} expected #{exp} but got #{got}" if got != exp
      {expected: :required}
    when 'delete'
      # Array#delete deletes them all; we only delete one.
      if (idx = list.index(op['value']))
        list.delete_at(idx)
      end
      {value: :required}
    else
      raise "bad operation #{op}"
    end

    expected_keys.merge!(operation: :required)
    raise "bad #{expected_keys}" unless expected_keys.values.all? { |v| %i(required optional).include?(v) }

    unexpected_keys = op.keys - expected_keys.keys.map(&:to_s)
    unless unexpected_keys.empty?
      raise "#{op} #{j} has unexpected keys #{unexpected_keys}"
    end
    missing_keys = expected_keys.filter_map { |k, v| k.to_s if v == :required } - op.keys
    unless missing_keys.empty?
      raise "#{op} #{j} has missing keys #{missing_keys}"
    end
  }
end

cases = JSON.parse(File.read(File.join(__dir__, 'canonical-data.json')))['cases']

# Not the best fit for verify, since we just dispatch on operation and raise, accepting if we get to the end.
# However, this will change with the schema anyway, so we don't care.
verify(cases, accept: ->(_, _) { true }, property: 'list') { |i, _|
  lst = []
  run_ops(lst, i['operations'])
}
```

SleeplessByte:
Sweet! Thank you so much @petertseng . The suggested changes you made are correct, so that verifier was worth it 🥇

petertseng (approved):

ErikSchierboom:
Thanks a lot @petertseng!

SleeplessByte:
@exercism/reviewers anyone wants to ✔️ that box?

neenjaw (approved):

PR #1737, opened 2020-12-16
Add configlet

SleeplessByte:
This adds configlet fetch script to the `bin` directory so that you can use `configlet` to generate uuids for the test cases.

PR #1746, opened 2020-12-18
list-ops: reimplement ambiguous tests

SleeplessByte:
The notation `(x, y) => x + y` for `foldl` and `foldr` is ambigious. Yes, there is a mathy standard for fold, but the fact is that this ordering is not consistent across languages, nor is it well known.

This attempts to make it completely un-ambigious what each parameter means, but since they _were_ ambiguous, it's technically a breaking change.

### DIFF
--- a/exercises/list-ops/canonical-data.json
+++ b/exercises/list-ops/canonical-data.json
@@ -280,18 +280,20 @@
       "description": "folds (reduces) the given list from the left with a function",
       "cases": [
         {
-          "uuid": "613b20b7-1873-4070-a3a6-70ae5f50d7cc",
+          "uuid": "36549237-f765-4a4c-bfd9-5d3a8f7b07d2",
+          "reimplements": "613b20b7-1873-4070-a3a6-70ae5f50d7cc",
           "description": "empty list",
           "property": "foldl",
           "input": {
             "list": [],
             "initial": 2,
-            "function": "(x, y) -> x * y"
+            "function": "(acc, el) -> el * acc"
           },
           "expected": 2
         },
         {
-          "uuid": "e56df3eb-9405-416a-b13a-aabb4c3b5194",
+          "uuid": "7a626a3c-03ec-42bc-9840-53f280e13067",
+          "reimplements": "e56df3eb-9405-416a-b13a-aabb4c3b5194",
           "description": "direction independent function applied to non-empty list",
           "property": "foldl",
           "input": {
@@ -302,23 +304,26 @@
               4
             ],
             "initial": 5,
-            "function": "(x, y) -> x + y"
+            "function": "(acc, el) -> el + acc"
           },
           "expected": 15
         },
         {
-          "uuid": "d2cf5644-aee1-4dfc-9b88-06896676fe27",
+          "uuid": "d7fcad99-e88e-40e1-a539-4c519681f390",
+          "reimplements": "d2cf5644-aee1-4dfc-9b88-06896676fe27",
           "description": "direction dependent function applied to non-empty list",
           "property": "foldl",
           "input": {
             "list": [
+              1,
               2,
-              5
+              3,
+              4
             ],
-            "initial": 5,
-            "function": "(x, y) -> x / y"
+            "initial": 24,
+            "function": "(acc, el) -> el / acc"
           },
-          "expected": 0
+          "expected": 64
         }
       ]
     },
@@ -326,18 +331,20 @@
       "description": "folds (reduces) the given list from the right with a function",
       "cases": [
         {
-          "uuid": "aeb576b9-118e-4a57-a451-db49fac20fdc",
+          "uuid": "17214edb-20ba-42fc-bda8-000a5ab525b0",
+          "reimplements": "aeb576b9-118e-4a57-a451-db49fac20fdc",
           "description": "empty list",
           "property": "foldr",
           "input": {
             "list": [],
             "initial": 2,
-            "function": "(x, y) -> x * y"
+            "function": "(acc, el) -> el * acc"
           },
           "expected": 2
         },
         {
-          "uuid": "c4b64e58-313e-4c47-9c68-7764964efb8e",
+          "uuid": "e1c64db7-9253-4a3d-a7c4-5273b9e2a1bd",
+          "reimplements": "c4b64e58-313e-4c47-9c68-7764964efb8e",
           "description": "direction independent function applied to non-empty list",
           "property": "foldr",
           "input": {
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):
Great!

SleeplessByte (commented):
DON'T MERGE THIS YET. Need to delete `bin/configlet.exe`  which snuck in because of #1737 not yet being merged.

valentin-p (approved):

exercises/list-ops/canonical-data.json
```diff
@@ -186,16 +314,80 @@
...
+            "list": [
+              1,
+              2,
+              3,
+              4
+            ],
+            "initial": 24,
+            "function": "(acc, el) -> el / acc"
+          },
+          "expected": 64
```
petertseng, line 382: I see that this is correct, though I'd like to note for my records that it needs the `/` to preserve fractions. Integer division will not work here, since it would compute `1 / 24 = 0`. I will take it into account when using this case.

exercises/list-ops/canonical-data.json
```diff
@@ -224,11 +421,72 @@
...
```
petertseng, line 489: By my understanding of foldr, this should be 9 instead. Can you confirm?

* acc = 24, el = 4. new acc = 1/6
* acc = 1/6, el = 3. new acc = 18
* acc = 18, el = 2. new acc = 1/9
* acc = 1/9, el = 1. new acc = 9

exercises/list-ops/canonical-data.json
SleeplessByte: Yes! I should add that as a comment.

I think it was the same with the previous case?

exercises/list-ops/canonical-data.json
SleeplessByte: It's 9 in my test suite 🙈 . You're correct.

SleeplessByte:

exercises/list-ops/canonical-data.json
petertseng: I think the previous case asked for integer division, since its expected value is zero and the only way to get a zero out of division is if you started with a zero. Since there is no zero present in the list nor in the initial value, I reason that the zero must have been introduced via integer division.

SleeplessByte: Right. I couldn't get the test case to work in some languages I work in _because_ they do fractional division. I think commenting here is the right approach, and something we should have had on the previous case. I _will_ add it there too!

exercises/list-ops/canonical-data.json
```diff
@@ -222,13 +282,67 @@
...
           "uuid": "be396a53-c074-4db3-8dd6-f7ed003cce7c",
           "description": "direction dependent function applied to non-empty list",
+          "comments": [
+            "Expects integer division (expects / to discard fractions).       "
+          ],
           "property": "foldr",
           "input": {
             "list": [2, 5],
             "initial": 5,
             "function": "(x, y) -> x / y"
```
petertseng, line 292: note for posterity: The possible ambiguous interpretation for this function were:

* acc, el -> acc / el
    * acc: 5, el: 5. new acc 1
    * acc: 1, el: 2. new acc 0
* acc, el -> el / acc
    * acc: 5, el: 5. new acc 1
    * acc: 1, el: 2. new acc 2
 
So clearly the **latter** was intended. Note that this is different from the other case, where the **former** was intended. The fact that these two same strings require two different interpretations is what's confusing and is the motivation for this PR.

exercises/list-ops/canonical-data.json
```diff
@@ -184,18 +187,75 @@
...
           "uuid": "d2cf5644-aee1-4dfc-9b88-06896676fe27",
           "description": "direction dependent function applied to non-empty list",
+          "comments": [
+            "Expects integer division (expects / to discard fractions).       "
+          ],
           "property": "foldl",
           "input": {
             "list": [2, 5],
             "initial": 5,
             "function": "(x, y) -> x / y"
```
petertseng, line 197: note for posterity: The possible ambiguous interpretation for this function were:

* acc, el -> acc / el
    * acc: 5, el: 2. new acc 0
    * acc: 0, el: 5. new acc 0
* acc, el -> el / acc
    * acc: 5, el: 2. new acc 0
    * acc: 0, el: 5. division by zero.

So clearly the **former** was intended. Note that this is different from the other case, where the **latter** was intended. The fact that these two same strings require two different interpretations is what's confusing and is the motivation for this PR.

exercises/list-ops/canonical-data.json
```diff
@@ -222,13 +282,67 @@
...
+            "will not work here, since it would compute 4 / 24 = 1 / 6. Use   ",
+            "the original test values (be396a53-c074-4db3-8dd6-f7ed003cce7c)  ",
+            "if integer division is expected / required."
+          ],
+          "description": "direction dependent function applied to non-empty list",
+          "property": "foldr",
+          "input": {
+            "list": [1, 2, 3, 4],
+            "initial": 24,
+            "function": "(acc, el) -> el / acc"
```
petertseng, line 343: The follow comment concerns **notation** only. It does not change the expected values of any test, or change the semantics of how `foldr` should operate.

I understand that some track maintainers may disagree about whether the function argument list should be **represented as a string** as `(acc, el)` or `(el, acc)`, but I would like to say to those track maintainers that I think we are just going to have to pick one and run with it. Again, I emphasise that either way there would not be a change to the operation of `foldr`: The implementing track should be sure to pass the element as `el` and the accumulator as `acc`, no matter the string representation here.

I advise implementing tracks to be aware of the conventions common to their language do what makes sense for their language. Why do I bring this up? Observe:

* Ramda (a functional library for Javascript): [foldr](https://ramdajs.com/0.9/docs/#foldr) is `(a, b -> a) -> a -> [b] -> a`. The passed function takes its accumulator first and the list element second, and `foldr` takes initial accumulator first and the list second.
* OCaml [List.fold_right](https://caml.inria.fr/pub/docs/manual-ocaml/libref/List.html#VALfold_right) is `('a -> 'b -> 'b) -> 'a list -> 'b -> 'b`. The passed function takes its list element first and the accumulator second, and `fold_right` takes the list first and accumulator second.
* Haskell [Prelude foldr](https://hackage.haskell.org/package/base-4.14.1.0/docs/Prelude.html#v:foldr) is `class Foldable t where foldr :: (a -> b -> b) -> b -> t a -> b`. The passed function takes its list element first and accumulator second, and `foldr` takes the initial accumulator first and the list second.

As you can see, I've listed three different languages doing things in three different ways, so I advise track maintainers looking to squabble over this that there is no universal convention, so we should just pick one for problem-specifications and run with it.

petertseng (approved):
The new cases add are all correct. I have added some notes for posterity.

petertseng:
EBWODP:

> * acc, el -> acc / el
>    * acc: 5, el: 2. new acc 0

ah, my apologies. I meant to say new acc 2.

petertseng:
By the way, I think the recommendation should change to be that these changes **should** be squashed. It will mean there will be no executable that got added and a subsequent commit removing it. And there are no formatting changes that need a separate commit, etc.

exercises/list-ops/canonical-data.json
SleeplessByte: This is great advice and I agree 100% with it. I'll link this comment in the comments in the canonical data.

petertseng (approved):
just approving to let you know I saw the new advice added

PR #1747, opened 2020-12-18
Update README.md

ErikSchierboom:


SleeplessByte (approved):

README.md
```diff
@@ -39,7 +39,8 @@ Test data can be incorporated into a track's test suites manually or extracted b
 - Exercises _must_ contain tests that cover the public interface of the exercise (also thought of as "application tests").
 - Exercises _may_ contain tests that cover the private or lower-level interface of the exercise (sometimes refered to as "library tests").
 
-- Test cases are immutable, which means that once a test case has been added, it never changes. There are two exceptions:
+- Test cases are immutable, which means that once a test case has been added, it never changes. There are a couple of exceptions:
+  - The `description` field _can_ be mutated and thus does not require adding a new test case when changing its value.
```
iHiD, line 43:
```suggestion
  - The `description` field _can_ be mutated to fix clarify the test or fix typos, in which case it does not require adding a new test case when changing its value. If the material meaning of the description changes, a new test is required.
```

SaschaMann (approved):

README.md
ErikSchierboom: I'd rather not add this, as I feel it is vague enough to potentially lead to interpretation issues. There is also not much risk involved in renaming a description for the following reasons:

1. Each PR requires three people to approve before it can be merged. This significantly decreases the chance of weird things being merged.
2. Tracks can now use the `uuid` property to identify a test case. In fact, allowing the description to be changed might discourage tracks from doing so :)

So I'd like to merge this _as is_. If it turns out this leads to any problematic situations, we can always make the rule more strict.

SleeplessByte: I think #1746 vs #1744 are good examples and would be caught by at least one person, I think.

angelikatyborska:
I think this change was already merged to `main` so this PR could be closed:
https://github.com/exercism/problem-specifications/blob/main/README.md#L42-L45

However, the merged change does not fix the part that says "There are two exceptions".

ErikSchierboom:
This can be closed

Issue #1749, opened 2020-12-18
Handling optionality of test cases that require additions to the description.md


SaschaMann:
## Problem

In #1674 the notion that every test case is optional was introduced. However, some optional test cases may require additions to the `description.md` file. There may also be situations where one will choose between two contradicting test cases of which both require different instructions.

Tracks could add these descriptions in the track-specific README inserts but these explanations are likely not track-specific and would benefit from being synced across repos.

### Example

\#1736 introduces a new test case for the `rational-numbers` exercise that introduces tests for the ordering of rational numbers. The ordering relation is not trivial and needs to be explained to students who are not familiar with rational numbers:

```diff
diff --git a/exercises/rational-numbers/description.md b/exercises/rational-numbers/description.md
index 55ebaeb4b..39f9ed475 100644
--- a/exercises/rational-numbers/description.md
+++ b/exercises/rational-numbers/description.md
@@ -18,9 +18,12 @@ Exponentiation of a rational number `r = a/b` to a real (floating-point) number
 
 Exponentiation of a real number `x` to a rational number `r = a/b` is `x^(a/b) = root(x^a, b)`, where `root(p, q)` is the `q`th root of `p`.
 
+Ordering of rational numbers: `a/b < c/d` if and only if `a * sign(b) * |d| < |b| * c * sign(d)`.
+
 Implement the following operations:
  - addition, subtraction, multiplication and division of two rational numbers,
  - absolute value, exponentiation of a given rational number to an integer power, exponentiation of a given rational number to a real (floating-point) power, exponentiation of a real number to a rational number.
+ - the ordering of rational numbers.
 
 Your implementation of rational numbers should always be reduced to lowest terms. For example, `4/4` should reduce to `1/1`, `30/60` should reduce to `1/2`, `12/8` should reduce to `3/2`, etc. To reduce a rational number `r = a/b`, divide `a` and `b` by the greatest common divisor (gcd) of `a` and `b`. So, for example, `gcd(12, 8) = 4`, so `r = 12/8` can be reduced to `(12/4)/(8/4) = 3/2`.
```

Tracks that choose to not implement this test would be left behind with an irrelevant stub in the description.

SaschaMann:
## Potential solution

This is the first potential solution that came to my mind. I don’t like it too much.

Introduce a new optional key to `canonical-data.json` that contains `description.md` inserts:

```json
{
    "uuid": "bd1eab16-7603-4a77-ae4e-23c930fe39cd",
[...]

PR #1753, opened 2020-12-19
Bob: Add test cases

SaschaMann:
should be rebased, not squashed

### DIFF
--- a/exercises/bob/canonical-data.json
+++ b/exercises/bob/canonical-data.json
@@ -123,6 +123,15 @@
       "expected": "Calm down, I know what I'm doing!"
     },
     {
+      "uuid": "e90bb162-bbfd-438c-820f-a6b09f5cdf99",
+      "description": "Bob in Vienna",
+      "property": "response",
+      "input": {
+        "heyBob": "OIDA?"
+      },
+      "expected": "Calm down, I know what I'm doing!"
+    },
+    {
       "uuid": "a20e0c54-2224-4dde-8b10-bd2cdd4f61bc",
       "description": "shouting numbers",
       "property": "response",
@@ -164,6 +173,15 @@
       "property": "response",
       "input": {
         "heyBob": "I HATE THE DENTIST"
+      },
+      "expected": "Whoa, chill out!"
+    },
+    {
+      "uuid": "2a6431bd-43e3-495d-9fff-8834a387329a",
+      "description": "Bob in Vienna",
+      "property": "response",
+      "input": {
+        "heyBob": "OIDA!"
       },
       "expected": "Whoa, chill out!"
     },
@@ -249,6 +267,15 @@
       "expected": "Sure."
     },
     {
+      "uuid": "452ac3a1-daf2-473f-9ace-3de3a41dbdf8",
+      "description": "Bob in Vienna",
+      "property": "response",
+      "input": {
+        "heyBob": "Oida?"
+      },
+      "expected": "Sure."
+    },
+    {
       "uuid": "72bd5ad3-9b2f-4931-a988-dce1f5771de2",
       "description": "other whitespace",
       "property": "response",
@@ -263,6 +290,15 @@
       "property": "response",
       "input": {
         "heyBob": "This is a statement ending with whitespace      "
+      },
+      "expected": "Whatever."
+    },
+    {
+      "uuid": "16f43a00-acbe-4e36-a323-d7f48d9f29c0",
+      "description": "Bob in Vienna",
+      "property": "response",
+      "input": {
+        "heyBob": "Oida."
       },
       "expected": "Whatever."
     }
--- a/exercises/bob/canonical-data.json
+++ b/exercises/bob/canonical-data.json
@@ -29,11 +29,31 @@
       "expected": "Whoa, chill out!"
     },
     {
+      "uuid": "c10699d2-7a89-40c8-9cd3-ed277fe1c14a",
+      "description": "shouting multilingual gibberish",
+      "scenarios": ["unicode"],
+      "property": "response",
+      "input": {
+        "heyBob": "FCÄEÜCÖDFCẞAB"
+      },
+      "expected": "Whoa, chill out!"
+    },
+    {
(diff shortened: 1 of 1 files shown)


exercises/bob/canonical-data.json
```diff
@@ -28,6 +28,16 @@
       },
       "expected": "Whoa, chill out!"
     },
+    {
+      "uuid": "c10699d2-7a89-40c8-9cd3-ed277fe1c14a",
+      "description": "shouting multilingual gibberish",
+      "scenarios": ["unicode"],
+      "property": "response",
+      "input": {
+        "heyBob": "FCÄEÜCÖDFCẞAB"
```
SaschaMann, line 37: It's not clear in GitHub's fonts, but this is actually an uppercase ẞ, not a lowercase ß.

exercises/bob/canonical-data.json
```diff
@@ -127,6 +176,15 @@
       },
       "expected": "Whoa, chill out!"
     },
+    {
+      "uuid": "2a6431bd-43e3-495d-9fff-8834a387329a",
+      "description": "Bob in Vienna",
+      "property": "response",
+      "input": {
+        "heyBob": "OIDA!"
```
iHiD, line 184: Let's stick to English for non-unicode examples.

exercises/bob/canonical-data.json
```diff
@@ -127,6 +176,15 @@
       },
       "expected": "Whoa, chill out!"
     },
+    {
+      "uuid": "2a6431bd-43e3-495d-9fff-8834a387329a",
+      "description": "Bob in Vienna",
```
iHiD, line 181: This doesn't tell me what is actually being tested in this test.

iHiD:
Presuming the purpose of these tests is to check for unicode, I don't think we need to be as exhaustive. I think one or two would surfice. For the others, the purpose of them isn't clear to me. Could you explain?

SaschaMann:
> Presuming the purpose of these tests is to check for unicode, I don't think we need to be as exhaustive. I think one or two would surfice. 

There's one case per response. Many solutions define one method per response type, so it makes sense to check every response.

> For the others, the purpose of them isn't clear to me. Could you explain?

More variety, and also they show off very concisely what the rules are since only case and punctuation change.

exercises/bob/canonical-data.json
SaschaMann: Why? Can you tell me an equivalent phrase/word in English then?

iHiD:
> There are no existing cases that manage to explain the rules of bob this concisely.

I think my concern here is that these tests aren't adding any new test-behaviour, at least that I can see. They seem to duplicate the existing tests. Is there something that would pass the existing tests, but fail in these extra ones?

exercises/bob/canonical-data.json
iHiD: > Why?

Because Exercism's exercises are written in English. Once we cross the mutlilingual bridge, we might consider how this changes, but for now, they're in English.

---

> Can you tell me an equivalent phrase/word in English then?

No, because I don't know what the word means, but I can refer you to the [first google result](https://www.urbandictionary.com/define.php?term=Oida):

Screenshot 2020-12-21 at 16 30 45

SaschaMann:
> I think my concern here is that these tests aren't adding any new test-behaviour, at least that I can see. They seem to duplicate the existing tests. Is there something that would pass the existing tests, but fail in these extra ones?

I was under the impression that #1674 meant we could use this repo as a place to share test cases used by the tracks again without having to justify each individual addition. Apparently I was wrong. I don't feel like spending my time on discussions like this.

iHiD:
I think if you're basically forking an exercise in a different direction, then it's best to fork it locally. If we end up with lots of variations of the same exercise in one repo, it's going to make it very hard for the tracks to know what set of tests sit together, and very hard for tracks to know whether they should be implementing the new tests etc without reading the whole body of PRs. I think if you're basically wanting to rewrite the tests, there's a fair argument for making a whole new exercise and putting it in problem specs, or just changing a language's specific one to do that. We really want exercises here to be the shared set of exercises that tracks can pick up and quickly implement. 

For the unicode tests in this PR though, I'm a :+1: on adding at least a couple of those, so if you want to reopen this for those, that would be welcomed as it's a simple toggle on/off for other tracks (supporting the unicode scenario).

>  a place to share test cases used by the tracks again without having to justify each individual addition.

The definition of the purpose of this repo in that issue is: "Problem Specifications exists to cover the "happy path". That means it should work for most tracks most of the time. In situations that aren't in the "happy path", it is down to individual tracks to solve those issues locally.". For an exercise like Bob, which is very old in Exercism-terms, there's a clear happy path defined. Adding the unicode tests adds extra value to the exercise, so would be welcome. But having alternative sets of tests proposed by each maintainer will lead to everything being very messy/complex, for no real value (as tracks can specify those alternative/additional tests using the mechanism we created for that purpose).

> I don't feel like spending my time on discussions like this.

Entirely reasonable 🙂

iHiD:
Having spoken to @SaschaMann more about this, I can see the argument for the "vienna" tests if we could put them into more of some sort of "Bonus" tests (as per #1749). I also am (and was) happy for the unicode tests to be added, although I think we probably don't need all of them (specifically the gibberish one breaks the "someone speaking" metaphor a bit for me - but I don't have a hugely strong opinion either way.

Sascha isn't interested in discussing this anymore, but is happy for us to reopen and merge if we want to.

SaschaMann:
> specifically the gibberish one breaks the "someone speaking" metaphor a bit for me - but I don't have a hugely strong opinion either way

The spec already has `fffbbcbeab`, `4?", ":) ?`, `Okay if like my  spacebar  quite a bit?   `, `FCECDFCAAB`, `ZOMG THE %^*@#\$(*^ ZOMBIES ARE COMING!!11!!1!`, `          `, `\t\t\t\t\t\t\t\t\t\t`, `\n\r \t`, `\nDoes this cryogenic chamber make me look fat?\nno`, `         hmmmmmmm...`, and `This is a statement ending with whitespace      `. I don't know how to pronounce `\n` or `\t` personally.

(for some reason GitHub removes the whitespace in those...)

PR #1757, opened 2020-12-24
queen attack: remove ambiguity and support non-zero indexed tracks

SleeplessByte:
This PR attempts three things:

- Remove ambiguity about row and column indices. Because it doesn't matter for _queens_ if the representation is top-to-bottom or bottom-to-top (or ltr/rtl for that matter), it doesn't matter _until_ a track wants to implement non-standard tests. Clarifying what is meant removes the ambiguity.
- Supports non-zero indexed tracks by using zero-indexed in the description. Before, the instructions made little sense.
- Provide standard notation example and board "layout"/direction (white at the bottom is standard)


ErikSchierboom (approved):
This really helps! Thanks

petertseng:
what could have happened such that that `q` file got added in the last commit?

SleeplessByte:
> what could have happened such that that `q` file got added in the last commit?

Me doing this stuff at 0500 in the morning.

Likely laptop lagged behind, and when I press q to quit git log (vim mode), it wrote to q instead. I'll fix!

SleeplessByte:
@exercism/reviewers pretty please

kotp (approved):

exercises/queen-attack/description.md
```diff
@@ -1,27 +1,24 @@
...
+3 _ _ _ _ _ _ _ _ 3
+2 _ _ _ _ _ B _ _ 2
+1 _ _ _ _ _ _ _ _ 1
+  a b c d e f g h
 ```
 
-You'd also be able to answer whether the queens can attack each other.
-In this case, that answer would be yes, they can, because both pieces
-share a diagonal.
+You'd also be able to answer whether the queens can attack each other. In this case, that answer would be yes, they can, because both pieces share a diagonal.
```
wolf99, line 24:
```suggestion
You'd also be able to answer whether the queens can attack each other. 
In this case, that answer would be yes, they can, because both pieces share a diagonal.
```

kotp: You would or you could... no contraction here, as it becomes unclear, potentially.

SleeplessByte: Actually think this makes more sense.

exercises/queen-attack/description.md
```diff
@@ -1,27 +1,25 @@
...
 
-In the game of chess, a queen can attack pieces which are on the same
-row, column, or diagonal.
+In the game of chess, a queen can attack pieces which are on the same row, column, or diagonal.
 
 A chessboard can be represented by an 8 by 8 array.
 
-So if you're told the white queen is at (2, 3) and the black queen at
-(5, 6), then you'd know you've got a set-up like so:
+So if you're told the white queen is at `c5` (zero-indexed at column 2, row 3) and the black queen at `f2` (zero-indexed at column 5, row 6), then you'd know you've got a set-up like so:
```
SleeplessByte, line 9:
```suggestion
So if you are told the white queen is at `c5` (zero-indexed at column 2, row 3) and the black queen at `f2` (zero-indexed at column 5, row 6), then you know that the set-up is like so:
```

wolf99 (approved):

PR #1760, opened 2020-12-27
grade-school: Change a UUID to be version 4

ee7:
Commit 03529d08e743 (#1735) added a new test case for the `grade-school`
exercise, but the UUID added was version 1 rather than version 4. This
was not caught by CI because the regex in the schema validator was
too permissive (fixed by #1759).

Some reasons to prefer version 4 UUIDs:
- Version 4 is the right version for a fully random and unique
  identifer; we don't want to indicate anything about the machine that
  generated the UUID, and version 1 UUIDs have a timestamp and MAC
  address component.
- All of the other UUIDs currently in `problem-specifications` are valid
  version 4 UUIDs.
- `configlet uuid` generates a version 4 UUID.
- A reader familiar with the UUID specification (RFC 4122) might see a
  version 1 UUID and infer that the timestamp or MAC address component
  is useful, causing them to wonder why the others are version 4.
- A user who generates a version 1 UUID might unintentionally leak their
  MAC address.

See also:
- https://tools.ietf.org/html/rfc4122.html

Note: the CI check is expected to fail.
```
The following tests have been deleted illegally:
 - dece43c8-3ba5-11eb-8fdf-7f8daeaeb5f2
 ```

Ping @SleeplessByte.

### DIFF
--- a/exercises/grade-school/canonical-data.json
+++ b/exercises/grade-school/canonical-data.json
@@ -15,7 +15,7 @@
             "expected": ["Aimee"]
         },
         {
-          "uuid": "dece43c8-3ba5-11eb-8fdf-7f8daeaeb5f2",
+          "uuid": "c125dab7-2a53-492f-a99a-56ad511940d8",
           "description": "A student can't be in two different grades",
           "property": "roster",
           "input": {


SleeplessByte (approved):

ErikSchierboom (approved):

ErikSchierboom:
Perhaps someone from @exercism/reviewers could take a look at this PR?

NobbZ:
Wouldn't this change also mean, that we have adjust all tracks configurations that implement `grade-school`?

Or the exercises toml at least?

Even though I second that UUIDv4 is to be prefered over v1, a given UUID should never change for an exercise/test.

Instead we should create CI that makes it impossible to add *new* entries with non desired UUID versions, while whitelisting the "bad" ones already slipped in.

(And for the paranoid, whitelist them per file to avoid the once in a millenium danger of a clash)

ErikSchierboom:
> Wouldn't this change also mean, that we have adjust all tracks configurations that implement grade-school?
> 
> Or the exercises toml at least?

All true, but @ee7 checked if any tracks did already use this UUID and they didn't. So I'm hoping this is an exception.

ee7:
> Wouldn't this change also mean, that we have adjust all tracks configurations that implement `grade-school`?

Yes, we have to update all tracks that added this new test case. However, only the `javascript` track has done that so far. See [this search query](https://github.com/search?q=org%3Aexercism+dece43c8-3ba5-11eb-8fdf-7f8daeaeb5f2&type=code) - the UUID is only in the `problem-specifications` repo right now.

If we merge this PR then we can also merge https://github.com/exercism/problem-specifications/pull/1759, which fixes the UUID checking in CI (without us having to add an allowlist of "bad" UUIDs).

UUIDs are supposed to be immutable, but I think it's fine to make an exception here and fix it while no other track has implemented it.

NobbZ:
If it is indeed only used within one track and therefore the amount of resulting changes is low, I think it is a "better now than later" change.

NobbZ (approved):

ErikSchierboom:
Thanks for the review @NobbZ!

PR #1762, opened 2020-12-30
Hamming: remove tests that don't make sense per 1761

### LINKED ISSUE #1761: Hamming tests for empty input don't make sense
In the `canonical-data.json` for the hamming exercise there are three tests that appear to be in conflict with each other (or testing one things when intending to test separate things).

This question has impeded our update alignment of the C track's implementation of this exercise to match the current version of this problem specification (https://github.com/exercism/c/pull/556) .

```json
//...
    {
      "uuid": "919f8ef0-b767-4d1b-8516-6379d07fcb28",
      "description": "disallow first strand longer",
      "property": "distance",
      "input": {
        "strand1": "AATG",
        "strand2": "AAA"
      },
      "expected": {"error": "left and right strands must be of equal length"}
    },
    {
      "uuid": "8a2d4ed0-ead5-4fdd-924d-27c4cf56e60e",
      "description": "disallow second strand longer",
      "property": "distance",
      "input": {
        "strand1": "ATA",
        "strand2": "AGTG"
      },
      "expected": {"error": "left and right strands must be of equal length"}
    },
    {
      "uuid": "5dce058b-28d4-4ca7-aa64-adfe4e17784c",
      "description": "disallow left empty strand",
      "property": "distance",
      "input": {
        "strand1": "",
        "strand2": "G"
      },
      "expected": {"error": "left strand must not be empty"}
    },
    {
      "uuid": "38826d4b-16fb-4639-ac3e-ba027dec8b5f",
      "description": "disallow right empty strand",
      "property": "distance",
      "input": {
        "strand1": "G",
        "strand2": ""
      },
      "expected": {"error": "right strand must not be empty"}
    }
```

I read "empty strand" to mean a string of zero length, rather than a `NULL` or equivalent absence of any string at all (see #905 ).
However, given that input of both strands empty are allowed, this means that the latter two tests will fail due to unmatched string length even if empty string checking is not implemented. Thus I think that the latter tests will not test what is intended for them to test.

Can someone confirm my logic here?

I will suggest simply removing the latter two tests as a resolution, would this be acceptable?

PS: the following seems also related #1666

wolf99:
fixes #1761 

I note that removing test cases does not show up as a change for the canonical data syncer. Is this an issue?

### DIFF
--- a/exercises/hamming/canonical-data.json
+++ b/exercises/hamming/canonical-data.json
@@ -79,26 +79,6 @@
         "strand2": "AGTG"
       },
       "expected": {"error": "left and right strands must be of equal length"}
-    },
-    {
-      "uuid": "5dce058b-28d4-4ca7-aa64-adfe4e17784c",
-      "description": "disallow left empty strand",
-      "property": "distance",
-      "input": {
-        "strand1": "",
-        "strand2": "G"
-      },
-      "expected": {"error": "left strand must not be empty"}
-    },
-    {
-      "uuid": "38826d4b-16fb-4639-ac3e-ba027dec8b5f",
-      "description": "disallow right empty strand",
-      "property": "distance",
-      "input": {
-        "strand1": "G",
-        "strand2": ""
-      },
-      "expected": {"error": "right strand must not be empty"}
     }
   ]
 }
--- a/exercises/hamming/canonical-data.json
+++ b/exercises/hamming/canonical-data.json
@@ -79,6 +79,50 @@
         "strand2": "AGTG"
       },
       "expected": {"error": "left and right strands must be of equal length"}
+    },
+    {
+      "uuid": "5dce058b-28d4-4ca7-aa64-adfe4e17784c",
+      "description": "disallow left empty strand",
+      "property": "distance",
+      "input": {
+        "strand1": "",
+        "strand2": "G"
+      },
+      "expected": {"error": "left strand must not be empty"}
+    },
+    {
+      "uuid": "db92e77e-7c72-499d-8fe6-9354d2bfd504",
+      "description": "disallow left empty strand",
+      "comments": ["Normalises error messages"],
+      "reimplements": "5dce058b-28d4-4ca7-aa64-adfe4e17784c",
+      "property": "distance",
+      "input": {
+        "strand1": "",
+        "strand2": "G"
+      },
+      "expected": {"error": "left and right strands must be of equal length"}
+    },
+    {
+      "uuid": "38826d4b-16fb-4639-ac3e-ba027dec8b5f",
+      "description": "disallow right empty strand",
+      "property": "distance",
+      "input": {
+        "strand1": "G",
+        "strand2": ""
+      },
+      "expected": {"error": "right strand must not be empty"}
+    },
+    {
+      "uuid": "920cd6e3-18f4-4143-b6b8-74270bb8f8a3",
+      "description": "disallow right empty strand",
+      "comments": ["Normalises error messages"],
+      "reimplements": "38826d4b-16fb-4639-ac3e-ba027dec8b5f",
+      "property": "distance",
+      "input": {
+        "strand1": "G",
+        "strand2": ""
+      },
+      "expected": {"error": "left and right strands must be of equal length"}
     }
   ]
 }


wolf99:
Updated PR, per the discussion on #1761, so that instead of removing tests, the errors are now normalised so that it is easier to see the desired result is the same across the 4 test cases.

petertseng (approved):

SleeplessByte (approved):

exercises/hamming/canonical-data.json
```diff
@@ -79,6 +79,50 @@
...
+    },
+    {
+      "uuid": "5dce058b-28d4-4ca7-aa64-adfe4e17784c",
+      "description": "disallow left empty strand",
+      "property": "distance",
+      "input": {
+        "strand1": "",
+        "strand2": "G"
+      },
+      "expected": {"error": "left strand must not be empty"}
```
kotp, line 91: Why can't it be empty?  As long as the right strand is also empty, representing that it is not there, the distance could still be said to be 0.  And if there is a length of not 0 for the right hand strand, the distance would equal the length of the second strand?  (I will have to go look at the discussions to find out if this is coming from there, though)

exercises/hamming/canonical-data.json
```diff
@@ -79,6 +79,50 @@
...
+      "uuid": "db92e77e-7c72-499d-8fe6-9354d2bfd504",
+      "description": "disallow left empty strand",
+      "comments": ["Normalises error messages"],
+      "reimplements": "5dce058b-28d4-4ca7-aa64-adfe4e17784c",
+      "property": "distance",
+      "input": {
+        "strand1": "",
+        "strand2": "G"
+      },
+      "expected": {"error": "left and right strands must be of equal length"}
```
kotp, line 103: This appears to be a test that is identical to the test above, just different words for the same situation.  And the reliance that a strand can not be 0 length.

exercises/hamming/canonical-data.json
```diff
@@ -79,6 +79,50 @@
...
+      "uuid": "920cd6e3-18f4-4143-b6b8-74270bb8f8a3",
+      "description": "disallow right empty strand",
+      "comments": ["Normalises error messages"],
+      "reimplements": "38826d4b-16fb-4639-ac3e-ba027dec8b5f",
+      "property": "distance",
+      "input": {
+        "strand1": "G",
+        "strand2": ""
+      },
+      "expected": {"error": "left and right strands must be of equal length"}
```
kotp, line 125: See prior comments, it is just mirrored here.

exercises/hamming/canonical-data.json
wolf99: The case for both strands being empty is allowed as a distance of 0, as you mention, per the case specified on ln 14, but the ambiguity that prompts your question is partly why this change is proposed 😃

exercises/hamming/canonical-data.json
wolf99: Yes, this is a reimplementation (see ln 97 `reimplement` property) of the test on ln 84 that normalises the error message to demonstrate that the logical result is actually the same as non-empty strands of unequal length (such as on ln 81).

kotp: I see in #97 but it doesn't seem to be related.

kotp: I see #1762 though.  I think this is what you meant be "In 97"?

kotp: Nope, that doesn't make sense.  Sorry very confused at the moment.

kotp: In ln  Sorry, apparently the font on my browser doesn't allow the visual distinction between what I thought was a capital "I" (eye) and a lower case "l" (ell) and though it was a typographical error, wrong case for "in".

kotp: We really should use "line" when we mean to say "line" to avoid confusion.

wolf99: Sorry, my bad! Will use "line" going forward

kotp (approved):

SleeplessByte:
Thank you @wolf99 . Looking great.

PR #1774, opened 2021-01-29
Use main branch for configlet-ci

ErikSchierboom:
See https://github.com/exercism/github-actions/issues/14

### DIFF
--- a/workflows/configlet.yml
+++ b/workflows/configlet.yml
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Fetch configlet
-        uses: exercism/github-actions/configlet-ci@master
+        uses: exercism/github-actions/configlet-ci@main
 
       - name: Configlet Linter
-        run: configlet lint .
+        run: configlet lint
--- a/workflows/README.md
+++ b/workflows/README.md
@@ -125,7 +125,7 @@ node_js:
   - lts/*
 script:
   - bin/fetch-configlet
-  - bin/configlet lint .
+  - bin/configlet lint
   - bin/build.sh
 ```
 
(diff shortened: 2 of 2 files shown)


ee7 (approved):

ErikSchierboom:
Pinging @exercism/reviewers for a review

workflows/configlet.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Fetch configlet
-        uses: exercism/github-actions/configlet-ci@master
+        uses: exercism/github-actions/configlet-ci@main
 
       - name: Configlet Linter
         run: configlet lint .
```
ee7, line 18:
```suggestion
        run: configlet lint
```

ee7 (commented):
Just spotted a leftover `configlet lint .` here. For others: please use just `configlet lint` without the `.` from now on, or
```
configlet lint --track-dir=foo
configlet lint -t=foo
```
(We can also use a colon or a space instead of `=` as the delimiter, and write the option before `lint` if preferred).

Could you also change it in the `README.md` too?
https://github.com/exercism/problem-specifications/blob/main/workflows/README.md

I think we can it this in this PR - just change the commit message to
```
Update configlet instructions
```
or similar.

This also clarifies that this PR isn't changing the CI in this repo. It changes a file in `workflows`, not `.github/workflows`.

But now that we've made the PR that ensures every track has a `configlet` workflow, should we just delete this configlet template? It might become outdated. We could just point users towards the configlet workflow on their track.

ErikSchierboom:
> But now that we've made the PR that ensures every track has a configlet workflow, should we just delete this configlet template? It might become outdated. We could just point users towards the configlet workflow on their track.

We can't, because these instructions are intended for new tracks, not existing ones.

ee7:
> > But now that we've made the PR that ensures every track has a configlet workflow, should we just delete this configlet template? It might become outdated. We could just point users towards the configlet workflow on their track.
> 
> We can't, because these instructions are intended for new tracks, not existing ones.

Wouldn't a new track be initialized with this?
https://github.com/exercism/request-new-language-track/blob/main/.github/workflows/configlet.yml

ErikSchierboom:
I'm not sure. I'll check.

kotp (approved):

iHiD (approved):

PR #1777, opened 2021-01-31
Improve descriptions to make sense per #1769

### LINKED ISSUE #1769: roman-numerals: [bug] canonical-data descriptions do not match data
The `canoncial-data.json` for the `roman-numerals` exercise is all kinds of wrong
For a lot of the latter tests, the data does not match the description

```json
    {
      "uuid": "46b46e5b-24da-4180-bfe2-2ef30b39d0d0",
      "description": "90, being 100 - 10, is XC",
      "property": "roman",
      "input": {
        "number": 93
      },
      "expected": "XCIII"
    },
```

For almost all of the tests, the descriptions are too wordy to be able to make "nice" matching test-case function names,

```json
    {
      "uuid": "b374a79c-3bea-43e6-8db8-1286f79c7106",
      "description": "3 is three I's",             // A more simple "3 is III" ?
      "property": "roman",
      "input": {
        "number": 3
      },
      "expected": "III"
    },
```

For some of the latter tests, the descriptions include maths and explanations rather than just what the test is.

```json
    {
      "uuid": "607ead62-23d6-4c11-a396-ef821e2e5f75",
      "description": "49 is not 40 + 5 + 4 but rather 50 - 10 + 10 - 1",     // Better to be "49 is XLIX" ? Describe test rather than explain how maths works
      "property": "roman",
      "input": {
        "number": 49
      },
      "expected": "XLIX"
    },
```

Once I get done with my current work on the C track I may make a PR for this myself, but if someone wants to take it up in the mean time please be my guest 😄

wolf99:
Fixes #1769 

There were cases where the descriptions where the data did not numerically match the descriptions, such as

```json
      "description": "1000 is a single M",
      "input": {
        "number": 1024
      },
      "expected": "MXXIV"
```

1. Leave as-is
2. Change the description to match the data
2. Change the data to match the description

There is a slight semantic difference between 2 and 3 if you consider that what this case is testing is not just explicitly that 1024 is MXXIV, but that absent any other test to the effect it also implicitly tests that 1000 is M as the original description states.

I chose 2 as it accurately describes what the test is doing, but 3 is probably as valid a change

### DIFF
--- a/exercises/roman-numerals/canonical-data.json
+++ b/exercises/roman-numerals/canonical-data.json
@@ -3,18 +3,7 @@
   "cases": [
     {
       "uuid": "19828a3a-fbf7-4661-8ddd-cbaeee0e2178",
-      "description": "1 is a single I",
-      "property": "roman",
-      "input": {
-        "number": 1
-      },
-      "expected": "I"
-    },
-    {
-      "uuid": "61797ec8-10dd-48fd-b55f-c106e21d1457",
-      "reimplements": "19828a3a-fbf7-4661-8ddd-cbaeee0e2178",
       "description": "1 is I",
-      "comments": ["Improve description"],
       "property": "roman",
       "input": {
         "number": 1
@@ -23,18 +12,7 @@
     },
     {
       "uuid": "f088f064-2d35-4476-9a41-f576da3f7b03",
-      "description": "2 is two I's",
-      "property": "roman",
-      "input": {
-        "number": 2
-      },
-      "expected": "II"
-    },
-    {
-      "uuid": "c1894cb6-b251-420a-b9b8-151f8e0e24fa",
-      "reimplements": "f088f064-2d35-4476-9a41-f576da3f7b03",
       "description": "2 is II",
-      "comments": ["Improve description"],
       "property": "roman",
       "input": {
         "number": 2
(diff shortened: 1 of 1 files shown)


ErikSchierboom (commented):
@wolf99 I completely support these changes. They nicely follow our documentation that still says that the description is immutable. We've had some discussion on that in another issue (do you remember which issue @SleeplessByte?) and the conclusion there is that changing the description is fine too. Could you update the PR accordingly? I'll submit a different PR to update the documentation.

wolf99:
>  We've had some discussion on that in another issue

Is in the review comments on #1747 I believe.

I will update my PR per your suggestion this evening then.

exercises/roman-numerals/canonical-data.json
```diff
@@ -93,7 +98,7 @@
     },
     {
       "uuid": "d5b283d4-455d-4e68-aacf-add6c4b51915",
-      "description": "50 is a single L",
+      "description": "59 is a single LIX",
```
ErikSchierboom, line 101: Do we need the `is a single` bit in the description? It looks like it was removed from other test cases.

wolf99: Whoops, I missed this in the update.
Well caught Erik

wolf99, line 101:
```suggestion
      "description": "59 is LIX",
```

ErikSchierboom (approved):
LGTM

valentin-p (approved):
LGTM

iHiD (approved):

Issue #1770, opened 2021-01-26
anagram: Replace test case that contains "master"


ee7:
Exercism renamed the main branch of every repo from `master` to `main`.

Given that we did that, we should probably also replace the below test case: 

https://github.com/exercism/problem-specifications/blob/c76d358b75c32468b8a6541484c50f8144e91beb/exercises/anagram/canonical-data.json#L14-L23

which we should do by adding a new test case that `reimplements` it.

One candidate:

```diff
+    {
+      "uuid": "803a4de9-9136-46d2-8f1d-a603f95409f0",
+      "reimplements": "b3cca662-f50a-489e-ae10-ab8290a09bdc",
+      "description": "detects two anagrams",
+      "comments": "Reimplemented to replace master with main :)",  
+      "property": "findAnagrams",
+      "input": {
+        "subject": "main",
+        "candidates": ["amin", "pigeon", "mina"]
+      },
+      "expected": ["amin", "mina"]
+    },
```

But even if that wasn't too silly, "amin" and "mina" are probably too uncommon.

PR #1780, opened 2021-02-06
anagram: reimplement test case containing 'master'

ee7:
Exercism has now renamed the main branch of every repo from `master` to
`main`, which makes it logically consistent to also remove references to
`master` in exercise tests. The only one was in the `anagram` exercise:

https://github.com/exercism/problem-specifications/blob/ef9e1e17c84721ee6e9d5a65c8dd3ba2122eac91/exercises/anagram/canonical-data.json#L14-L23 

This commit adds a test case that tries to stay close to the one that it
reimplements, maintaining these properties:
- the `subject` is a six-letter word
- the `candidates` are three six-letter words
- the first and last `candidates` as anagrams
- the non-anagram has one letter in common with the `subject`

Minor hypothetical benefit: if all tracks magically changed to use this test
case it would make it easier to search for 'master' Exercism-wide and find
outdated links (which do redirect, however).

Closes: #1770

### DIFF
--- a/exercises/anagram/canonical-data.json
+++ b/exercises/anagram/canonical-data.json
@@ -22,6 +22,20 @@
       "expected": ["stream", "maters"]
     },
     {
+      "uuid": "03eb9bbe-8906-4ea0-84fa-ffe711b52c8b",
+      "reimplements": "b3cca662-f50a-489e-ae10-ab8290a09bdc",
+      "comments": [
+        "Reimplemented to be consistent about removing references to 'master'"
+      ],
+      "description": "detects two anagrams",
+      "property": "findAnagrams",
+      "input": {
+        "subject": "solemn",
+        "candidates": ["lemons", "cherry", "melons"]
+      },
+      "expected": ["lemons", "melons"]
+    },
+    {
       "uuid": "a27558ee-9ba0-4552-96b1-ecf665b06556",
       "description": "does not detect anagram subsets",
       "property": "findAnagrams",


exercises/anagram/canonical-data.json
```diff
@@ -21,6 +21,20 @@
...
+      "uuid": "03eb9bbe-8906-4ea0-84fa-ffe711b52c8b",
+      "reimplements": "b3cca662-f50a-489e-ae10-ab8290a09bdc",
+      "comments": [
+        "Reimplemented to be consistent about removing references to 'master'"
+      ],
+      "description": "detects two anagrams",
+      "property": "findAnagrams",
+      "input": {
+        "subject": "solemn",
+        "candidates": ["lemons", "pigeon", "melons"]
```
valentin-p, line 34: Other ideas for non-anagrams: model, novel, pleon

valentin-p (approved):

exercises/anagram/canonical-data.json
ee7: "pigeon" is there because I was trying to keep it as close to the test case that it `reimplements` as possible. That is: three six-letter candidates, where the first and last are anagrams. I should've written that, sorry - I've edited the top post to show that existing test case.

But now that I'm looking at "pigeon" again: previously, it had only one letter in common with the `subject`, but now it has three. So, depending on how strictly we interpret `reimplements`, I think we should change "pigeon" to another six-letter word that only has one letter in common with "solemn". Perhaps "parrot" or "puffin", to stay in the same genre.

This test case reimplemented by this PR is the second one, so I think it's supposed to avoid testing edge cases. This exercise should definitely test the kind of near-anagrams that you suggest, but from a quick look at the `canonical-data.json` I think we already do that in these two cases:
https://github.com/exercism/problem-specifications/blob/ef9e1e17c84721ee6e9d5a65c8dd3ba2122eac91/exercises/anagram/canonical-data.json#L24-L43

And in this test case:
https://github.com/exercism/problem-specifications/blob/ef9e1e17c84721ee6e9d5a65c8dd3ba2122eac91/exercises/anagram/canonical-data.json#L121-L130

Any thoughts?

valentin-p: I forgot the S for the sixth letter, they are all plurals. Parrot is fine.
Another: Timber

wolf99 (approved):

exercises/anagram/canonical-data.json
ee7: Maybe `cherry`?

Extreme bikeshedding, but I didn't love that:
- "parrot" is also a verb, so "lemons parrot melons" can be read strangely.
- With the input `["lemons", "parrot", "melons"]` (or with `"timber"`) producing `["lemons", "melons"]`, the function could be a "fruit detector function" rather than an anagram detector. Although with  `["lemons", "cherry", "melons"]` it could be a plural detector...

valentin-p: Then, we could pick a plural fruit: olives, grapes, apples

SleeplessByte (approved):

SleeplessByte:
Thanks @ee7

PR #1787, opened 2021-02-26
forth: added test to check redefining negative numbers

tgujar:
The original version of the test only checks for positive numbers.

```if(/^\d*$/.test(functionName)) throw new Error("Invalid definition");```

An implementation like above will work for positive numbers, but not for negative numbers.


wolf99 (commented):
Should line number 431 also be changed to read "cannot redefine positive numbers", to explicitly describe the difference for that test case?

tgujar:
> Should line number 431 also be changed to read "cannot redefine positive numbers", to explicitly describe the difference for that test case?

Yup I think so, but maybe it should be "cannot redefine non-negative numbers"

SleeplessByte:
cc @exercism/reviewers please review; I have a PR dependent on this waiting to be 💥 merged 🗡️

SleeplessByte (approved):

ErikSchierboom (approved):

ericbalawejder (approved):

ErikSchierboom:
Thanks @tgujar!

PR #1799, opened 2021-05-04
Update sync documentation

ErikSchierboom:
The current documentation for `tests.toml` files was still referring to the obsolete canonical data syncer. This PR updates the documentation to use configlet's `sync` command instead. It also updates the contents of the example `tests.toml` file to the latest spec.

Closes https://github.com/exercism/configlet/issues/124

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,7 +225,7 @@ or the short form `configlet sync -e foo -u -mi`.
 The `sync` command operates on Practice Exercises that are in the track-level `config.json` file.
 If you are adding an exercise that has canonical data to a track, first add that exercise to the track-level `config.json`, and then run a `configlet sync` command to create the corresponding `tests.toml` file.
 
-To download the latest version of the `configlet` tool, please run the [`fetch-configlet`](https://github.com/exercism/configlet/blob/main/scripts/fetch-configlet) script or the [`fetch-configlet.ps1`](https://github.com/exercism/configlet/blob/main/scripts/fetch-configlet.ps1) script. At least one of these scripts should already exist in every track repo's `bin` directory - the script will also download `configlet` to this location. You can then sync the tests by running `./bin/configlet sync` (on Linux or macOS) or `.\bin\configlet.exe sync` (on Windows).
+To download the latest version of the `configlet` tool, please run the [`fetch-configlet`](https://github.com/exercism/configlet/blob/main/scripts/fetch-configlet) bash script or the [`fetch-configlet.ps1`](https://github.com/exercism/configlet/blob/main/scripts/fetch-configlet.ps1) PowerShell script (Windows only). At least one of these scripts should already exist in every track repo's `bin` directory - the script will also download `configlet` to this location. You can then sync the tests by running `./bin/configlet sync` (on macOS/Linux/similar) or `.\bin\configlet.exe sync` (on Windows).
 
 ## Conventions
 
--- a/README.md
+++ b/README.md
@@ -213,6 +213,17 @@ A plain `configlet sync` performs no changes, and just compares the tests specif
 To interactively update the `tests.toml` files, use `configlet sync --update`.
 For each missing test, this prompts the user to choose whether to include/exclude/skip it, and updates the corresponding `tests.toml` file accordingly.
 
+To non-interactively include every missing test for an exercise `foo`, use:
+
+```
+configlet sync --exercise foo --update --mode include
+```
+
+or the short form `configlet sync -e foo -u -mi`.
+
+The `sync` command operates on Practice Exercises that are in the track-level `config.json` file.
+If you are adding an exercise that has canonical data to a track, first add that exercise to the track-level `config.json`, and then run a `configlet sync` command to create the corresponding `tests.toml` file.
+
 To download the latest version of the `configlet` tool, please run the [`fetch-configlet`](https://github.com/exercism/configlet/blob/main/scripts/fetch-configlet) script or the [`fetch-configlet.ps1`](https://github.com/exercism/configlet/blob/main/scripts/fetch-configlet.ps1) script. At least one of these scripts should already exist in every track repo's `bin` directory - the script will also download `configlet` to this location. You can then sync the tests by running `./bin/configlet sync` (on Linux or macOS) or `.\bin\configlet.exe sync` (on Windows).
 
 ## Conventions
--- a/README.md
+++ b/README.md
@@ -190,27 +190,26 @@ If a track implements an exercise for which test data exists, the exercise _must
 A `tests.toml` file for a track's `two-fer` exercise looks like this:
 
 ```toml
-[canonical-tests]
+[19709124-b82e-4e86-a722-9e5c5ebf3952]
+description = "no name given"
 
-# no name given
-"19709124-b82e-4e86-a722-9e5c5ebf3952" = true
+[3451eebd-123f-4256-b667-7b109affce32]
+description = "a name given"
+include = false
 
-# a name given
-"3451eebd-123f-4256-b667-7b109affce32" = true
-
-# another name given
-"653611c6-be9f-4935-ab42-978e25fe9a10" = false
+[653611c6-be9f-4935-ab42-978e25fe9a10]
+description = "another name given"
 ```
 
-In this case, the track has chosen to implement two of the three available tests.
+In this case, the track has chosen to implement two of the three available tests (`include = true` is the default and is omitted).
 
 If a track uses a _test generator_ to generate an exercise's test suite, it _must_ use the contents of the `tests.toml` file to determine which tests to include in the generated test suite.
 
 ### Track Test Data Tooling
 
-To make it easy to keep the `tests.toml` up to date, tracks can use the [`canonical_data_syncer` application](https://github.com/exercism/canonical-data-syncer). This application is a small, standalone binary that will compare the tests specified in the `tests.toml` files against the tests that are defined in the exercise's canonical data. It then interactively gives the maintainer the option to include or exclude test cases that are currently missing, updating the `tests.toml` file accordingly.
+To make it easy to keep the `tests.toml` up to date, tracks can use the [`configlet` application](https://github.com/exercism/configlet)'s `sync` command. This command will compare the tests specified in the `tests.toml` files against the tests that are defined in the exercise's canonical data. It then interactively gives the maintainer the option to include or exclude test cases that are currently missing, updating the `tests.toml` file accordingly.
 
-To use the canonical data syncer tool, tracks should copying the [`fetch-canonical_data_syncer`](https://github.com/exercism/canonical-data-syncer/blob/main/scripts/fetch-canonical_data_syncer) and/or [`fetch-canonical_data_syncer.ps1`](https://github.com/exercism/canonical-data-syncer/blob/main/scripts/fetch-canonical_data_syncer.ps1) scripts into their repository. Then, running either of these scripts will download the latest version of the tool to the track's `bin` directory. The tool can be run using `./bin/canonical_data_syncer` or `.\bin\canonical_data_syncer.exe`, depending on your operating system.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -190,27 +190,26 @@ If a track implements an exercise for which test data exists, the exercise _must
...
 
-To use the canonical data syncer tool, tracks should copying the [`fetch-canonical_data_syncer`](https://github.com/exercism/canonical-data-syncer/blob/main/scripts/fetch-canonical_data_syncer) and/or [`fetch-canonical_data_syncer.ps1`](https://github.com/exercism/canonical-data-syncer/blob/main/scripts/fetch-canonical_data_syncer.ps1) scripts into their repository. Then, running either of these scripts will download the latest version of the tool to the track's `bin` directory. The tool can be run using `./bin/canonical_data_syncer` or `.\bin\canonical_data_syncer.exe`, depending on your operating system.
+To use the configlet tool, tracks should copy the [`fetch-configlet`](https://github.com/exercism/configlet/blob/main/scripts/fetch-configlet) and/or [`fetch-configlet.ps1`](https://github.com/exercism/configlet/blob/main/scripts/fetch-configlet.ps1) scripts into their repository's `bin` directory. Then, running either of these scripts will download the latest version of the tool to the same `bin` directory. The tool can then be run using `./bin/configlet sync` or `.\bin\configlet.exe sync`, depending on your operating system.
```
ee7, line 212:
```suggestion
To download the latest version of the `configlet` tool, please run the [`fetch-configlet`](https://github.com/exercism/configlet/blob/main/scripts/fetch-configlet) script or the [`fetch-configlet.ps1`](https://github.com/exercism/configlet/blob/main/scripts/fetch-configlet.ps1) script. At least one of these scripts should already exist in every track repo's `bin` directory - the script will also download `configlet` to this location. You can then sync the tests by running `./bin/configlet sync` (on Linux or macOS) or `.\bin\configlet.exe sync` (on Windows).
```

Non-blocking suggestion.

This suggested wording could be improved, but the idea is that it's better to tell the user to run the one that's already in the track repo. If we tell them to download the fetch script themselves, the user still has to save it to the `bin` directory - most of the time this just means they'll be prompted about overwriting the existing one.

Referring to `configlet sync` as "the tool" also felt a bit awkward - I'd say that "the tool" is `configlet` itself.

README.md
```diff
@@ -190,27 +190,26 @@ If a track implements an exercise for which test data exists, the exercise _must
...
 
 If a track uses a _test generator_ to generate an exercise's test suite, it _must_ use the contents of the `tests.toml` file to determine which tests to include in the generated test suite.
 
 ### Track Test Data Tooling
 
-To make it easy to keep the `tests.toml` up to date, tracks can use the [`canonical_data_syncer` application](https://github.com/exercism/canonical-data-syncer). This application is a small, standalone binary that will compare the tests specified in the `tests.toml` files against the tests that are defined in the exercise's canonical data. It then interactively gives the maintainer the option to include or exclude test cases that are currently missing, updating the `tests.toml` file accordingly.
+To make it easy to keep the `tests.toml` up to date, tracks can use the [`configlet` application](https://github.com/exercism/configlet)'s `sync` command. This command will compare the tests specified in the `tests.toml` files against the tests that are defined in the exercise's canonical data. It then interactively gives the maintainer the option to include or exclude test cases that are currently missing, updating the `tests.toml` file accordingly.
```
ee7, line 210: Non-blocking comment: the default `sync` behavior will stop being interactive in the future. We should remember to update this later.

ee7 (approved):
LGTM.

I did some nitpicking, partly unrelated to the `canonical_data_syncer` to `configlet sync` change - but is still in the scope of "update sync documentation"

And I've checked that, with this PR, running
```
$ git grep 'syncer'
```
in this repo produces no output.

README.md
ErikSchierboom: We should

wolf99 (approved):

ee7:
Pinging @exercism/reviewers for a review.

This is an important docs update - the current wording has probably already caused a lot of confusion.

ErikSchierboom:
@SleeplessByte Care to review?

ee7:
This PR should no longer be merged as-is - I'll update it to reflect [the latest changes](https://github.com/exercism/configlet/releases/tag/4.0.0-alpha.21) to `configlet sync`. A plain `configlet sync` is no longer interactive.

ee7:
I've also added some extra docs.

In the future, it would be nice if we didn't have to maintain 3 sets of `configlet` docs (in `exercism/configlet`, `exercism/docs`, and `exercism/problem-specifications`).

ErikSchierboom:
Thanks! Ping @exercism/reviewers

hiljusti (changes requested):

hiljusti:
Minor suggestions aside... lgtm

hiljusti (approved):

ee7:
@hiljusti: Thanks for the review. You raise some good points about the wording, but it's a little tricky to find a great one.

Some things to consider:
1. A user can run a bash script on Windows, and when running on Windows, the `fetch-configlet` script correctly downloads the Windows release of `configlet`.
1. We might prefer to point Windows users towards the bash script because:
   1. The bash script is in every track repo, but the PowerShell script is only in 11 of 79 track repos. (The repos are: `cpp`, `csharp`, `dart`, `fsharp`, `javascript`, `red`, `solidity`, `typescript`, `wren`, `z3`, `zig`).
   1. Many Windows users are more familiar with bash than PowerShell.
   1. The bash script has been reviewed much more thoroughly than the PowerShell script.
1. A user can theoretically run a PowerShell script on a non-Windows OS. However, the `fetch-configlet.ps1` always downloads the Windows release of `configlet`, so we should indeed communicate "only run the `.ps1` script on Windows".
1. We currently only build `configlet` releases for Linux, macOS, and Windows (all `x86_64` only).
1. The `fetch-configlet` script does run without error on the BSDs (as long as bash is installed). It downloads the Linux binary.
1. The Linux binary works fully on FreeBSD and NetBSD if Linux binary compatibility is enabled, as described e.g. [here](https://docs.freebsd.org/en/books/handbook/linuxemu/). However, I have not documented this fact, and we don't test it during CI. But I **am** willing to document it - I think it will stay working.
1. The Linux binary is statically linked, which makes it easier to run on the BSDs that provide Linux binary compatibility.
1. On BSDs that don't provide Linux binary compatibility (e.g. OpenBSD) a user who wants to use `configlet` must compile it themselves, or run a release binary in a VM.
1. I have tested that `configlet` compiles and runs properly on at least FreeBSD, OpenBSD, and NetBSD.

@ErikSchierboom any thoughts on the wording here?

Also: in the future, can we have a single source of truth for `configlet` installation instructions? If so, should that be in `exercism/docs`, or `exercism/configlet`? Then the instructions in `exercism/problem-specifications` can be minimal, and point elsewhere for more details.

hiljusti:
All sounds reasonable to me

ErikSchierboom:
> Many Windows users are more familiar with bash than PowerShell.

In my experience, I've not found this to be true. Bash isn't even installed by default on Windows, so I don't think we can assume Windows users will just be able to run the Bash command.

> Also: in the future, can we have a single source of truth for configlet installation instructions? If so, should that be in exercism/docs, or exercism/configlet? Then the instructions in exercism/problem-specifications can be minimal, and point elsewhere for more details.

I think having the documentation be in exercism/configlet and referring to that from the docs and prob-specs repos makes the most sense.

README.md
```diff
@@ -187,30 +187,44 @@ This is an example of what a re-implementation looks like:
...
+description = "a name given"
+include = false
 
-# a name given
-"3451eebd-123f-4256-b667-7b109affce32" = true
-
-# another name given
-"653611c6-be9f-4935-ab42-978e25fe9a10" = false
+[653611c6-be9f-4935-ab42-978e25fe9a10]
+description = "another name given"
```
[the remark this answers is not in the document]
ErikSchierboom: I like the suggestion. @ee7?

hiljusti:
I'm fine with not further specifying OSs. It's kinda minor; I think Windows developers will recognize a `.ps1` as a powershell script, and non-windows developers will ignore it because it's foreign. BSD developers are used to software (outside of ports) not working out-of-the-box

Rambly thoughts follow:

Since this targets track maintainers, I think it's less important what the OS image comes with and more important what developers will have installed. I'm not a Windows expert, though I do have one machine that runs it, and I even ran a Windows machine for a few weeks recently for work... though I just used it as a shell to reach other remote hosts I could use for development

Anyway - I know Bash won't be installed on Windows by default, but what is the developer setup these days for a Windows dev?  Not all Linux distros will have `git` installed by default, but we expect people to have at least gotten that far if they're maintaining tracks. I thought the Cygwin/MinGW/GitBash days were mostly over since Windows Subsystem Linux 2.0 can be run as a (non-virtualized) peer to windows now. Not sure if that's what people are using, though. I've heard of some difficulties when you want to do deeper Linux networking things, since Windows can gate the Linux environment.

ErikSchierboom:
> Since this targets track maintainers, I think it's less important what the OS image comes with and more important what developers will have installed. 

Correct, until we centralize the documentation in one place, in which case it will become aimed at students too.

> Anyway - I know Bash won't be installed on Windows by default, but what is the developer setup these days for a Windows dev?

Unless you're running things exclusively in WSL, Windows users will _probably_ have installed Git locally too. That means they'll have Git bash, but I never used in my 10+ years of running Git on Windows.

SleeplessByte:
I use git bash all the time, but it's important to note that those bash scripts _only_ work in a git bash environment, which you're not always in.

I would prefer keeping both the "cross-env-if-you're-bash-like" and "always-works-on-windows" scripts.

README.md
```diff
@@ -187,30 +187,44 @@ This is an example of what a re-implementation looks like:
...
 
-In this case, the track has chosen to implement two of the three available tests.
+In this case, the track has chosen to implement two of the three available tests (`include = true` is the default and is omitted).
 
 If a track uses a _test generator_ to generate an exercise's test suite, it _must_ use the contents of the `tests.toml` file to determine which tests to include in the generated test suite.
 
 ### Track Test Data Tooling
 
-To make it easy to keep the `tests.toml` up to date, tracks can use the [`canonical_data_syncer` application](https://github.com/exercism/canonical-data-syncer). This application is a small, standalone binary that will compare the tests specified in the `tests.toml` files against the tests that are defined in the exercise's canonical data. It then interactively gives the maintainer the option to include or exclude test cases that are currently missing, updating the `tests.toml` file accordingly.
+To make it easy to keep the `tests.toml` up to date, tracks can use the [`configlet` application](https://github.com/exercism/configlet)'s `sync` command.
```
kotp, line 210:
```suggestion
To make it easy to keep the `tests.toml` up to date, tracks should use the [`configlet` application](https://github.com/exercism/configlet)'s `sync` command.
```

Should we encourage this behavior?  So "should" instead of "can"?

ErikSchierboom: I think we should (pun intended)! @ee7?

ee7 (commented):
I mostly liked @hiljusti's suggestion, I just wanted to:
- avoid the possible reading of "the bash script cannot be run on Windows"
- clarify that the PowerShell script is Windows-only

As a final thing before we merge this - how about this?

README.md
```diff
@@ -187,30 +187,44 @@ This is an example of what a re-implementation looks like:
...
 
-To use the canonical data syncer tool, tracks should copying the [`fetch-canonical_data_syncer`](https://github.com/exercism/canonical-data-syncer/blob/main/scripts/fetch-canonical_data_syncer) and/or [`fetch-canonical_data_syncer.ps1`](https://github.com/exercism/canonical-data-syncer/blob/main/scripts/fetch-canonical_data_syncer.ps1) scripts into their repository. Then, running either of these scripts will download the latest version of the tool to the track's `bin` directory. The tool can be run using `./bin/canonical_data_syncer` or `.\bin\canonical_data_syncer.exe`, depending on your operating system.
+To download the latest version of the `configlet` tool, please run the [`fetch-configlet`](https://github.com/exercism/configlet/blob/main/scripts/fetch-configlet) script or the [`fetch-configlet.ps1`](https://github.com/exercism/configlet/blob/main/scripts/fetch-configlet.ps1) script. At least one of these scripts should already exist in every track repo's `bin` directory - the script will also download `configlet` to this location. You can then sync the tests by running `./bin/configlet sync` (on Linux or macOS) or `.\bin\configlet.exe sync` (on Windows).
```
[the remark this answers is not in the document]
ErikSchierboom: I like it! In the future, we could consider modifying the PowerShell script to also download the binaries for non-Windows OS'es (PowerShell is cross-platform nowadays).

ErikSchierboom:
Thanks everyone!

PR #545, opened 2017-02-11
difference-of-squares: tidy up description

stkent:
Updating this for legibility based on watching a new programmer solve the exercise.

PR #1808, opened 2021-06-14
pythagorean-triplets: Fix maths notation in instructions

SaschaMann:
`**` is not commonly used for exponentiation. I went with `²` but `^` would also be a more common choice than `**`. In v3, this should probably be updated to use Mathjax/KaTeX notation instead. (see https://github.com/exercism/website/issues/476)

### DIFF
--- a/exercises/pythagorean-triplet/description.md
+++ b/exercises/pythagorean-triplet/description.md
@@ -4,7 +4,7 @@ A Pythagorean triplet is a set of three natural numbers, {a, b, c}, for
 which,
 
 ```text
-a**2 + b**2 = c**2
+a² + b² = c²
 ```
 
 and such that,
@@ -16,7 +16,7 @@ a < b < c
 For example,
 
 ```text
-3**2 + 4**2 = 9 + 16 = 25 = 5**2.
+3² + 4² = 9 + 16 = 25 = 5².
 ```
 
 Given an input integer N, find all Pythagorean triplets for which `a + b + c = N`.


ee7 (commented):
I don't know if I'd prefer superscripts or `^`. Perhaps a screenreader does better with superscripts?

But maybe the screenreader does best with something like Mathjax. Ideally, a user would hear e.g.:
> a squared plus b squared equals c squared

So I'll just suggest:
1. If we change from `**` then we should probably do the same in `diffie-hellman` (which has the only other occurrences):

https://github.com/exercism/problem-specifications/blob/df6420b279d15629f2459b4b6c3d709d3016407b/exercises/diffie-hellman/description.md#L22-L35

2. If we prefer superscripts to `^` then we should consider changing the other exercises that use `^`. This could be done in a separate PR, if desired. `**` is indeed rarer than `^` currently:

```
affine-cipher/description.md
22:  `D(y) = a^-1(y - b) mod m`
24:  -  it is important to note that `a^-1` is the modular multiplicative inverse

all-your-base/description.md
20:(4 * 10^1) + (2 * 10^0)
24:(1 * 2^5) + (0 * 2^4) + (1 * 2^3) + (0 * 2^2) + (1 * 2^1) + (0 * 2^0)
28:(1 * 3^3) + (1 * 3^2) + (2 * 3^1) + (0 * 3^0)

armstrong-numbers/description.md
7:- 9 is an Armstrong number, because `9 = 9^1 = 9`
8:- 10 is *not* an Armstrong number, because `10 != 1^2 + 0^2 = 1`
9:- 153 is an Armstrong number, because: `153 = 1^3 + 5^3 + 3^3 = 1 + 125 + 27 = 153`
10:- 154 is *not* an Armstrong number, because: `154 != 1^3 + 5^3 + 4^3 = 1 + 125 + 64 = 190`

binary/description.md
21:- The rightmost digit gets multiplied by 10^0 = 1
22:- The next number gets multiplied by 10^1 = 10
24:- The *n*th number gets multiplied by 10^*(n-1)*.
27:So: `23 => 2*10^1 + 3*10^0 => 2*10 + 3*1 = 23 base 10`
31:So: `101 => 1*2^2 + 0*2^1 + 1*2^0 => 1*4 + 0*2 + 1*1 => 4 + 1 => 5 base 10`.

complex-numbers/description.md
3:A complex number is a number in the form `a + b * i` where `a` and `b` are real and `i` satisfies `i^2 = -1`.
7:The absolute value of a complex number `z = a + b * i` is a real number `|z| = sqrt(a^2 + b^2)`. The square of the absolute value `|z|^2` is the result of multiplication of `z` by its complex conjugate.
17:`1 / (a + i * b) = a/(a^2 + b^2) - b/(a^2 + b^2) * i`.
20:`(a + i * b) / (c + i * d) = (a * c + b * d)/(c^2 + d^2) + (b * c - a * d)/(c^2 + d^2) * i`.
22:Raising e to a complex exponent can be expressed as `e^(a + i * b) = e^a * e^(i * b)`, the last term of which is given by Euler's formula `e^(i * b) = cos(b) + i * sin(b)`.

gigasecond/description.md
6:A gigasecond is 10^9 (1,000,000,000) seconds.

octal/description.md
23:- The rightmost digit gets multiplied by 10^0 = 1
24:- The next number gets multiplied by 10^1 = 10
26:- The *n*th number gets multiplied by 10^*(n-1)*.
33: = 2*10^2 + 3*10^1 + 3*10^0
43: = 2*8^2 + 3*8^1 + 3*8^0

rational-numbers/description.md
15:Exponentiation of a rational number `r = a/b` to a non-negative integer power `n` is `r^n = (a^n)/(b^n)`.
17:Exponentiation of a rational number `r = a/b` to a negative integer power `n` is `r^n = (b^m)/(a^m)`, where `m = |n|`.
19:Exponentiation of a rational number `r = a/b` to a real (floating-point) number `x` is the quotient `(a^x)/(b^x)`, which is a real number.
21:Exponentiation of a real number `x` to a rational number `r = a/b` is `x^(a/b) = root(x^a, b)`, where `root(p, q)` is the `q`th root of `p`.

trinary/description.md
17:1*3^5 + 0*3^4 + 2*3^3 + 0*3^2 + 1*3^1 + 2*3^0    # the value
```

petertseng (approved):
As I was in https://github.com/exercism/problem-specifications/pull/545, I am in favour.

SaschaMann:
>     If we prefer superscripts to ^ then we should consider changing the other exercises that use ^. This could be done in a separate PR, if desired. ** is indeed rarer than ^ currently:

In #1655 people brought up that superscripts that aren't a number, e.g. <sup>n</sup> aren't as easy to read (the LaTeX-based rendering would likely address this), so we'd likely still have a mix of `^`  and superscripts for now.

angelikatyborska (approved):
Out of the three options (`**`, `^`, and superscript), `**` is the one that I am least likely to associate with exponentiation, so I'm in favor.

## Side rant, feel free to ignore

The first time I saw the documentation of affine-cipher, I couldn't figure out the order of operations for:
```
D(y) = a^-1(y - b) mod m
```

This needs either superscripts or some more parentheses, or both. As the equations were totally new to me and their syntax is not formalized anywhere (right?), I didn't know which one of the interpretations is correct:

1)
	  ```
	  D(y) = ((a^-1) * (y - b)) mod m
	  ```

2)
	  ```
	  D(y) = (a^(-1 * (y - b))) mod m
	  ```

3)
	  ```
	  D(y) = a^((-1 * (y - b)) mod m)
	  ```

Or maybe there's a 4th option 🙃

SaschaMann:
We should probably do a pass over all instructions once/if LaTeX support of some sort is added. That should make it clearer.

ErikSchierboom (approved):

PR #1782, opened 2021-02-09
Add h1 heading to exercise description

ErikSchierboom:
This is a follow-up of https://github.com/exercism/problem-specifications/pull/1779

PR #1811, opened 2021-06-22
yacht: Make top-level header consistent

ee7:
This was the only `description.md` file that did not begin with
`# Description`.

---

@ErikSchierboom do we want this change? 

PR #1782 added `# Description` to the `description.md` of every exercise in `problem-specifications` apart from `yacht`, which already had a top-level header.

On `main`:
```
$ cd problem-specifications/exercises
$ find -name 'description.md' | wc -l
135
$ rg --glob 'description.md' '# Description' --files-with-matches | wc -l
134
$ rg --glob 'description.md' '# Description' --files-without-match
yacht/description.md
```

### DIFF
--- a/exercises/yacht/description.md
+++ b/exercises/yacht/description.md
@@ -1,4 +1,4 @@
-# Score a single throw of dice in *Yacht*
+# Description
 
 The dice game [Yacht](https://en.wikipedia.org/wiki/Yacht_(dice_game)) is from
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):
I like things being consistent, so I'm in favor. It's also worth noting that we don't show the heading so it doesn't _really_ matter, but I'm still in favor.

angelikatyborska (approved):

junedev (approved):

PR #1819, opened 2021-07-09
grade-school: fix student cannot be in multiple grades logic

### LINKED ISSUE #1763: grade-school test for student in multiple grades does not make sense
https://github.com/exercism/problem-specifications/blob/eb6c32b421ad3e4da9623fa504543e69d9323f57/exercises/grade-school/canonical-data.json#L19

Firstly this test has `desiredGrade` as an input, but is a test of the `roster` property, this input is otherwise only used with the `grade` property. This does not make sense in terms of the functionality required for the properties

```c
students_t roster(input); // this case, and only this case, implies adding desired_grade argument here also, which does not make sense
students_t grade(input, desired_grade);
```

This seems like possibly just a mistake of copy-pasta `roster` when `grade` was intended. Thus the solution would be simply  to change the property under test from `roster` to `grade`. I can add a PR to resolve this if I have understood correctly?

A second issue is that the `students` input is intended, per the `README.md`, one by one. This implies some order of addition. The order used does not matter, just that there must be one.
However, the `desiredGrade` is then queried later.
So a function used to add students (e.g. `add_student(student_t student)` ) could check inputs for duplicate names, but it can only prevent second and third additions. It cannot know in advance of adding the first name that a second name will be added that will conflict with it.

Thus the input of the following cannot make sense given that the `desiredGrade`, that should return an empty roster, is the _first_ student to be added:
```json
            "students": [
              ["Aimee", 2],
              ["Aimee", 1]
            ],
            "desiredGrade": 2
```

If instead, we say that addition of students should not be filtered, but instead the `grade()` function should filter what gets output from the data that has been added; then yes, this can work, but it is ugly and requires poor implementation in just about any language I know (true, I don't know a lot of languages 😄 ).

A solution to this problem can be that the `README.md` is added to remove suggestions about what implementation should be - just describe the problem and let languages implement a solution in an idiomatic manner. Is this correct understanding, maybe I have missed something?

wolf99:
Fixes #1763, see the discussion there for explanation

### DIFF
--- a/exercises/grade-school/canonical-data.json
+++ b/exercises/grade-school/canonical-data.json
@@ -33,7 +33,7 @@
           "comments": [
             "Reimplemented to be logically consistent"
           ],
-          "description": "A student can only be added to the same grade in the sorted roster more than once",
+          "description": "A student can only be added to the same grade in the roster once",
           "property": "roster",
           "input": {
             "students": [["Aimee", 2], ["Aimee", 2]]
--- a/exercises/grade-school/canonical-data.json
+++ b/exercises/grade-school/canonical-data.json
@@ -82,6 +82,23 @@
               "desiredGrade": 1
             },
             "expected": []
+        },
+        {
+          "uuid": "a0c7b9b8-0e89-47f8-8b4a-c50f885e79d1",
+          "reimplements": "c125dab7-2a53-492f-a99a-56ad511940d8",
+          "comments": [
+            "Reimplemented to be logically consistent"
+          ],
+          "description": "A student can't be in two different grades",
+          "property": "grade",
+          "input": {
+            "students": [
+              ["Aimee", 2],
+              ["Aimee", 1]
+            ],
+            "desiredGrade": 1
+          },
+          "expected": []
         }
     ]
 }
--- a/exercises/grade-school/canonical-data.json
+++ b/exercises/grade-school/canonical-data.json
@@ -28,6 +28,19 @@
           "expected": []
         },
         {
+          "uuid": "a0c7b9b8-0e89-47f8-8b4a-c50f885e79d1",
+          "reimplements": "c125dab7-2a53-492f-a99a-56ad511940d8",
+          "comments": [
+            "Reimplemented to be logically consistent"
+          ],
+          "description": "A student can only be added to the same grade in the sorted roster more than once",
+          "property": "roster",
+          "input": {
+            "students": [["Aimee", 2], ["Aimee", 2]]
+          },
+          "expected": ["Aimee"]
+        },
+        {
             "uuid": "233be705-dd58-4968-889d-fb3c7954c9cc",
             "description": "Adding more students adds them to the sorted roster",
             "property": "roster",
@@ -44,6 +57,19 @@
               "students": [["Chelsea", 3], ["Logan", 7]]
             },
             "expected": ["Chelsea", "Logan"]
+        },
+        {
+          "uuid": "6a03b61e-1211-4783-a3cc-fc7f773fba3f",
+          "reimplements": "c125dab7-2a53-492f-a99a-56ad511940d8",
+          "comments": [
+            "Reimplemented to be logically consistent"
+          ],
+          "description": "A student cannot be added to more than one grade in the sorted roster",
+          "property": "roster",
+          "input": {
+            "students": [["Aimee", 2], ["Aimee", 1]]
+          },
+          "expected": ["Aimee"]
         },
         {
             "uuid": "a3f0fb58-f240-4723-8ddc-e644666b85cc",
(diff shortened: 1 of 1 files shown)


petertseng (commented):
To me, I find this new case a0c7b9b8-0e89-47f8-8b4a-c50f885e79d1 equally as unsatisfying as the case that it reimplements c125dab7-2a53-492f-a99a-56ad511940d8 because both of these cases can be passed by an implementation that declares this:

* a student that is added twice shall be considered to belong to **neither** of the grades they were attempted to be added to.

In other words, it does not seem to me that either the added case nor the case that it reimplements distinguishes between a desired behaviour and my above-described behaviour (and I presume my above-described behaviour is not what is intended/desired by the https://github.com/exercism/problem-specifications/blob/main/exercises/grade-school/description.md).

Thus, this newly-added case cannot receive my endorsement (and thus my GitHub Approval). But since there is a valid implementation (the one I described above) that passes it, I do not declare it incorrect, so it will also not receive my refusal. If others decide this case is the best solution to https://github.com/exercism/problem-specifications/issues/1763 and the right way forward given the problems described therein, I do not oppose that determination - in such a case, please proceed without my Approval.

petertseng:
EBWODP: clarify what I mean by my last sentence.

If others decide this case is the best solution to https://github.com/exercism/problem-specifications/issues/1763 and the right way forward given the problems described therein, I do not oppose that determination - in such a case, please proceed without my Approval.

petertseng (approved):
Upon review, I specifically revoke the part of my previous comment where I claim the added case is equally as unsatisfying as the case it reimplements. I now notice that the reimplemented case is a `roster` with a `desiredGrade`, which is very much against the https://github.com/exercism/problem-specifications/blob/main/exercises/grade-school/description.md.

My standard for a GitHub Approval shall be whether a change is an improvement. The added case is unsatisfying (my reasons for declaring it such are unchanged), but it is less unsatisfying than the case it reimplements. Therefore, it receives an Approval.

wolf99:
>both of these cases can be passed by an implementation that declares this:

> - a student that is added twice shall be considered to belong to neither of the grades they were attempted to be added to.

I agree that this is an issue. I will move this PR to draft and look at solutions.
Thanks @petertseng !

wolf99:
I realise this still doesn't cover all all aspects of this condition. This is because the case is really an error condition for the action of *adding* students rather than querying the roster... Fixing this will require the entire spec to be changed. I'm not averse to doing that but would like to get progress on #1763 and https://github.com/exercism/c/pull/555 first though.

petertseng (approved):
We have an interesting situation: If we wish to signal explicitly that a case in canonical-data.json is no longer recommended, we can add a case that `reimplements` it. But I believe this is a situation where we might simply wish for a way to signal that c125dab7-2a53-492f-a99a-56ad511940d8 is no longer recommended and **not** provide a case that `reimplements` it. But we have no way to do that. I'll continue to Approve any valid and consistent cases that improve the situation. These two cases fit that criterion. Perfection isn't a requirement.

The reason I wished for a way to not provide a replacement has already been pointed out - we don't currently have a good way to be clear about the fact that this is an error that should happen at `add` time, not at query (`roster`/`grade`) time. We could:

1. Use a different `property` to signal that
2. Just put an `error` in the `roster` cases and expect tracks to do what makes the most sense for them
3. Choose not to specify any such cases in this canonical-data.json at all
4. Other solution I didn't think of

And, since c125dab7-2a53-492f-a99a-56ad511940d8 exists, doing 3 involves a way to signal its un-recommendation.

Note that we used to have a similar problem for bowling - any errors in response to invalid rolls should happen at `roll` time, not `score` time. This had to be fixed in https://github.com/exercism/problem-specifications/pull/536.

petertseng:
EBWODP:

> any invalid rolls should happen at `roll` time

This should instead be "if there are any invalid rolls, the errors should be signaled at `roll` time"

exercises/grade-school/canonical-data.json
```diff
@@ -27,6 +27,19 @@
           },
           "expected": []
         },
+        {
+          "uuid": "a0c7b9b8-0e89-47f8-8b4a-c50f885e79d1",
+          "reimplements": "c125dab7-2a53-492f-a99a-56ad511940d8",
+          "comments": [
+            "Reimplemented to be logically consistent"
+          ],
+          "description": "A student can only be added to the same grade in the sorted roster more than once",
```
coriolinus, line 36:
```suggestion
          "description": "A student can only be added to the same grade in the roster once",
```

- Whether or not the roster is sorted is irrelevant here.
- A student _cannot_ be added more than once.

coriolinus (approved):

ErikSchierboom:
> And, since c125dab7-2a53-492f-a99a-56ad511940d8 exists, doing 3 involves a way to signal its un-recommendation.
> 
> Note that we used to have a similar problem for bowling - if there are any invalid rolls, the errors should be signaled at roll time, not score time. This had to be fixed in #536.

Interesting problem. We indeed don't have a way to signify this "just don't use this test case anymore" scenario. Will perhaps be useful in the future. The only thing thats I could come up with are:

- Add comments (which don't show up in `tests.toml` files)
- Add a suffix to the title (which _does_ show up in `tests.toml` files)

SleeplessByte (approved):

SleeplessByte:
Thanks @wolf99. I like the new canonical data, and thanks everyone else for chiming in.

PR #1834, opened 2021-09-09
Clarify problem description of two-bucket

### LINKED ISSUE #1832: Two-Buckets ambiguous wording in problem statement
In [the problem statement](https://github.com/exercism/problem-specifications/blob/main/exercises/two-bucket/description.md) there is a paragraph that reads:

>To help, the tests provide you with which bucket to fill first. That means, when starting with the larger bucket full, you are NOT allowed at any point to have the smaller bucket full and the larger bucket empty (aka, the opposite starting point); that would defeat the purpose of comparing both approaches!

This assumption holds until the test cases added ~ 10 mos ago in this ~commit: https://github.com/exercism/problem-specifications/commit/14131a0ad4f4d0ebfe6b348828a4dac1b7645a00~ Pr: #1580 

Now it requires you to forbid any traversal through a state where the starting bucket is 0 and the other bucket is max capacity.

Thoughts? I think the problem statement could be clarified with clearer rules about the intended solution.

Happy to PR if people agree.

neenjaw:
Followup to #1832.

Goals:
- maintain current problem with current limitations/scope of the exercise
- clarify the requirements of programs when generating a solution (or finding there is no solution)

close #1832

### DIFF
--- a/exercises/two-bucket/description.md
+++ b/exercises/two-bucket/description.md
@@ -1,12 +1,12 @@
 # Description
 
-Given two buckets of different size and which bucket to fill first, determine how many actions are required to measure an exact number of liters by strategically transferring fluid between the buckets. Since this mathematical problem is subject to a fair amount of interpretation and an individual's approach, the tests have been written specifically to expect one overarching solution.
+Given two buckets of different size and which bucket to fill first, determine how many actions are required to measure an exact number of liters by strategically transferring fluid between the buckets.
 
 There are some rules that your solution must follow:
 
 - You can only do one action at a time.
 - There are only 3 possible actions:
-  1. Pouring one bucket into the other bucket until:
+  1. Pouring one bucket into the other bucket until either:
      a) the first bucket is empty
      b) the second bucket is full
   2. Emptying a bucket and doing nothing to the other.
@@ -29,6 +29,6 @@ Your program should determine:
 Note: any time a change is made to either or both buckets counts as one (1) action.
 
 Example:
-Bucket one can hold up to 7 liters, and bucket two can hold up to 11 liters. Let's say at a given step, bucket one is holding 7 liters and bucket two is holding 8 liters (7,8). If you empty bucket one and make no change to bucket two, leaving you with 0 liters and 8 liters respectively (0,8), that counts as one "move". Instead, if you had poured from bucket one into bucket two until bucket two was full, resulting in 4 liters in bucket one and 11 liters in bucket two (4,11), that would also only count as one "move".
+Bucket one can hold up to 7 liters, and bucket two can hold up to 11 liters. Let's say at a given step, bucket one is holding 7 liters and bucket two is holding 8 liters (7,8). If you empty bucket one and make no change to bucket two, leaving you with 0 liters and 8 liters respectively (0,8), that counts as one action. Instead, if you had poured from bucket one into bucket two until bucket two was full, resulting in 4 liters in bucket one and 11 liters in bucket two (4,11), that would also only count as one action.
 
 Written with <3 at [Fullstack Academy](http://www.fullstackacademy.com/) by Lindsay Levine.
--- a/exercises/two-bucket/description.md
+++ b/exercises/two-bucket/description.md
@@ -31,4 +31,7 @@ Note: any time a change is made to either or both buckets counts as one (1) acti
 Example:
 Bucket one can hold up to 7 liters, and bucket two can hold up to 11 liters. Let's say at a given step, bucket one is holding 7 liters and bucket two is holding 8 liters (7,8). If you empty bucket one and make no change to bucket two, leaving you with 0 liters and 8 liters respectively (0,8), that counts as one action. Instead, if you had poured from bucket one into bucket two until bucket two was full, resulting in 4 liters in bucket one and 11 liters in bucket two (4,11), that would also only count as one action.
 
+Another Example:
+Bucket one can hold 3 liters, and bucket two can hold up to 5 liters.  You are told you must start with bucket one.  So your first action is to fill bucket one.  You choose to empty bucket one for your second action.  For your third action, you may not fill bucket two, because this violates the third rule -- you may not end up in a state after any action where the starting bucket is empty and the other bucket is full.
+
 Written with <3 at [Fullstack Academy](http://www.fullstackacademy.com/) by Lindsay Levine.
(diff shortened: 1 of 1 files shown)


neenjaw:
Pinged you three for reviews since you all :+1: my last comment on the issue.  Happy to consider revisions and your input.  Thanks in advance for your time!

exercises/two-bucket/description.md
```diff
@@ -1,30 +1,34 @@
...
+Note: any time a change is made to either or both buckets counts as one (1) action.
 
-To conclude, the only valid moves are:
-- pouring from either bucket to another
-- emptying either bucket and doing nothing to the other
-- filling either bucket and doing nothing to the other
+Example:
+Bucket one can hold up to 7 liters, and bucket two can hold up to 11 liters. Let's say at a given step, bucket one is holding 7 liters and bucket two is holding 8 liters (7,8). If you empty bucket one and make no change to bucket two, leaving you with 0 liters and 8 liters respectively (0,8), that counts as one "move". Instead, if you had poured from bucket one into bucket two until bucket two was full, resulting in 4 liters in bucket one and 11 liters in bucket two (4,11), that would also only count as one "move".
```
petertseng, line 32: other parts of this commit have replaced quoted `"move"` with unquoted `action`. Perhaps the same replacement can be done here as well.

exercises/two-bucket/description.md
```diff
@@ -1,30 +1,34 @@
...
+Given two buckets of different size and which bucket to fill first, determine how many actions are required to measure an exact number of liters by strategically transferring fluid between the buckets. Since this mathematical problem is subject to a fair amount of interpretation and an individual's approach, the tests have been written specifically to expect one overarching solution.
 
-Since this mathematical problem is fairly subject to interpretation / individual approach, the tests have been written specifically to expect one overarching solution.
+There are some rules that your solution must follow:
 
-To help, the tests provide you with which bucket to fill first. That means, when starting with the larger bucket full, you are NOT allowed at any point to have the smaller bucket full and the larger bucket empty (aka, the opposite starting point); that would defeat the purpose of comparing both approaches!
+- You can only do one action at a time.
+- There are only 3 possible actions:
+  1. Pouring one bucket into the other bucket until:
```
petertseng, line 9: adding something like "either" here may help to clarify that it's not "both"

petertseng (approved):
This is an improvement; I Approve (GitHub sense of the word). Here are some minor improvements you may also consider.

exercises/two-bucket/description.md
```diff
@@ -1,30 +1,34 @@
 # Description
 
-Given two buckets of different size, demonstrate how to measure an exact number of liters by strategically transferring liters of fluid between the buckets.
+Given two buckets of different size and which bucket to fill first, determine how many actions are required to measure an exact number of liters by strategically transferring fluid between the buckets. Since this mathematical problem is subject to a fair amount of interpretation and an individual's approach, the tests have been written specifically to expect one overarching solution.
```
jiegillet, line 3: I don't really understand this statement "Since this mathematical problem is subject to a fair amount of interpretation and an individual's approach...". Yes, obviously, before we explain the rules, things are ambiguous, and yes, obviously, tests will expect specific answers. So I would like to hear your thoughts about getting rid of it. I don't think anything is lost.
 
```suggestion
Given two buckets of different size and which bucket to fill first, determine how many actions are required to measure an exact number of liters by strategically transferring fluid between the buckets.
```

exercises/two-bucket/description.md
neenjaw: Agree, missed this instance

neenjaw:
```suggestion
Bucket one can hold up to 7 liters, and bucket two can hold up to 11 liters. Let's say at a given step, bucket one is holding 7 liters and bucket two is holding 8 liters (7,8). If you empty bucket one and make no change to bucket two, leaving you with 0 liters and 8 liters respectively (0,8), that counts as one action. Instead, if you had poured from bucket one into bucket two until bucket two was full, resulting in 4 liters in bucket one and 11 liters in bucket two (4,11), that would also only count as one action.
```

exercises/two-bucket/description.md
neenjaw: Agree

neenjaw:
```suggestion
  1. Pouring one bucket into the other bucket until either:
```

exercises/two-bucket/description.md
neenjaw: It was there before, my first draft attempt was to clarify but retain some of the original authors voice in the copy. I agree though, it is superfluous

glennj:
IMO, the new text could use an example to demonstrate "you may not arrive at a state where the starting bucket is empty and the other bucket is full."

jiegillet (approved):

exercises/two-bucket/description.md
```diff
@@ -1,30 +1,34 @@
...
+Note: any time a change is made to either or both buckets counts as one (1) action.
 
-To conclude, the only valid moves are:
-- pouring from either bucket to another
-- emptying either bucket and doing nothing to the other
-- filling either bucket and doing nothing to the other
+Example:
+Bucket one can hold up to 7 liters, and bucket two can hold up to 11 liters. Let's say at a given step, bucket one is holding 7 liters and bucket two is holding 8 liters (7,8). If you empty bucket one and make no change to bucket two, leaving you with 0 liters and 8 liters respectively (0,8), that counts as one action. Instead, if you had poured from bucket one into bucket two until bucket two was full, resulting in 4 liters in bucket one and 11 liters in bucket two (4,11), that would also only count as one action.
```
neenjaw, line 32:
```suggestion
Bucket one can hold up to 7 liters, and bucket two can hold up to 11 liters. Let's say at a given step, bucket one is holding 7 liters and bucket two is holding 8 liters (7,8). If you empty bucket one and make no change to bucket two, leaving you with 0 liters and 8 liters respectively (0,8), that counts as one action. Instead, if you had poured from bucket one into bucket two until bucket two was full, resulting in 4 liters in bucket one and 11 liters in bucket two (4,11), that would also only count as one action.

Another Example:
Bucket one can hold 3 liters, and bucket two can hold up to 5 liters.  You are told you must start with bucket one.  So your first action is to fill bucket one.  You choose to empty bucket one for your second action.  For your third action, you may not fill bucket two, because this violates the third rule -- you may not end up in a state after any action where the starting bucket is empty and the other bucket is full.
```

neenjaw:
@glennj, see https://github.com/exercism/problem-specifications/pull/1834/commits/3d3c0c56c36b9a63b4865cfc88983fbf3bd93cab

glennj (approved):

PR #1835, opened 2021-09-09
Add test to `rational-numbers`: reduce places minus sign on denominator

### LINKED ISSUE #1833: Proposing a new test case for `rational-numbers`
We received a nice contribution [here](https://github.com/exercism/elixir/pull/892) that added a new test:

```json
        {
          "uuid": "7",
          "description": "Reduce places the minus sign on the numerator",
          "property": "reduce",
          "input": {
            "r": [3, -4]
          },
          "expected": [-3, 4]
        }
``` 
There are tests that _imply_ that this is a property of reduce, but it's nice to have an explicit one.

jiegillet:
Closes #1833. There were no comment except for a thumbs up from @glennj (please review 💟), so I'm moving forward.

We received a [nice contribution](https://github.com/exercism/elixir/pull/892) on the Elixir track adding this test, so I thought other tracks should benefit as well.

This test makes explicit a standard behavior of `reduce`: the minus sign should end up on top. 

This is already implied by other tests, in particular, the test "Divide a positive rational number by a negative rational number" produces a minus sign on the denominator when the instructions are followed to the letter and the test expects the minus sign on top. Therefore implementations shouldn't need to change.

Happy to hear your opinions.

### DIFF
--- a/exercises/rational-numbers/canonical-data.json
+++ b/exercises/rational-numbers/canonical-data.json
@@ -420,6 +420,15 @@
             "r": [13, 13]
           },
           "expected": [1, 1]
+        },
+        {
+          "uuid": "5ed6f248-ad8d-4d4e-a545-9146c6727f33",
+          "description": "Reduce places the minus sign on the numerator",
+          "property": "reduce",
+          "input": {
+            "r": [3, -4]
+          },
+          "expected": [-3, 4]
         }
       ]
     }


glennj (approved):
Looks good.

jiegillet:
Bump? Do requested reviewers get a notification when I add a comment? :)
Should I request a review from more people?

angelikatyborska (approved):
Good addition, but I'm wondering if it should be added before the test `3b92ffc0-5b70-4a43-8885-8acee79cdaaf` which also depends on this behavior, so that the simpler test case (the new one) is first.

SleeplessByte:
(Won't merge because @jiegillet might want to address first)

PR #1836, opened 2021-09-11
Fix a triangle inequality note

Steffan153:
image

The description said "the sum of the lengths of any two sides length of the third side." I fixed this here. Feel free to reword it. :smile:

### DIFF
--- a/exercises/triangle/description.md
+++ b/exercises/triangle/description.md
@@ -13,7 +13,8 @@ A _scalene_ triangle has all sides of different lengths.
 ## Note
 
 For a shape to be a triangle at all, all sides have to be of length > 0, and if `a`, `b`, and `c` are sides of the triangle, all three of the following expressions must be true:
-```
+
+```text
 a + b >= c
 b + c >= a
 a + c >= b
--- a/exercises/triangle/description.md
+++ b/exercises/triangle/description.md
@@ -12,9 +12,14 @@ A _scalene_ triangle has all sides of different lengths.
 
 ## Note
 
-For a shape to be a triangle at all, all sides have to be of length > 0, and
-the sum of the lengths of the first two smallest sides must be greater than or equal to the
-length of the biggest side. See [Triangle Inequality](https://en.wikipedia.org/wiki/Triangle_inequality).
+For a shape to be a triangle at all, all sides have to be of length > 0, and if `a`, `b`, and `c` are sides of the triangle, all three of the following expressions must be true:
+```
+a + b >= c
+b + c >= a
+a + c >= b
+```
+
+See [Triangle Inequality](https://en.wikipedia.org/wiki/Triangle_inequality).
 
 ## Dig Deeper
 
--- a/exercises/triangle/description.md
+++ b/exercises/triangle/description.md
@@ -13,8 +13,8 @@ A _scalene_ triangle has all sides of different lengths.
 ## Note
 
 For a shape to be a triangle at all, all sides have to be of length > 0, and
-the sum of the lengths of any two sides must be greater than or equal to the
-length of the third side. See [Triangle Inequality](https://en.wikipedia.org/wiki/Triangle_inequality).
+the sum of the lengths of the first two smallest sides must be greater than or equal to the
+length of the biggest side. See [Triangle Inequality](https://en.wikipedia.org/wiki/Triangle_inequality).
 
 ## Dig Deeper
 
--- a/exercises/triangle/description.md
+++ b/exercises/triangle/description.md
@@ -12,12 +12,16 @@ A _scalene_ triangle has all sides of different lengths.
 
 ## Note
 
-For a shape to be a triangle at all, all sides have to be of length > 0, and if `a`, `b`, and `c` are sides of the triangle, all three of the following expressions must be true:
+For a shape to be a triangle at all, all sides have to be of length > 0, and the sum of the lengths of any two sides must be greater than or equal to the length of the third side.
+
+In equations:
+
+Let `a`, `b`, and `c` be sides of the triangle. Then all three of the following expressions must be true:
 
 ```text
-a + b >= c
-b + c >= a
-a + c >= b
+a + b ≥ c
+b + c ≥ a
+a + c ≥ b
 ```
 
 See [Triangle Inequality](https://en.wikipedia.org/wiki/Triangle_inequality).


siebenschlaefer:
Previously that sentence said: "*For a shape to be a triangle at all, [...] the sum of the lengths of any two sides must be greater than or equal to the length of the third side.*"  
As a non-native English speaker I interpret that "any" as: a+b>=c *and* b+c>=a *and* c+a>=b

The proposed new sentence says: "*For a shape to be a triangle at all, [...] the sum of the lengths of the first two smallest sides must be greater than or equal to the length of the biggest side*".  
I don't understand the word "first" in that context. Without it I would interpret that as: first(a, b, c) + second_smallest(a, b, c) >= biggest(a, b, c)

Doesn't that mean the same? If yes, then I'd prefer the current wording because it sounds "more mathematical" to me.

Steffan153:
Yes, that means the same. But the old wording is kind of confusing imo, because it (imo) sounds like you can take any old two sides, and if they're greater than or equal to the remaining side, the triangle is valid. Which is not correct.

If you think there's a better way to word it, let me know.

jiegillet:
I prefer the old description. It's a standard math description: "the sum of the lengths of any two sides must be greater than or equal to the length of the third side" which means "for any choice of two sides, the sum is greater than that the third one".

The word "first" in your proposed sentence implies an order that wasn't specified before, so that might be more confusing.

Steffan153:
> "for any choice of two sides, the sum is greater than that the third one"

The problem is, that's incorrect. Take the triangle `[2, 5, 2]`. Let's say we take the choice `[2, 5]`. So, `7 >= 2`. So the triangle is valid, right? Nope. `2 + 2 < 7`. To me, the current description sounds like you can take just any old random pair and it'll work.

jiegillet:
I understand your point, you are saying that "any" means "at least one", but what I'm saying is that the standard meaning when it comes to mathematics is "all".
"it must work with any pair" = "it must work with all pairs", not "it must work with one pair".
Another way to see this is "there cannot be any pair that doesn't satisfy the condition."

angelikatyborska:
I have to agree with @jiegillet, this is also how I was taught basic math:

> I understand your point, you are saying that "any" means "at least one", but what I'm saying is that the standard meaning when it comes to mathematics is "all".

The old description is more correct in my opinion.

petertseng:
Understood by me that by understanding of how "any" is usually used mathematically, current description is correct.

But, could it be that the existence of this PR gives us evidence that the description could nevertheless be improved, despite its current correctness? https://math.stackexchange.com/questions/402020/whats-the-difference-between-any-all-and-some is another person from the internet who is unsure about "any".

But, I must admit it's not easy to come up with an acceptable improvement. A possible rewording that I attempted seems too wordy; the attempt is below:

> for each pair of two sides, the sum of the lengths of those two sides must be greater than or equal to the
length of the third side

Perhaps it could be restated in the opposite order, like:

> the length of each side must be less than or equal to the sum of the length of the other two sides

Steffan153:
That second wording sounds awesome to me. What confuses me is that mathematically, "any" = "all" supposedly, when in Ruby, `any?` means "at least one" (and [English](https://www.dictionary.com/browse/any)).

siebenschlaefer:
@Steffan153 In a mathematical context "any" means the predicate holds for *any* element, i.e. "you could pick *any* element" or "whatever element you pick": the predicate is always true.

jmrunkle:
Why don't we bypass this whole issue and just explicitly write our the three cases using math symbols?

```
a + b >= c
b + c >= a
a + c >= b
```

Leaves no room for confusion in the English no matter how you interpret the and vs any.

Steffan153:
That better?

exercises/triangle/description.md
```diff
@@ -12,9 +12,14 @@ A _scalene_ triangle has all sides of different lengths.
 
 ## Note
 
-For a shape to be a triangle at all, all sides have to be of length > 0, and
-the sum of the lengths of any two sides must be greater than or equal to the
-length of the third side. See [Triangle Inequality](https://en.wikipedia.org/wiki/Triangle_inequality).
+For a shape to be a triangle at all, all sides have to be of length > 0, and if `a`, `b`, and `c` are sides of the triangle, all three of the following expressions must be true:
+```
```
wolf99, line 16:
```suggestion

```text
```

wolf99:
Not sure why the tests think that a case in `luhn` has been changed.
I am not familiar with the CICD in this repo, @ErikSchierboom can you take a look?

ErikSchierboom:
@wolf99 I don't see any linting error.

SaschaMann (changes requested):
> Why don't we bypass this whole issue and just explicitly write our the three cases using math symbols?

I think this makes it a lot harder to tell what is actually required at a glance. One of the first things I was taught in maths lectures was that more symbols don't always make things more readable, quite often words will be clearer. I think this is also the case here.

Perhaps we can have both?

````markdown
For a shape to be a triangle at all, all sides have to be of length > 0, and the sum of the lengths of any two sides must be greater than or equal to the length of the third side.

In equations:

```text
a + b ≥ c
b + c ≥ a
a + c ≥ b
```

See [Triangle Inequality](https://en.wikipedia.org/wiki/Triangle_inequality).
````

szabgab:
Saying that all the sides must be greater than 0 is good, but should the tests also check if an input that violates this will make the functions return false? Right now there is one test-case 'all zero sides is not a triangle' that checks the `equilateral` with such input.

IMHO, to verify that the student indeed understood the restrictions, there should be other test cases with 0 for the other functions and negative numbers like the ones I mentioned in #1871.

BTW Maybe an even better response would be to raise an exception when any of the sides is negative or 0.

siebenschlaefer:
@szabgab There's a section "Dig Deeper":
> The case where the sum of the lengths of two sides _equals_ that of the
> third is known as a _degenerate_ triangle - it has zero area and looks like
> a single line. Feel free to add your own code/tests to check for degenerate triangles.

Adding tests for degenerate triangles would either force users to handle them as "regular" triangles and thereby contradict that section, or require handling them as a special case (e.g. with exceptions or as 4th "flavor") and thereby make that section non-optional.

szabgab:
@siebenschlaefer I am not sure how that quote about degenerate triangles relates to functions calls where one or more of the numbers is 0 or less.

ErikSchierboom:
@Steffan153 Are you interested in continuing to work on this PR? I think that if you apply the suggestion made by @SaschaMann (see https://github.com/exercism/problem-specifications/pull/1836#pullrequestreview-793443657), we might be able to get this merged.

ErikSchierboom:
@Steffan153 Small bump

Steffan153:
Sure. I missed your previous comment, sorry about that.

kotp:
I only want to leave a comment here about historical [conversations regarding degenerate triangles][1].  It appears some tracks allow and some tracks disallow. 

No changes requested, only wanted to keep a note that this has been discussed over the years as well.

[1]: https://github.com/search?q=org%3Aexercism+degenerate+triangle&type=issues

SaschaMann (approved):

PR #1837, opened 2021-09-13
Fix grade-school

### LINKED ISSUE #1824: Grade school: new test raises a question
Regarding PR #1819 
```json
        {
          "uuid": "6a03b61e-1211-4783-a3cc-fc7f773fba3f",
          "reimplements": "c125dab7-2a53-492f-a99a-56ad511940d8",
          "comments": [
            "Reimplemented to be logically consistent"
          ],
          "description": "A student cannot be added to more than one grade in the sorted roster",
          "property": "roster",
          "input": {
            "students": [["Aimee", 2], ["Aimee", 1]]
          },
          "expected": ["Aimee"]
        },
```
All well and good, but what grade is Aimee in? 
1. grade 2 because that's what came first, or
2. grade 1 because that's what came last?

I think we're missing the property:grade test for this scenario.

wolf99:
Example implementation here: https://github.com/exercism/c/pull/683

Fixes #1824

Wasn't sure how to replace the existing `grade-school` exercise so have just used `grade-school2` (should this be `grade-school-2`?) as had been suggested in slack (https://exercism-team.slack.com/archives/CARRG4MNC/p1628017550027100?thread_ts=1627502454.403900&cid=CARRG4MNC)

If accepted the intention is that the existing `grade-school` should be deprecated as the spec is too problematic. 
I can add that as part of this PR if there is an agreed method?

### DIFF
--- a/exercises/grade-school/canonical-data.json
+++ b/exercises/grade-school/canonical-data.json
@@ -1,244 +1,241 @@
 {
-    "exercise": "grade-school",
-    "comments": [
-        "Given students' names along with the grade that they are in, ",
-        "create a roster for the school."
-    ],
-    "cases": [
-        {
-            "uuid": "a3f0fb58-f240-4723-8ddc-e644666b85cc",
-            "description": "Roster is empty when no student is added",
-            "property": "roster",
-            "input": {
-              "students": []
-            },
-            "expected": []
-        },
-        {
-            "uuid": "9337267f-7793-4b90-9b4a-8e3978408824",
-            "description": "Add a student",
-            "property": "add",
-            "input": {
-              "students": [["Aimee", 2]]
-            },
-            "expected": [true]
-        },
-        {
-            "uuid": "6d0a30e4-1b4e-472e-8e20-c41702125667",
-            "description": "Student is added to the roster",
-            "property": "roster",
-            "input": {
-              "students": [["Aimee", 2]]
-            },
-            "expected": ["Aimee"]
-        },
-        {
-            "uuid": "73c3ca75-0c16-40d7-82f5-ed8fe17a8e4a",
-            "description": "Adding multiple students in the same grade in the roster",
-            "property": "add",
-            "input": {
-              "students": [["Blair", 2], ["James", 2], ["Paul", 2]]
-            },
-            "expected": [true, true, true]
-        },
-        {
-            "uuid": "233be705-dd58-4968-889d-fb3c7954c9cc",
-            "description": "Multiple students in the same grade are added to the roster",
-            "property": "roster",
-            "input": {
-              "students": [["Blair", 2], ["James", 2], ["Paul", 2]]
-            },
-            "expected": ["Blair", "James", "Paul"]
-        },
-        {
-            "uuid": "87c871c1-6bde-4413-9c44-73d59a259d83",
-            "description": "Cannot add student to same grade in the roster more than once",
-            "property": "add",
-            "input": {
-              "students": [["Blair", 2], ["James", 2], ["James", 2], ["Paul", 2]]
-            },
-            "expected": [true, true, false, true]
-        },
-        {
-            "uuid": "c125dab7-2a53-492f-a99a-56ad511940d8",
-            "description": "A student can't be in two different grades",
-            "property": "roster",
-            "input": {
-              "students": [
-                ["Aimee", 2],
-                ["Aimee", 1]
-              ],
-              "desiredGrade": 2
-            },
-            "expected": []
-        },
-        {
-            "uuid": "a0c7b9b8-0e89-47f8-8b4a-c50f885e79d1",
-            "reimplements": "c125dab7-2a53-492f-a99a-56ad511940d8",
-            "comments": [
-              "Reimplemented to be logically consistent"
-            ],
-            "description": "A student can only be added to the same grade in the roster once",
-            "property": "roster",
-            "input": {
-              "students": [["Aimee", 2], ["Aimee", 2]]
-            },
-            "expected": ["Aimee"]
-        },
-        {
-            "uuid": "d7982c4f-1602-49f6-a651-620f2614243a",
-            "description": "Student not added to same grade in the roster more than once",
-            "reimplements": "a0c7b9b8-0e89-47f8-8b4a-c50f885e79d1",
-            "comments": [
-              "Ensure the implementation does not affect students already added",
-              "Ensure the implementation does not prevent addition of other students after preventing the error case"
-            ],
-            "property": "roster",
-            "input": {
-              "students": [["Blair", 2], ["James", 2], ["James", 2], ["Paul", 2]]
-            },
-            "expected": ["Blair", "James", "Paul"]
-        },
-        {
-            "uuid": "e70d5d8f-43a9-41fd-94a4-1ea0fa338056",
-            "description": "Adding students in multiple grades",
-            "property": "add",
-            "input": {
-              "students": [["Chelsea", 3], ["Logan", 7]]
-            },
-            "expected": [true, true]
-        },
-        {
-            "uuid": "75a51579-d1d7-407c-a2f8-2166e984e8ab",
-            "description": "Students in multiple grades are added to the roster",
-            "property": "roster",
-            "input": {
-              "students": [["Chelsea", 3], ["Logan", 7]]
-            },
-            "expected": ["Chelsea", "Logan"]
-        },
-        {
-            "uuid": "7df542f1-57ce-433c-b249-ff77028ec479",
-            "description": "Cannot add same student to multiple grades in the roster",
-            "property": "add",
-            "input": {
-              "students": [["Blair", 2], ["James", 2], ["James", 3], ["Paul", 3]]
-            },
-            "expected": [true, true, false, true]
-        },
-        {
-            "uuid": "6a03b61e-1211-4783-a3cc-fc7f773fba3f",
-            "reimplements": "c125dab7-2a53-492f-a99a-56ad511940d8",
-            "comments": [
-              "Reimplemented to be logically consistent"
-            ],
-            "description": "A student cannot be added to more than one grade in the sorted roster",
-            "property": "roster",
-            "input": {
-              "students": [["Aimee", 2], ["Aimee", 1]]
-            },
-            "expected": ["Aimee"]
-        },
-        {
-            "uuid": "c7ec1c5e-9ab7-4d3b-be5c-29f2f7a237c5",
-            "reimplements": "6a03b61e-1211-4783-a3cc-fc7f773fba3f",
-            "comments": [
-              "Ensure the implementation does not affect students already added",
-              "Ensure the implementation does not prevent addition of other students after preventing the error case"
-            ],
-            "description": "Student not added to multiple grades in the roster",
-            "property": "roster",
-            "input": {
-              "students": [["Blair", 2], ["James", 2], ["James", 3], ["Paul", 3]]
-            },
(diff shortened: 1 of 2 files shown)


junedev:
@wolf99 In the JavaScript track someone suggested an additional (or changed) test case for this exercise: https://github.com/exercism/javascript/issues/1337

Could you maybe have look whether that could be covered with your new system and whether you want to include that case?

exercises/grade-school2/canonical-data.json
```diff
@@ -0,0 +1,155 @@
...
+            "students": [["Aimee", 2], ["Aimee", 2]]
+          },
+          "expected": ["Aimee"]
+        },
+        {
+          "uuid": "e70d5d8f-43a9-41fd-94a4-1ea0fa338056",
+          "description": "Adding students in multiple grades",
+          "property": "add",
+          "input": {
+            "students": [["Chelse", 3], ["Logan", 7]]
```
petertseng, line 76: Although it does not affect correctness since the name does not appear in the output, if the other instance of Chelse is changed, it may be prudent to change this one as well for consistency.

petertseng (changes requested):
As of this time, here is what my review has covered:

* I see that the description.md, metadata.toml, and metadata.yml are exactly the same as the existing ones of grade-school.
* The proposed API given by `add`, `roster`, and `grade` are agreeable to me.
* The test cases' correctness, about which I have things to say. The verifier used is at https://github.com/petertseng/exercism-problem-specifications/blob/grade1837-ver/exercises/grade-school2/verify.rb and will be updated if any significant changes to the API are implemented.

Here are things I have no opinion on, and so review of that is left to others (I will not review these things):

* The mechanism by which grade-school should be replaced, if at all.

petertseng:
Here are some other cases you might consider. I consider these optional; I would still approve a correct PR that does not contain these cases.

* add: Some case where there should be a `false` followed by a `true` - ensures that a failed addition doesn't prevent future successful additions.
* roster: Analogous to the above. Add Aimee, Aimee, Bethany, and ensure that Bethany is in the roster.
* grade: Analogous to the above.

petertseng:
EBWODP: Here is something that I think should be a necessary component of this: It should be clearly specified in the description that we are not allowing two students to have the same name.

ErikSchierboom:
Reading #1824, the main problem mentioned was that the duplicate add behaviour was not well-specified. You also mention:

> Intention after the mentioned PR was to rewrite the spec completely such that it covers the operation of adding students, as this clears up this and a lot of other issues.

What other issues are you referring to?

wolf99:
Thanks for the comments all.

@petertseng  I will update these now. The metadata etc are the same as grade-school as the intention is to specify the same problem, but in such a way as it can be solved without logical inconsistencies.

@junedev I will make up a commit to add this shortly

@ErikSchierboom was being a bit enthusiastic there I think. Main thing is to resolve the logical inconsistency in #1763 and #1824.  #1819 was intended to resolve #1763, but I felt it wasn't the best solution to the issue.

ErikSchierboom:
I'm having a hard time figuring out what exactly changed :) The issue was with the add method, right? And the fix was to return a boolean indicating if the value could be added?

wolf99:
Hi @ErikSchierboom , 
The issue for one of the existing `add` test cases was that I had specified that it expected `true` output, whereas going byt hte other `add` test cases I should have specified `[true]` (singuar vs array of one). 
So the change for this comment was quite small. 
It is on line 24 in the diff for that commit: https://github.com/exercism/problem-specifications/pull/1837/commits/e43e0b876b970719db1d1579b1e4431196b25275

wolf99:
My new commit just now should, I hope, add the test cases suggested by @petertseng  and @junedev

exercises/grade-school2/canonical-data.json
```diff
@@ -0,0 +1,185 @@
...
+        },
+        {
+          "uuid": "66c8e141-68ab-4a04-a15a-c28bc07fe6b9",
+          "description": "Student not added to multiple grades",
+          "property": "grade",
+          "input": {
+            "students": [["Blair", 2], ["James", 2], ["James", 3], ["Paul", 3]],
+            "desiredGrade": 5
+          },
+          "expected": ["Blair", "James", "Paul"]
```
petertseng, line 172: I don't think it can be proper for grade 5 to have a nonzero number of students when no students were added to grade 5.

petertseng (changes requested):

exercises/grade-school2/canonical-data.json
wolf99: Whoops, copy-pasta error

wolf99:
The review comment from @petertseng  here: https://github.com/exercism/problem-specifications/pull/1837#discussion_r713393315 bings up another corner case.

In the test case referred to in the comment, the intention is to ensure that James is not added to two grades.
But grade only returns one grade. So this test actually only confirms that James is in the selected grade, it does not confirm that James ins not also in other grades. Not sure if there is a good way to specify a test that would confirm what is actually intended.

Any thoughts?

petertseng (dismissed):

petertseng:
> So this test actually only confirms that James is in the selected grade, it does not confirm that James ins not also in other grades. Not sure if there is a good way to specify a test that would confirm what is actually intended.

Indeed, not with what we currently have. The way that fits in most naturally with the existing schema may simply be to have another case with the same inputs and different `desiredGrade`, right? It seems like a bit much to create a new `property` just to express that.

petertseng:
EBWODP:

Delete following unclear text:

> ~~It seems like a bit much to create a new `property` just to express that.~~

replace with new clearer text:

Another way to do it would be to create a new `property`, but that seems a bit much, so the new test case seems like a more natural fit.

ErikSchierboom:
With the risk of sounding dumb: was the old exercise unsalvageable? As in: couldn't we reimplement some test cases and add new ones with proper behavior?

petertseng:
> With the risk of sounding dumb: was the old exercise unsalvageable? As in: couldn't we reimplement some test cases and add new ones with proper behavior?

`roster` and `grade` semantics **are compatible** between the two.

Raw data supporting the above statement:

The verifier for grade-school2 gets the correct result on all[1] grade-school cases without modification to either the cases or the verifier.
The verifier for grade-school gets the correct result on all grade-school2 cases with only one modification to the verifier (must only show any name only once for a `grade` answer, instead of potentially having duplicate names) and the only modification to the cases being correcting the incorrect case pointed out in https://github.com/exercism/problem-specifications/pull/1837#pullrequestreview-760262186, which also needs to be corrected for the grade-school2 verifier to get the correct result anyway.

[1]: excludes the one reimplemented grade-school case

`add` cases are new, so certainly they don't present a compatibility problem.

Based on these above statements about `add`, `roster`, and `grade`, right now it doesn't seem to me that a new exercise is needed. 
I would defer to a reasoned and well-informed judgment by interested individuals if new information comes up that would indicate it's needed.
I'm not particularly for or against, but the above are the suggestions I would make to the interested individuals who are, and I will continue to make new suggestions if new information comes to light.

wolf99:
TBH, I find it very hard to keep track of whats going on in a canonical-data.json when there are more than one or two deprecated test cases, thus, where we have several here, I though it better/easier to start afresh.

However, if you think it better to continue in the original `grade-school` rather than create `grade-school2` I'm happy enough to attempt to "merge" the two together once we are happy with the new test cases here.

junedev:
I was under the impression the two version are not compatible. But if they are, then I would vote for salvaging the existing exercise even if that means just adding another "re-implements" thingy for all test cases. As far as I know there is no (semi-automated) way to signal to all tracks that have the old version "this exercise is deprecated, implement this new exercise instead" so the chances the tracks apply the fixes is probably higher when they can re-sync the existing exercise instead (even if they need to adjust the code a bit).

wolf99:
Have merged `grade-school2` into `grade-school`, removing `grade-school2`.
This required some description changes in existing tests, addition of new tests, re-ordering some existing tests so as to make more sense, and re-implementing a few tests.

exercises/grade-school/canonical-data.json
```diff
@@ -1,113 +1,234 @@
...
+      "reimplements": "c125dab7-2a53-492f-a99a-56ad511940d8",
+      "comments": [
+        "Reimplemented to be logically consistent"
+      ],
+      "description": "A student cannot be added to more than one grade in the sorted roster",
+      "property": "roster",
+      "input": {
+        "students": [["Aimee", 2], ["Aimee", 1]]
+      },
+      "expected": ["Aimee"]
```
ErikSchierboom, line 140: Just to be clear: for this test case to be implemented, we expect the test suite to rewrite this to something like:

```csharp
var gradeSchool= new GradeSchool();
gradeSchool.Add("Aimee", 2);
gradeSchool.Add("Aimee", 1);
gradeSchool.Roster() // Returns ["Aimee"]
```

Right?

exercises/grade-school/canonical-data.json
```diff
@@ -1,113 +1,234 @@
...
+      "description": "Students are sorted by name in the roster",
+      "property": "roster",
+      "input": {
+        "students": [["Peter", 2], ["Zoe", 2], ["Alex", 2]]
+      },
+      "expected": ["Alex", "Peter", "Zoe"]
+    },
+    {
+      "uuid": "180a8ff9-5b94-43fc-9db1-d46b4a8c93b6",
+      "description": "Students are sorted by grades and names in the roster",
```
ErikSchierboom, line 176: Maybe rewrite slightly to make the sorting order clearer?

```suggestion
      "description": "Students are sorted by grade and then by name in the roster",
```

exercises/grade-school/canonical-data.json
```diff
@@ -1,113 +1,234 @@
...
+      },
+      "expected": ["Blair", "James", "Paul"]
+    },
+    {
+      "uuid": "66c8e141-68ab-4a04-a15a-c28bc07fe6b9",
+      "description": "Student not added to multiple grades",
+      "property": "grade",
+      "input": {
+        "students": [["Blair", 2], ["James", 2], ["James", 3], ["Paul", 3]],
+        "desiredGrade": 2
```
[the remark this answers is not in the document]
wolf99: Well caught, this one got lost in the merge!

exercises/grade-school/canonical-data.json
```diff
@@ -1,113 +1,241 @@
...
+      "description": "A student cannot be added to more than one grade in the sorted roster",
+      "property": "roster",
+      "input": {
+        "students": [["Aimee", 2], ["Aimee", 1]]
+      },
+      "expected": ["Aimee"]
+    },
+    {
+      "description": "Student not added to multiple grades in the roster",
+      "uuid": "c7ec1c5e-9ab7-4d3b-be5c-29f2f7a237c5",
```
petertseng, line 141: this case appears to uniquely be the one where `uuid` isn't first. obviously this isn't a problem as the keys can come in any order, but it is unusual.

exercises/grade-school/description.md
```diff
@@ -21,8 +21,11 @@ In the end, you should be able to:
   and Jim in grade 5.
   So the answer is: Anna, Barb, Charlie, Alex, Peter, Zoe and Jim"
 
-Note that all our students only have one name.  (It's a small town, what
-do you want?)
+Note that all our students only have one name  (It's a small town, what
+do you want?) and each student cannot be added more than once to a grade or the
+roster.
+In fact, when a test attempts to add the same student more than once, your
+implementation should indicate that this is incorrect
```
petertseng, line 28: this sentence could use a period at the end.

ErikSchierboom:
Thanks @wolf99!

PR #1842, opened 2021-09-18
Two-fer: add title to metadata

loziniak:
Because it would be possible otherwise to generate title "Two Fer" by automatic exercise generator. Proper title punctuation is taken from description.md.


petertseng (approved):
I approved because "Two Fer" doesn't seem correct. "Two-fer" seems correct.

ErikSchierboom (approved):

angelikatyborska (approved):

PR #1843, opened 2021-09-18
Resistor-color-duo: one-digit solution

loziniak:
With "black" as first band.

Perhaps instructions should also be updated, as not every output is two-digit?


loziniak:
Looks similar to #1605.

SaschaMann (approved):

ErikSchierboom (approved):

wolf99 (approved):

PR #1845, opened 2021-09-19
Luhn: A new test case with odd number of digits

fahadhk:
A new positive test case with odd number of digits and non-zero first digit.
I have seen a solution in Golang which passes all the tests for Luhn containing odd number of digits and never reads the first digit.
It happens because in all such cases the first digit is zero . This test case will fix the problem.

### DIFF
--- a/exercises/luhn/canonical-data.json
+++ b/exercises/luhn/canonical-data.json
@@ -146,6 +146,15 @@
       "expected": true
     },
     {
+      "uuid": "ab56fa80-5de8-4735-8a4a-14dae588663e",
+      "description": "valid luhn with an odd number of digits and non zero first digit",
+      "property": "valid",
+      "input": {
+        "value": "109"
+      },
+      "expected": true
+    },
+    {
       "comments": [
         "Convert non-digits to their ascii values and then offset them by 48 sometimes accidentally declare an invalid string to be valid.",
         "This test is designed to avoid that solution."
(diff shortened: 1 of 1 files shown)


petertseng (approved):
That is actually quite unfortunate that the current cases miss this.

So here is the impl you would add to https://github.com/petertseng/exercism-problem-specifications/blob/verify/exercises/luhn/verify.rb to check this:

```ruby
  {
    # https://github.com/exercism/problem-specifications/pull/1845
    name: 'ignore first digit',
    should_fail: true,
    f: ->(i, _) {
      no_space = i['value'].delete(' ')
      valid_luhn?(no_space.size.odd? ? no_space[1..] : no_space)
    }   
  }, 
```

And indeed, without this test case, that implementation passes all tests (when it shouldn't)

```
== anti-verify ignore first digit valid ==
/home/pt/src/exercism/problem-specifications/verify.rb:130:in `block (2 levels) in multi_verify': ignore first digit: 0 failed, 0 errored in a test of valid (RuntimeError)
```

Only after adding this case does that implementation fail a test

```
== anti-verify ignore first digit valid ==
17. input digit 9 is correctly converted to output digit 9: Got false instead of true
```

We keep finding incorrect implementations that the tests don't catch. Some may find this troubling.

ErikSchierboom (approved):

jiegillet (approved):

PR #1847, opened 2021-09-20
Additional test case for luhn exercise

moshegood:
Often, students attempt to convert the entire string to a number to validate
that it contains only digits. This is not a good approach, as you may overflow.


exercises/luhn/canonical-data.json
```diff
@@ -145,6 +145,15 @@
       },
       "expected": true
     },
+    {
+      "uuid": "b9887ee8-8337-46c5-bc45-3bcab51bc36f",
+      "description": "very long input isn't converted to a single number to validate all digits",
```
wolf99, line 150: The description seem very focused on the one aspect of implementation.
How about something like this?

```suggestion
      "description": "very long input is valid",
      "comments": [
        "Designed to prevent converting input to a single number as this approach can risk an overflow"
        ],
```

ErikSchierboom (approved):

PR #1848, opened 2021-09-21
Remove old metadata.yml files

ErikSchierboom:
- Remove metadata.yml files
- Fix blurb script to use TOML


bin/transfer_blurb_to_description.rb
```diff
@@ -33,9 +33,8 @@ def blurb
   end
 
   def metadata
-    # TODO: Update to parse TOML instead
     metadata_filename = File.join(path, 'metadata.yml')
```
angelikatyborska, line 36: I'm not 100% sure what this script does, does it read the `metadata.yml` files from this repos, or from tracks? If from this repo, the filename need changing to `metadata.toml`?

ErikSchierboom: The filename needed to be changed. It was there to transfer blurbs to description. We can _probably_ remove that file, but I didn't want to do that in this PR. I'm not sure of the removal though.

ErikSchierboom:
Pinging @exercism/reviewers

PR #1854, opened 2021-10-04
fix typo for bash raindrops exercise

khendrikse:
I saw this little error while doing the exercise. I am hoping this reads a bit easier!

Going from:
`The simplest way to test if a one number is a factor of another is`

to 

`The simplest way to test if one number is a factor of another is`

### DIFF
--- a/exercises/raindrops/description.md
+++ b/exercises/raindrops/description.md
@@ -1,6 +1,6 @@
 # Description
 
-Your task is to convert a number into a string that contains raindrop sounds corresponding to certain potential factors. A factor is a number that evenly divides into another number, leaving no remainder. The simplest way to test if a one number is a factor of another is to use the [modulo operation](https://en.wikipedia.org/wiki/Modulo_operation).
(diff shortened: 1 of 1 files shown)


SleeplessByte (approved):

SleeplessByte:
cc @exercism/reviewers

neenjaw (approved):

BethanyG (approved):

glennj:
OK, now what has to happen so the change is reflected in each track?

BethanyG:
@glennj  -- I think its a `configlet` thing.  But the docs are [sparse](https://github.com/exercism/docs/blob/main/building/configlet/generating-documents.md)  @SleeplessByte -- any insight into the syncing action?

angelikatyborska:
Once https://github.com/exercism/configlet/issues/179 gets implemented, track maintainers will be able to automatically update their tracks so that the practice exercises all have up to date instructions. Otherwise it's not possible at the moment, not without @ErikSchierboom running a script that would do this.

It happened once this year for all exercises with Erik's script.

SleeplessByte:
Until then, track maintainers can manually update using configlet sync (or generate).

ErikSchierboom:
We (well @ee7) are indeed working on adding this functionality. We hope to be releasing this sometime soon-ish.

PR #1856, opened 2021-10-10
rational-numbers: Clarify that denominator can be 0

SaschaMann:
In some languages the denominator can be 0 and therefore the track's implementation of this exercise may expect it. Stating `b != 0` in the description may be misleading if the track expects it.

### DIFF
--- a/exercises/rational-numbers/description.md
+++ b/exercises/rational-numbers/description.md
@@ -1,6 +1,6 @@
 # Description
 
-A rational number is defined as the quotient of two integers `a` and `b`, called the numerator and denominator, respectively, where `b != 0`.
+A rational number is defined as the quotient of two integers `a` and `b`, called the numerator and denominator, respectively.
 
 The absolute value `|r|` of the rational number `r = a/b` is equal to `|a|/|b|`.
 


ErikSchierboom (approved):

SaschaMann:
In some languages used for numerical maths, 0 as denominator is allowed and behaves similar to `Inf` in floats.

The track shouldn't diverge from that in a reimplementation exercise like this, imo.

wolf99:
I see two sides

On one side, this seems innocuous enough.

On the other hand, this could be a little confusing for languages where this is *not* the case.
My understanding is that 
specifications should aim to be the lowest common denominator, i.e that it can sets expectations for behavior across the widest set of languages possible. If languages wish to deviate from that for things that are specific to that language, that's fine but that language-specificness should not be upstreamed into the canonical spec.

This is my thoughts only. On the whole I don't think it matters too much as long as:

1. It is clearly explained, which the latest proposal appears to be.
2. This doesn't then lead to language-specific aspects being upstreamed into specs in general.

angelikatyborska:
> I don't think the mathematical definition is the only relevant consideration. After all, we also limit solutions to rational numbers representable by integers, and don't expect solutions to use arbitrary precision types. 

In this specific case (rational numbers) it is in the mathematical definition that they are represented by integers only.

> with behaviour similar to +/-Inf in floating point numbers

This is language-specific IMO, Elixir has no notion of infinity and you cannot divide by the floating point `0.0` either.

I'm still convinced that what you're trying to add @SaschaMann would better fit in the `instructions.append.md` file in the Julia track than here.

PR #1857, opened 2021-10-12
affine-cipher: Updates to the exercise instructions file

onpikono:
Hi, I have tried to clean up the instructions to this exercise a bit and possibly make them a bit more informative by:
* Changing the markdown text from plain to `code` for important values to let them stand out
* Adding the Encryption and Decryption section headers for improved clarity
* Adding guidance about digits and spaces as inputs to match the test cases
* Updating the Decryption formula to avoid exponent ambiguity
* Updated the MMI property formula to be more consistent with the resource (Wiki), and subjectively easier to grasp
* Adding a hint about what coprimes (or relative primes) are for improved clarity
* Removing one example of MMI which seemed to be redundant

Please review the proposed updates and let me know what you think - they are subjective so it's perfectly fine if only some / none of them will be accepted. Thank you for your time!

PR #1858, opened 2021-10-13
Amend unclear hyphen use

rneilsen:
Two hyphens in a row, one acting as a dash and one as a minus sign, was confusing. Amend to use English word 'to' and parens to make the subtraction clearer.

### DIFF
--- a/exercises/affine-cipher/description.md
+++ b/exercises/affine-cipher/description.md
@@ -13,7 +13,7 @@ because it has many more keys.
 The encryption function is:
 
   `E(x) = (ax + b) mod m`
-  -  where `x` is the letter's index from 0 - length of alphabet - 1
+  -  where `x` is the letter's index from 0 to (length of alphabet - 1)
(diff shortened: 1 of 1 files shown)


RNeilsen:
It does, although I would argue that my way of doing it is a little clearer than the way in that PR:

Original: `where `x` is the letter's index from 0 - length of alphabet - 1`
Mine: `where `x` is the letter's index from 0 to (length of alphabet - 1)`
[That PR](https://github.com/exercism/problem-specifications/pull/1857/files): `where `i` is the letter's index from `0` to the length of the alphabet - 1`

However I'm extremely new to this (this is my second PR ever!) so not sure what correct protocol or procedure is here!

RNeilsen:
I'll just close this one and do a new one, since it looks like they also changed the indexing letter so this version would need modification anyway (and also I don't know what rebasing or force-pushing are).

ErikSchierboom:
Sure!

Issue #1844, opened 2021-09-19
Crypto square in c# track : can't get the same result following the spec


vjau:
In test ```Number_8_character_plaintext_results_in_3_chunks_the_last_one_with_a_trailing_space``` of the ```crypto square``` problem in the c# track.
We work with the following string : 

```var plaintext = "Chill out.";```

normalized, it gives ```chillout``` which is a string of length 8.
Specification says 
```
The plaintext should be organized in to a rectangle. The size of the rectangle (r x c) should be decided by the length of the message, such that c >= r and c - r <= 1, where c is the number of columns and r is the number of rows.
```
Following this spec, i get a c of 4 and a r of 3.
Which gives these segments 
```
"chil"
"lout"
""
Specification then says 
```
The coded message is obtained by reading down the columns going left to right.

Which gives following string : 
```clhoiult```

BUT the test says that you should get a result strings of 
```var expected = "clu hlt io ";```
Which correspond to encoded string of 
```"cluhltio"```
Which is not at all what i'm getting following spec.
So at this point the problem is either that i have had a bad understanding of the specification, so perhaps it should be improved to be easier to grasp, or the test is wrong or the specification is wrong.

Thank you

siebenschlaefer:
You're right, I think there's something missing in the instructions.

> The plaintext should be organized in to a rectangle.

Your rectangle `c`*`r` is `4`3`.  Check.

> [...] such that c >= r

`4 >= 3` Check.

> [...] and c - r <= 1

`4 - 3 <= 1` Check.

IMHO there should be a requirement for `c` to be as small as possible. Because there's a smaller rectangle `3 * 3` that would satisfy the other requirements, too, and then you solve the exercise, right?

vjau:
Right !

PR #1859, opened 2021-10-13
Crypto square: mention that the task is to minimize c

angelikatyborska:
Closes https://github.com/exercism/problem-specifications/issues/1844

This adds the keywords "smallest possible integer" that were missing in know exactly how to find `c`. I had to rephrase the whole description a bit because I had no idea how to fit those extra keywords in the original one in a way that would sound good to me.

### DIFF
--- a/exercises/crypto-square/description.md
+++ b/exercises/crypto-square/description.md
@@ -23,15 +23,14 @@ is normalized to:
 ```
 
 The plaintext should be organized in to a rectangle. The size of the
-rectangle (`r x c`) should be decided by the length of the message.
+rectangle should be decided by the length of the message.
 
-Find the smallest possible integer `c` such that:
+If `c` is the number of columns and `r` is the number of rows, then for
+the rectangle `r x c` find the smallest possible integer `c` such that:
 
 - `r x c >= length(message)`,
 - and `c >= r`,
 - and `c - r <= 1`.
-
-`c` is the number of columns and `r` is the number of rows in the rectangle. 
 
 Our normalized text is 54 characters long, dictating a rectangle with
 `c = 8` and `r = 7`:


exercises/crypto-square/description.md
```diff
@@ -22,10 +22,16 @@ is normalized to:
...
+The plaintext should be organized in to a rectangle. The size of the
+rectangle (`r x c`) should be decided by the length of the message.
+
+Find the smallest possible integer `c` such that:
+
+- `r x c >= length(message)`,
+- and `c >= r`,
+- and `c - r <= 1`.
+
+`c` is the number of columns and `r` is the number of rows in the rectangle.
```
wolf99, line 34: Perhaps this line could go  earlier, before the usages of `c` and `r`.
Maybe somewhere around line # 26?

ErikSchierboom (approved):

exercises/crypto-square/description.md
```diff
@@ -22,10 +22,15 @@ is normalized to:
...
-rectangle (`r x c`) should be decided by the length of the message,
-such that `c >= r` and `c - r <= 1`, where `c` is the number of columns
-and `r` is the number of rows.
+The plaintext should be organized in to a rectangle. The size of the
+rectangle should be decided by the length of the message.
+
+If `c` is the number of columns and `r` is the number of rows, then for
+the rectangle `r x c` find the smallest possible integer `c` such that:
+
+- `r x c >= length(message)`,
```
SaschaMann, line 31: What is `x` here?

SaschaMann (approved):

exercises/crypto-square/description.md
siebenschlaefer: I guss it's a multiplication sign. And if you think that's not necessarily obvious I would agree.

angelikatyborska: Yeah, it's multiplication. It was used as "by" ("r by c") to express the dimensions of the square, and here I used it as multiplication. Should I change both to `*`, or just this one? I remember we wanted to be consistent with the symbol usage thorough the whole repo, but I'm not sure which one is the preferred one.

SaschaMann: I would either use `*` or `⋅`. This could also be understood as`r times x times c`.

Probably belongs in a separate PR/issue, I don't mean to block this PR because of this.

> I remember we wanted to be consistent with the symbol usage thorough the whole repo, but I'm not sure which one is the preferred one.

I'm not sure either. I think ideally we'd have https://github.com/exercism/website/issues/476 and use proper notation instead of these workarounds.

Issue #1852, opened 2021-09-29
[Complex Numbers]: Proposal to Test mixed operations (complex number with real number)


dawidl022:
Hello, after having spoken to my mentor on exercism, I thought I would contribute to the tests to make them check operations such as (1+2i) - 2, as currently they only check operations on two complex numbers. I am not sure if the creators of this exercise want this complexity though, so that's why I'm asking here beforehand.

PR #1860, opened 2021-10-17
Add test cases for operations between complex and real numbers

dawidl022:
As discussed here: https://github.com/exercism/problem-specifications/issues/1852

Added one test for each of the 4 basic operations: addition, subtraction, multiplication and division. The first operand is a complex number and the second is an integer.

### DIFF
--- a/exercises/complex-numbers/canonical-data.json
+++ b/exercises/complex-numbers/canonical-data.json
@@ -446,6 +446,16 @@
             "z2": 10
           },
           "expected": [1, 10]
+        },
+        {
+          "uuid": "9a867d1b-d736-4c41-a41e-90bd148e9d5e",
+          "description": "Divide real number by complex number",
+          "property": "div",
+          "input": {
+            "z1": 5,
+            "z2": [1, 1]
+          },
+          "expected": [2.5, -2.5]
         }
       ]
     }
--- a/exercises/complex-numbers/canonical-data.json
+++ b/exercises/complex-numbers/canonical-data.json
@@ -365,7 +365,7 @@
       ]
     },
     {
-      "description": "Operations between real numbers (int/float) and complex numbers",
+      "description": "Operations between real numbers and complex numbers",
       "cases": [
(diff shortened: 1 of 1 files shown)


exercises/complex-numbers/canonical-data.json
```diff
@@ -363,6 +363,51 @@
           "expected": [-2, 0]
         }
       ]
+    },
+    {
+      "description": "Operations between real numbers (int/float) and complex numbers",
```
SaschaMann, line 368:
```suggestion
      "description": "Operations between real numbers and complex numbers",
```

Number types vary between languages, so I think it's best to stick to `real numbers` here and let the tracks map that to whatever type fits their language.

SaschaMann (commented):
Thanks for adding these!

I don't know if it makes a difference for any tracks but it might be useful to have the same tests in reverse too, so subtracting a complex number from a real number etc. Or perhaps that should be left to the implementation/track maintainers. I'm not sure.

dawidl022:
> I don't know if it makes a difference for any tracks but it might be useful to have the same tests in reverse too, so subtracting a complex number from a real number etc. Or perhaps that should be left to the implementation/track maintainers. I'm not sure.

I am honestly not sure how to implement this the other way around (in python anyways). For example I am able to get `my_complex_number * 5` to work no problem, but not `5 * my_complex_number`.

SaschaMann:
> I am honestly not sure how to implement this the other way around (in python anyways).

That sounds like a good reason to have them the other way around as a commutativity check.

dawidl022:
> IIRC you could implement `__rmul__(self, x)` (see [docs](https://docs.python.org/3/reference/datamodel.html#object.__rmul__)).

Thank you! I did not know these methods existed.

> That sounds like a good reason to have them the other way around as a commutativity check.

Yes, I agree. It would help people like me to learn a new thing along the way.

wolf99 (approved):

PR #1861, opened 2021-10-18
Do not use 'x' for multiplication

angelikatyborska:
A follow up to https://github.com/exercism/problem-specifications/pull/1859#discussion_r730428002

`x` in an equation shouldn't be used for multiplication because it could be easily mistaken for a variable named `x`.

I tried to found all usages of `x` by searching for `\d\s?x\s?\d`. All the remaining ones are not used in equations, but to give dimensions (e.g. " a 5 x 4 board") and I think there "x" still makes sense.


SaschaMann (approved):
Thanks!

PR #1863, opened 2021-10-21
Update description for sublist exercise

onpikono:
Major updates to the description for sublist exercise. I think the updated version is a bit easier to understand and contains guidance on *edge* case of empty list(s).

The updates are subjective though, so if you don't think they would be an improvement, feel free to either edit or completely discard them.

Thank you!

### DIFF
--- a/exercises/sublist/description.md
+++ b/exercises/sublist/description.md
@@ -4,10 +4,10 @@ Given any two lists `A` and `B`, determine if:
 
 - List `A` is equal to list `B`; or
 - List `A` contains list `B` (`A` is a superlist of `B`); or
-- List `B` contains list `A` (`A` is a sublist of `B`); or
+- List `A` is contained by list `B` (`A` is a sublist of `B`); or
 - None of the above is true, thus lists `A` and `B` are unequal
 
-Specifically, list `A` is equal to list `B` if both contain only the same sequence of values.
+Specifically, list `A` is equal to list `B` if both lists have the same values in the same order.
 List `A` is a superlist of `B` if `A` contains a sub-sequence of values equal to `B`.
 List `A` is a sublist of `B` if `B` contains a sub-sequence of values equal to `A`.
 
@@ -22,3 +22,4 @@ Examples:
 - If `A = [1, 2, 3]` and `B = [1, 2, 3]`, then `A` and `B` are equal
 - If `A = [1, 2, 3, 4, 5]` and `B = [2, 3, 4]`, then `A` is a superlist of `B`
 - If `A = [1, 2, 4]` and `B = [1, 2, 3, 4, 5]`, then `A` and `B` are unequal
+- If `A = [1, 2, 3]` and `B = [1, 3, 2]`, then `A` and `B` are unequal


exercises/sublist/description.md
```diff
@@ -1,18 +1,24 @@
...
-list, if the second list is contained within the first list, if both lists are
-contained within each other or if none of these are true.
+Given any two lists `A` and `B`, determine if:
 
-Specifically, a list A is a sublist of list B if by dropping 0 or more elements
-from the front of B and 0 or more elements from the back of B you get a list
-that's completely equal to A.
+* List `A` is equal to list `B`; or
+* List `A` contains list `B` (`A` is a superlist of `B`); or
+* List `B` contains list `A` (`A` is a sublist of `B`); or
```
glennj, line 7: Perhaps:
```none
* List `A` is contained by list `B` (`A` is a sublist of `B`); or
```
keeping A consistently before B would aid comprehension, I think

exercises/sublist/description.md
```diff
@@ -1,18 +1,24 @@
...
+* None of the above is true, thus lists `A` and `B` are unequal
+
+Specifically, a list `A` is equal to list `B` if both contain only the same sequence of values.
```
glennj, line 10: Do we need to specify that "sequence" means: the same values _in the same order_?

exercises/sublist/description.md
```diff
@@ -1,18 +1,24 @@
...
 
-* A = [1, 2, 3], B = [1, 2, 3, 4, 5], A is a sublist of B
-* A = [3, 4, 5], B = [1, 2, 3, 4, 5], A is a sublist of B
-* A = [3, 4], B = [1, 2, 3, 4, 5], A is a sublist of B
-* A = [1, 2, 3], B = [1, 2, 3], A is equal to B
-* A = [1, 2, 3, 4, 5], B = [2, 3, 4], A is a superlist of B
-* A = [1, 2, 4], B = [1, 2, 3, 4, 5], A is not a superlist of, sublist of or equal to B
+* If `A = []` and `B = []` (both lists are empty), then `A` and `B` are equal
+* If `A = [1, 2, 3]` and `B = []`, then `A` is a superlist of `B`
+* If `A = []` and `B = [1; 2; 3]`, then `A` is a sublist of `B`
```
glennj, line 18: semicolons?

exercises/sublist/description.md
onpikono: What do you think, which version would work better?

1. ...both contain only the same values in the same order (i.e. they are the same exact sequence of values).
2. ...both are the same exact sequence of values, i.e. they only contain the same values in the same order.
3. ...both contain only the same values in the same order, i.e. the same sequence.

I would probably lean towards no. 2.

exercises/sublist/description.md
onpikono: @wolf99 I would lean towards my original suggestion, if it was up to me. Just don't want to waste too much of your time :).

wolf99: I don't feel that either should block this PR from being merged.
Just a counterpoint to the suggestion from @glennj.
Perhaps Glenn say if my argument has swayed his opinion; if it has not I'm OK if you go with his suggestion if you agree.

exercises/sublist/description.md
wolf99: 2 looks good to me

onpikono (commented):
All seems fine from my point of view

glennj:
I'd prefer 2 PR's for changes to 2 exercises.

PR #1875, opened 2021-10-31
Add left, right, up, up case to zipper canon

### LINKED ISSUE #1872: Zipper needs a test to go deep and return to the top
As currently written, the Zipper problem specification doesn't require testing a key feature of the zipper structure, which is the ability to navigate all the way back to the top of the tree from any depth. I recently completed this exercise on the Ruby track, and my first iteration passed all tests even though it did not properly implement this feature.

Operations for this new case would be `left`, `right`, `up`, `up`, `value`.

I've opened a PR on the Ruby track that would implement this case: https://github.com/exercism/ruby/pull/1261 to address the issue in that track: https://github.com/exercism/ruby/issues/1260

moveson:
This PR adds a case to the Zipper canonical data. The new case checks whether the zipper implementation can find its way back to the top node from deep within the tree.

Resolves #1872

### DIFF
--- a/exercises/zipper/canonical-data.json
+++ b/exercises/zipper/canonical-data.json
@@ -281,7 +281,7 @@
     },
     {
       "uuid": "b9aa8d54-07b7-4bfd-ab6b-7ff7f35930b6",
-      "description": "left, right, up, and up",
+      "description": "test ability to descend multiple levels and return",
       "property": "expectedValue",
       "input": {
         "initialTree": {
--- a/exercises/zipper/canonical-data.json
+++ b/exercises/zipper/canonical-data.json
@@ -280,6 +280,51 @@
       }
     },
     {
+      "uuid": "b9aa8d54-07b7-4bfd-ab6b-7ff7f35930b6",
+      "description": "left, right, up, and up",
+      "property": "expectedValue",
+      "input": {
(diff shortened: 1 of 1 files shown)


glennj:
If you use "fixes" or "resolves" instead of "addresses", then the issue is automatically linked to the PR.

exercises/zipper/canonical-data.json
```diff
@@ -279,6 +279,51 @@
         "value": 3
       }
     },
+    {
+      "uuid": "b9aa8d54-07b7-4bfd-ab6b-7ff7f35930b6",
+      "description": "left, right, up, and up",
```
wolf99, line 284: Would be good for the description to say what aspect is being tested, rather than just list the operations.

(I realise that _some_ other test descriptions in this file follow the same format, but no reason to continue a less clear approach)

moveson: Good idea. Done.

wolf99 (approved):

petertseng (approved):
OK. case matches existing schema (I mean the exercise-specific schema) and the answer is correct.

ErikSchierboom (approved):

ErikSchierboom:
Thanks @moveson!

Issue #1873, opened 2021-10-30
[Grep] Discrepancy between instructions and tests about output format


vjau:
The instructions says ```Your task is to implement the `grep` function, which should read the contents
of the specified files, find the lines that match the specified pattern
and then output those lines as a single string.```
By single string, i understand that if there are multiple lines matching the pattern, those lines should be concatenated to a single string. 
However, at least in Rust, a test for multiple matches is written like this : 
```rust
test_one_file_several_matches_no_flags(
        pattern = "may",
        flags = [],
        files = ["midsummer_night.txt"],
        expected = [
            "Nor how it may concern my modesty,",
            "But I beseech your grace that I may know",
            "The worst that may befall me in this case,"
        ]
    )
```
Where matching lines are not concatenated.
I don't know if this discrepancy is only found in Rust or in all languages that includes the Grep exercise.

wolf99:
I am not very familiar with this exercise. However IIRC, grep (the tool) operates per the description you have quoted. Thus I would imagine that that was the intended result. I.E. that the output strings are concatenated.

glennj:
I'm guessing some tracks split the output into a list of lines mostly to make the test suite is easier to read.

vjau:
> I'm guessing some tracks split the output into a list of lines mostly to make the test suite is easier to read.

The problem with the Rust track is that the test fails because it's really expecting to get a vector of strings. 
I guess this is more a problem with the rust track than with the meta specification.

vjau:
Another problem i see it that the specs never describe how the output should be formatted when greping through multiple files.
However the tests (at least in Rust and Javascript), seems to imply that the name of the file where the pattern was found should be a prefix of the result.

wolf99:
Possibly the original implementer on the rust track is quite familiar with the grep tool and implemented the exercise to closely match the tool.
Having read the exercise `description.md`, it seems that the expected output is fairly well specified.

Maybe someone more familiar with either, or both, the tool and the exercise could suggest some addition to the description text that would indicate if the exercise is intended to fully mimic the tool or not?

petertseng:
I don't see a reason to force tracks to pick one or the other (one string of multiple lines, or list of lines).

vjau:
> I don't see a reason to force tracks to pick one or the other (one string of multiple lines, or list of lines).

It's not about forcing tracks. But the tests should follow the specification.

glennj:
[...]

Issue #1850, opened 2021-09-22
hello-world: description forces implementation


ErikSchierboom:
The [description.md file](https://github.com/exercism/problem-specifications/blob/main/exercises/hello-world/description.md) of the `hello-world` exercise contains the following text:

> Write a function that returns the string "Hello, World!"

While this may seem innocuous, not all tracks are using a function to return the string "Hello, World!". 

Case in point: [F#](https://github.com/exercism/fsharp/blob/main/exercises/practice/hello-world/HelloWorld.fs)

While we _could_ make F# use a function, I'd rather not to as it adds unnecessary overhead. 

Apart from that, some tracks only have methods (which are a special type of functions), so the terminology seems wrong there too. 

Maybe we could come up with an alternative description?

petertseng:
When we removed the language of "Write a program to do X" in https://github.com/exercism/problem-specifications/issues/321 we considered alternatives:

* Write a function to do X
* Write code to do X
* Do X

It seems that in most cases, "Do X" was used in https://github.com/exercism/problem-specifications/pull/528

If we were to take that approach here, then perhaps we're looking at something like...

> Produce the string "Hello, World!"

It's not easy to come up with a lowest common denominator wording, that's for sure.

petertseng:
EBWODP:

> Do X

 (to clarify, the verb differs based on what the exercise is; it's not literally the word "do"; often it will be "implement X", etc.)

petertseng:
EBWODP: 

Here's another idea, based on the fact that (to my knowledge) we've generally agreed that stubs should be provided:

> Modify the provided code so that it produces the string "Hello, World!"

ErikSchierboom:
> Here's another idea, based on the fact that (to my knowledge) we've generally agreed that stubs should be provided:

> > Modify the provided code so that it produces the string "Hello, World!"

I really like this suggestion. We have indeed mandated stubs, so this fits with that nicely.

peterloane:
Would the following wording be clearer?

Change
> Modify the provided code so that it produces the string "Hello, World!"
to
Modify the provided code so that it displays the string "Hello, World!"

glennj:
If the language wants the string _returned_, that's not really "displaying". "Producing" is more apropos.

peterloane:
Yes, displaying is not an improvement.

Producing just seems a bit opaque to me here.
Would "output" be better? 

Modify the provided code so that it outputs the string "Hello, World!"

[...]

PR #1876, opened 2021-11-02
grep: remove mentions of "output" and "single string"

petertseng:
The type implied by "single string" is not part of the essence of this
problem. Some tracks wish to ask for an ordered collection of lines,
rather than a single string containing the lines, and it seems that
there is no reason for the specification to forbid this.

The specific word "output" could also be misunderstood to mean "print on
standard output", though we're not yet currently aware of any track
tests that do indeed require this.

Closes https://github.com/exercism/problem-specifications/issues/1873

### DIFF
--- a/exercises/grep/description.md
+++ b/exercises/grep/description.md
@@ -12,11 +12,11 @@ The `grep` command takes three arguments:
 2. Zero or more flags to customize the matching behavior.
 3. One or more files in which to search for matching lines.
 
-Your task is to implement the `grep` function, which should read the contents
-of the specified files, find the lines that match the specified pattern
-and then output those lines as a single string. Note that the lines should
-be output in the order in which they were found, with the first matching line
-in the first file being output first.
+Your task is to implement the `grep` function: given a list of files, find all
+lines that match the specified pattern.
+Return the lines in the order they appear in the files.
+You'll also have to handle options (given as flags), which control how matching
+is done and how the results are to be reported.
 
 As an example, suppose there is a file named "input.txt" with the following contents:
 
@@ -26,7 +26,7 @@ world
 hello again
 ```
 
-If we were to call `grep "hello" input.txt`, the returned string should be:
+If we were to call `grep "hello" input.txt`, the result should be:
 
 ```text
 hello


exercises/grep/description.md
```diff
@@ -13,10 +13,9 @@ The `grep` command takes three arguments:
 3. One or more files in which to search for matching lines.
 
 Your task is to implement the `grep` function, which should read the contents
-of the specified files, find the lines that match the specified pattern
-and then output those lines as a single string. Note that the lines should
-be output in the order in which they were found, with the first matching line
-in the first file being output first.
+of the specified files and find the lines that match the specified pattern.
+The lines should be in the order in which they were found, with the first
+matching line in the first file being first.
```
petertseng, line 18: Check for clarity of this sentence. Perhaps it may not be clear that it's referring to the order of lines that the tested `grep` function returns (or other appropriate verb). Though I'm not quite sure what it might be misinterpreted to mean.

petertseng, line 18: > with the first matching line in the first file being first

This could be clearer. It doesn't say anything about subsequent files or subsequent lines!

ErikSchierboom (approved):
This is already a good improvement! You've raised interesting considerations.

exercises/grep/description.md
glennj: Perhaps
> Your task is to implement the `grep` function: given a list of files, find all lines that match the specified pattern. Return the lines in the order they appear in the files. You'll also have to handle options, which control how matching is done, and how the results are to be reported.

wolf99: I like that @glennj  suggestion includes "return". In the current change it's not clear what the implementation is supposed to do with the lines it finds.

petertseng: I can agree. I will make this change now. I had to think about why this was different from https://github.com/exercism/problem-specifications/issues/1850 (hello-world). The difference there is that some tracks won't necessarily implement hello-world as a function, whereas I think with most tracks grep will be a function, so we are fine here.

petertseng: Agreed and implemented suggestion, with thanks.

I had to think about why this was different from https://github.com/exercism/problem-specifications/issues/1850 (`hello-world`). I now see that the difference is that some tracks won't necessarily implement `hello-world` as a function so we didn't want to use the word **return**, whereas I think with most tracks `grep` will be a function, so we are fine to use the word **return** for `grep`. I don't think there will be objections to that.

wolf99 (approved):

glennj (approved):

PR #1880, opened 2021-11-05
Add reviewing info

SaschaMann:
For context, watch the weekly call from today.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Pull Requests to this repo require three approving reviews to be merged.
 - Please also use the **Request Changes** button if you are unsure about how a concern of yours could be resolved.
   - This will block the PR from being merged until the change request has been addressed / overruled by an admin.
   - It ensures that your thoughts will be heard and can be addressed rather than being overruled by a majority.
-- When a discussion goes stale or there are several opposing views and a compromise cannot be reached, ping [@ErikSchierboom](https://github.com/ErikSchierboom/) to resolve the discussion.
+- When a discussion goes stale or there are several opposing views and a compromise cannot be reached, ping [@exercism/maintainers-admin](https://github.com/orgs/exercism/teams/maintainers-admin/members) to resolve the discussion.
 
 ## Exercise Metadata
 
--- a/README.md
+++ b/README.md
@@ -6,6 +6,16 @@ Shared metadata for Exercism exercises.
 
 Please see the [contributing guide](./CONTRIBUTING.md).
 
+## Reviewing Guidelines
+
+Pull Requests to this repo require three approving reviews to be merged.
+
+- If you think that a PR should not be merged, either as is or in general, please leave a review **requesting changes**.
+  - This will block the PR from being merged until the change request has been addressed or overruled by an admin.
+  - Please also use the Request Changes button if you are unsure about how a concern of yours could be resolved.
+  - This ensures that your thoughts will be heard and can be addressed rather than being overruled by a majority.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,6 +6,16 @@ Shared metadata for Exercism exercises.
...
 Please see the [contributing guide](./CONTRIBUTING.md).
 
+## Reviewing Guidelines
+
+Pull Requests to this repo require three approving reviews to be merged.
+
+- If you think that a PR should not be merged, either as is or in general, please leave a review **requesting changes**.
+  - This will block the PR from being merged until the change request has been addressed or overruled by an admin.
+  - Please also use the Request Changes button if you are unsure about how a concern of yours could be resolved.
+  - This ensures that your thoughts will be heard and can be addressed rather than being overruled by a majority.
```
SleeplessByte, line 16: This list has items which are not of equal "list-indentness" in terms of information. I don't know how to solve, but right now it seems wrong 😅

SleeplessByte (changes requested):
See inline

README.md
BethanyG: How about... Might be too much, but nested does seem a bit better for which points are "main" vs "follow-on"

-----

Pull Requests to this repo require three approving reviews to be merged.

- If you think that a PR should not be merged, either as is or in general, please leave a review **requesting changes**.
- Please also use the **Request Changes** button if you are unsure about how a concern of yours could be resolved.
    - This will block the PR from being merged until the change request has been addressed / overruled by an admin.

    - It ensures that your thoughts will be heard and can be addressed rather than being overruled by a majority.
    - When a discussion goes stale or there are several opposing views and a compromise cannot be reached, ping [@ErikSchierboom](https://github.com/ErikSchierboom/) to resolve the discussion.

SaschaMann: > When a discussion goes stale or there are several opposing views and a compromise cannot be reached, ping @ErikSchierboom to resolve the discussion.

This point is not linked to change requests, imo, so it should be on the higher level. Otherwise that change would be fine with me.

SleeplessByte (approved):

SleeplessByte:
Can you hear me now GIF

I feel ~~hurt~~ heard.

iHiD (approved):
Thanks for this :)

README.md
```diff
@@ -6,6 +6,16 @@ Shared metadata for Exercism exercises.
...
 
+## Reviewing Guidelines
+
+Pull Requests to this repo require three approving reviews to be merged.
+
+- If you think that a PR should not be merged, either as is or in general, please leave a review **requesting changes**.
+- Please also use the **Request Changes** button if you are unsure about how a concern of yours could be resolved.
+  - This will block the PR from being merged until the change request has been addressed / overruled by an admin.
+  - It ensures that your thoughts will be heard and can be addressed rather than being overruled by a majority.
+- When a discussion goes stale or there are several opposing views and a compromise cannot be reached, ping [@ErikSchierboom](https://github.com/ErikSchierboom/) to resolve the discussion.
```
iHiD, line 17: Let's make it the maintainers-admin group to futureproof having other people in that alongside Erik, and also so things get to me if he's on holiday etc. I don't think there's a great link for that, so I've removed it, but I have no problem with some link being added back if appropriate.

```suggestion
- When a discussion goes stale or there are several opposing views and a compromise cannot be reached, ping @exercism/maintainers-admin to resolve the discussion.
```

SaschaMann: I've added a link to the team page

BethanyG (approved):
I like it.  😄

PR #1881, opened 2021-11-05
[POV] Change description from graph to tree

### LINKED ISSUE #1874: [POV] Graph vs. Tree confusion
The exercise [description](https://github.com/exercism/problem-specifications/blob/main/exercises/pov/description.md) consistently uses the term `graph` and only mentions `family tree` as an example. However, the [canonical test data](https://github.com/exercism/problem-specifications/blob/main/exercises/pov/canonical-data.json#L5) uses the term `tree` and the test cases only contain trees. Most languages that implemented this exercise also already call the data structure `Tree` not `Graph` from what I could see.

A graph is defined as an arbitrary set of nodes and edges that connect two nodes: https://en.wikipedia.org/wiki/Graph_(discrete_mathematics)
A tree is a special case of graph where any two nodes are connected by exactly one path (no cycles, no "loose" nodes, [etc](https://en.wikipedia.org/wiki/Tree_(graph_theory)#Tree)): https://en.wikipedia.org/wiki/Tree_(graph_theory)

I think the exercise of re-rooting/re-parenting does make the most sense in the context of trees, not in the context of generic graphs (even though the concept of e [rooted graph](https://en.wikipedia.org/wiki/Rooted_graph) exists apparently).

Sidenote: There was already confusion about something similar before (https://github.com/exercism/problem-specifications/issues/377).

My suggestion how to address this issue would be to change the description to use the term `tree` with short explanation what a tree is and a like to the Wikipedia page.

junedev:
Fixes #1874. Please see issue description for more details.

Note that this PR focuses on the issue linked above. (I think the description has some more unrelated issues.)

### DIFF
--- a/exercises/pov/description.md
+++ b/exercises/pov/description.md
@@ -1,8 +1,12 @@
 # Description
 
-Reparent a graph on a selected node.
+Reparent a tree on a selected node.
 
-This exercise is all about re-orientating a graph to see things from a different
+A [tree][wiki-tree] is a special type of [graph][wiki-graph] where all nodes
+are connected but there are no cycles. That means, there is exactly one path to
+get from one node to another for any pair of nodes.
+
+This exercise is all about re-orientating a tree to see things from a different
 point of view. For example family trees are usually presented from the
 ancestor's perspective:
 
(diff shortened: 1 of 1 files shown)


BethanyG (approved):
I think this solves the issue quite nicely.  Having recently worked on this exercise for the Python track where everything is referred to as a `tree`, this description is much better.  😄

SleeplessByte (approved):

petertseng (approved):

Issue #1640, opened 2020-02-19
Scale Generator task: wording and tests


b10s:
1. It is difficult to understand which part of the sentence the word `major` refers to. I have feeling there is missed period or new line:
![Selection_152](https://user-images.githubusercontent.com/4618288/71672434-464ef200-2db9-11ea-8365-13283af5dc3b.png)
_**How could we interpret the `major` word in the aforementioned sentences?**_

2. Test cases have interval pattern e.g. `MmMMmMM` as input. To build scale with of 7 intervals we need 8 points, as per my naive understanding. However test expects only 7 points:

   ```
   46     description: "Minor scale with sharps",                                      
   47     tonic:       "f#",                                                           
   48     interval:    "MmMMmMM",                                                      
   49     expected:    []string{"F#", "G#", "A", "B", "C#", "D", "E"},                 
   50   },
   ```
   Thus we can ignore last interval and will pass all tests.
What says wiki somehow confirms my theory about 8 points in a scale to represent 7 intervals:
![Selection_154](https://user-images.githubusercontent.com/4618288/71673130-3fc17a00-2dbb-11ea-889c-00b3905dd8ac.png)
_**In solution, how would you expect us to interpret the last interval character in an interval pattern?**_

yawpitch:
I'm absolutely _not_ a musically-minded person, and I agree that the wording there is _terribly_ confusing, but if I understand it correctly than it should _probably_ be represented and understood as a table:

|Sharps/Flats | Major | Minor|
|--|--|--|
|Neither | C | a|
|Sharps| G, D, A, E, B, F# | e, b, f#, c#, g#, d#|
|Flats | F, Bb, Eb, Ab, Db, Gb | d, g, c, f, bb, eb |

Unfortunately I will **have** to defer to someone with more knowledge of the subject to say if that actually _helps_ make it easier to understand.

As for the notation: the major and minor scales are heptatonic; there are 7 distinct notes, 1 and 8 in the notation described in the wikipedia article are both the same tonic note (albeit 1 octave apart). So they essentially represent modular wrapping around back to the front of the list of distinct notes.

So the given example represents the individual notes found in the minor scale (with sharps) starting (and thus ending) with the tonic of f#:

|F# | G# | A | B | C# | D | E |
|--|--|--|--|--|--|--|
| 1 | 2 | b3 | 4 | 5 | b6 | 7 |
| 8 | | | | | | |

glennj:
[...]

PR #1884, opened 2021-11-16
scale-generator: add clear directives

dzegel:
The instructions are very unclear about what the task at hand requires so I'm adding clear directives for the student to follow.

See https://github.com/exercism/problem-specifications/issues/1640

### DIFF
--- a/exercises/scale-generator/description.md
+++ b/exercises/scale-generator/description.md
@@ -38,8 +38,8 @@ Given a tonic, generate the 12 note chromatic scale starting with the tonic.
 starting with the given tonic.
 - For the given tonic, determine if the scale is to be returned with flats
 or sharps.
-- Return the major scale (i.e. uppercase notes) irrespective of if the
-given tonic is a major or minor note (i.e. uppercase or lowercase respectively).
+- Return all notes in uppercase letters (except fo the `b` for flats)
+irrespective of the casing of the given tonic.
 
 ## Diatonic Scales
 
--- a/exercises/scale-generator/description.md
+++ b/exercises/scale-generator/description.md
@@ -64,6 +64,6 @@ intervals, but they will not figure into this exercise.
 Given a tonic and a set of intervals, generate the musical scale starting with
 the tonic and following the specified interval pattern.
 
-This is similar generating chromatic scales except that instead of returning 12
-notes you need to return one note for each interval in the given pattern
+This is similar to generating chromatic scales except that instead of returning
+12 notes you need to return one note for each interval in the given pattern by
 skipping the number of notes indicated by each interval in the pattern.
--- a/exercises/scale-generator/description.md
+++ b/exercises/scale-generator/description.md
@@ -1,26 +1,24 @@
 # Description
 
-Given a tonic, or starting note, and a set of intervals, generate
-the musical scale starting with the tonic and following the
-specified interval pattern.
+## Chromatic Scales
 
 Scales in Western music are based on the chromatic (12-note) scale. This
 scale can be expressed as the following group of pitches:
 
-A, A#, B, C, C#, D, D#, E, F, F#, G, G#
+> A, A#, B, C, C#, D, D#, E, F, F#, G, G#
 
 A given sharp note (indicated by a #) can also be expressed as the flat
 of the note above it (indicated by a b) so the chromatic scale can also be
 written like this:
 
-A, Bb, B, C, Db, D, Eb, E, F, Gb, G, Ab
+> A, Bb, B, C, Db, D, Eb, E, F, Gb, G, Ab
 
 The major and minor scale and modes are subsets of this twelve-pitch
 collection. They have seven pitches, and are called diatonic scales.
 The collection of notes in these scales is written with either sharps or
-flats, depending on the tonic. Here is a table indicating whether the 
-flat expression or sharp expression of the scale would be used for a given
-note:
+flats, depending on the tonic (starting note). Here is a table indicating
+whether the flat expression or sharp expression of the scale would be used for
+a given tonic:
 
 | Key Signature | Major                 | Minor                |
 | ------------- | --------------------- | -------------------- |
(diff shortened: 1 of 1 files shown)


SaschaMann (commented):
These are track-/language-specific tasks. They should go into the track repo.

The instructions in problem-specifications are generally kept very broad which allows them to be used across 50+ languages. Even something like "implement a method" or "write a program" will be too specific.

If you feel that these instructions are unclear, perhaps check out other instructions in this repo first and then try to match their style.

ErikSchierboom:
Yes, that's correct. What this PR should do is to integrate the suggestions made in https://github.com/exercism/problem-specifications/issues/1640 and try to get a version merged that is less confusing to the student.

dzegel:
I updated it to be language agnostic

exercises/scale-generator/description.md
```diff
@@ -1,26 +1,24 @@
 # Description
 
-Given a tonic, or starting note, and a set of intervals, generate
-the musical scale starting with the tonic and following the
-specified interval pattern.
```
dzegel, line 5: Seems to me we should give the lay of the land before giving the problem statement.
Additionally there are two tasks for this exercise but IMO that was not clear from this statement.

I moved the task directives down below the relevant background info and spelled out the requirements of each task.

exercises/scale-generator/description.md
```diff
@@ -1,26 +1,24 @@
...
-A, Bb, B, C, Db, D, Eb, E, F, Gb, G, Ab
+> A, Bb, B, C, Db, D, Eb, E, F, Gb, G, Ab
 
 The major and minor scale and modes are subsets of this twelve-pitch
 collection. They have seven pitches, and are called diatonic scales.
 The collection of notes in these scales is written with either sharps or
-flats, depending on the tonic. Here is a table indicating whether the 
-flat expression or sharp expression of the scale would be used for a given
-note:
+flats, depending on the tonic (starting note). Here is a table indicating
```
dzegel, line 19: this is now the first reference of the term "tonic" so it requires definition

exercises/scale-generator/description.md
```diff
@@ -1,26 +1,24 @@
...
+whether the flat expression or sharp expression of the scale would be used for
+a given tonic:
```
dzegel, line 21: "note" -> "tonic"
I think that's more explicit

wolf99:
Looks pretty clear to me but I don't know anything about music so I cannot say if it is musically correct.

dzegel:
> Looks pretty clear to me but I don't know anything about music so I cannot say if it is musically correct.

I'm also unfamiliar with these musical concepts but my agenda with this change is just to make the task easier to understand.

There is some discussion in https://github.com/exercism/problem-specifications/issues/1640 about the accuracy of the concepts and I'm leaving it to people who know something about it to hash out.

wolf99 (approved):

exercises/scale-generator/description.md
```diff
@@ -47,3 +58,12 @@ Non-diatonic scales can contain other intervals.  An "augmented second"
...
 intervals, but they will not figure into this exercise.
+
+### Task
+
+Given a tonic and a set of intervals, generate the musical scale starting with
+the tonic and following the specified interval pattern.
+
+This is similar to generating chromatic scales except that instead of returning
+12 notes you need to return one note for each interval in the given pattern by
+skipping the number of notes indicated by each interval in the pattern.
```
petertseng, line 69: Here is where I would move the sentence from above. But probably modified to something like this:

```suggestion
skipping the number of notes indicated by each interval in the pattern.

As a reminder, notes are still only ever referred to with uppercase letters.
Even though minor scales are sometimes named using lowercase letters (such as
"f# minor"), the first note of f# minor is still F# (written in uppercase), and 
all other notes in the scale are also written in uppercase.
```

exercises/scale-generator/description.md
```diff
@@ -32,6 +30,19 @@ Note that by common music theory convention the natural notes "C" and "a"
...
+### Task
+
+Given a tonic, generate the 12 note musical scale starting with the tonic.
+
+- Shift the base scale appropriately so that all 12 notes are returned
+starting with the given tonic.
+- For the given tonic, determine if the scale is to be returned with flats
+or sharps.
+- Return the major scale (i.e. uppercase notes) irrespective of if the
+given tonic is a major or minor note (i.e. uppercase or lowercase respectively).
```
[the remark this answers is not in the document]
dzegel: being that there are `Interval()` tests with lowercase input and it seems to me the lack of such tests for `Chromatic()` is an oversight (or maybe as you point out the existence of such tests at all is the oversight) so lowercase needs to be accounted for. When you get to the `Interval()` task the only additional logic you should need to account for is the interval, `Chromatic()` should just be a special case of `Interval()` with a specific interval pattern so I think it makes sense to mention it here.

That said I can update the language to avoid the major/minor terminology (i clearly don't know anything about it LOL).

exercises/scale-generator/description.md
dzegel: as I said above IMO this second task should only introduce the complexity of the pattern input. I think if you introduce this casing thing here students might be misled to try and deal with casing differently for the two tasks.

exercises/scale-generator/description.md
```diff
@@ -32,6 +30,19 @@ Note that by common music theory convention the natural notes "C" and "a"
...
 
+### Task
+
+Given a tonic, generate the 12 note chromatic scale starting with the tonic.
+
+- Shift the base scale appropriately so that all 12 notes are returned
+starting with the given tonic.
+- For the given tonic, determine if the scale is to be returned with flats
+or sharps.
+- Return all notes in uppercase letters (except fo the `b` for flats)
```
petertseng, line 41:
```suggestion
- Return all notes in uppercase letters (except for the `b` for flats)
```

petertseng (approved):
This is good now. It's a good improvement to state clearly the task . Thanks for writing it. I'm approving it now but I do want to see that missing letter added before it gets merged.

dzegel:
@ErikSchierboom @SaschaMann any more comments?

ErikSchierboom:
Not really. It might be okay 🤷 @exercism/reviewers thoughts?

dzegel:
@ErikSchierboom can we move forward with this?
I'm happy to account for any suggestions.

ErikSchierboom (approved):
I'll approve. Maybe later on we can improve things further, but this is already a good improvement.

petertseng:
I'm actually wondering whether we might want to add back an executive summary of the task at the top - most other exercises have that! But that can be discussed later on.

petertseng:
EBWODP: Ah, no, not all tasks have an executive summary. Maybe one day we'll decide we want standards for how tasks should be stated (Should the description start with an executive summary? Should there be a bolded **Task**: somewhere in the description? etc.), but until then, it's what feels right.

PR #1885, opened 2021-11-18
queen-attack: Add test case to catch an incomplete diagonal check

shrugalic:
# TL;DR
This is an extra test case for the queen attack exercise. It catches an issue with an erroneous diagonal check that the existing tests do not catch.

The PR I submitted to the Rust track was already merged (https://github.com/exercism/rust/pull/1398). The maintainer suggested I should add this test case here as well.

This is my first PR to this particular project, and my second PR on Github ever, so I appreciate constructive feedback on how to improve.

### Details about why I think this test case might be useful
I started with a condition of
```rust
self.row + self.column == other.row + other.column
```
for the `/` diagonal check, and
```rust
self.row - self.column == other.row - other.column
```
for the `\` diagonal check.

The latter obviously fails when column > row [edit: because I chose unsigned integers as row and column, and Rust will panic on subtract with overflow]. So the code might become something like this:
```rust
if self.row >= self.column && other.row >= other.column
	self.row - self.column == other.row - other.column
else if self.row >= self.column && other.row < other.column
	self.row - self.column == other.column - other.row
else if self.row < self.column && other.row >= other.column
	self.column - self.row == other.row - other.column
else // self.row < self.column && other.row < other.column
	self.column - self.row == other.column - other.row
```

(BTW I didn't use match here on purpose, because everybody understands if/else)

Because this is … inelegant at best, one might come up with something like this:
```rust
max(self.row, self.column) - min(self.column, self.row)
	== max(other.row, other.column) - min(other.column, other.row)    
```

While It's more concise, and did solve all the previous test cases, it's still wrong. It's easy to notice on paper so I created a test case for it. :)

### DIFF
--- a/exercises/queen-attack/canonical-data.json
+++ b/exercises/queen-attack/canonical-data.json
@@ -242,7 +242,7 @@
           "expected": true
         },
         {
-          "uuid": "33ae4113-d237-42ee-bac1-e1e699c0c007",
+          "uuid": "543f8fd4-2597-4aad-8d77-cbdab63619f8",
           "description": "cannot attack with equal difference",
           "property": "canAttack",
           "input": {
--- a/exercises/queen-attack/canonical-data.json
+++ b/exercises/queen-attack/canonical-data.json
@@ -243,7 +243,7 @@
         },
         {
           "uuid": "543f8fd4-2597-4aad-8d77-cbdab63619f8",
-          "description": "cannot attack with equal difference",
+          "description": "cannot attack if falling diagonals are only the same when reflected across the longest falling diagonal",
           "property": "canAttack",
           "input": {
             "white_queen": {
--- a/exercises/queen-attack/canonical-data.json
+++ b/exercises/queen-attack/canonical-data.json
@@ -240,6 +240,26 @@
             }
           },
           "expected": true
+        },
+        {
+          "uuid": "33ae4113-d237-42ee-bac1-e1e699c0c007",
+          "description": "cannot attack with equal difference",
+          "property": "canAttack",
+          "input": {
+            "white_queen": {
+              "position": {
+                "row": 4,
+                "column": 1
+              }
+            },
+            "black_queen": {
+              "position": {
+                "row": 2,
+                "column": 5
+              }
+            }
+          },
+          "expected": false
         }
       ]
     }


shrugalic:
I just noticed that I copy/pasted the UUID from the other cannot attack case, without chaning it. I suppose I should have added a random new one? Here's one: `543f8fd4-2597-4aad-8d77-cbdab63619f8`

SleeplessByte:
Hi there!

Yes, each test case requires a unique UUID.

shrugalic:
> Yes, each test case requires a unique UUID.

Hi, thanks, good thing I added one in the second commit. According to the search box it's unique across GitHub.

exercises/queen-attack/canonical-data.json
```diff
@@ -240,6 +240,26 @@
             }
           },
           "expected": true
+        },
+        {
+          "uuid": "543f8fd4-2597-4aad-8d77-cbdab63619f8",
+          "description": "cannot attack with equal difference",
```
wolf99, line 246: Not sure I understand what is expected by the test from this description. Is it possible to rephrase it in chess board terms?

siebenschlaefer:
Can you help me understand the issue?

> I started with a condition of
> ```
> self.row + self.column == other.row + other.column
> ```
> for the / diagonal check

That looks correct, both additions of the `==` compute the Manhattan distance from (row=0, col=0):
```
 0  1  2  3  4  5  6  7
 1  2  3  4  5  6  7  8
 2  3  4  5  6  7  8  9
 3  4  5  6  7  8  9 10
 4  5  6  7  8  9 10 11
 5  6  7  8  9 10 11 12
 6  7  8  9 10 11 12 13
 7  8  9 10 11 12 13 14
```

> [...] and
> ```
> self.row - self.column == other.row - other.column
> ```
> for the \ diagonal check.

Looks correct, too, both subtractions compute the Manhattan distance from (row=0, col=7), minus 7:
```
 0 -1 -2 -3 -4 -5 -6 -7
+1  0 -1 -2 -3 -4 -5 -6
+2 +1  0 -1 -2 -3 -4 -5
+3 +2 +1  0 -1 -2 -3 -4
+4 +3 +2 +1  0 -1 -2 -3
+5 +4 +3 +2 +1  0 -1 -2
+6 +5 +4 +3 +2 +1  0 -1
+7 +6 +5 +4 +3 +2 +1  0
```

> The latter obviously fails when either column > row.

I don't understand that. Is that specific to the Rust track? For which cases does the latter fail?

> So the code might become something like this:
> ```
> if self.row >= self.column && other.row >= other.column
> 	self.row - self.column == other.row - other.column
> else if self.row >= self.column && other.row < other.column
> 	self.row - self.column == other.column - other.row
> else if self.row < self.column && other.row >= other.column
> 	self.column - self.row == other.row - other.column
> else // self.row < self.column && other.row < other.column
> 	self.column - self.row == other.column - other.row
> ```

I think something got lost during that transformation. Now three of the tests fail:
- Test "can attack on first diagonal" with the white queen on (2, 2) and the black queen on (0, 4)
    It matches the second condition (`2 >= 2 && 0 < 4`) and returns false (`2 - 2` == `4 - 0`)
- Test "can attack on second diagonal" with the white queen on (2, 2) and the black queen on (3, 1)
    It matches the first condition (`2 >= 2 && 3 >= 1`) and returns false (`2 - 2` == `3 - 1`)
- Your proposed additional test "cannot attack with equal difference" with the white queen on (4, 1) and the black queen on (2, 5)
    It matches the second condition (`4 >= 1 && 2 < 5`) and returns false (`4 - 1` == `5 - 2`)

> one might come up with something like this:
> ```
> max(self.row, self.column) - min(self.column, self.row)
> 	== max(other.row, other.column) - min(other.column, other.row)    
> ```
> While It's more concise, and did solve all the previous test cases [...]

That still does not pass the same three tests:
- Test "can attack on first diagonal" with the white queen on (2, 2) and the black queen on (0, 4)
    `max(2, 2) - min(2, 2) == max(0, 4) - min(4, 0)`
    `2 - 2 == 4 - 0`
   `false`
- Test "can attack on second diagonal" with the white queen on (2, 2) and the black queen on (3, 1)
    `max(2, 2) - min(2, 2) == max(3, 1) - min(1, 3)`
    `2 - 2 == 3 - 1`
    `false`
- Your proposed additional test "cannot attack with equal difference" with the white queen on (4, 1) and the black queen on (2, 5)
    `max(4, 1) - min(1, 4) == max(2, 5) - min(5, 2)`
    `4 - 1 == 5 - 2`
    `false`

tl;dr: I think your first simple approach was correct, then the more complex code with the four `if` statements fails two of the existing tests and the next snippet with `max()` and `min()` fails the same tests.
Can you present an invalid solution that would pass the existing tests?

petertseng:
Please note that the additional iterations only replace the code that calculates the falling diagonal.

Thus, please note this implementation, which passes all existing test cases:

```ruby
multi_verify(json['cases'][1]['cases'], property: 'canAttack', implementations: [
  {
    name: 'incorrectly using abs to calculate falling diagonal',
    should_fail: true,
    f: ->(i, _) {
      w, b = %w(white black).map { |color|
        i["#{color}_queen"]['position'].values_at('row', 'column')
      }   
      diffs = w.zip(b).map { |wb| wb.reduce(:-).abs }
      same_rising = [w, b].map(&:sum).uniq.size == 1
      same_falling = [w, b].map { |wb| wb.reduce(:-).abs }.uniq.size == 1
      diffs.include?(0) || same_rising || same_falling
    },  
  },  
])
```

As you can see if you add that to https://github.com/petertseng/exercism-problem-specifications/blob/verify/exercises/queen-attack/verify.rb and run it, it passes all test cases, which causes an error.

```
== anti-verify incorrectly using abs to calculate falling diagonal canAttack ==
/home/probspecs/repo/verify.rb:130:in `block (2 levels) in multi_verify': incorrectly using abs to calculate falling diagonal: 0 failed, 0 errored in a test of canAttack (RuntimeError)
```

petertseng:
EBWODP: modifying the incorrect code to this (with a better variable name), but both versions still incorrectly pass all existing tests.

```ruby
multi_verify(json['cases'][1]['cases'], property: 'canAttack', implementations: [
  {
    name: 'incorrectly using abs to calculate falling diagonal',
    should_fail: true,
    f: ->(i, _) {
      w, b = %w(white black).map { |color|
        i["#{color}_queen"]['position'].values_at('row', 'column')
      }   
      same_rank_or_file = w.zip(b).any? { |wb| wb.reduce(:==) }
      same_rising = [w, b].uniq(&:sum).size == 1
      same_falling = [w, b].uniq { |wb| wb.reduce(:-).abs }.size == 1
      same_rank_or_file || same_rising || same_falling
    },  
  },  
])
```

exercises/queen-attack/canonical-data.json
petertseng: Some suggestions, ordered by length

* cannot attack if falling diagonals are only the same when reflected across the longest falling diagonal
* cannot attack if falling diagonals are on the two opposite sides of the longest falling diagonal
* cannot attack on opposite falling diagonals
    * Shorter isn't necessarily better. It is not immediately obvious what an "opposite" diagonal is, if it's not made clear that it's in relation to the longest falling diagonal.

petertseng (approved):
Since there is an incorrect implementation that passes all existing tests but fails the new one, it seems correct to add this new test. If it needs a better name, I'll put in a few ideas.

exercises/queen-attack/canonical-data.json
shrugalic: > cannot attack if falling diagonals are on the two opposite sides of the longest falling diagonal

FWIW: ^ this description is most understandable to me, and definitely a lot more descriptive than my initial one.

shrugalic:
>> The latter obviously fails when either column > row.
>
> I don't understand that. Is that specific to the Rust track? For which cases does the latter fail?

The values in the Rust track are unsigned integers, so `row - column` fails with 'attempt to subtract with overflow' whenever `column > row`.

> I think your first simple approach was correct

Not with unsigned integers.

> Can you present an invalid solution that would pass the existing tests

I did, the one  in the new test.

siebenschlaefer:
> Thus, please note this implementation, which passes all existing test cases:

Oh, I don't know Rust at all and I can't decipher the code. I leave this whole issue to others

> The values in the Rust track are unsigned integers, so `row - column` fails with 'attempt to subtract with overflow' whenever `column > row`.
>  [...]
> It does not work with unsigned integers though, which is a problem in Rust (and other languages that make this distinction).

Interesting, I'm used to languages where overflow of unsigned integer types is valid (and therefore that would work) or languages where all integers are signed.

Thanks for the reply.

shrugalic:
> Oh, I don't know Rust at all and I can't decipher the code. I leave this whole issue to others

I can't properly grok that Ruby code either. But allow me two post the complete Rust version, including more sensible implementations:
```rust
fn is_same_diagonal(&self, other: &ChessPosition) -> bool {
    // This first part is for rising diagonals
    self.row + self.column == other.row + other.column
        ||
    // and this second part is for falling diagonals (and incorrect!)
    if self.row >= self.column && other.row >= other.column {
        self.row - self.column == other.row - other.column
    } else if self.row >= self.column && other.row < other.column {
        self.row - self.column == other.column - other.row
    } else if self.row < self.column && other.row >= other.column {
        self.column - self.row == other.row - other.column
    } else {
        self.column - self.row == other.column - other.row
    }
        
    // All of the following code is for illustrative purposes only, I did not
    // comment it out only to get syntax highlighting for easier reading
        ||
    // A shorter (but equally wrong) version of the second part is this:
    max(self.row, self.column) - min(self.column, self.row)
        == max(other.row, other.column) - min(other.column, other.row)
        ||
    // A correct version of the second part would be the following, using signed 8-bit integers (i8). 
    // My bad for not using signed integers as coordinates in the first place,
    // since I figured they cannot be negative – which is true – but their diffs sure can be
    self.row as i8 - self.column as i8 == other.row as i8 - other.column as i8
        ||
    // Both parts of the diagonal check could be combined into this single check:
    (self.row as i8 - other.row as i8).abs() == (self.column as i8 - other.column as i8).abs()
}
```

> Interesting, I'm used to languages where overflow of unsigned integer types is valid (and therefore that would work) or languages where all integers are signed.

Yeah, Rust is really strict that way.

exercises/queen-attack/canonical-data.json
shrugalic: I'll expand a bit on the situation under test, to hopefully make it clearer.

You'll note the two positions (4,1) and (2,5) in the new test are on opposite sides of the longest falling diagonal – the one from the top left corner to the bottom right corner of the board.

Reflecting position (4,1) to the other side of this longest falling diagonal would make it (1,4), which is diagonally next to (2,5).

Similarly, reflecting (2,5) to the other side of this longest falling diagonal would make it (5,2), which is diagonally next to (4,1).

A buggy implementation, such as the one I listed, would falsely claim that (4,1) and (2,5) are in a diagonal, because the reflection of either point _is_ diagonal of the respective other point.

This new test makes sure that an implementation with this particular bug does not make it through.

Does this help?

Also, since I used "reflect" a lot, allow me to change my mind and now think the first suggestion is the better one:
> cannot attack if falling diagonals are only the same when reflected across the longest falling diagonal

ErikSchierboom (approved):
> Since there is an incorrect implementation that passes all existing tests but fails the new one, it seems correct to add this new test.

I agree with this.

> If it needs a better name, I'll put in a few ideas.

I must admit that the current description is lightly hard to parse for me.

ErikSchierboom:
CC @exercism/reviewers

wolf99:
> I must admit that the current description is lightly hard to parse for me.

Same here, but not a blocking issue now I think.

wolf99 (approved):

SleeplessByte:
Updating the description can be done in a later PR. This is a good addition. Thanks everyone for your input <3

PR #1887, opened 2021-11-25
fix sgf canonical-data

dzegel:
fixing typo in the test data

### DIFF
--- a/exercises/sgf-parsing/canonical-data.json
+++ b/exercises/sgf-parsing/canonical-data.json
@@ -179,7 +179,7 @@
             },
             "expected": {
                 "properties": {
-                    "A": ["]b\\nc\\nd  e \\n]"]
+                    "A": ["]b\\nc\\nd\\t\\te \\n]"]
                 },
                 "children": []
             }


exercises/sgf-parsing/canonical-data.json
```diff
@@ -179,7 +179,7 @@
             },
             "expected": {
                 "properties": {
-                    "A": ["]b\\nc\\nd  e \\n]"]
+                    "A": ["]b\\nc\\nd\\t\\te \\n]"]
```
dzegel, line 182: see line 178 there are tabs between the `d` and `e` not spaces

dzegel:
@ErikSchierboom what's the process for updating the exercise in all the various tracks after a fix like this is merged?

petertseng:
What I find odd about the test case as it currently stands:

For better or worse, it is different than the original implementation of the exercise in https://github.com/exercism/exercism/commit/7a5075bf96e2cafb29f2f883ff98b708be5594a6. The current test case is `"(;A[\\]b\\nc\\nd\\t\\te \\n\\]])"`, which is `(;A[\]b\nc\nd\t\te \n\]])`. But the original test case was `"(;A[\\]b\nc\\\nd\t\te\\\\ \\\n\\]])"` which is 

```
(;A[\]b
c\
d		e\\ \
\]])
```

In either case, let's look at what https://www.red-bean.com/sgf/sgf4.html has to say on the matter, shall we? All emphasis was added by me.

> ## 3.2. Text
>
> Text is a formatted text. **White spaces other than linebreaks are converted to space** (e.g. no tab, vertical tab, ..).
> 
> ## 3.3. SimpleText
> SimpleText is a simple string. **Whitespaces other than space must be converted to space**, i.e. there's no newline!

So, expecting spaces in the output instead of tabs is consistent with the specifications. I have not checked whether the handling of other characters is consistent with the specifications yet, but since the tabs are in question in this PR, that was the question to answer first.

dzegel:
@petertseng that's really interesting, in that case I would argue the issue is the lack of direction on handling tabs in the instructions. I don't propose that the instructions should detail every aspect of the SGF specification but being that tabs are in fact included in the canonical data it should be clarified. I'll adjust this PR to make that change

petertseng:
I'm also not sure our expectations for the `\\n` are correct (why wouldn't we write it as just `\n` in the expected output, instead of `\\n`?), but there is at least an interpretation where the current case is correct, so I won't make a big deal out of that for the time being.

dzegel:
> I'm also not sure our expectations for the `\\n` are correct (why wouldn't we write it as just `\n` in the expected output, instead of `\\n`?), but there is at least an interpretation where the current case is correct, so I won't make a big deal out of that for the time being.

yea I wondered about that as well, however the generated test file has `\n` so its being converted

kotp:
On 21/11/25 03:29PM, dzegel wrote:
> yea I wondered about that as well, however the generated test file has
> `\n` so its being converted
> 

That is correct, we are escaping the backslash so that it will show.
Otherwise it will escape the `n` and be a newline.

exercises/sgf-parsing/description.md
```diff
@@ -53,6 +53,13 @@ A key can have multiple values associated with it. For example:
 
 Here `AB` (add black) is used to add three black stones to the board.
 
+Being that parentheses, square brackets and semicolons are integral parts of
+the SGF format they must be escaped by preceding the character with a backslash
+(i.e. `\(`, `\;` and etc.) when found in text. Some other characters that
+require escaping are newlines, tabs and backslashes (`\n`, `\t`, `\\`). In
```
petertseng, line 59: These four lines are relevant to someone who wants to output a string in the SGF format. Students are not asked to do that (students are parsing, not outputting), so this information is more information for them to read which ultimately does not help them.

However, it is useful to discuss how to deal with escape sequences when parsing. So perhaps instead just a short bit like "characters preceded by a backslash (the escape character) appear in the output as-is, without the backslash" ?

Finally, if these four lines are instead kept as-is, I cannot verify their truthfulness, because they contradict what's said in https://www.red-bean.com/sgf/sgf4.html

> Escaping: "\" is the escape character. Any char following "\" is inserted verbatim (exception: whitespaces still have to be converted to space!). Following chars have to be escaped, when used in Text: "]", "\" and ":" (only if used in compose data type).

petertseng (approved):
I'm going to approve it since it's an improvement over the original, in that it tells us what to do with whitepsace, which it did not previously do.

But, there is one major way in which I think it could be better. I would appreciate if that would be considered.

exercises/sgf-parsing/description.md
dzegel: 1. I can fix the language to explain how/when escaped characters will be encountered instead of saying that chars need to be escaped.
2. I can update the content to conform to what the spec says
3. one thing I don't find clear in the spec is how to handle line-breaks, the expectation in this exercise is clearly to replace the C# string `"\\n"` with the char `\n` so I'm going to go with that for now

ErikSchierboom:
> @ErikSchierboom what's the process for updating the exercise in all the various tracks after a fix like this is merged?

There isn't one. Tracks have to do that manually. In the future, we can automatically post issues/updates to track repos whenever metadata changes.

exercises/sgf-parsing/description.md
```diff
@@ -53,6 +53,19 @@ A key can have multiple values associated with it. For example:
 
 Here `AB` (add black) is used to add three black stones to the board.
 
+According the the SGF specification `\` is the escape character. When `\` is
+encountered during parsing it indicates that the following character should be
+printed as is. The exception to this is whitespace, as is common with many
+encoding schemas the character sequences `"\n"`, `"\r"` and `"\t"` are the
+escaped forms of the newline, carriage return and space characters
+respectively. When parsing you can expect some characters to always be escaped
+including parentheses, square brackets, semicolons and colons.
```
ErikSchierboom, line 62: Maybe shorten this a bit.

```suggestion
printed as is. The exception to this is whitespace, with newline, carriage return and 
tabs encoded as `"\n"`, `"\r"` and `"\t"` respectively. When parsing you can expect some characters to _always_ be escaped, including parentheses, square brackets, semicolons 
and colons.
```

There was also a small mistake, where is said the `\t` encoded spaces.

exercises/sgf-parsing/description.md
dzegel: updated, what do you think?

ErikSchierboom (approved):

exercises/sgf-parsing/description.md
```diff
@@ -53,6 +53,18 @@ A key can have multiple values associated with it. For example:
...
+encountered during parsing, it indicates that the following character should be
+printed as is. The exception to this is whitespace, with newline, carriage
+return and tab encoded as `"\n"`, `"\r"` and `"\t"` respectively. When parsing
+you can expect some characters to _always_ be escaped, including parentheses,
+square brackets, semicolons and colons.
+
+When parsing SGF files whitespace gets special handling, with the exception of
+newlines any whitespace character encountered is replaced with a single space
+character irrespective of if the original character is found in its escaped
+from or its unescaped form.
```
petertseng, line 66:
```suggestion
form or its unescaped form.
```

exercises/sgf-parsing/description.md
```diff
@@ -53,6 +53,18 @@ A key can have multiple values associated with it. For example:
 
 Here `AB` (add black) is used to add three black stones to the board.
 
+According the the SGF specification `\` is the escape character. When `\` is
+encountered during parsing, it indicates that the following character should be
+printed as is. The exception to this is whitespace, with newline, carriage
+return and tab encoded as `"\n"`, `"\r"` and `"\t"` respectively. When parsing
+you can expect some characters to _always_ be escaped, including parentheses,
+square brackets, semicolons and colons.
```
petertseng, line 61: I'm unsure about the sentence beginning with "When parsing you can expect..." for two reasons:

1. https://www.red-bean.com/sgf/sgf4.html does not say that parentheses or semicolons will always be escaped, so if we make this claim, we are going against the specification... but that is okay! It's okay for this exercise's description to be **more restrictive** than the specification. It just means that we'll never test some inputs that would be legal, and that's fine. If we were **less restrictive** than the specification, then we would test inputs that would be illegal, and we should never do that. So if we're keeping this sentence as-is, let us just agree here that we have decided to deliberately be more restrictive than the specification, to make things easier on the students. Please confirm.
2. It's not clear to me that the student's job is made easier by knowing this. I guess it means that the student can immediately search for the next non-text character they expect to see (that would typically be `]` in the inputs we're testing), and if they find it's been escaped, they can just search for the next one, repeating until they find a non-escaped one? Sure I guess, that's fine. I suppose it doesn't hurt.

exercises/sgf-parsing/description.md
petertseng, line 66: While I think the idea of this distinction between "its escaped form or its unescaped form" contradicts https://www.red-bean.com/sgf/sgf4.html, fixing this would be a larger undertaking and also involves changing the case currently in the canonical-data.json, so I'll be silent about it for now and refrain from making that a part of this PR.

exercises/sgf-parsing/description.md
dzegel: 1. yes the specification didn't say that semicolon or parenthesis must be escaped, but I don't understand how you can practically include those chars in the text without escaping them (you may be able to so closing parens without escaping but not semicolon or open parens) so I assumed its a typo. the spec does say that colons must be escaped which I suspect is a type intending semicolon but I could be wrong there so I included colons.
2. it seems to me the student needs to know how to handle these chars when encountered in text. if in theory these chars could appear unescaped it make the task much more complex so I think this information is very helpful

exercises/sgf-parsing/description.md
dzegel: I thought it was consistent with the spec.

The spec is not explicit but it states
- "White spaces other than linebreaks are converted to space"
- "Any char following "\" is inserted verbatim (exception: whitespaces still have to be converted to space!)."
- "Following chars have to be escaped, when used in Text: "]", "\" and ":" (only if used in compose data type)."

which seems to indicate:
1. that escaped WS is interpreted as WS (so`"\t"` is a tab and not an escaped `t`) and should be replaced with a space character.
2. that whitespace does not have to be escaped so unescaped whitespace would also need to be converted to spaces (save for line break)

exercises/sgf-parsing/description.md
petertseng: Huh, sounds like they would be interesting test cases to add, then... This statement is **not** a request that the test case should be added as a part of this PR, it was just a statement in passing of what might make a good future addition.

The reason you can have an unescaped semicolon inside a Text is that Texts are only closed by the specific character `]`, so if you've encountered a `;` before a `]` you know the semicolon is part of the Text and not the semicolon that indicates a new node. Same reasoning for the other characters.

exercises/sgf-parsing/description.md
petertseng: You have already pointed out the crucial statement: "Any char following `"\"` is inserted verbatim". Therefore, the SGF parser, when seeing the sequence of characters `\t`, should not emit a tab or a space but just a `t`.

Since we have a test in the canonical-data.json that does otherwise, **our canonical-data.json is wrong**. But, we can fix that later - since the test has been wrong for years, no additional harm is caused by it remaining wrong for some time longer.

The improvement to the canonical-data.json would probably be having individual tests for some of the individual required behaviours (replace whitespace with a space, replace an escaped newline with nothing at all, all other escaped characters).

So, let's not worry about it for this PR and just stick with what's there.

dzegel: `"Any char following "\" is inserted verbatim (exception: whitespaces still have to be converted to space!)."` you need to read what they have in the parentheses, if `"\t"` was to be replaced with `"t"` then how do you understand that "exception"?

exercises/sgf-parsing/description.md
dzegel: i see your point

exercises/sgf-parsing/description.md
petertseng: A backslash followed by a tab should become a space. And a backslash followed by a t should become a t.

petertseng: Important to distinguish between:

* strings as they appear in a in JSON or a given programming language's string literal
* strings as they are presented to the SGF parser

If the string presented to the SGF parser contains a backslash followed by a tab (which will typically be written in JSON and/or in a language's string literal as `"\\\t"` that should be converted to a space, because "(exception: whitespaces still have to be converted to space)".

If the string presented to the SGF parser contains a backslash followed by a t (which will typically be written in JSON and/or in a language's string literal as "\\t" that should be converted to a t, because "Any char following `"\"` is inserted verbatim"

petertseng: Important to distinguish between:

* strings as they appear in a in JSON or a given programming language's string literal
* strings as they are presented to the SGF parser

If the string presented to the SGF parser contains a backslash followed by a tab (which will typically be written in JSON and/or in a language's string literal as `"\\\t"` that should be converted to a space, because "(exception: whitespaces still have to be converted to space)".

If the string presented to the SGF parser contains a backslash followed by a t (which will typically be written in JSON and/or in a language's string literal as `"\\t"` that should be converted to a t, because "Any char following `"\"` is inserted verbatim"

petertseng: If the string presented to the SGF parser contains a tab (written as "\t"), that should be converted to a space because "White spaces other than linebreaks are converted to space"

kytrinyx (approved):

Issue #1891, opened 2021-12-02
space-age: cover with test correct handling of unknow planet passed as argument


mstevanic:
Implementation could trigger divide by zero panic when unknown planet is passed as input to a function. Do we want to cover this case with test?

wolf99:
Could be good to have a test case(s) that cover any non-existent planet.

Would you like to add a PR for this?

mstevanic:
sure

PR #1892, opened 2021-12-02
space-age: add nonexistent planet to tests

mstevanic:
resolves  exercism/problem-specifications#1891

### DIFF
--- a/exercises/space-age/canonical-data.json
+++ b/exercises/space-age/canonical-data.json
@@ -80,6 +80,16 @@
         "seconds": 1821023456
       },
       "expected": 0.35
+    },
+    {
+      "uuid": "57b96e2a-1178-40b7-b34d-f3c9c34e4bf4",
+      "description": "invalid planet cause error",
+      "property": "age",
+      "input": {
+        "planet": "Sun",
+        "seconds": 680804807
+      },
+      "expected": {"error": "not a planet"}
     }
   ]
 }
--- a/exercises/space-age/canonical-data.json
+++ b/exercises/space-age/canonical-data.json
@@ -83,7 +83,7 @@
     },
     {
       "uuid": "57b96e2a-1178-40b7-b34d-f3c9c34e4bf4",
-      "description": "invalid planet cause error",
+      "description": "invalid planet causes error",
       "property": "age",
       "input": {
         "planet": "Sun",


exercises/space-age/canonical-data.json
```diff
@@ -80,6 +80,16 @@
...
+    },
+    {
+      "uuid": "57b96e2a-1178-40b7-b34d-f3c9c34e4bf4",
+      "description": "age on Pluto",
+      "property": "age",
+      "input": {
+        "planet": "Pluto",
+        "seconds": 680804807
+      },
+      "expected": 0
```
petertseng, line 92: I can't agree with the answer being 0. Your age isn't 0 on Pluto. Instead it should be `{"error": "not a planet"}`, or something similar.

exercises/space-age/canonical-data.json
```diff
@@ -80,6 +80,16 @@
         "seconds": 1821023456
       },
       "expected": 0.35
+    },
+    {
+      "uuid": "57b96e2a-1178-40b7-b34d-f3c9c34e4bf4",
+      "description": "age on Pluto",
```
petertseng, line 86:
```suggestion
      "description": "age on a non-planet",
```

Since the fact that it's Pluto specifically isn't important, and instead the important part is the fact that it's a non-planet, I suggest that this description better describes the true purpose of this test.

exercises/space-age/canonical-data.json
```diff
@@ -80,6 +80,16 @@
...
+      "property": "age",
+      "input": {
+        "planet": "Pluto",
```
junedev, line 89: Building on what was said above https://github.com/exercism/problem-specifications/pull/1892/files#r761412772 I would additionally recommend to use some other name than Pluto here to avoid confusion for people that didn't follow the whole "Is Pluto a planet or not?" discussion some time ago. Maybe something like "Milky Way" ...

exercises/space-age/canonical-data.json
```diff
@@ -80,6 +80,16 @@
         "seconds": 1821023456
       },
       "expected": 0.35
+    },
+    {
+      "uuid": "57b96e2a-1178-40b7-b34d-f3c9c34e4bf4",
+      "description": "age on Sun",
```
angelikatyborska, line 86: If I'm not mistaken, "sun" being a common noun needs a definitive article when we're talking about our sun in our solar system.
```suggestion
      "description": "age on the Sun",
```

exercises/space-age/canonical-data.json
kotp: Why not use something nonsensical to test "unknown" or "not handled" such as "banana"?  No need to worry about it being a planet or a comet or anything that might orbit the sun?  The focus is not really "some nonexistent planet" but more "unknown input".

exercises/space-age/canonical-data.json
```diff
@@ -80,6 +80,16 @@
         "seconds": 1821023456
       },
       "expected": 0.35
+    },
+    {
+      "uuid": "57b96e2a-1178-40b7-b34d-f3c9c34e4bf4",
+      "description": "age on the Sun",
```
wolf99, line 86: Description could maybe indicate more the purpose of the test case, i.e. that only planets in our solar system are acceptable. 
The sun is only improtant because it reflects that case.
Maybe something like "Other heavenly bodies cause error" ?

Other than that this looks good to me. 👍

wolf99 (changes requested):

wolf99 (approved):

exercises/space-age/canonical-data.json
```diff
@@ -80,6 +80,16 @@
         "seconds": 1821023456
       },
       "expected": 0.35
+    },
+    {
+      "uuid": "57b96e2a-1178-40b7-b34d-f3c9c34e4bf4",
+      "description": "invalid planet cause error",
```
petertseng, line 86:
```suggestion
      "description": "invalid planet causes error",
```

petertseng (approved):
I think it's more correct grammar to say an invalid planet causes an error, but I'll take either way.

angelikatyborska (approved):

wolf99:
Yay, thanks @mstevanic 👍

PR #1893, opened 2021-12-08
Fix dice examples description

artamonovkirill:


exercises/dnd-character/description.md
```diff
@@ -13,7 +13,7 @@ from your character's constitution, divide by 2 and round down.
 
 Write a random character generator that follows the rules above.
 
-For example, the six throws of four dice may look like:
+For example, the four throws of 6-sided dice may look like:
```
siebenschlaefer, line 16: Are you sure about that? One throw is to take multiple dice and roll them.
The following six lines list **six** throws, each with **four** dice.

(Disclaimer: I'm not a native English speaker.)

artamonovkirill: No, I think you are right. Although I was initially confused by how it's worded.

artamonovkirill: In my head it was one dice that you throw 4 times, not 4 dice that you throw at once

siebenschlaefer: If you still find it (somewhat) confusing then it will probably confuse others, too.
So it might be worth improving the description. (But I can't think of a better alternative ATM.)

kotp: > A throw of four 6-sided die may look like:

The problem may be the singular version of "dice" instead of the plural word "die".

artamonovkirill: Isn't it the opposite - plural "dice" and singular "die" or "dice"? At least that's what [Wikipedia](https://en.wikipedia.org/wiki/Dice) claims. But it's the first time I hear "die" and not "dice" for singular.

kotp: Yes, the typographical error was indeed me missing the "c" in the word.  A single die is cut, for a pair, when you cut a plate as well.  So in my message, it was a typo, indeed you take may throw multiple dice and check the value of each die.

kotp: Definitely my English as not first language peeking out.  I forgot the C when I wrote the sentence and then made the comment based on what I wrote.

PR #1820, opened 2021-07-12
Collatz Conjecture: test for non-integers

javidlakha:
The Collatz Conjecture holds only for positive integers. Whilst for strongly-typed languages such as Rust, the compiler ensures that the input must be an integer, for dynamically-typed languages (e.g. Python) or languages in which there is no integer type (e.g. TypeScript), the input must be explicitly tested. This commit adds a test case for a positive non-integer and modifies the expected error message from 'Only positive numbers are allowed' to 'Only positive integers are allowed' so that the student can test the input using a single if statement.

PR #1897, opened 2021-12-15
Revert "Collatz Conjecture: test for non-integers"

SleeplessByte:
Reverts exercism/problem-specifications#1820

@wolf99 did you want to do this??

### DIFF
--- a/exercises/collatz-conjecture/canonical-data.json
+++ b/exercises/collatz-conjecture/canonical-data.json
@@ -47,17 +47,6 @@
       "expected": { "error": "Only positive numbers are allowed" }
     },
     {
-      "uuid": "2187673d-77d6-4543-975e-66df6c50e2da",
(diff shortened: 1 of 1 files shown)


junedev (approved):

SaschaMann (approved):

wolf99:
No, sorry these tests should stay. It's the `f7176864-e35d-41eb-ba80-0242ac130004` test that should be removed.

Per: https://github.com/exercism/problem-specifications/pull/1820#pullrequestreview-828551671

wolf99 (changes requested):
Negative and zero tests were OK, its the non-integer (i.e. floating point) test that should be removed.

wolf99:
Oh! looking now it seems like it is already removed!
This PR is not needed at all then.

PR #1898, opened 2021-12-23
isbn-verifier: Enhance invalid characters test

lcarva:
Add a test case to cover the scenario where an invalid character is
nested inside an otherwise valid ISBN.

Signed-off-by: Luiz Carvalho <lucarval@redhat.com>

### DIFF
--- a/exercises/isbn-verifier/canonical-data.json
+++ b/exercises/isbn-verifier/canonical-data.json
@@ -150,6 +150,17 @@
             "expected": false
         },
         {
+            "uuid": "daad3e58-ce00-4395-8a8e-e3eded1cdc86",
+            "description": "invalid characters are not ignored",
+            "comments": ["Catch invalid characters in an otherwise valid isbn"],
+            "reimplements": "ed6e8d1b-382c-4081-8326-8b772c581fec",
+            "property": "isValid",
+            "input": {
+                "isbn": "3598P215088"
+            },
+            "expected": false
+        },
+        {
             "uuid": "fb5e48d8-7c03-4bfb-a088-b101df16fdc3",
             "description": "input is too long but contains a valid isbn",
             "property": "isValid",


exercises/isbn-verifier/canonical-data.json
```diff
@@ -149,6 +149,15 @@
             },
             "expected": false
         },
+        {
+            "uuid": "6ec400c2-6421-11ec-97c5-8c1645feb539",
+            "description": "extraneous invalid characters are not ignored",
```
wolf99, line 154: To better distinguish this case from the preceding case (  
ed6e8d1b-382c-4081-8326-8b772c581fec​), it may be good to follow the same pattern used in other cases for the description:

E.g.: `"description": "invalid characters are not ignored in valid isbn"`

lcarva: Makes sense. Updated the description as suggested.

exercises/isbn-verifier/canonical-data.json
```diff
@@ -149,6 +149,15 @@
...
         },
+        {
+            "uuid": "6ec400c2-6421-11ec-97c5-8c1645feb539",
+            "description": "invalid characters are not ignored in valid isbn",
+            "property": "isValid",
+            "input": {
+                "isbn": "3598P215088"
+            },
+            "expected": false
+        },
```
junedev, line 160: I like this test case but I find it a bit difficult to differentiate from the one above. I am thinking whether the best course of action might be to call this one `"invalid characters are not ignored"` and make it a [re-implementation](https://github.com/exercism/problem-specifications#changing-tests) of the one above.

There is no way to tell with the test whether the test case above (line 144) actually failed because of the invalid character or just because the isbn was invalid. Because of that, the new test case would cover that "not ignoring" much better.

lcarva: Yes, a single test makes more sense. Updated accordingly.

junedev (approved):

wolf99 (approved):

exercises/isbn-verifier/canonical-data.json
```diff
@@ -149,6 +149,17 @@
             },
             "expected": false
         },
+        {
+            "uuid": "6ec400c2-6421-11ec-97c5-8c1645feb539",
```
angelikatyborska, line 153: The validation is failing on this UUID. I think it's because you generated a version 1 UUID, but we expect a version 4 UUID.

wolf99: Good catch!

FYI to the OP that the appropriate UUID can be generated using Exercism's configlet tool.

angelikatyborska: (Or on https://www.uuidgenerator.net/)

lcarva: Went with a third option by using the cli fom [OSSP UUID](https://github.com/exercism/problem-specifications/pull/1898): `uuid -v 4`,

ErikSchierboom (approved):

wolf99:
Thanks for contributing @lcarva !

petertseng:
Since these two tests, `3132P34035` and `3598P215088`, reject two *different* incorrect implementations, my recommendation is that tracks include **both**.

`3132P34035` rejects those implementations that silently drop invalid letters, with the length requirement of 10 checked **before** dropping.

`3598P215088` rejects those implementations that silently drop invalid letters, with the length requirement of 10 checked **after** dropping.

Running https://github.com/petertseng/exercism-problem-specifications/blob/verify/exercises/isbn-verifier/verify.rb#L114-L131 shows that these two implementations each get the correct answer on all other tests and only get the wrong answer on the specific input that rejects them. Here is the output.

```
== anti-verify invalid middle -> dropped, length checked before drop isValid ==
15. invalid characters are not ignored: Got true instead of false
== anti-verify invalid middle -> dropped, length checked after drop isValid ==
16. invalid characters are not ignored: Got true instead of false
```

Since the first implementation only fails test 15 and not 16, while the second implementation only fails test 16 and not 15, my recommendation is that tracks use **both**.

PR #1901, opened 2021-12-24
Markdown style for Exercism

kotp:
This addresses:

* single sentence per line
* minor formatting changes
* minor tyopgraphical fixes

Message-Id: <20211224204253.1630934-3-keeperotphones@gmail.com>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,20 +1,15 @@
 # Contributing
 
-It's pretty tricky to keep all the language tracks in good working order,
-and it's not uncommon that people discover typos in the README or the test
-suites, inconsistencies between the README and the tests, missing edge cases,
-factual errors, logical errors, and confusing ambiguities.
+It is pretty tricky to keep all the language tracks in good working order, and it is not uncommon that people discover typos in the README or the test suites, inconsistencies between the README and the tests, missing edge cases, factual errors, logical errors, and confusing ambiguities.
 
-We welcome contributions of all sorts and sizes, from reporting issues to
-submitting patches, to hanging out in the support chat to help people get
-unstuck.
+We welcome contributions of all sorts and sizes, from reporting issues to submitting patches, to hanging out in the support chat to help people get unstuck.
 
 We are grateful for any help in making Exercism better!
 
 -----
 
 This guide covers several common scenarios pertaining to **improving the language tracks**
-themselves. [There are other guides about contributing to other parts of the Exercism ecosystem.](https://exercism.org/contributing)
+themselves. [There are other guides about contributing to other parts of the Exercism ecosystem](https://exercism.org/contributing).
 
 ## Table of Contents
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -83,12 +71,11 @@ We use the following terminology:
 
 ### Language Track Repositories
 
-We've given each language track an ID, which is a url-friendly version of the
-language name. For example, _C++_ has the ID `cpp`. This ID is used throughout
-the Exercism ecosystem.
+We have given each language track an ID, which is an url-friendly version of the language name.
```
wolf99, line 74: `an url-` vs `a URL-`?
The latter seems more correct to me. But it's not a big deal.

wolf99 (approved):

kotp:
On 21/12/25 01:17AM, wolf99 wrote:
> > @@ -83,12 +71,11 @@ We use the following terminology:
>  
> `an url-` vs `a URL-`?
> The latter seems more correct to me. But it's not a big deal.

I believe that the use of "an" is based on the following sound.  So if
"universal" is as a consonant "y" sound, then yes, I believe it should
be "a universal resource…" rather than "an universal resource…".  But if
it is pronounced as a vowel, then I think it is correct to say "an".
Even with a literal consonant, we say "an", such as "an hour" rather
than "a hour", and write it the same way, again, because of it being
based on the vowel or consonant sound.

Please do let me know if in English that is a "y" consonant sound for
"universal".  Write me an "h", write me an "i", write me a "u".

wolf99:
I have the same understanding of the grammatical guide for a vs an... My point isabout how people prefer to use "url" and the resulting choice of a or an.

URL is an acronym, usually spelled out when read out, like "you are ell". Meaning that "a" should be used. 

If we consider it instead to be a word in its own right, "url" (which I think is what you are following @kotp), then it should be "an".

I guess I have no strong feeling as to which is chosen, moreso pointing out that it _could_ be seen as more correct with the acronym way if we are pedantic about acronyms vs word for URL/url 😄.

kotp:
On 21/12/25 01:57PM, wolf99 wrote:
> If we consider it instead to be a word in its own right, "url" (which
> I think is what you are following @kotp), then it should be "an".

No, I am not thinking that you would say "oorl", for example, but I also
do not have experience in how people say it out loud.  I assume they
read acronyms as what they stand for.  Such as reading lol, as "laugh
out loud" as I have never heard anyone actually say "lol", or "rotf",
rather than "rolling on the floor".

I will remove the "an" and restore the "a" because for uniform, that has
the verbal "y" constant sound as far as I know in English, and so
reading it out loud as either the word or the individual letters should
be correct.

Thank you for the confirmation.

angelikatyborska (approved):

ErikSchierboom (approved):

kotp:
On 21/12/28 04:37AM, wolf99 wrote:
> Merged #1901 into main.

Thank you @wolf99 for merging!

PR #1904, opened 2021-12-28
isbn-verifier: Distinguish two different invalid character tests

petertseng:
One (3132P34035) tests what happens if the invalid character is ignored
after checking length, the other (3598P215088) tests what happens if the
invalid character is ignored before checking length.

These are two different incorrect implementations, and they get the
correct answer on the other tests. So only including one of the two
would fail to reject the other, and tracks would lose coverage.

Therefore, my recommendation is that tracks include both tests.

I have delivered the full recommendation in
https://github.com/exercism/problem-specifications/pull/1898#issuecomment-1002061905

Instead of having one reimplement the other, distinguish the two by
description.

### DIFF
--- a/exercises/isbn-verifier/canonical-data.json
+++ b/exercises/isbn-verifier/canonical-data.json
@@ -142,7 +142,7 @@
         },
         {
             "uuid": "ed6e8d1b-382c-4081-8326-8b772c581fec",
-            "description": "invalid characters are not ignored",
+            "description": "invalid characters are not ignored after checking length",
             "property": "isValid",
             "input": {
                 "isbn": "3132P34035"
@@ -151,9 +151,8 @@
         },
         {
             "uuid": "daad3e58-ce00-4395-8a8e-e3eded1cdc86",
-            "description": "invalid characters are not ignored",
+            "description": "invalid characters are not ignored before checking length",
             "comments": ["Catch invalid characters in an otherwise valid isbn"],
-            "reimplements": "ed6e8d1b-382c-4081-8326-8b772c581fec",
             "property": "isValid",
             "input": {
                 "isbn": "3598P215088"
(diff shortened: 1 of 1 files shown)


wolf99 (approved):

petertseng:
Going to propose that I also clarify the purpose of the other invalid test to further distinguish it from these two - the other one tests that P isn't treated as zero.

junedev (approved):

junedev:
@petertseng Thanks for this. I should have looked closer at the implementation before suggesting to replace the existing test. Sorry for that!

petertseng:
No harm done at all and no fault of yours!

One of the things that the canonical data doesn't do a great job of right now is noting whether there is a specific incorrect implementation that a particular input is supposed to test. It's not easy for me to see how we could do that, so in my view no one is to blame.

When we added 3132P34035 we could not have foreseen that there was another similar implementation that was incorrect in a related but different way, so we could not have thought to make the description reflect this. But we are getting better descriptions out of the deal now, so that is good.

ErikSchierboom (approved):

exercises/isbn-verifier/canonical-data.json
```diff
@@ -151,9 +151,8 @@
         },
         {
             "uuid": "daad3e58-ce00-4395-8a8e-e3eded1cdc86",
-            "description": "invalid characters are not ignored",
+            "description": "invalid characters are not ignored before checking length",
             "comments": ["Catch invalid characters in an otherwise valid isbn"],
```
ErikSchierboom, line 155: Now that the description will be updated, does the comment provide any additional information?

petertseng: Hmm, there may still be some value in noting that the ISBN was otherwise valid! Or one might argue that "in an otherwise-valid ISBN" should be moved to the description instead? Might make it a bit too long though. Might leave it as-is for that reason.

ErikSchierboom: Okay, just checking.

PR #1907, opened 2022-01-06
Update style guide link

ErikSchierboom:


petertseng (approved):
the previous link doesn't exist anymore so it would be good to have this, so that we will then have a link that does work

kotp (approved):

wolf99 (approved):

PR #1911, opened 2022-01-08
Affine markdown style guide applied

kotp:
Markdown guidelines applied

- Affine description: One sentence per line
- Reference style links for Affine

### DIFF
--- a/exercises/affine-cipher/description.md
+++ b/exercises/affine-cipher/description.md
@@ -1,14 +1,12 @@
 # Description
 
-Create an implementation of the affine cipher,
-an ancient encryption system created in the Middle East.
+Create an implementation of the affine cipher, an ancient encryption system created in the Middle East.
 
-The affine cipher is a type of mono-alphabetic substitution cipher.
-Each character is mapped to its numeric equivalent, encrypted with
-a mathematical function and then converted to the letter relating to
-its new numeric value. Although all mono-alphabetic ciphers are weak,
-the affine cipher is much stronger than the atbash cipher,
-because it has many more keys.
+The affine cipher is a type of monoalphabetic substitution cipher.
+Each character is mapped to its numeric equivalent, encrypted with a mathematical function and then converted to the letter relating to its new numeric value.
+Although all monoalphabetic ciphers are weak, the affine cipher is much stronger than the atbash cipher, because it has many more keys.
+
+[comment]: # ( monoalphabetic as spelled by Merriam-Webster, compare to polyalphabetic )
 
 ## Encryption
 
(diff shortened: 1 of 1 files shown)


exercises/affine-cipher/description.md
```diff
@@ -1,14 +1,12 @@
...
 
-The affine cipher is a type of mono-alphabetic substitution cipher.
-Each character is mapped to its numeric equivalent, encrypted with
-a mathematical function and then converted to the letter relating to
-its new numeric value. Although all mono-alphabetic ciphers are weak,
-the affine cipher is much stronger than the atbash cipher,
-because it has many more keys.
+The affine cipher is a type of monoalphabetic substitution cipher.
+Each character is mapped to its numeric equivalent, encrypted with a mathematical function and then converted to the letter relating to its new numeric value.
+Although all monoalphabetic ciphers are weak, the affine cipher is much stronger than the atbash cipher, because it has many more keys.
```
petertseng, line 7: is this PR the place to discuss potential changes to the text, or does that go elsewhere because this PR was only meant to change formatting?

In particular, I'd like to call attention to the word "it"  - someone reading it might ask "which of the two ciphers is being referred to by 'it' here?"

and therefore, I have one of two suggestions.

```suggestion
Although all monoalphabetic ciphers are weak, the affine cipher is much stronger than the atbash cipher, because the former has many more keys.
```
```suggestion
Although all monoalphabetic ciphers are weak, the affine cipher is much stronger than the atbash cipher, because the former has many more keys than the latter.
```

exercises/affine-cipher/description.md
```diff
@@ -41,9 +40,10 @@ The decryption function is:
 D(y) = (a^-1)(y - b) mod m
 ```
 
-- where `y` is the numeric value of an encrypted letter, ie. `y = E(x)`
-- it is important to note that `a^-1` is the modular multiplicative inverse (MMI)
-  of `a mod m`
+Where:
```
petertseng, line 43: I was initially uncertain that the "Where" should be lifted out of the list, because only the term `y` is being defined anew, and the other things are just additional notes, but after thinking about it I see that it does make sense if each individual item is prefixed with the word "Where", so that seems fine. Even though the only term being defined is `y`, the operation `a^-1` is being defined as well.

exercises/affine-cipher/description.md
```diff
@@ -1,14 +1,12 @@
...
-Each character is mapped to its numeric equivalent, encrypted with
-a mathematical function and then converted to the letter relating to
-its new numeric value. Although all mono-alphabetic ciphers are weak,
-the affine cipher is much stronger than the atbash cipher,
-because it has many more keys.
+The affine cipher is a type of monoalphabetic substitution cipher.
+Each character is mapped to its numeric equivalent, encrypted with a mathematical function and then converted to the letter relating to its new numeric value.
+Although all monoalphabetic ciphers are weak, the affine cipher is much stronger than the atbash cipher, because it has many more keys.
+
+[comment]: # ( monoalphabetic as spelled by Merriam-Webster, compare to polyalphabetic )
```
petertseng, line 9: I wouldn't have required that this comment be preserved in the file, but I'm fine either way.

petertseng (approved):

exercises/affine-cipher/description.md
kotp: I think it is common for native speakers to infer properly, while perhaps English as a second and third language may miss it a bit.

But yeah, not quite right PR for this, but if I did this, it would be a different commit, so that if necessary it could be reverted without taking the formatting away.  (So it would not be squashed, only rebased.)  But I can not necessarily control how people merge this in, so I would almost prefer a different PR.

exercises/affine-cipher/description.md
kotp: I added this in the file, because it is not the first time this word has been modified.  Having the comment in the file hopefully gives the reasoning as to why it is the way it is.  (when changing the spelling again, it is more likely this reference will be seen.)

exercises/affine-cipher/description.md
kotp: I think making a different PR might also bring abut a "policy" to tend to more clearly tend to pronouns across the exercises?

wolf99 (approved):

ErikSchierboom (approved):

PR #1912, opened 2022-01-08
forth: format: instructions all on one line

petertseng:
Originally a part of
https://github.com/exercism/problem-specifications/pull/1742

---

My own commentary on this is that I prefer not to have this change. I think the original version without this change is more readable, since it's helpful to see the individual instructions each on their own line.

Is it possible to change the formatting rules of whatever made this change so that it doesn't make this change anymore?

Or I could relent on this. Individual tracks can still format the elements of `instructions` as they see fit. So that is fine with me if it's not possible or not desirable to change the formatting rules.

### DIFF
--- a/exercises/forth/canonical-data.json
+++ b/exercises/forth/canonical-data.json
@@ -343,10 +343,7 @@
           "description": "can consist of built-in words",
           "property": "evaluate",
           "input": {
-            "instructions": [
-              ": dup-twice dup dup ;",
-              "1 dup-twice"
-            ]
+            "instructions": [": dup-twice dup dup ;", "1 dup-twice"]
           },
           "expected": [1, 1, 1]
         },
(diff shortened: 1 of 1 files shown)


kotp (commented):
I vote for per line instructions as well.  Easier to read in my opinion as well.

wolf99:
Ditto, per line seems clearer.

ErikSchierboom:
Agreed.

petertseng:
As we consider formatting tools it seems like the most prominent point that varies from exercise to exercise is whether we prefer elements of an array to be on one line or many. It will be interesting to see if we can come up with a one-size-fits-all rule (something like "if each element of the array is small, put the entire array on one line, else put each element on its own line") or if we have no choice but to configure it per exercise.

kotp:
Nested arrays, "small" being subjective, and yeah, I find I have not yet found a "fits every time" model for this.

axtens:
The other benefit of one op per line (assuming I understand the conversion) is that it leaves room for a comment describing stack behaviour, viz
```
stop
drop
roll // you're on fire!
```

PR #1914, opened 2022-01-08
`rational-numbers` Add additional test cases

### LINKED ISSUE #1909: `rational-numbers` missing test case for negative rational exponent
The problem description states:
> Exponentiation of a rational number `r = a/b` to a negative integer power `n` is `r^n = (b^m)/(a^m)`, where `m = |n|`.

However this is not checked for in the [tests](https://github.com/exercism/problem-specifications/blob/main/exercises/rational-numbers/canonical-data.json#L268).

As such, the tests pass even if the implementation does something like
```
def rational_exponentiation(r, n):
  return Fraction(r.a ** n, r.b ** n)
```

Would my understanding be correct?  I'm happy to raise PR and add a test, just that I don't know how to generate UUID for it :sweat_smile:

(thanks @angelikatyborska for suggesting raising an issue)

neduard:
Fixes #1909 .  Any feedback welcome!

Note I have not ran the schema check.  Relying on CI.

### DIFF
--- a/exercises/rational-numbers/canonical-data.json
+++ b/exercises/rational-numbers/canonical-data.json
@@ -288,6 +288,26 @@
           "expected": [-1, 8]
         },
         {
+          "uuid": "c291cfae-cfd8-44f5-aa6c-b175c148a492",
+          "description": "Raise a positive rational number to a negative integer power",
+          "property": "exprational",
+          "input": {
+            "r": [3, 5],
+            "n": -2
+          },
+          "expected": [25, 9]
+        },
+        {
+          "uuid": "45cb3288-4ae4-4465-9ae5-c129de4fac8e",
+          "description": "Raise a negative rational number to a negative integer power",
+          "property": "exprational",
+          "input": {
+            "r": [-3, 5],
+            "n": -2
+          },
+          "expected": [-25, 9]
+        },
+        {
           "uuid": "e2f25b1d-e4de-4102-abc3-c2bb7c4591e4",
           "description": "Raise zero to an integer power",
           "property": "exprational",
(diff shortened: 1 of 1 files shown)


exercises/rational-numbers/canonical-data.json
```diff
@@ -287,6 +287,26 @@
...
+        },
+        {
+          "uuid": "45cb3288-4ae4-4465-9ae5-c129de4fac8e",
+          "description": "Raise a negative rational number to a negative integer power",
+          "property": "exprational",
+          "input": {
+            "r": [-3, 5],
+            "n": -2
+          },
+          "expected": [-25, 9]
```
petertseng, line 308: as I understand it, the result should be positive here, because a negative times a negative (we square the number) will be positive.

However, if we had taken the power -3, it would remain negative.

SaschaMann:
```suggestion
          "expected": [25, 9]
```

We could have cases for both `n = -2` and `n = -3` too.

SaschaMann (changes requested):

exercises/rational-numbers/canonical-data.json
neduard: D'oh :facepalm: you are right!  Fixed the test and added a new one with an odd negative power.

petertseng (approved):
OK. But, you would do better to use a word in https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue so that it automatically closes the issue

SaschaMann (approved):

ErikSchierboom (approved):

Issue #1910, opened 2022-01-07
largest-series-product: Should the description discuss expected results for a window size of 0?


petertseng:
For a long time we've had tests that the expected result for a window size of 0 should be 1 (the identity for multiplication). We never mentioned it in the description though. Should we?

Issue #1638, opened 2020-02-17
Largest Series Product: why LargestSeriesProduct(, 0) wants 1 ?


b10s:
Hi,

the `LargestSeriesProduct(, 0)` wants `1` as a result but product of nothing can not be 1, isn't it ?

SaschaMann:
The canonical data explains it: https://github.com/exercism/problem-specifications/blob/master/exercises/largest-series-product/canonical-data.json#L107:L120

Wikipedia has a more in-depth justification for why the empty product is 1 by convention: https://en.wikipedia.org/wiki/Empty_product

b10s:
@SaschaMann wiki says: `In mathematics, an empty product, is by convention zero` as my naive expectation:

<img width="1008" alt="Screen Shot 2020-02-19 at 0 22 01" src="https://user-images.githubusercontent.com/4618288/74749731-ef8c7180-52ad-11ea-9ed3-ed975766c436.png">

Did I miss something?

SaschaMann:
> Did I miss something?

The highlighted part below refers to the empty product. The rest of the sentence is about the empty sum.

> **In mathematics, an empty product, or nullary product, is the result of multiplying no factors. It is by convention equal to the multiplicative identity** (assuming there is an identity for the multiplication operation in question), just as the empty sum—the result of adding no numbers—is by convention zero, or the additive identity.

The multiplicative identity for real numbers is `1` because for any real number x, `1 * x = x`.

The justification is outlined [below](https://en.wikipedia.org/wiki/Empty_product#Justification):

> Allowing a "product" with only one or zero factors reduces the number of cases to be considered in many mathematical formulas. Such "products" are natural starting points in induction proofs, as well as in algorithms. For these reasons, the "empty product is one" convention is common practice in mathematics and computer programming.

In other words, it's 1 by convention because then it doesn't break things without requiring special cases. It's also consistent with other behaviour that the English Wikipedia article doesn't mention, for example in combinatorics. It's a convenient definition to have.

yawpitch:
That wikipedia article's first line could really use some editing for grammar.

b10s:
That is amazing!
Thank you for explaining and sorry for bothering

PR #1915, opened 2022-01-10
largest-series-product: Describe window size 0

petertseng:
Closes https://github.com/exercism/problem-specifications/issues/1910

We established the tests for window size 0 via
https://github.com/exercism/problem-specifications/pull/170

Window size 0 has caused questions in the past, including but not
limited to:

* https://github.com/exercism/problem-specifications/issues/1638
* https://github.com/exercism/go/issues/349

We have comments on the test cases, but it's not necessarily the case
that everyone will see those comments.

Let's try adding the reasoning to the description.

### DIFF
--- a/exercises/largest-series-product/description.md
+++ b/exercises/largest-series-product/description.md
@@ -12,3 +12,7 @@ in the input; the digits need not be *numerically consecutive*.
 
 For the input `'73167176531330624919225119674426574742355349194934'`,
 the largest product for a series of 6 digits is 23520.
+
+For a series of zero digits, the largest product is 1 because 1 is the multiplicative identity.
+(You don't need to know what a multiplicative identity is to solve this problem;
+it just means that multiplying a number by 1 gives you the same number.)


exercises/largest-series-product/description.md
```diff
@@ -12,3 +12,5 @@ in the input; the digits need not be *numerically consecutive*.
 
 For the input `'73167176531330624919225119674426574742355349194934'`,
 the largest product for a series of 6 digits is 23520.
+
+For a series of zero digits, the largest product is 1 because 1 is the identity for multiplication.
```
petertseng, line 16: To discuss: Should a link be added? Candidates:

* https://en.wikipedia.org/wiki/Identity_element - note that this lists identities for operations other than multiplication as well.
* https://en.wikipedia.org/wiki/Empty_product - note that when reading the first paragraph you don't want to get mixed up between the additive identity 0 versus the multiplicative identity 1.

I'm not sure there should be a link here because I don't want to imply that reading the link is required to solve the problem. We could solve the problem by saying "You don't have to know what a multiplicative identity is, but if you wanted to, you could read X"

petertseng, line 16: To discuss: Should a link be added? Candidates:

* https://en.wikipedia.org/wiki/Identity_element - note that this lists identities for operations other than multiplication as well.
* https://en.wikipedia.org/wiki/Empty_product - note that when reading the first paragraph you don't want to get mixed up between the additive identity 0 versus the multiplicative identity 1.

I'm not sure there should be a link here because I don't want to imply that reading the link is required for the student to solve this problem.

However, if that is a concern, we could avoid implying the link is required by instead saying "You don't have to know what a multiplicative identity is, but if you wanted to, you could read X"

SaschaMann: Wikipedia articles on maths topics are often rather overwhelming for people who aren't at least vaguely familiar with the topic. I think a link would help but perhaps there's a simpler site to link to.

SaschaMann, line 16:
```suggestion
For a series of zero digits, the largest product is 1 because 1 is the multiplicative identity.
```

SaschaMann (approved):

exercises/largest-series-product/description.md
wolf99: > You don't have to know what a multiplicative identity is, but if you wanted to, you could read X"

I like this suggestion because although technically correct, "multiplicative identity" is a fancy maths-y sounding term that may put people off, especially people that may be struggling with non-native English; and really instead of learning maths all the student needs to know is that for empty sets their implementation needs to return 1.

ErikSchierboom (approved):

exercises/largest-series-product/description.md
petertseng: Based on that concern (jargon may be offputting), I think it's good to say this anyway, even if we don't have a link to use. I'll try this:

"You don't need to know what a multiplicative identity is to solve this problem; it just means that multiplying a number by 1 gives you the same number."

wolf99 (approved):

SaschaMann:
I saw the last change too late but this might also be a good candidate to make us of admonitions, e.g.

```md

For a series of zero digits, the largest product is 1.

~~~exercism/advanced
This is the case because 1 is the [multiplicative identity](https://en.wikipedia.org/wiki/Identity_element) of [natural numbers](https://en.wikipedia.org/wiki/Natural_number).
~~~
```

That way we can have both an accurate motivation but the admonition allows us to avoid the issue with "scary" jargon as this case is exactly what they were made for.

kotp:
I do not think it was too later, as I only just saw the response to it now. :)

This might be a good quick PR.  I like the idea.

PR #1916, opened 2022-01-11
Format canonical data using jq

ErikSchierboom:
While looking into re-ordering the keys across all `canonical-data.json` files, I found that there was a lot of inconsistency in the formatting between files.
This isn't a big problem, but _some_ consistency is nice.
For this draft PR, I've used jq to reformat all the canonical data files, as its style of formatting seemed to be slightly closer to our current style than prettier.

The goal of this file is to check which formatting changes are desirable, and more important: which are not.


exercises/accumulate/canonical-data.json
```diff
@@ -23,43 +23,85 @@
       "description": "accumulate squares",
       "property": "accumulate",
       "input": {
-        "list": [1, 2, 3],
+        "list": [
```
SaschaMann, line 26: These don't seem desirable. They make reading and copy and pasting cases harder when manually implementing a solution.

exercises/accumulate/canonical-data.json
```diff
@@ -23,43 +23,85 @@
...
+        [
+          "b1",
+          "b2",
+          "b3"
+        ],
+        [
+          "c1",
+          "c2",
+          "c3"
+        ]
```
SaschaMann, line 104: This is way harder to read than the matrix on the left.

exercises/all-your-base/canonical-data.json
```diff
@@ -18,87 +18,147 @@
...
-        "digits": [1, 0, 1, 0, 1, 0],
+        "digits": [
+          1,
+          0,
+          1,
+          0,
+          1,
+          0
+        ],
         "outputBase": 10
```
SaschaMann, line 78: This doesn't read like a binary number anymore.

exercises/binary-search/canonical-data.json
```diff
@@ -4,11 +4,9 @@
     "The error object is used to indicate that the value is not included in the array.",
     "It should be replaced with the respective expression that is idiomatic",
     "for the language that implements the tests.",
-
```
SaschaMann, line 7: Should probably be replaced with a `"",` or ignored then because the newline seems intentional.

exercises/dominoes/canonical-data.json
```diff
@@ -68,7 +78,20 @@
...
+          ],
+          [
+            3,
+            1
+          ],
+          [
+            2,
+            3
+          ]
+        ]
```
SaschaMann, line 94: An even worse example than the ones above 😅

SaschaMann (commented):
Exploding arrays onto several lines seems undesirable in most cases.

ErikSchierboom:
Totally agree on the array examples. Are _all_ arrays rewrites bad, or just some?

SaschaMann:
I've only skimmed and commented on a few. I think for the inputs/outputs, they might not all be _that_ bad but I think manual formatting depending on the context would be better.

As for reformatting the comment arrays, that seems pointless but also not harmful.

exercises/grains/canonical-data.json
```diff
@@ -69,7 +69,7 @@
           "input": {
             "square": 64
           },
-          "expected": 9223372036854775808
+          "expected": 9223372036854776000
```
petertseng, line 72: unfortunate. for this reason, we may not be able to use jq to format grains.

exercises/grains/canonical-data.json
```diff
@@ -105,7 +111,7 @@
       "description": "returns the total number of grains on the board",
       "property": "total",
       "input": {},
-      "expected": 18446744073709551615
+      "expected": 18446744073709552000
```
petertseng, line 114: unfortunate. for this reason, we may not be able to use jq to format grains.

ErikSchierboom:
I've also opened https://github.com/exercism/problem-specifications/pull/1917 to see what the formatting looks like when using prettier.

ErikSchierboom:
Closing this in favor of #1917

Issue #1705, opened 2020-10-13
Order of canonical-data keys


SaschaMann:
From @ee7 in https://github.com/exercism/problem-specifications/pull/1703#pullrequestreview-507842858

> But two questions:
> 
>  1. Do we have an official ordering that the keys are supposed to be in?
> 
>  2. If so, should we enforce the key order during CI? Maybe it helps with readability when adding tests. The ordering seemed pretty consistent before, but  there will be nearly double the number of keys now.
> 
> 
> I notice that the `reimplements` in this PR is after `uuid`, but when I made my recent PRs I instead ordered the keys as I saw them in #1674, that is:
> 
> ```json
>   {
>     "uuid": "82d32c2e-07b5-42d9-9b1c-19af72bae860",
>     "description": "two times one is two",
>     "comments": ["Expected value is changed to 2"],
>     "reimplements": "e46c542b-31fc-4506-bcae-6b62b3268537",
>     "property": "twice",
>     "input": {
>       "number": 1
>     },
>     "expected": 2
>   }
> ```

---

## Tasks:

- [ ] Decide if there should be an order
- [ ] Decide on an order
- [ ] Convert all data to the right order
- [ ] Add a CI check for it

ErikSchierboom:
I think the order for virtually all of the existing canonical data is:

1. "uuid"
1. "description"
1. "comments"
1. "property"
1. "input"
1. "expected"

I'm fine with keeping this as is. Maybe we should add the "reimplements" key directly after the "uuid" key?

1. "uuid"
1. "reimplements"
1. "description"
1. "comments"
1. "property"
1. "input"
1. "expected"

SaschaMann:
> Maybe we should add the "reimplements" key directly after the "uuid" key?

I'd like that because it keeps the unreadable part away from the rest

ee7:
@ErikSchierboom And `"scenarios"`? In #1674 it was written last in:
```json
{
  "uuid": "25a8463d-560f-4a42-9be7-b79d00cd28a4",
  "description": "64",
  "property": "square",
  "input": {
    "square": 64
  },
  "expected": 9223372036854775808,
  "scenarios": ["big-integers"]
}
```

ErikSchierboom:
Ah, yes, forgot about that 🤦 Scenarios is basically metadata, so it would perhaps make sense to group it with the other metadata. Maybe:

1. "uuid"
1. "reimplements"
1. "scenarios"
1. "description"
1. "comments"
[...]

PR #1917, opened 2022-01-12
Format using prettier

ErikSchierboom:
As a counterpart to https://github.com/exercism/problem-specifications/pull/1916, this PR formats the `canonical-data.json` files used prettier.


exercises/all-your-base/canonical-data.json
```diff
@@ -153,7 +153,7 @@
         "digits": [0],
         "outputBase": 10
       },
-      "expected": {"error": "input base must be >= 2"}
+      "expected": { "error": "input base must be >= 2" }
```
wolf99, line 156: Seems like prettier is a lot less opinionated about JSON than jq. That's good for arrays, but bad for cases like this where it would be nice to have fields (`error` in this case) on their own lines.

Does prettier have any options to force this ?

ErikSchierboom: I don't think so. Prettier has (purposefully) few options: https://prettier.io/docs/en/options.html

exercises/connect/canonical-data.json
```diff
@@ -43,12 +39,7 @@
       "description": "only edges does not make a winner",
       "property": "winner",
       "input": {
-        "board": [
-          "O O O X",
-          " X . . X",
-          "  X . . X",
-          "   X O O O"
-        ]
+        "board": ["O O O X", " X . . X", "  X . . X", "   X O O O"]
```
SaschaMann, line 42: This change looks undesirable as one can no longer read it as an actual board.

exercises/diamond/canonical-data.json
```diff
@@ -1,139 +1,127 @@
...
+      "expected": [" A ", "B B", " A "]
+    },
+    {
+      "uuid": "af8efb49-14ed-447f-8944-4cc59ce3fd76",
+      "description": "Smallest non-degenerate case with odd diamond side length",
+      "property": "rows",
+      "input": {
+        "letter": "C"
+      },
+      "expected": ["  A  ", " B B ", "C   C", " B B ", "  A  "]
```
SaschaMann, line 46: Same issue here.

exercises/dominoes/canonical-data.json
```diff
@@ -68,7 +68,11 @@
       "description": "three elements",
       "property": "canChain",
       "input": {
-        "dominoes": [[1, 2], [3, 1], [2, 3]]
+        "dominoes": [
+          [1, 2],
+          [3, 1],
+          [2, 3]
+        ]
```
SaschaMann, line 75: This removes the visual hint that they form a chain.

SaschaMann (commented):
I feel like array formatting should just be left to humans since neither tool seems to do a good job when formatting this kind of human-readable data.

SaschaMann:
Perhaps #1705 should also be considered when choosing a formatter since there was agreement on an order. It would be nice if the tool could also sort the data according to the spec.

ErikSchierboom:
> Perhaps #1705 should also be considered when choosing a formatter since there was agreement on an order. It would be nice if the tool could also sort the data according to the spec.

The funny thing is that I started working on that and have a working jq script to re-order fields. That script though, re-formatted a bunch of files in the process, which is why I started looking into the formatting of the files.

> I feel like array formatting should just be left to humans since neither tool seems to do a good job when formatting this kind of human-readable data.

Yes, I agree. I think the end conclusion is that we can do a one-time pass for consistent formatting, but with manual tweaks of the results.

exercises/book-store/canonical-data.json
```diff
@@ -145,27 +147,33 @@
...
       "description": "Check that groups of four are created properly even when there are more groups of three than groups of five",
-      "comments": ["Suggested grouping, [[1,2,3,4],[1,2,3,5],[1,2,3,4],[1,2,3,5],[1,2,3],[1,2,3]]."],
+      "comments": [
+        "Suggested grouping, [[1,2,3,4],[1,2,3,5],[1,2,3,4],[1,2,3,5],[1,2,3],[1,2,3]]."
+      ],
       "input": {
-        "basket": [1,1,1,1,1,1,2,2,2,2,2,2,3,3,3,3,3,3,4,4,5,5]
+        "basket": [
+          1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2, 3, 3, 3, 3, 3, 3, 4, 4, 5, 5
+        ]
```
SaschaMann, line 176:
```suggestion
        "basket": [1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2, 3, 3, 3, 3, 3, 3, 4, 4, 5, 5]
```

This makes copy and pasting harder for no reason.

exercises/change/canonical-data.json
```diff
@@ -11,183 +11,89 @@
...
-        100,
-        100,
-        100,
-        100,
-        100,
-        100,
-        100,
-        100
+        2, 2, 5, 20, 20, 50, 100, 100, 100, 100, 100, 100, 100, 100, 100
       ]
```
SaschaMann, line 69: Should be one line `[ .... ]` imo.

exercises/isogram/canonical-data.json
```diff
@@ -124,7 +124,9 @@
     {
       "uuid": "0d0b8644-0a1e-4a31-a432-2b3ee270d847",
       "description": "word with duplicated character and with two hyphens",
-      "comments": ["This test aims to catch buggy implementations that check for duplicate spaces or hyphens."],
+      "comments": [
+        "This test aims to catch buggy implementations that check for duplicate spaces or hyphens."
+      ],
```
SaschaMann, line 129: For consistency it'd make more sense to leave this as a one line comment, I think. The line isn't that long.

exercises/reverse-string/canonical-data.json
```diff
@@ -1,8 +1,8 @@
 {
   "exercise": "reverse-string",
   "comments": [
-        "If property based testing tools are available, a good property to test is reversing a string twice: reverse(reverse(string)) == string"
-      ],
+    "If property based testing tools are available, a good property to test is reversing a string twice: reverse(reverse(string)) == string"
+  ],
```
SaschaMann, line 5: Same as above.

exercises/variable-length-quantity/canonical-data.json
```diff
@@ -170,7 +170,9 @@
           "input": {
             "integers": [8192, 1193046, 268435455, 0, 16383, 16384]
           },
-          "expected": [192, 0, 200, 232, 86, 255, 255, 255, 127, 0, 255, 127, 129, 128, 0]
+          "expected": [
+            192, 0, 200, 232, 86, 255, 255, 255, 127, 0, 255, 127, 129, 128, 0
+          ]
```
SaschaMann, line 175: Same as the integer array above.

exercises/variable-length-quantity/canonical-data.json
```diff
@@ -238,14 +242,18 @@
...
         },
         {
           "uuid": "a91e6f5a-c64a-48e3-8a75-ce1a81e0ebee",
           "description": "multiple values",
           "property": "decode",
           "input": {
-            "integers": [192, 0, 200, 232, 86, 255, 255, 255, 127, 0, 255, 127, 129, 128, 0]
+            "integers": [
+              192, 0, 200, 232, 86, 255, 255, 255, 127, 0, 255, 127, 129, 128, 0
+            ]
```
SaschaMann, line 256: Same as the integer array above.

SaschaMann (commented):
I don't like the error change but that's personal preference; I don't have any real objections against it. I think for everything else, this is a lot better than what either tool produced without the manual tweaks.

ErikSchierboom:
> I don't like the error change but that's personal preference

I don't massively care either way, but I thought people would like this better 🤷 If someone else wants to chip in, I'm happy to change

exercises/bowling/canonical-data.json
```diff
@@ -19,255 +19,362 @@
...
-    "description": "all strikes is a perfect game",
-    "property": "score",
-    "input": {
-      "previousRolls": [10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10]
+    {
+      "uuid": "62ee4c72-8ee8-4250-b794-234f1fec17b1",
+      "description": "all strikes is a perfect game",
+      "property": "score",
+      "input": {
+        "previousRolls": [10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10]
```
petertseng, line 169: in `bowling`, is there some rule I can use to understand when the elements of `previousRolls` are (or should be) on the same line as the `[` and `]` versus when they are on different lines?

exercises/scale-generator/canonical-data.json
```diff
@@ -15,7 +15,20 @@
           "input": {
             "tonic": "C"
           },
-          "expected": [ "C", "C#", "D", "D#", "E", "F", "F#", "G", "G#", "A", "A#", "B" ]
+          "expected": [
```
petertseng, line 18: I can't come up with a rule for why the chromatic scale starting at C should be on separate lines but F should be on one line. They have the same number of notes and the same sum of string lengths. What is the rule?

exercises/word-search/canonical-data.json
```diff
@@ -125,13 +103,8 @@
...
       "property": "search",
       "input": {
-        "grid": [
-          "jefblpepre",
-          "tclojurerm"
-        ],
-        "wordsToSearchFor": [
-          "clojure"
-        ]
+        "grid": ["jefblpepre", "tclojurerm"],
```
petertseng, line 106: although putting them on one line is fine because this case doesn't ask to search for a word vertically, I'm kind of thinking that tracks might want to start presenting each on one line, to start introducing the idea that this should be a grid. And if tracks are going to start doing it, I suggest it's a good idea for the JSON to do it as well

exercises/bowling/canonical-data.json
ErikSchierboom: There isn't. I've made them consistent.

exercises/scale-generator/canonical-data.json
ErikSchierboom: Fixed

exercises/word-search/canonical-data.json
ErikSchierboom: Agreed. I've now applied the same multi-line formatting for _all_ array fields. Let me know what that looks like.

exercises/grade-school/canonical-data.json
```diff
@@ -37,7 +37,11 @@
       "description": "Adding multiple students in the same grade in the roster",
       "property": "add",
       "input": {
-        "students": [["Blair", 2], ["James", 2], ["Paul", 2]]
```
petertseng, line 40: had I been the one making this PR, I would have kept the existing formatting, but I have no disagreement with the decision made in this PR.

exercises/proverb/canonical-data.json
```diff
@@ -1,108 +1,92 @@
...
+      "input": {
+        "strings": [
+          "nail",
+          "shoe",
+          "horse",
+          "rider",
+          "message",
+          "battle",
+          "kingdom"
+        ]
```
petertseng, line 65: I see that unlike the other `"strings"` in this file, this one qualified to be on separate lines because it has many elements (so putting all on one line would make long). okay.

exercises/reverse-string/canonical-data.json
```diff
@@ -1,8 +1,6 @@
 {
   "exercise": "reverse-string",
-  "comments": [
-        "If property based testing tools are available, a good property to test is reversing a string twice: reverse(reverse(string)) == string"
-      ],
+  "comments": ["If property based testing tools are available, a good property to test is reversing a string twice: reverse(reverse(string)) == string"],
```
petertseng, line 3: showing a place where a long comment was chosen to be the same line as the `[` and `]`

exercises/isogram/canonical-data.json
```diff
@@ -124,7 +122,9 @@
...
```
petertseng, line 127: in contrast to a different place where a long comment was chosen to be on the same line as the `[` and `]`, I see that here it was chosen to be on different lines than `[` and `]`. I don't have an explanation, but I don't need one.

petertseng (approved):
Do I understand correctly that we cannot add CI checks because we are making case-by-case decisions for arrays? That's too bad, because that means there's a possibility of future changes being badly formatted. Given that, I'm not excited about this PR; I only feel completely ambivalent about it. But if someone really wants this, I support that desire, and therefore I approve.

ErikSchierboom:
> Do I understand correctly that we cannot add CI checks because we are making case-by-case decisions for arrays? That's too bad, because that means there's a possibility of future changes being badly formatted. Given that, I'm not excited about this PR; I only feel completely ambivalent about it. But if someone really wants this, I support that desire, and therefore I approve.

This PR was mostly a test to check whether the auto formatting used by prettier _could_ be used to auto-format all files and what that would look like. Looking at the review comments, it looked like there were multiple cases where the formatting was not what we wanted. The way I envision this PR going forward was:

1. Auto format all files using prettier. This will normalize the formatting of all files
2. Do a manual check to find the files which formatting we feel is wrong. Fix that formatting to what we want

Once we've manually updated the files, we can exclude those files from being auto-formatted. This means that we _would_ be able to have a CI check for the formatting, although there would still be a risk of the excluded files' formatting getting out of sync. Would this approach excite you slightly more?

exercises/hamming/canonical-data.json
```diff
@@ -1,5 +1,5 @@
 {
-"exercise": "hamming",
+  "exercise": "hamming",
   "comments": [
     "Language implementations vary on the issue of unequal length strands.",
     "A language may elect to simplify this task by only presenting equal",
```
wolf99, line 5: The leading spacing here looks a bit odd. But I'm on mobile so not sure and it is not worth blocking this PR for.

wolf99 (approved):

petertseng:
Well, having some files be checked is better than none, I think. So that is good.

I'm even getting some ideas such as: Even though it appears prettier doesn't allow configuring whether array elements should go on their own, line, perhaps there is something we can write that would do that... then, we can use that tool in conjunction with prettier. We'd have a fixed list of exercises that tool should be run on, and it would run after prettier. In that way, every file will be automatically formatted.

ErikSchierboom:
> I'm even getting some ideas such as: Even though it appears prettier doesn't allow configuring whether array elements should go on their own, line, perhaps there is something we can write that would do that... then, we can use that tool in conjunction with prettier. We'd have a fixed list of exercises that tool should be run on, and it would run after prettier. In that way, every file will be automatically formatted.

Ooh, that would be nice. Just on the off chance: would you potentially be interested in developing such a tool? Or do you have ideas on whether we could something like this with plain bash?

petertseng:
Remind me what prettier does? Does it put all the elements on one line?

Now this tool does need to respect indentation, so it may be a little involved for a shell script, though a shell script might be able to do it by counting the number of leading spaces in the line and adding two more.

Now let's see. We're already running Python in CI, in https://github.com/exercism/problem-specifications/blob/main/bin/check-immutability.py, but there's a difference here in that check-immutability.py mainly needs to be run in CI, whereas it may be handy for contributors to be able to run this formatting thing, so a decision should be made with that in mind.

I think it will be easier to get a better sense of what this should look like after formatting starts getting enforced in CI, with exercises excluded. Then, exercises can be un-excluded and this formatting tool can be tested. It's fine to keep talking about it in the meantime though.

One thing I'm thinking about is whether it'll have any problems with strings that contain either `[` or `,` in them, which may start posing more difficulty for a shell script, but I will need to think about that more.

ErikSchierboom:
> Remind me what prettier does? Does it put all the elements on one line?

I think it depends on the number of characters the array would take if presented on a single line.

SaschaMann (approved):

ErikSchierboom:
I've added a couple of commits:

- [Ignore manually formatted files](b11f5089292086049321b2e3149cc29ff3fae6aa)
- [Format non-exercise JSON files](8a0fae807af8b19f552da240494b3389a3af1e11)
- [Check JSON formatting in CI](0b470a11920cb634bd3070c821febf34de4832d1)

canonical-data.schema.json
```diff
@@ -1,166 +1,159 @@
...
-               { "error": { "maxProperties": 1 }
-               }
-          },
-
-      "error":
-          { "description": "A message describing an error condition"
-          , "type"       : "string"
-          }
-
+  "comments": [
```
SaschaMann, line 2: The schema previously used a strange commas-first format. I don't particularly care either way but I assume it was done that way for a reason.

ErikSchierboom: Perhaps, maybe someone can remember? To me it always looked very odd and it was widely different from the other files.

petertseng: No specific reason is given in https://github.com/exercism/problem-specifications/pull/602. That's just how all JSON in that PR was formatted. It does not appear to me to have any significance other than that some decision needed to be made for how to format the file, and that particular choice, equally as valid as any other, was chosen.

SaschaMann: In that case it's probably best to format it to the standard JSON format as well :)

ErikSchierboom:
> in the ideal world, the commit entitled "Allow testing of formatting" would have been made easier to review by putting the lines that add the integrity line and remove the URL portion after the # into their own commit, but the problem was not so serious that I was forced to not approve.

Noted. I'll try and do my best to make commits more granular.

> Please consider the experience for a contributor who wants to make an addition to the JSON files but does not format it the same way prettier would and therefore CI fails. Is the output informative enough to tell them how to remedy their problem? I will trust your judgment on what you decide is sufficient for you to say we have considered their experience.

The output looks like the following:

```shell
~/exercism/problem-specifications$ yarn test-json-formatting
yarn run v1.22.17
$ prettier --check **/*.json
Checking formatting...
[warn] exercises/bob/canonical-data.json
[warn] Code style issues found in the above file(s). Forgot to run Prettier?
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

It does _not_ make is clear what the actual problem is, only that we expect them to run prettier. I see a couple of options here:

1. We can have a Github action that formats the code when someone posts a comment with a certain text (like `/format`). We had this for the v3 repo and it worked fairly well.
2. We can add a `format-json` command that invokes prettier to format all files.

ErikSchierboom:
I've [added a yarn command to format the JSON](https://github.com/exercism/problem-specifications/pull/1917/commits/ddfc0f9f23b6d9fcbab06be98dae878388d898b0) and [updated the contributing documentation](https://github.com/exercism/problem-specifications/pull/1917/commits/fbd495f606e14ee8261c319cc0da5611d42a1a06).

ErikSchierboom:
With 3 approvals, I'm gonna merge this and do a GHA workflow formatting follow-up PR later. Thanks everyone!

iHiD (approved):
Approving on principle.

ErikSchierboom:
Thanks everybody!

Issue #974, opened 2017-10-27
meetup: Test descriptions are unhelpful.


Insti:
The [canonical-data for meetup](https://github.com/exercism/problem-specifications/blob/master/exercises/meetup/canonical-data.json) has a lot of test cases, each of which has a description that just echoes the test input values.

None of the descriptions say why these values are interesting or useful to test.

**Task:**

Update the descriptions so each description describes which particular property of the solution it is trying to test for.

References: [Meetup description](https://github.com/exercism/problem-specifications/blob/master/exercises/meetup/description.md)

ErikSchierboom:
The [test case descriptions](https://github.com/exercism/problem-specifications/blob/main/exercises/meetup/canonical-data.json) have since all been updated, so this issue can be closed.

petertseng:
There have not been any changes to the test case descriptions since the creation of meetup/canonical-data.json.

My comment doesn't recommend for or against closing this issue; it is only saying that if this issue is to be closed, it has to be closed based on an assertion like "we decided that it is okay to close it without updating the descriptions".

ErikSchierboom:
Hmmm, I misread then, my mistake. 

> None of the descriptions say why these values are interesting or useful to test.

Well, most of the values _are_ not interesting :) With meetup, we took a fairly exhaustive approach to the test cases, which makes most of them not very interesting, but mostly useful for checking the code for exhaustiveness. I'd argue that the current descriptions are fine, but I'm open to people disagreeing. 

The fact though that nobody has replied to this in 5 years may give some indication on how others view this, but maybe not.

petertseng:
But there's a pattern to the values. Look at the expected dates for teenth Mondays:

* 2013-05-13
* 2013-08-19
* 2013-09-16

Days in the teenth week can fall between 13th to the 19th inclusive, so the cases have been carefully chosen to test both endpoints and an arbitrary date in the middle. This pattern (testing 13th, 19th, and a date in the middle) is consistently followed for all seven days of the teenth week, so this was a deliberate choice. And failing to note this deliberate choice is losing information.

The choices are not as thorough for the other weeks, but that can be rectified by adding cases.

* first: 1-7
* second: 8-14
* third: 15-21
* fourth: 22-28
* last: several interesting cases here, depending on whether the month has five of that day or four, and also for a February in a leap year.

This does raise an interesting choice of how one would write the description though. Taking an example for teenth Monday of May 2013 which lands on 2013-05-13, would it be written as:

[...]

PR #1919, opened 2022-01-14
meetup: improve descriptions by saying why each case is tested

petertseng:
descriptions show whether a date is the first, last, or an arbitrary
middle date of the week. This helps understand why certain cases are
selected.

Closes https://github.com/exercism/problem-specifications/issues/974

Please note the code that generated the descriptions as well.


petertseng:
search for the string "another" (11 occurrences) where I was unable to discern a specific reason for a given test case. If you can figure out the reason, the descriptions for those can be improved.

Although analysing the new descriptions may indicate gaps in coverage, adding cases to fill those gaps is not for this PR to do.

ErikSchierboom (approved):
LGTM

Note: some tracks have issues with fairly long descriptions. I don't know if the current descriptions would be considered long for those tracks, but I wanted to note it.

exercises/meetup/canonical-data.json
```diff
@@ -1131,7 +1131,7 @@
     },
     {
       "uuid": "00c3ce9f-cf36-4b70-90d8-92b32be6830e",
-      "description": "first Friday of December 2012",
+      "description": "first Friday in a month where it falls on the 7th, the last day of the first week",
```
petertseng, line 1134: I don't really know why this case is all the way down here. but I don't think moving it should be the job of this PR

petertseng:
The longest description is "second Thursday in a month where it falls on another arbitrary day in the middle of the second week" at 99 characters. There are ways of shortening it; we'd just have to decide what information we're willing to sacrifice in the description, and/or what words can be safely dropped while losing no information.

for example, if we changed "in a month where" to "when", we'd have "second Thursday when it falls on another arbitrary day in the middle of the second week" and it'd save 12 characters.

Another solution is to group test cases together, for example all of the `second Tuesday` cases would be grouped together, the "second Tuesday" prefix removed from each of their descriptions, and the test runner would show the group in an appropriately-labeled way, such as

```
second Tuesday
  when it falls on the 8th, the first day of the second week - PASS
  when it falls on the 15th, the last day of the second week - PASS
  when it falls on an arbitrary day in the middle of the second week - PASS
```

that's an option that can be explored if needed, assuming we're comfortable relying on everyone doing the right thing with these descriptions that only make sense when combined with their group description.

ErikSchierboom:
> for example, if we changed "in a month where" to "when", we'd have "second Thursday when it falls on another arbitrary day in the middle of the second week" and it'd save 12 characters.

This is definitely an option. Would changing "when it falls on" to "is" also be an option?

> Another solution is to group test cases together

It is, but only for tracks that support nested testS. I think the track that had issues with longer test names was the C track, which might or might not support nested tests. @wolf99 am I correct in remembering that the C or C++ track had issues with long test names?

wolf99:
The C track does not "like" long test names. Technically they can be used but it results in quite ugly formatting.

For nested test cases, the C track usually concatenates the nesting name to avoid duplicate names.l across different nestings.

I am not sure about the C++ track.

ErikSchierboom:
> The C track does not "like" long test names. Technically they can be used but it results in quite ugly formatting, which can get confusing for humans.

Do you know what the limit is for a long test name?

> For nested test cases, the C track usually concatenates the nesting name to avoid duplicate names across different nestings.

Ah, this is also what I do for C# and F#, so that doesn't really help with the length of the tests.

petertseng:
Okay. So as a comparison:

    second Thursday in a month where it falls on another arbitrary day in the middle of the second week
    when second Thursday is another arbitrary day in the middle of the second week

that saves 21 characters.

ErikSchierboom:
Yeah, that sounds fairly sensible too me. Could "is another arbitrary" be shortened to "is an arbitrary"?

Is that still too much @wolf99?

wolf99:
This seems a good compromise (we don't want to get too convoluted purely to achieve terseness) 😊

petertseng:
For cases with "another", I cannot change to "an" because there is already a case using "an"

ErikSchierboom (approved):
LGTM

ErikSchierboom:
CC @exercism/reviewers

BethanyG (approved):
LGTM

kotp:
> "is another arbitrary"
Can be changed to "any" for either or both of them.  "Arbitrary" and "any" are often interchangeable in meaning (just arbitrary makes one sound smarter?)

petertseng:
thinking about it, you're right.

so for the test cases:

> when first Thursday is an arbitrary day in the middle of the first week
> when first Thursday is another arbitrary day in the middle of the first week

I would then change those two to:

> when first Thursday is some day in the middle of the first week
> when first Thursday is another day in the middle of the first week

thank you for the suggestion!

petertseng:
EBWODP: Your suggestion has been applied and it saves 8 characters in the first case and 10 characters in the second! Great improvement!

junedev (approved):

junedev:
I think the text length are very reasonable now. @petertseng Feel free to merge.

exercises/meetup/canonical-data.json
```diff
@@ -1107,7 +1107,7 @@
     },
     {
       "uuid": "fe0936de-7eee-4a48-88dd-66c07ab1fefc",
-      "description": "last Wednesday of December 2014",
+      "description": "last Wednesday in December that also happens to be the last day of the year",
```
kotp, line 1110: Similarly, yes, it "happens to be" but it also simply "is"

Perhaps:

```suggestion
      "description": "last Wednesday in December that is also the last day of the year",
```

petertseng:
I'd say it's a bit of a policy grey area whether I should be merging it at this time, because even though there are three approvals total, some changes came after some approvals.

In this case I'm using my judgment to say that overall the main changes have been looked by many, and the subsequent changes were small and all had at least two pairs of eyes on each (no one person has made them unilaterally and without review), and are not particularly likely to be contentious.

As you have seen before, with potentially contentious changes I would actually dismiss some prior approvals so that more approvals are needed.

However this is relying on a judgment call (what constitutes "potentially contentious"?), so if I turn out to be in error, let's discuss how to avoid future erroneous judgments in the future. For this one, I take responsibility for any problems that result.

ErikSchierboom:
Thanks @petertseng!

PR #1922, opened 2022-01-18
Remove mention of workflows from README

SaschaMann:
The templates have been moved to the docs a while ago in #1792.


ee7 (approved):

petertseng (approved):
In my review, I have **not** checked whether the new location is linked to in places where it should be, because that is not the job of this PR; the job of this PR is only to remove a nonexistent link.

PR #1924, opened 2022-01-19
Use uuid format in JSON schema

ErikSchierboom:
Closes https://github.com/exercism/problem-specifications/issues/1701﻿


ee7 (commented):
Before this PR, we required that each UUID is version 4.

With this PR, does CI pass for a UUID that isn't version 4? It looks like the answer might be "yes" (in which case, we shouldn't merge this as-is).

ErikSchierboom:
I've just checked a couple of v1/v4 uuids, and the changes in this PR validate both equally, so I'll close this PR (should have been a draft PR BTW 🤦 )

PR #1930, opened 2022-01-23
Updated license date to 2022

yusufadell:


LICENSE
```diff
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2019 Exercism
+Copyright (c) 2022 Exercism
```
kotp, line 3: We might need to have a range instead of a single date.  Copyright exists on original material at the time of creation, but those things that are created after the initial date, then are copyright beyond the first date.  For notice of intent of enforcement of rights, then, there may be a need to correctly indicate first publication year.  Since these things are not created and published at once, the range makes sense to me, even though I am not a lawyer.

I do not think we should be tossing out the earlier year as that might be deemed not on the up and up, really, since it is misleading. (Perhaps some of this information was created and even unmodified since 2013).

Like I said, I am not a lawyer, and surely not a lawyer that is familiar with the complexities of international copyright.  But it does not make sense that only by changing a year, we have extended the copyright of material by another 3 years of material that would have normally come under public domain because of time passing.

kotp (commented):
Confirmation from legal probably required.

LICENSE
yusufadell: Hi Kotp, Here I'm not changing terms of license  it self **License != Copyright**,  Copyright on code is renewable if we made material changes to the code and It is not strictly speaking, related to the license.

if we applied a range instead, it would be at the time of creation not the date it was published thought this isn't necessary.

kotp: To be clear, I did not state that the terms of license were being changed, only the notice of the copyright.  I also did not indicate that licensing is copyright, it is not.  The only thing changed on this line is the copyright notice.  I am asking to be careful that notice is properly given if we are giving notice at all (It is not required for copyright to be in effect, in my jurisdiction).

I did say "first publication year" but I meant "first copyright event" which would be creation.

rpottsoh, line 3:
```suggestion
Copyright (c) 2019-2022 Exercism
```

rpottsoh: An interesting read, https://techwhirl.com/updating-copyright-notices/

yusufadell: Well, thank you for clearing that up. Now i can see that by doing so i you’d be creating an inaccurate copyright notice.

junedev:
@ErikSchierboom Isn't the LICENSE template something we keep in org-wide-files?

ErikSchierboom:
@junedev Only for tracks: https://github.com/exercism/org-wide-files/blob/main/tracks-files/LICENSE The reason for that is that non-track repos sometimes have different licenses.

LICENSE
ErikSchierboom: Very interesting read. As I'm reading it, changing the existing year in the license to the current year is discouraged?

kotp: I think especially here, where things are version controlled, we can a bit more easily show which portions are copyright events.  So it may be appropriate to add dates, but I would not change the date, outright, only append a range or a new year, depending on change activity.

kotp:
I do not think we need to do anything with it, but we might add the year that v3 went online as an additional date, since a lot of new material came in (looking at Concept exercises).  I would be concerned with not notifying that some content is copyright at the earliest year, and losing that information in the notice.

kotp:
> @kotp I would be okay with that. Would the license text then become:
> 
> > Copyright (c) 2019, 2021 Exercism
> 
> Is that correct?

I believe that is almost correct, and for the reasons stated, as well as the license file acknowledging that it is also a copyright notice at that location.

The only problem is that the 2014 copyright notice was removed at some point, and that should still be there.  So, without doing a content review, but in terms of a general overall product copyright, I would probably do:

    Copyright (c) 2014, 2019, 2021 Exercism

ErikSchierboom:
I'm fine with that. @yusufadell could you update using the above suggestion?

PR #1933, opened 2022-01-28
Update "a**b" writting to "a^b" in diffie-hellman description

Gautzilla:
Creation of the diffie-hellman-description-simplifying branch
Rewriting from "a**b" to "a^b" for expressing exponentiation


ErikSchierboom (commented):
Thanks for the pull request! There seem to be some unrelated files in this PR. Could you remove them from the pull request?

ErikSchierboom (approved):

exercises/diffie-hellman/description.md
```diff
@@ -28,11 +28,11 @@ private key b.
...
-    s = B**a mod p
+    s = B^a mod p
 
 Bob calculates
 
-    s = A**b mod p
+    s = A^b mod p
 
 The calculations produce the same result!  Alice and Bob now share
 secret s.
```
rpottsoh, line 38: I would like to suggest adding a note be added at the end of this document that mentions that in some programming languages `**` is used as the method to indicate exponentiations instead of `^`.

I think just replacing one form for another doesn't necessarily clear up the confusion for everyone, just shifts it.

wolf99: Might it be better to remove any implementation dependent form and instead use the mathematical form? That is what is being described after all.

BethanyG: Hard agree with @wolf99 comment.  For example, in Python, `^` is used as a `set` operator, which is why `**` is used for exponentiation.  It varies widely by programming language.

rpottsoh: @wolf99 I like your suggestion, however I worry about how the special character will translate for users.  If there is no concern to worry about then 👍 from me too.

Gautzilla: I feel exactly like @rpottsoh : would `aᵇ` render properly in any IDE for people downloading the exercise through CLI? 


> I would like to suggest adding a note be added at the end of this document that mentions that in some programming languages `**` is used as the method to indicate exponentiations instead of `^`.

Well, the note might just mention that a^b means "a raised to the power of b", or "a*a*a*...*a" b times", so that the description only uses mathematics and no programming conventions?

SaschaMann: We have other exercises that use them. If an IDE can't display standard characters like that, it's most likely a deliberate choice by the student to use archaic IDE and make their life hard.

Gautzilla: Thus, I see no reason not to use this notation. I'll make a new commit.

kotp: > deliberate choice by the student to use archaic IDE

I use `ed` (not always) and it displays it, so I wonder if you mean "IDE that sucks" rather than archaic.

SaschaMann (approved):

kotp (approved):

petertseng:
Please squash when we have files like .vs that should never be in the history

kotp:
I remembered the request that it be removed from the pull request, and since I saw Erik's approval, I did not bother to check  (It was not removed from the pull request, only removed as a file).  I did not bother to check that it was actually removed from the request.

SaschaMann:
I'm surprised the repo settings even allow anything but squash-merges (cc @iHiD)

petertseng:
There will be instances where we do not want to squash. So both options should be available. Please do not make squash the only option.

For the non-squash option I prefer merge over rebase, but that ship has sailed apparently.

petertseng:
EBWODP: We should have the discussion of acceptable ways to merge elsewhere. It is not apparently from the title of this pull request that it may contain this discussion.

kotp:
Who is EBWODP?  I do not know what that means, and there seems to be no indication in this forum what that means.

ErikSchierboom:
> I remembered the request that it be removed from the pull request, and since I saw Erik's approval, I did not bother to check (It was not removed from the pull request, only removed as a file). I did not bother to check that it was actually removed from the request.

Ehm, but it was https://github.com/exercism/problem-specifications/pull/1933/commits/90617beb97a164f981f03678f9116de63005ecfe? There is no such thing as checking in a directory in git, a directory only ever exists if it has files in it.

kotp:
You are right Erik, the reference to the vs file was carried over from the prior reference to that as a file, rather than the files in the directory.

exercises/diffie-hellman/description.md
SleeplessByte: Instead of trying to abstract it away:

```text
s = (A to the power of b) mod p
 ```

SaschaMann: That is far more abstract and less readable

PR #1938, opened 2022-02-01
rational-numbers: additional test for abs value

mstranger:
According to the issue exercism/elixir#1052

### DIFF
--- a/exercises/rational-numbers/canonical-data.json
+++ b/exercises/rational-numbers/canonical-data.json
@@ -261,6 +261,15 @@
             "r": [0, 1]
           },
           "expected": [0, 1]
+        },
+        {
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):

SaschaMann:
This feels redundant and we don't have analogous cases for the other methods either. But I'm not opposed to adding the case.

mstranger:
In other cases it happens implicitly, as I understand. For example for `add([1, 2], [-1, 2])` is expected `[0, 1]`. But using the plain formula we get `[0, 4]`.

SaschaMann:
I think an equivalent and explicit case for `add` would be something like `add([1, 4], [0, 1])` where the identity is added. Same for multiplication etc.

jiegillet (approved):

PR #1941, opened 2022-02-01
Add action to format code

ErikSchierboom:
Depends on https://github.com/exercism/problem-specifications/pull/1917 to be merged

This is almost a carbon copy of the [workflow that the JavaScript track uses](https://github.com/exercism/javascript/blob/main/.github/workflows/action-format.yml).

@iHiD This repo now needs access to the SSH deploy key, like the `javascript` track.

### DIFF
--- a/.github/workflows/action-format.yml
+++ b/.github/workflows/action-format.yml
@@ -1,0 +1,130 @@
+name: Format
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  format:
+    name: 'Format JSON'
+    runs-on: ubuntu-latest
+    if: github.event.issue.pull_request != '' && contains(github.event.comment.body, '/format')
+    steps:
+      - name: 'Post acknowledgement that it will format code'
+        continue-on-error: true # Never fail the build if this fails
+        uses: actions/github-script@e3cbab99d3a9b271e1b79fc96d103a4a5534998c
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            github.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: 'The "Format code" action has started running.'
+            })
+
+      - name: 'Download PR data'
+        run: |
+          PR_DATA="/tmp/pr.json"
+
+          jq -r ".issue.pull_request.url" "$GITHUB_EVENT_PATH" | \
+            xargs curl --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' -o "$PR_DATA" --url
+
+      - name: 'Check fork status'
+        id: fork_status
+        run: |
+          IS_FORK="$(jq '.head.repo.fork' "/tmp/pr.json")"
+          echo "::set-output name=fork::$IS_FORK"
+
+      - name: 'Setup SSH deploy key'
+        if: steps.fork_status.outputs.fork == 'false'
+        run: |
+          mkdir ~/.ssh
+          echo "${{ secrets.DEPLOY_KEY }}" > ~/.ssh/id_ed25519
+          chmod 600 ~/.ssh/id_ed25519
+
+      - name: 'Checkout code'
+        run: |
+          PR_DATA="/tmp/pr.json"
+
+          HEAD_REF=$(jq -r ".head.ref" "$PR_DATA")
+
+          if [ ${{ steps.fork_status.outputs.fork }} == "false" ]; then
+            echo "::debug::Setting up repo using SSH"
+            HEAD_REPO=$(jq -r '.head.repo.ssh_url' "$PR_DATA")
+          else
+            echo "::debug::Setting up repo using HTTPS"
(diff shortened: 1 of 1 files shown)


SaschaMann:
> @iHiD This repo now needs access to the GITHUB_TOKEN secret, like the javascript track.

It has that by default, doesn't it? The secret it needs in addition to the defaults is an SSH deploy key so that pushing triggers CI.

ErikSchierboom:
Ah, makes sense.

.github/workflows/action-format.yml
```diff
@@ -0,0 +1,130 @@
...
+          else
+            echo "::debug::Setting up repo using HTTPS"
+            HEAD_REPO=$(jq -r '.head.repo.clone_url | sub("https://"; "https://x-access-token:${{ secrets.GITHUB_TOKEN }}@")' "$PR_DATA")
+          fi
+
+          git clone $HEAD_REPO .
+          git checkout -b "$HEAD_REF" "origin/$HEAD_REF"
+
+      - name: Use Node.js LTS (16.x)
+        uses: actions/setup-node@1f8c6b94b26d0feae1e387ca63ccbdc44d27b561
```
wolf99, line 64: Any updates needed for `.github/dependabot.yml` for the `uses` actions (such as this one, actions/cache, actions/github-script) ?

.github/workflows/action-format.yml
```diff
@@ -0,0 +1,130 @@
...
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            github.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: 'The "Format code" action has finished running.'
+            })
+
+      - name: 'Post reminder to trigger build manually'
```
wolf99, line 118: A question for my own understanding: in what use case does this step run?

It seems like when it is a fork, this is catered for (e.g. ln # 55 - 58), and there is nothing stopping the other steps running when it is a fork... yet this step seems to imply that the job should not work for a fork?

SleeplessByte: It actually says it in the message!

> For security reasons, `/format` does not trigger CI builds when the PR has been submitted from a fork. If checks were not passing due to code format, trigger a build to make the required checks pass, through one of the following ways:
> 
> - Push an empty commit to this branch: `git commit --allow-empty -m "Trigger builds"`.
> - Close and reopen the PR.
> - Push a regular commit to this branch.

Basically when you _write_ commits (so you use the SECRET token, which we do in the format _write_ action) the CI doesn't trigger. So it's not that you cannot run this action on a forked-pr (you can!), it's that _after_ running this action, it won't re-trigger a CI run from the commit that this action writes.

.github/workflows/action-format.yml
```diff
@@ -0,0 +1,130 @@
...
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      - name: Install dependencies
+        run: yarn install --pure-lockfile
+
+      - name: Verify that canonical-data.json files adhere to their json schema
+        run: yarn format-json
```
SleeplessByte, line 86: This doesn't verify. This writes. Right?

.github/workflows/action-format.yml
SaschaMann: > Basically when you write commits (so you use the SECRET token, which we do in the format write action) the CI doesn't trigger. So it's not that you cannot run this action on a forked-pr (you can!), it's that after running this action, it won't re-trigger a CI run from the commit that this action writes.

This is only true for commits written using the default GitHub Actions token. When the PR is from a branch within this repo, we can use an SSH deploy key to push to it and avoid this issue. That's why the workflow has two cases that are handled differently (fork vs non-fork).

SleeplessByte: Oh yeah, there's that :)

wolf99: Great, thanks for clarifying folks!

.github/workflows/action-format.yml
ErikSchierboom: No, this should be covered by https://github.com/exercism/problem-specifications/blob/main/.github/dependabot.yml#L6. 
See https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-the-dependency-graph#supported-package-ecosystems

.github/workflows/action-format.yml
ErikSchierboom: Correct. I've updated.

wolf99 (approved):

kotp (approved):

SaschaMann (approved):

iHiD:
I think I've done it but I don't really have any idea what I'm doing :)

SaschaMann:
@iHiD in that case, could you approve this PR so we can merge it and we'll try it out in a test PR? :)

iHiD (approved):

PR #1940, opened 2022-02-01
word-search: Add test case for false-positives

mariohuq:
Closes #1791.

PR #1944, opened 2022-02-02
word-search: Add cases checking for concatenation and wrapping

petertseng:
The author of this commit thinks that concatenation is highly unlikely,
but the wrapping might be useful to check in languages that allow
negative indices.

### DIFF
--- a/exercises/word-search/canonical-data.json
+++ b/exercises/word-search/canonical-data.json
@@ -1024,6 +1024,61 @@
         "abf": null,
         "cbd": null
       }
+    },
+    {
+      "uuid": "5b6198eb-2847-4e2f-8efe-65045df16bd3",
+      "description": "Should not concatenate different lines to find a horizontal word",
+      "property": "search",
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):
It is indeed unlikely, but the additional test cases don't hurt

petertseng:
excuse me for rebasing; it was necessary because https://github.com/exercism/problem-specifications/pull/1940 got merged.

kotp:
Rebasing and modifying the history of your own branch is not only normal git workflow, it is expected to happen, so no problems at all @petertseng .  It is the main branch where it causes problem.  Anyone following along would do a `fetch` and a `git reset --hard` to work with you on the branch.  (The review process itself indicates that the commits are not expected to be static, the approval and movement to the main branch is the signal that the history expected to be static, though expanding.)

BethanyG (approved):

SleeplessByte (approved):

PR #1955, opened 2022-02-07
Fix bowling game copy

### LINKED ISSUE #1954: "The Bowling Game Kata at but UncleBob" sounds wrong
The copy for the `source` of the bowling exercise sounds wrong:

> "The Bowling Game Kata **at but UncleBob**"

https://github.com/exercism/problem-specifications/blob/main/exercises/bowling/metadata.toml

The url is `butunclebob.com`, which is presumably why the copy reads the way it does

ceddlyburge:
Fixes #1954

### DIFF
--- a/exercises/bowling/metadata.toml
+++ b/exercises/bowling/metadata.toml
@@ -1,3 +1,3 @@
 blurb = "Score a bowling game."
-source = "The Bowling Game Kata at but UncleBob"
+source = "The Bowling Game Kata from UncleBob"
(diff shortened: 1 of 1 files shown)


SleeplessByte (approved):

angelikatyborska (approved):

ErikSchierboom (approved):
I find it slightly hilarious that we've never noticed this :)

kotp:
I did notice it, checked it, and that is how the name of the site is presented, so I considered it correct.  Notice the top left hand presented as a break in the wording.

However, either way is probably fine.  I just do not consider the original to be an error.

PR #1958, opened 2022-02-09
Reduced rational nr. should be in standard form.

dmolin:
The current instructions fail to mention that a reduced rational number should always be rendered in standard form (without any negative value at the denominator).


ErikSchierboom (approved):

SleeplessByte (approved):

SaschaMann (approved):

SaschaMann:
Thank you!

PR #1959, opened 2022-02-10
phone number: only one problem per test input

astrieanna:
Because the area code is not allowed to start with 0 or 1, inputs designed to elicit other errors should not use area codes that start with either of those digits.

### DIFF
--- a/exercises/phone-number/canonical-data.json
+++ b/exercises/phone-number/canonical-data.json
@@ -98,7 +98,7 @@
     {
       "uuid": "eb8a1fc0-64e5-46d3-b0c6-33184208e28a",
       "reimplements": "63f38f37-53f6-4a5f-bd86-e9b404f10a60",
-      "comment": "make input invalid only due to letters; area code may not start with 1",
+      "comments": "make input invalid only due to letters; area code may not start with 1",
       "description": "invalid with letters",
       "property": "clean",
       "input": {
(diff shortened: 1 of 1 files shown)


SleeplessByte (changes requested):
I do like this suggestion. What @astrieanna seems to be suggesting is that by having the other errors use a format that will also be caught by the first error, is that an incorrect implementation would actually be able to go through, or at least make it more difficult for a student to figure out where they went amiss.

@astrieanna As you can see [the CI failed](https://github.com/exercism/problem-specifications/runs/5136636383?check_suite_focus=true). That is because tests in this definiton are _immutable_. Instead, what you'll want to do, if you're up for it, is _duplicate_ the test cases, generate new ids (You can use https://www.uuidgenerator.net/version4) and add `reimplements` as key with the _old_ UUID. 

This will indicate that a new test has replaced an old one. We usually use the `comments` field to indicate _why_.

Feel free to dismiss this review once you've made the changes.

@ErikSchierboom @petertseng your input is welcome.

ErikSchierboom:
This is a great idea!

As @SleeplessByte said, the test cases are immutable, so could you apply what @SleeplessByte suggested and create new, duplicated test cases with the input changed and their `reimplements` key set to the corresponding old test case's `uuid` value?

astrieanna:
Thanks for interpreting my terse description correctly. 😅 My thought was that checking multiple errors at once both:

1. Rejects some otherwise correct implementations by requiring a particular hierarchy of error messages.
2. Gives confusing feedback to some broken implementations (since there's an "incorrect" error message would also be true description of the input).

I decided that one of the three I'd originally changed didn't make sense. In that test, the input is a 9-digit number that starts with a 1; the starting 1 could be interpreted as the country code, so there's no other error that would make sense to return instead of an error about too few digits.

ErikSchierboom (approved):
LGTM!

angelikatyborska (approved):
Very good idea 👏

SaschaMann (approved):

SleeplessByte:
It definitively was ok. Thanks everyone <3

PR #1960, opened 2022-02-11
Reorder keys

ErikSchierboom:
This PR makes the order of the keys in the `canonical-data.json` files consistent and adds a CI check to verify its order.
This is just a minor improvement, so if people are not convinced this is useful, I'm happy to not go through with this.

See https://github.com/ErikSchierboom/problem-specifications/runs/5158116886?check_suite_focus=true for an example of the output.

Closes https://github.com/exercism/problem-specifications/issues/1705

### DIFF
--- a/bin/check_key_order.rb
+++ b/bin/check_key_order.rb
@@ -14,12 +14,13 @@ end
 
 exit_code = 0
 
-Dir.glob('exercises/*/canonical-data.json').each do |path|
+root_dir = "#{__dir__}/.."
+Dir.glob("#{root_dir}/exercises/*/canonical-data.json").each do |path|
   json = JSON.parse(File.read(path))
   invalid_tests = find_tests(json).select {|test| (CORRECT_ORDER & test.keys) != test.keys}
   next if invalid_tests.empty?
 
-  puts "The following tests in #{path} use the wrong key order:"
+  puts "The following tests in #{path.delete_prefix("#{root_dir}/")} use the wrong key order:"
   invalid_tests.each do |test|
     puts "- Test: #{test['uuid']} (#{test['description']})"
     puts "  Actual: #{test.keys}"
--- a/bin/check_key_order.rb
+++ b/bin/check_key_order.rb
@@ -1,0 +1,33 @@
+#!/usr/bin/env ruby
+
+CORRECT_ORDER = %w[uuid reimplements description comments scenarios property input expected]
+
+require 'json'
+
+def find_tests(json)
+  if json['cases']
+    json['cases'].flat_map {|cases| find_tests(cases)}
+  elsif json['uuid']
+    [json]
+  end
+end
+
+exit_code = 0
+
+Dir.glob('exercises/*/canonical-data.json').each do |path|
+  json = JSON.parse(File.read(path))
+  invalid_tests = find_tests(json).select {|test| (CORRECT_ORDER & test.keys) != test.keys}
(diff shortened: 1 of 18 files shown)


bin/check_key_order.rb
```diff
@@ -0,0 +1,33 @@
...
+  elsif json['uuid']
+    [json]
+  end
+end
+
+exit_code = 0
+
+Dir.glob('exercises/*/canonical-data.json').each do |path|
+  json = JSON.parse(File.read(path))
+  invalid_tests = find_tests(json).select {|test| (CORRECT_ORDER & test.keys) != test.keys}
```
SleeplessByte, line 19: Nice

bin/check_key_order.rb
```diff
@@ -0,0 +1,33 @@
...
+  if json['cases']
+    json['cases'].flat_map {|cases| find_tests(cases)}
+  elsif json['uuid']
+    [json]
+  end
+end
+
+exit_code = 0
+
+Dir.glob('exercises/*/canonical-data.json').each do |path|
```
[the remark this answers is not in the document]
ErikSchierboom: Good idea!

SaschaMann (approved):
Can't read the Ruby script but otherwise this looks good

wolf99 (approved):
I don't know about the Ruby, but the resulting JSON looks good.

bin/check_key_order.rb
```diff
@@ -0,0 +1,33 @@
...
+  json = JSON.parse(File.read(path))
+  invalid_tests = find_tests(json).select {|test| (CORRECT_ORDER & test.keys) != test.keys}
+  next if invalid_tests.empty?
+
+  puts "#{path} has tests with the wrong key order:"
+  invalid_tests.each do |test|
+    puts "- Test: #{test['uuid']}"
```
SaschaMann, line 24: Perhaps this should include the description similarly to the mutation-check so that it's easier to find without CTRL F

ErikSchierboom:
@SaschaMann I've added the description to the output.

How do we want to merge this? As on squashed commit, two commits or many individual commits? My preference would be for the second option, but I'm fine with whatever.

SaschaMann:
I like it if the resulting changes are bundled with the script, that way it makes it easier to understand what actually changed if you can't read workflows or Ruby files. 1st and 2nd option both seem fine though, the 3rd seems a bit too noisy.

SleeplessByte:
I would just squash it.

bin/check_key_order.rb
petertseng, line 17: this has made it so that this can only be run from the root of the `problem-specifications` repo. If you want to instead make it runnable from anywhere, you should express the path relative from `bin`. 

```suggestion
Dir.glob("#{__dir__}/../exercises/*/canonical-data.json").each do |path|
```

But, I do not actually expect wanting to run it from anywhere else to be something that is often desired, so I'm Approving (in the GitHub sense of the word) anyway regardless of whether this change is made.

petertseng (approved):
No particular preference between two commits vs one.

bin/check_key_order.rb
ErikSchierboom: I've applied the change, with a minor change to allow the printed path to be slightly clearer

iHiD (approved):

ErikSchierboom:
Thanks everyone!

PR #1963, opened 2022-02-15
bank-account: remove additional notes

ErikSchierboom:
These additional notes are, while useful in general, not something we should be adding to a single exercise. For that a track can used its shared files.


SaschaMann (approved):
Can't have fun with bank accounts anymore, huh? :(

BethanyG (approved):

PR #1965, opened 2022-02-16
Make queen-attack clearer for non-chess players

T-O-R-U-S:
- Add an explanation for what B and W are (the queens)
- Add examples of queens seeing each other, and not interacting.

### DIFF
--- a/exercises/queen-attack/description.md
+++ b/exercises/queen-attack/description.md
@@ -8,6 +8,7 @@ A chessboard can be represented by an 8 by 8 array.
 
 So if you are told the white queen is at `c5` (zero-indexed at column 2, row 3) and the black queen at `f2` (zero-indexed at column 5, row 6), then you know that the set-up is like so:
 
+###### B and W are the queens, standing for Black and White. This is not relevant info to the puzzle.
 ```text
   a b c d e f g h
 8 _ _ _ _ _ _ _ _ 8
@@ -22,4 +23,36 @@ So if you are told the white queen is at `c5` (zero-indexed at column 2, row 3) 
 ```
 
 You are also be able to answer whether the queens can attack each other.
-In this case, that answer would be yes, they can, because both pieces share a diagonal.
+In this case, that answer would be yes; they can, because both pieces share a diagonal.
+
+### Examples of queens attacking:
+
+```text
+_ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
+_ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _     _ _ _ W _ _ _ _
+_ _ _ W _ _ _ _     _ B _ _ _ W _ _     _ _ _ _ _ _ _ _
+_ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
+_ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
+_ _ _ _ _ _ B _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
+_ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _     _ _ _ B _ _ _ _
+_ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
+```
+
+### Examples of queens not interacting:
+
+```text
+
+_ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
+_ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _     _ _ _ W _ _ _ _
+_ _ _ _ _ _ _ _     _ _ _ _ _ W _ _     _ _ _ _ _ _ _ _
+_ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
+_ _ _ _ W _ _ _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
+_ _ _ _ _ _ B _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
+_ _ _ _ _ _ _ _     _ _ B _ _ _ _ _     _ _ _ _ _ _ B _
+_ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
+```
+
+B and W stand for **Black** and **White**, the two sides competing
+against each other in a game of chess, which is why each queen
+is labelled B and W; though you do not need to know which side
+the queens are on to solve this challenge.
--- a/exercises/queen-attack/description.md
+++ b/exercises/queen-attack/description.md
@@ -52,7 +52,7 @@ _ _ _ _ _ _ _ _     _ _ B _ _ _ _ _     _ _ _ _ _ _ B _
 _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
 ```
 
-B and W stand for **Black** and **White**, the two sides competing
-against each other in a game of chess, which is why each queen
-is labelled B and W; though you do not need to know which side
-the queens are on to solve this challenge.
+~~~~exercism/note
+B and W are the queens, standing for Black and White.
+This is not relevant info to the puzzle.
+~~~~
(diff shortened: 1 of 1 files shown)


SleeplessByte:
cc @exercism/reviewers

exercises/queen-attack/description.md
```diff
@@ -8,6 +8,7 @@ A chessboard can be represented by an 8 by 8 array.
 
 So if you are told the white queen is at `c5` (zero-indexed at column 2, row 3) and the black queen at `f2` (zero-indexed at column 5, row 6), then you know that the set-up is like so:
 
+###### B and W are the queens, standing for Black and White. This is not relevant info to the puzzle.
```
SaschaMann, line 11: This would make more sense as an [admonition](https://exercism.org/docs/building/markdown/markdown#h-special-blocks-sometimes-called-admonitions).

SaschaMann (commented):
I don't personally see the value of adding info to the description that is, as you say yourself, not relevant information to the puzzle, but if we decide to add it, it should be done as an advanced admonition. It meets the description for that kind of admonition perfectly:

> **Information that is only relevant for people who want to dig more deeply into something** or are expected to have more advanced knowledge.

exercises/queen-attack/description.md
```diff
@@ -22,4 +23,36 @@ So if you are told the white queen is at `c5` (zero-indexed at column 2, row 3)
...
+_ _ _ _ W _ _ _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
+_ _ _ _ _ _ B _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
+_ _ _ _ _ _ _ _     _ _ B _ _ _ _ _     _ _ _ _ _ _ B _
+_ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
+```
+
+B and W stand for **Black** and **White**, the two sides competing
+against each other in a game of chess, which is why each queen
+is labelled B and W; though you do not need to know which side
+the queens are on to solve this challenge.
```
SaschaMann, line 58: The [Style Guide](https://exercism.org/docs/building/markdown/markdown) requires one sentence per line.

T-O-R-U-S:
> I don't personally see the value of adding info to the description that is, as you say yourself, not relevant information to the puzzle, but if we decide to add it, it should be done as an advanced admonition imo. It meets the description for that kind of admonition perfectly:
> 
> 
> 
> > **Information that is only relevant for people who want to dig more deeply into something** or are expected to have more advanced knowledge.

I believe that it has value though, as B and W may be confusing to people trying to solve the puzzle. Perhaps B and W should just be changed to Q, standing for queen? Thought this might break some prior solutions to the problem.

exercises/queen-attack/description.md
SleeplessByte: Yes, this should not be a heading. I think having the content here is a good idea by @T-O-R-U-S , but I would wrap it in:

`````markdown
~~~~exercism/note
B and W are the queens, standing for Black and White.
This is not relevant info to the puzzle.
~~~~
`````

File: exercises/queen-attack/description.md [hunk 3]
```diff
@@ -22,4 +23,36 @@ So if you are told the white queen is at `c5` (zero-indexed at column 2, row 3)
...
+```text
+_ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
+_ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _     _ _ _ W _ _ _ _
+_ _ _ W _ _ _ _     _ B _ _ _ W _ _     _ _ _ _ _ _ _ _
+_ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
+_ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
+_ _ _ _ _ _ B _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
+_ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _     _ _ _ B _ _ _ _
+_ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
+```
```
SleeplessByte, line 39: This could use an explanation of what you're seeing.

exercises/queen-attack/description.md
```diff
@@ -22,4 +23,36 @@ So if you are told the white queen is at `c5` (zero-indexed at column 2, row 3)
...
+
+_ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
+_ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _     _ _ _ W _ _ _ _
+_ _ _ _ _ _ _ _     _ _ _ _ _ W _ _     _ _ _ _ _ _ _ _
+_ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
+_ _ _ _ W _ _ _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
+_ _ _ _ _ _ B _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
+_ _ _ _ _ _ _ _     _ _ B _ _ _ _ _     _ _ _ _ _ _ B _
+_ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _     _ _ _ _ _ _ _ _
+```
```
SleeplessByte, line 53: This could use an explanation of what you're seeing.

exercises/queen-attack/description.md
angelikatyborska: @T-O-R-U-S I see that you already added the proposed note, but also make sure to remove this heading.

exercises/queen-attack/description.md
angelikatyborska: An example of queens attacking already exists in the description. I am not convinced that we need more. The textual description `In the game of chess, a queen can attack pieces which are on the same row, column, or diagonal.` sounds clear enough to me.

However, if we do want more examples, then:
- They should all use the same notation, the existing ones have labels "a b c..." and "1 2 3...".
- They should be grouped together, not scattered in two different places.
- +1 to @SleeplessByte that each example should come with an explanation.

angelikatyborska (commented):
I think adding a note explaining "B" and "W" is a nice idea, especially for non-native speakers.

PS: Make sure to fix all of the linting issues.

exercises/queen-attack/description.md
ErikSchierboom: I'm also of the opinion that these examples are not needed. I do like the explaining of B and W.

exercises/queen-attack/description.md
kotp: Indeed, we could use `Q` for both, since it is explained that the side does not matter, the black nor the white makes a difference.

exercises/queen-attack/description.md
```diff
@@ -9,17 +9,41 @@ A chessboard can be represented by an 8 by 8 array.
...
+  a b c d e f g h       a b c d e f g h       a b c d e f g h
+8 _ _ _ _ _ _ _ _ 8   8 _ _ _ _ _ _ _ _ 1   8 _ _ _ _ _ _ _ _ 8
+7 _ _ _ _ _ _ _ _ 7   7 _ _ _ _ _ _ _ _ 2   7 _ _ _ W _ _ _ _ 7
+6 _ _ _ W _ _ _ _ 6   6 _ B _ _ _ W _ _ 3   6 _ _ _ _ _ _ _ _ 6
+5 _ _ _ _ _ _ _ _ 5   5 _ _ _ _ _ _ _ _ 4   5 _ _ _ _ _ _ _ _ 5
+4 _ _ _ _ _ _ _ _ 4   4 _ _ _ _ _ _ _ _ 5   4 _ _ _ _ _ _ _ _ 4
+3 _ _ _ _ _ _ B _ 3   3 _ _ _ _ _ _ _ _ 6   3 _ _ _ _ _ _ _ _ 3
+2 _ _ _ _ _ _ _ _ 2   2 _ _ _ _ _ _ _ _ 7   2 _ _ _ B _ _ _ _ 2
+1 _ _ _ _ _ _ _ _ 1   1 _ _ _ _ _ _ _ _ 8   1 _ _ _ _ _ _ _ _ 1
+  a b c d e f g h      a b c d e f g h        a b c d e f g h
```
angelikatyborska, line 21: @T-O-R-U-S You have modified the first example of queens attacking so that it no longer matches the text before it. Not only the positions are different than in the text (the example shows d6 and g3, not c5 and f2 as the text says), but the text is clearly written for a single example, not for 3. Maybe we can just revert this part to the original and only leave the three new examples of queens not attacking?

Additionally, the CI is complaining about linting problems. Those need to be fixed before we can approve the PR.

safwansamsudeen:
@T-O-R-U-S are you still willing to work on this or can I take it up?

IsaacG:
There's been multiple comments calling out that the docs are pretty clear and that this change may not add much value. Unless anyone feels there's a need to clarify things, I propose leaving well enough alone and closing this PR. Or, rather, given how long this PR has been open, I'll just close it out. Feel free to reopen if you think the problem would benefit from clarifications.

PR #1966, opened 2022-02-17
allow prettier to format more files

petertseng:
It's my opinion that for these 15 files, we should not have a strong
opinion on the formatting, and therefore should just accept what
prettier gives us.

This leaves only 18 files ignored. That leaves much less manual work
needed.


petertseng (commented):
you will notice that some of these appear to cause internal inconsistency, in that some JSON files an array will go on a single line, while in the same JSON file some other arrays will have each element on their own line. While I regret the inconsistency, I believe it's simpler to let `prettier` do as much as it can, because whatever tool we write to reformat the remaining files is maintained by us, and the less work it has to do, the less work it is to maintain it.

None of the files formatted in this PR are files where the array elements are representing rows of a 2D grid, so the formatting doesn't affect our ability to understand the input.

If the idea is that test cases are immutable, we don't run into problems where adding an element to an array causes a larger diff than necessary, since we simply cannot add elements to those arrays.

exercises/spiral-matrix/canonical-data.json
```diff
@@ -52,10 +52,10 @@
         "size": 4
       },
       "expected": [
-        [ 1,  2,  3, 4],
+        [1, 2, 3, 4],
```
petertseng, line 55: while I regret that we are losing the padding, I do not think it is important enough to preserve. I do not think losing the padding is going to cause trouble understanding this expected output.

exercises/book-store/canonical-data.json
```diff
@@ -171,7 +171,9 @@
       ],
       "property": "total",
       "input": {
-        "basket": [1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2, 3, 3, 3, 3, 3, 3, 4, 4, 5, 5]
+        "basket": [
+          1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2, 3, 3, 3, 3, 3, 3, 4, 4, 5, 5
+        ]
```
SaschaMann, line 176: See #1917 why this is not desirable.

exercises/dominoes/canonical-data.json
```diff
@@ -136,7 +161,13 @@
...
       "property": "canChain",
       "input": {
-        "dominoes": [[1, 2], [2, 3], [3, 1], [2, 4], [2, 4]]
+        "dominoes": [
+          [1, 2],
+          [2, 3],
+          [3, 1],
+          [2, 4],
+          [2, 4]
+        ]
```
SaschaMann, line 170: See #1917 why this is not desirable.

exercises/pascals-triangle/canonical-data.json
```diff
@@ -75,7 +82,18 @@
...
+        [1, 1],
+        [1, 2, 1],
+        [1, 3, 3, 1],
+        [1, 4, 6, 4, 1],
+        [1, 5, 10, 10, 5, 1],
+        [1, 6, 15, 20, 15, 6, 1],
+        [1, 7, 21, 35, 35, 21, 7, 1],
+        [1, 8, 28, 56, 70, 56, 28, 8, 1],
+        [1, 9, 36, 84, 126, 126, 84, 36, 9, 1]
+      ]
```
SaschaMann, line 96: Perhaps these should _all_ be triangles in this file.

SaschaMann (dismissed):

exercises/book-store/canonical-data.json
ErikSchierboom: Are you referring to:

> This makes copy and pasting harder for no reason.

If so, I don't really see why the new format is much harder, nor why copy-pasting is important here.

ErikSchierboom (approved):
I approve of all these changes.

exercises/dominoes/canonical-data.json
petertseng: If I understand correctly, the idea here is that seeing `[1, 2], [2, 3], [3, 1]` all next to each other makes it more visually apparent that they could connect to each other, and that is lost if putting them on separate lines.

I have thought about it and I am sufficiently convinced.

May I remind from the standpoint of a person intending to write the formatting tool that I expect the extra maintenance cost incurred by this to be small. Given it is small, I think it's worth it.

exercises/pascals-triangle/canonical-data.json
petertseng: I will note this for consideration when writing the array formatter. After it is incorporated into CI I will try applying that to pascals-triangle and see how it looks. For the time being, I let prettier do as it will.

exercises/book-store/canonical-data.json
SaschaMann: > If so, I don't really see why the new format is much harder, nor why copy-pasting is important here.

Line breaks make it annoying to copy into a REPL or into the code file (which is likely `foo([...])`). It's important because it saves a lot of time when manually implementing an exercise. For tracks that use generators, the formatting is irrelevant, so I don't see why we shouldn't optimise it for human-use whenever possible. I don't see any advantage of the new format for reading or implementing the exercise, only disadvantages.

exercises/dominoes/canonical-data.json
SaschaMann: > If I understand correctly, the idea here is that seeing [1, 2], [2, 3], [3, 1] all next to each other makes it more visually apparent that they could connect to each other, and that is lost if putting them on separate lines.

Indeed.

exercises/pascals-triangle/canonical-data.json
SaschaMann: Sounds good.

exercises/book-store/canonical-data.json
SaschaMann: > However. I do need to gain clarification on which exercises this idea needs to be applied to. I apologise for engaging in what-about-ism, but understanding this concern is key to making sure we are helping track maintainers and our future selves. It seems to me like book-store, bowling, change, and variable-length-quantity would fall under the same category here - that their inputs would all benefit from being on the same line as their []. That would be four exercises. Please confirm whether my understanding is correct.

Yes, that sounds right. I'm not sure if there are other exercise but those are the main ones where it makes a big difference.

Sometimes the added line breaks add readability (see Pascal's Triangle) but for a list of numbers like in these exercises, I don't see that being the case.

SaschaMann (approved):

exercises/book-store/canonical-data.json
ErikSchierboom: > Line breaks make it annoying to copy into a REPL or into the code file
> It's important because it saves a lot of time when manually implementing an exercise.

Well, many tracks don't _have_ a REPL nor do I think that "it saves a lot of time" is necessarily true (I feel that is an assumption). 
That said, we _should_ indeed optimize for human-readability, so if it is do-able for Peter to make these exceptional cases, then I am fine with it.

SaschaMann: > Well, many tracks don't have a REPL nor do I think that "it saves a lot of time" is necessarily true (I feel that is an assumption).

Do those tracks gain anything from spreading it out over several lines? Maybe there are advantages, I just can't think of any.


> (I feel that is an assumption).

Try it and measure it yourself if you want, I don't see how it wouldn't be slower to remove line breaks for every paste compared to just pasting them. 🤷

ErikSchierboom:
Requesting a review from @exercism/reviewers

kotp (approved):
Nothing looks awry, the formatting seems to be reasonable.  The things that are meant to be formatted different appear to be not affected in a harmful way.

PR #1967, opened 2022-02-17
auto-format portions of some JSON files

petertseng:
this commit was made by having prettier format these files and
committing portions of the changes.

For now we want prettier to continue to ignore these files, but
eventually the plan would be to use prettier but to reformat specific
portions.

The portions in this commit are *not* the portions we would want to
reformat, so for these specific portions we are satisfied to accept
prettier's changes.

Issue #1934, opened 2022-01-31
Decide merge options for repo


petertseng:
A discussion was starting in https://github.com/exercism/problem-specifications/pull/1933, but it should be moved into its own discussion otherwise we won't be able to find it from the title.

Originally, I reacted harshly to the idea that squashing should be the only option. But after considering it, I realise I should not be so harsh. It is still my _preference_ that a non-squash option remain available, in case there are instances where we do not want to squash. However, given this is a repo shared among many, if the consensus is that the advantage of only enabling squashes outweighs the drawbacks, I accept that consensus.

Advantages: simplifies decisions and avoids questions of "should this be squashed" - if it's the only choice, no more questions!
Drawbacks: Makes contributors' lives harder in instances where they wish to contribute commits that should not be squashed - they'll be forced to make one PR for each.

SaschaMann:
> Originally, I reacted harshly to the idea that squashing should be the only option. But after considering it, I realise I should not be so harsh. It is still my preference that a non-squash option remain available, in case there are instances where we do not want to squash. However, given this is a repo shared among many, if the consensus is that the advantage of only enabling squashes outweighs the drawbacks, I accept that consensus.

In other larger repos like the former v3 repo it was rare that a PR was in a state where it shouldn't have been squashed. Given that the vast majority of PRs in this repo only edit one exercise, I imagine it'll be similar here. However, in those cases there's still the option to have it rebase-merged by an admin as they can override the settings that are available.

ErikSchierboom:
> It is still my preference that a non-squash option remain available, in case there are instances where we do not want to squash.

This is also my preference, given that we do sometimes have PRs with multiple commits that we'd like to merge as individual commits.

> Advantages: simplifies decisions and avoids questions of "should this be squashed" - if it's the only choice, no more questions!

A key question is: how often does this occur (having a discussion on the merge strategy)?

> However, in those rare cases there's still the option to have it rebase-merged by an admin as they can override the settings that are available.

This could also work., but would effectively make admins a bottleneck. Not to say that I think that won't work in practice, but just wanted to point it out.

SaschaMann:
> A key question is: how often does this occur (having a discussion on the merge strategy)?

I think the main concern is the opposite. This discussion doesn't happen, someone merges without squashing, and the commit history gets cluttered (see the recent commits)

ErikSchierboom:
Ah, that _is_ a very good argument.

ErikSchierboom:
[...]

PR #1968, opened 2022-02-19
add format-array (formats JSON arrays)

petertseng:
format-array expresses our per-file preferences for array formatting, in
cases where the preference differs from prettier.

To see that this is true, the following procedure can
be used:

1. Delete or truncate .pretterignore and run prettier on JSON files.
   prettier will have reformatted them.
2. run format-array. It will restore the files to their original state.

---

This is ready for review.

The below are historical blockers that had to be solved before it is ready for review, but they are now solved:

* ~~#1966 and #1967 need to be merged first (notice that this PR includes commits from both of those PRs). Reviews of this PR should be limited solely to https://github.com/exercism/problem-specifications/pull/1968/commits/0196b9e844f4834418351589cbdae095ca9053f0 and https://github.com/exercism/problem-specifications/pull/1968/commits/c384a84d5326cf6f32108cb89260cb4c6c950f53. Reviews of other commits should go in their respective PRs.~~ done
* ~~#1978 needs to be merged first~~ done
* ~~It should be incorporated into CI before being merged (and that will need review when the time comes as well)~~ done

suggested items for review:

* comment on whether any portions of the code could be made clearer.
* comment on where the per-file preferences should be. I see two main options:
    * in a single file (currently, just directly in format-array). Advantage: Most easily allows us to see in one place all the exceptions we're making to prettier and where they are. analogous to current .prettierignore where all the exceptions are in that one file.
    * somewhere associated with each exercise (for example, the preferences for `book-store` could be in `exercises/book-store/canonical-data.json` or some other file in `exercises/book-store`). Advantage: If I'm just interested in editing a single exercise's JSON file, I only need to look in that JSON file to find the rules, rather than knowing that I need to look in `bin/format-array.rb` as well.
* a reminder that the reason for the existence of format-array was because we could not find a tool that allowed us to specify array formatting, so we had to do it ourselves. If you do know of such a tool, you should let us know so that we can explore using it instead of format-array. Items to consider:
  * https://www.npmjs.com/package/prettier-plugin-multiline-arrays - a recent creation (2022-02-07), which is why we had not heard of it before making this PR. However we would need to conditionally include it based on which file was being formatted, and we also don't really have equivalents for `padded`, and it can't correctly format `dominoes`, `triangle`, or `saddle-points`, where the multiline-ness should descend only one level deep into an array of arrays.
  * the `--print-width` flag for prettier: I haven't explored this to find out whether we can specify a value for `print-width` that correctly does what we want. I suspect it would be a little difficult, since our decision is not based on line length or width.


petertseng:
There are a few `multi_line` that I feel could be changed into `multi_line_unless_single`, but for now I'm keeping the existing formatting how it is. If you feel some of the `multi_line` can be changed to `multi_line_unless_single`, I suggest enacting those changes in a separate PR, where that idea can have its dedicated area for discussion. If your PR gets merged before mine, I will gladly rebase mine and change the rule accordingly. If my PR gets merged first, you will have to rebase so you have format-array, and you wil have to change the rule. I do not plan to enact any `multi_line` -> `multi_line_unless_single` changes (or any other rule changes, in fact) in this particular PR.

bin/format-array.rb
```diff
@@ -0,0 +1,180 @@
+require 'json'
+
+# format-array expresses (and enacts) our per-file preferences for array formatting,
+# in cases where the preference differs from prettier.
+#
+# It has only one mode of operation:
+# Run it without any arguments to format the files.
```
petertseng, line 7: You may suggest that I could add a `--check` flag to have it exit 1/0 depending on whether files need to be formatted, so that it can be incorporated into CI

Actually, my plan to incorporate it into CI was just to run it and check `git diff`. The advantage of that is that we could even show the `git diff` output in the CI, which will give contributors an idea of what changes they need to make. That seems to me to be more useful than just saying "you need to make some changes".

SaschaMann: We've done the same in the past in the v3-repo, you can probably copy parts of the format workflows there to set it up.

exercises/ocr-numbers/canonical-data.json
```diff
@@ -68,7 +68,9 @@
           "    "
         ]
       },
-      "expected": {"error": "Number of input columns is not a multiple of three"}
+      "expected": {
+        "error": "Number of input columns is not a multiple of three"
+      }
```
ErikSchierboom, line 73: Should this be (similar to the above change):

```suggestion
      "expected": { "error": "Number of input columns is not a multiple of three" }
```

exercises/pascals-triangle/canonical-data.json
```diff
@@ -66,7 +66,14 @@
       "input": {
         "count": 6
       },
-      "expected": [[1], [1, 1], [1, 2, 1], [1, 3, 3, 1], [1, 4, 6, 4, 1], [1, 5, 10, 10, 5, 1]]
+      "expected": [
```
ErikSchierboom, line 69: Is it possible to also have the other expected values formatted with one element per line? That way one really gets a visual indication of the triangle shape

bin/format-array.rb
```diff
@@ -0,0 +1,178 @@
...
+    'grid' => :multi_line,
+    'wordsToSearchFor' => :multi_line,
+  },
+}.each_value(&:freeze).freeze
+
+def single_line_arrays(contents, key)
+  # matches things of the form "key": [1, 2, 3]
+  # because this is not a multi-line regex,
+  # . does NOT match newlines.
+  contents.scan(/^( +)"#{key}": (\[.*\],?$)/)
```
ErikSchierboom, line 94: Theoretically speaking, this (using a regex) could also match the text in a comment, but that is both highly unlikely and pretty noticeable in a PR.

petertseng: This is a fascinating case to consider. Presumably it would be something like

```json
{
  "comment": [
    "consider the following input",
    "foo\": [1, 2, 3]"
  ]
}
```

That doesn't particularly look good in JSON, and for this to match you have to pass a `key` that includes a `\`... it's indeed pretty noticeable on both ends.

exercises/ocr-numbers/canonical-data.json
petertseng: that would not be in line with `prettier`. `format-array` has no intention of formatting objects or other non-array items

exercises/pascals-triangle/canonical-data.json
petertseng: with `format-array`, very possible! I will add it to the list of things to be done after this PR. It will be as simple as having `expected` be `multi_line`.

petertseng:
PR description has been edited to add:

Reviews of this PR should be limited solely to https://github.com/exercism/problem-specifications/pull/1968/commits/0196b9e844f4834418351589cbdae095ca9053f0 and https://github.com/exercism/problem-specifications/pull/1968/commits/c384a84d5326cf6f32108cb89260cb4c6c950f53. Reviews of other commits should go in their respective PRs.

ErikSchierboom (approved):
Approving based on the two mentioned commits.

petertseng:
Now that the other two PRs are merged (I have rebased it on main now that they are, so this PR now only contains the relevant commits), we could go about this two possible ways:

1. I could take this PR out of drafts right now, and if it gets approved, it can be merged, without waiting for it to be incorporated into CI. The reason to do this would be so reviewers have to review fewer changes at one time.
2. Or I could leave it as a draft until such time as I do incorporate it into CI. The reason to do this would be that it's less churn if it so happens that I find out only while trying to incorporate it into CI that I need to make a change, and so that reviewers can see the whole picture of how this fits in when reviewing (be able to see how the trees make up the forest, so to speak) 

I prefer the latter so I will be doing that unless I hear otherwise.

petertseng:
amendment: https://github.com/exercism/problem-specifications/pull/1978 will also need to be merged before I would take this PR out of drafts, but I will look toward incorporating this into CI in the coming days.

petertseng:
I now think this is ready for review. See https://github.com/exercism/problem-specifications/runs/5430556582?check_suite_focus=true, which is the CI run for #1981, to see what happens now if a JSON file's formatting does not match prettier + format-array.

petertseng:
EBWODP: The CI run is now https://github.com/exercism/problem-specifications/runs/5430593239?check_suite_focus=true instead

CONTRIBUTING.md
```diff
@@ -336,4 +336,5 @@ Note: if you use VS Code as your editor, you can install the [prettier plugin][p
 [improve-exercise-metadata]: https://github.com/exercism/legacy-docs/blob/main/you-can-help/improve-exercise-metadata.md
 [legacy-docs]: https://github.com/exercism/legacy-docs
 [prettier]: https://prettier.io/
+[format-array]: https://github.com/exercism/problem-specifications/blob/main/bin/format-array.rb
```
petertseng, line 339: notice! this link does not work yet for the obvious reason that `format-array` doesn't exist on `main` yet! I am not sure if I have a better solution to offer for review, other than "you will just have to check this link very carefully"

Or, this change could be removed from this PR and only submitted in a separate PR after format-array.rb has been merged, so that all can verify that the link is correct, but I was not convinced that was a better solution.

petertseng:
notice! I have just found out that `format-array.rb` uses hyphens to separate words in its filename, whereas other files in https://github.com/exercism/problem-specifications/tree/main/bin are inconsistent in whether they use hyphens or underscores, but crucially the other Ruby script [`check_key_order.rb`](https://github.com/exercism/problem-specifications/blob/main/bin/check_key_order.rb) uses underscores. If this is a problem, just say the word and I will change `format-array.rb` to `format_array.rb`.

petertseng:
Recall that I also put out a call for anyone to put forth alternative ways of formatting arrays. I have added discussion of two potential options to the initial PR comment (https://www.npmjs.com/package/prettier-plugin-multiline-arrays, and the `--print-width` flag), but I was not convinced at this time they are suitable.

ErikSchierboom (commented):
Looks good!

ErikSchierboom:
This has been lying dormant for a while. @exercism/reviewers could you maybe review this PR?

SaschaMann (approved):

iHiD (approved):

petertseng:
All right! So, I have a weak preference for not squashing these commits. Nobody else who reviewed expressed an opinion either way, so that could mean nobody has an opinion or just that it wasn't on the forefront of their minds at the time, and that's fine. I know that I could ask Erik to merge it without squashing. So @ErikSchierboom  I'm going to tentatively ask for that. If it seems better to just squash just say the word and that's fine with me.

kotp (approved):

ErikSchierboom:
Ah, it turns out I don't have enough permissions to do a merge commit. @iHiD could you merge this without squashing?

iHiD:
I don't have permission for that either :) Can I rebase+merge?

SaschaMann:
It's still inked in the UI, I think. But I'd just squash it tbh

ErikSchierboom:
@petertseng ?

ee7:
I can't claim to know @petertseng's current thoughts, but I remember them writing [a nice post](https://github.com/exercism/problem-specifications/pull/1501#issuecomment-482646134) on this topic years ago.

To expand a little, I'd agree that the best practice is a merge commit in this situation. Ideally after rebasing the PR branch on the most recent `main` commit, so that:
- The history is as simple as possible (`git log --graph` has minimal noise).
- CI tests the merge. Despite the fact that "checks have passed" on this PR, CI hasn't actually checked the result of pressing the big green button - it's only checked the result of merging with a past `main` commit. This problem isn't limited to this PR - it exists whenever the state of `main` at merge-time is different from the state of `main` when CI ran on the PR. See e.g. [bors](https://github.com/bors-ng/bors-ng#a-merge-bot-for-github-pull-requests) for a fuller explanation. An important takeaway: "no merge conflicts" does not guarantee that a merge is OK.

I think that using the merge commit is the best of all worlds here because:
- The commits stay granular and easy to understand, with specific commit messages (lacking with "squash and merge").
- The git log can contain the PR number and context (lacking with "rebase and merge", unless the commits are written in advance to contain that information).
- The commits are applied as a whole (lacking in "rebase and merge", where in general the `main` branch can then contain some intermediate broken state that e.g. affects the ability to `git bisect`). And again in general, it also means that if the merge unexpectedly breaks prod (because of suboptimal test coverage or merge practices), you can immediately revert the entire merge commit and ask questions later. That's more painful with "rebase and merge", where again, reverting one commit might produce a broken intermediate state, and so you revert the range of commits (and the fact that you probably want a squashed commit or a merge commit when reverting shows that "rebase and merge" was suboptimal in the first place).

But please consider this post entirely non-blocking. There's a tradeoff between allowing the "best" merge method for every situation, and avoiding a "bad" merge method for the average situation.

petertseng:
Thanks for linking to my previous post! I can also help the situation a little - I'll simply include a link to the pull request in the commit messages. That will solve my complaint with rebase+merge (loses link to pull request), and I would say it'd be okay to rebase+merge in that case! Don't want to use up too much of everyone's valuable time trying to get things figured out. If we can't figure things out, just squashing is all right.

Aside: To allow a merge commit, I am going to hazard a guess that one thing to check would be to see whether "Require linear history" is checked in https://github.com/exercism/problem-specifications/settings/branches (the protection rule for `main`). If it's checked, it would need to be unchecked to allow merge commits. I don't know if there are other things to check.

petertseng:
Thanks for linking to my previous post! I think it generally still accurately describes my views. I can also help the situation a little - I have now included a link to the pull request in the commit messages. That solves my complaint with rebase+merge (loses link to pull request), and I would say it'd be okay to rebase+merge in this case! Don't want to use up too much of everyone's valuable time trying to get things figured out. If we can't figure things out, just squashing is all right.

Aside: To allow a merge commit, I am going to hazard a guess that one setting to look at would be to see whether "Require linear history" is checked in https://github.com/exercism/problem-specifications/settings/branches (the protection rule for `main`). If it's checked, it would need to be unchecked to allow merge commits. I don't know if there are other settings to look at.

SaschaMann:
I thought you were talking about a rebase merge. I thought we agreed that merge commits are not wanted at all some time ago?

ee7:
>  I thought we agreed that merge commits are not wanted at all some time ago?

I think the conclusion was https://github.com/exercism/problem-specifications/issues/1934#issuecomment-1032621135:

> Okay. Having heard from a lot of people (thanks all!), let's make this a squash-only repository. If we ever want to merge/rebase, ping me and I can override it.

SaschaMann:
No, I'm thinking of something older. Merge commits have (thankfully) been disabled across pretty much the entire org quite some time ago.

SaschaMann:
> The commits stay granular, easy to understand, and easy to reference later - with specific commit messages (lacking with "squash and merge").

I strongly disagree with this. If a PR contains enough unrelated changes where this is important, it should be split into several PRs.

ErikSchierboom:
@ihid Could you rebase merge this?

PR #1969, opened 2022-02-24
Exercise allergies test case

karstenj:
Added test case. As mentioned in instructions(https://github.com/karstenj/problem-specifications/blob/main/exercises/allergies/description.md) the program shall only report the eggs (1) allergy for a given score of 257. This case is currently not fully covered by the existing test cases.

### DIFF
--- a/exercises/allergies/canonical-data.json
+++ b/exercises/allergies/canonical-data.json
@@ -566,6 +566,17 @@
             "pollen",
             "cats"
           ]
+        },
+        {
+          "uuid": "93c2df3e-4f55-4fed-8116-7513092819cd",
+          "description": "no allergen score parts without highest valid score",
+          "property": "list",
+          "input": {
+            "score": 257
+          },
+          "expected": [
+            "eggs"
+          ]
         }
       ]
     }


wolf99:
Nice!!

I'm not sure about the description. I understand the teat data but I think my brain just isn't parsing the description correctly. Can you comment with an explanation?

karstenj:
Hi @wolf99 ,

yes, sure. The score is build out of a bit vector. Each bit represents an allergy. This is the list of possible values, where the bits are already converted to an integer:

    eggs (1)
    peanuts (2)
    shellfish (4)
    strawberries (8)
    tomatoes (16)
    chocolate (32)
    pollen (64)
    cats (128)

In the new test case the score is generated out of: Bit 8 = 256 and Bit 0 = 1. This results in the given input score of 257. Due to the fact, that the Bit8(256) is not assigned to an allergy it shall be ignored and only "egg" shall be returned in the list of allergies. Compared to the existing test "no allergen score parts", the highest valid bit Bit7 is not set in the input score.

The main rational for adding this test was also, that this input was mentioned in the instructions(https://github.com/karstenj/problem-specifications/blob/main/exercises/allergies/description.md) , but not considered in the test.

angelikatyborska:
@karstenj
> Compared to the existing test "no allergen score parts", the highest valid bit Bit7 is not set in the input score.

Does this really make a difference? Are there implementations that would pass the existing test for 509 (0001 1111 1101) and all the other tests, but wouldn't pass for 257 (0001 0000 0001)?

angelikatyborska (approved):

angelikatyborska:
Alright, in that case it's a useful change. I'm approving.

> I had one student who provided a solution where the 509 was succesfully executed, but the 257 would fail.

Could you show us that solution? I'm really curious 😅

karstenj:
Yes, you can have a look https://exercism.org/mentoring/discussions/b93c6b3f7e9d44d7809d54db691208d0

angelikatyborska:
@karstenj I am unable to access this link because mentoring discussions are private, you would need to copy-paste the code.

karstenj:
#include "allergies.h"
#include <string.h>
#include <stdio.h>  

/*
subtract numbers until you exhaust the score.
flag the subtracted number along with increasing count.
*/

static int find_less_or_equal(int *arr, int len, int num) {
    // find a index in the array which is less that or eual to number
    for (int i = len-1; i >= 0 ; --i)
        if (arr[i] <= num)
            return i;
    return -1;
}

allergen_list_t get_allergens(int score) {
    int scores[] = {1, 2, 4, 8, 16, 32, 64, 128};
    allergen_list_t my_list;
    memset(my_list.allergens, 0, sizeof(my_list.allergens));
    my_list.count = 0;

    int temp = score;
    while (temp > 0) {
        int idx = find_less_or_equal(scores, ALLERGEN_COUNT, temp); // find the one to subtract
        if (!my_list.allergens[idx]) {
            my_list.allergens[idx] = 1;
            my_list.count++;
        }
        temp -= scores[idx];
    }
    return my_list;
}

bool is_allergic_to(allergen_t a, int score) {
    allergen_list_t result = get_allergens(score);
    if (result.allergens[a])
        return true;
    return false;
}





#ifndef ALLERGIES_H
#define ALLERGIES_H

#include <stdbool.h>

typedef enum {
   ALLERGEN_EGGS = 0,
   ALLERGEN_PEANUTS,
   ALLERGEN_SHELLFISH,
   ALLERGEN_STRAWBERRIES,
   ALLERGEN_TOMATOES,
   ALLERGEN_CHOCOLATE,
   ALLERGEN_POLLEN,
   ALLERGEN_CATS,
   ALLERGEN_COUNT,
} allergen_t;

typedef struct {
   int count;
   bool allergens[ALLERGEN_COUNT];
} allergen_list_t;

allergen_list_t get_allergens(int score);
bool is_allergic_to(allergen_t a, int score);

#endif

wolf99 (approved):

wolf99:
ping @exercism/reviewers 🙂

SaschaMann (approved):

Issue #568, opened 2017-02-13
linked-list: Implement canonical-data.json


kytrinyx:
We want to have a standard set of test inputs and outputs for each exercise to
make it easier to port them to new languages, as well as to help keep the
exercises in sync.

The Linked List exercise can be found in the [./exercises/linked-list/][exercise-dir] directory.

The step-by-step instructions for how to compile a canonical-data.json file is
described in [this section of the contributing guide][contributing].

See http://exercism.io/contribute/canonical-data/linked-list for the up-to-date list of
language tracks that have an implementation of the Linked List exercise.

The [tracking issue][tracking-issue] has more context.

[contributing]: https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#improving-consistency-by-extracting-shared-test-data
[tracking-issue]: https://github.com/exercism/x-common/issues/552
[exercise-dir]: https://github.com/exercism/x-common/tree/master/exercises/linked-list

stkent:
When working on this, it would be nice if `simple-linked-list` ended up being, ahem, simpler, than `linked-list`. I'm imagining that the functionality demanded by `simple-linked-list` should be a subset of the functionality demanded by `linked-list`, and that the tests for `simple-linked-list` should similarly be a subset of the tests for `linked-list`.

IsaacG:
Closed by #1745

PR #1974, opened 2022-03-02
simple-linked-list: add canonical data

loziniak:
references #568 and #587

### DIFF
--- a/exercises/simple-linked-list/canonical-data.json
+++ b/exercises/simple-linked-list/canonical-data.json
@@ -1,0 +1,23 @@
+{
+  "exercise": "simple-linked-list",
+  "cases": [
+    {
+      "uuid": "3b72d23a-4487-4632-8682-59684c875dd9",
+      "description": "Convert an array to linked list, and then from linked list back to array",
+      "property": "fromArrayAndBack",
+      "input": {
+        "array": [3, 6, 1]
+      },
+      "expected": [3, 6, 1]
+    },
+    {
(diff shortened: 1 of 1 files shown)


angelikatyborska:
I am a bit surprised by the property names you're suggesting in this PR, `fromArrayAndBack` and `convertReverseConvertBack`? Did you base that on some existing implementation of this exercise?

I'm looking at the implementations from some of the most popular tracks:
- [JavaScript](https://github.com/exercism/javascript/blob/main/exercises/practice/simple-linked-list/.meta/proof.ci.js)
- [Python](https://github.com/exercism/python/blob/main/exercises/practice/simple-linked-list/.meta/example.py)
- [Ruby](https://github.com/exercism/ruby/blob/main/exercises/practice/simple-linked-list/.meta/example.rb)

They all include `push` and `reverse`, two of them also `pop`. Should those properties be part of the problem specification?

I think the challenge in writing problem specs for this exercise is to choose some representation of linked lists that will be different from a JSON array, to be able to write specifications for properties something like `fromArray` and `toArray`. I'm looking at the problem specification of `linked-list` (https://github.com/exercism/problem-specifications/blob/main/exercises/linked-list/canonical-data.json) and there the problem is worked around by only specifying chains of operations that result in an empty list 🤔.

I'm not really sure what should happen in problem specs for this exercise but I wanted to kick off a discussion so that this PR doesn't go ignored. @exercism/reviewers anyone would like to chime in?

IsaacG:
Superseded by #2635

Issue #1971, opened 2022-02-26
Darts - wrong and lack of help


mohmad-null:
The darts problem description has a couple of problems:

a) There's a link to a wolfram alpha page for "concentric circles" that is unlikely to make much sense to most people, and even if it does, won't actually be helpful because that level of knowledge is not important to the problem anyway (I don't have a clue what the page is saying but still solved it).

b) There should be a link explaining how to calculate the distance between points. https://www.mathsisfun.com/algebra/distance-2-points.html for example is suitable for a broad audience, but there may be others.

kotp:
Thanks for the title update.

Any PR upcoming for this from you?

mohmad-null:
Sorry, no. That would require I do the whole Git thing. My skills lie in other areas. 🙂

SaschaMann:
Spoiler admonitions as suggested in https://github.com/exercism/exercism/issues/6194 could be useful here.

PR #1976, opened 2022-03-03
darts: link to explanations for math lingo

SaschaMann:
see also #1971 even if this change doesn't resolve the issue.

### DIFF
--- a/exercises/darts/description.md
+++ b/exercises/darts/description.md
@@ -14,4 +14,4 @@ In our particular instance of the game, the target rewards with 4 different amou
 
 The outer circle has a radius of 10 units (This is equivalent to the total radius for the entire target), the middle circle a radius of 5 units, and the inner circle a radius of 1. Of course, they are all centered to the same point (That is, the circles are [concentric](http://mathworld.wolfram.com/ConcentricCircles.html)) defined by the coordinates (0, 0).
 
-Write a function that given a point in the target (defined by its `real` cartesian coordinates `x` and `y`), returns the correct amount earned by a dart landing in that point.
+Write a function that given a point in the target (defined by its [real](https://www.mathsisfun.com/numbers/real-numbers.html) [cartesian coordinates](https://www.mathsisfun.com/data/cartesian-coordinates.html) `x` and `y`), returns the correct amount earned by a dart landing in that point.


SaschaMann:
> this is certainly an improvement, so I think that is reason to Approve. I do wonder however if it is not easy to tell that there are two separate links right next to each other.

I was wondering the same. I'm not sure.

Another solution would be a small icon that indicates it's a link to an external website.

ErikSchierboom:
I agree with the usefulness of visually separating the two links somehow. 

Maybe: 

> "defined by its x and y cartesian coordinates (which are both real numbers)"

exercises/darts/description.md
```diff
@@ -14,4 +14,4 @@ In our particular instance of the game, the target rewards with 4 different amou
 
 The outer circle has a radius of 10 units (This is equivalent to the total radius for the entire target), the middle circle a radius of 5 units, and the inner circle a radius of 1. Of course, they are all centered to the same point (That is, the circles are [concentric](http://mathworld.wolfram.com/ConcentricCircles.html)) defined by the coordinates (0, 0).
 
-Write a function that given a point in the target (defined by its `real` cartesian coordinates `x` and `y`), returns the correct amount earned by a dart landing in that point.
+Write a function that given a point in the target (defined by its [real](https://www.mathsisfun.com/numbers/real-numbers.html) [cartesian coordinates](https://www.mathsisfun.com/data/cartesian-coordinates.html) `x` and `y`), returns the correct amount earned by a dart landing in that point.
```
SaschaMann, line 17:
```suggestion
Write a function that given a point in the target (defined by its [cartesian coordinates](https://www.mathsisfun.com/data/cartesian-coordinates.html) `x` and `y`, where `x` and `y` are [real](https://www.mathsisfun.com/numbers/real-numbers.html)), returns the correct amount earned by a dart landing in that point.
```

SaschaMann: How about this? @petertseng @ErikSchierboom

ErikSchierboom: :+1

ErikSchierboom (approved):

exercises/darts/description.md
petertseng: any opinions on "real numbers" versus just "real"?

kotp: How about leaving that out altogether and just use "Cartesian coordinates"  If they can pass the tests using not numbers at all, it would be mentor-worthy.

SaschaMann: That they're real numbers may be useful as a hint that the student needs to use floating point numbers.

SaschaMann: > any opinions on "real numbers" versus just "real"?

I'm fine with either 🤷

SaschaMann: Please explain how a test case with `-0.1` is gonna work with integers.

kotp: Many languages allow for float comparison with integers.  As a value on a dart board, this value seems invalid, because if it is the distance from center, it does not make sense.  So I suppose I would need more information.

In Ruby, the test for where we land looks like this, for my current solution:

```ruby
  BOARD = {
    10 => 0..1,
     5 => 1..5,
     1 => 5..10
  }
```

Integers used all around, not floats or other numeric types found in the code.

But as a practice exercise, it is up to the student to solve this with whatever concepts they would like to explore, such as how floats can or can not work with integers in their language.

kotp: That is poorly worded (the test for) but this is the object that is fundamentally used to test for the location being given a score.  How this is applied is in a different area, but it also involves no floats, though a float may be provided as a distance among other numeric types.

SaschaMann: Talking about this test case for example:  https://github.com/exercism/ruby/blob/95de4127a188821da217fe1e4bf4212dee841bd7/exercises/practice/darts/darts_test.rb#L36-L40

```ruby
  def test_near_the_centre
    skip
    darts = Darts.new(-0.1, -0.1)
    assert_equal(10, darts.score)
  end
```

So mentioning that the coordinates are real is useful, as it gives a hint that these values are allowed (and most likely floats) and need to be considered. Even if the solution doesn't explicitly require them.

kotp: > Please explain how a test case with -0.1 is gonna work with integers.

Then my explanation would involve the phrase "just fine".  I am assuming the test that you have shown does not blow up if integers are compared in the language.

A deeper understanding of how it will work with integers very likely involves coercion to the base most common type, in this case, likely the equivalent integer in the form of a float.

SaschaMann:
@exercism/reviewers

exercises/darts/description.md
```diff
@@ -14,4 +14,4 @@ In our particular instance of the game, the target rewards with 4 different amou
 
 The outer circle has a radius of 10 units (This is equivalent to the total radius for the entire target), the middle circle a radius of 5 units, and the inner circle a radius of 1. Of course, they are all centered to the same point (That is, the circles are [concentric](http://mathworld.wolfram.com/ConcentricCircles.html)) defined by the coordinates (0, 0).
 
-Write a function that given a point in the target (defined by its `real` cartesian coordinates `x` and `y`), returns the correct amount earned by a dart landing in that point.
+Write a function that given a point in the target (defined by its [cartesian coordinates](https://www.mathsisfun.com/data/cartesian-coordinates.html) `x` and `y`, where `x` and `y` are [real](https://www.mathsisfun.com/numbers/real-numbers.html)), returns the correct amount earned by a dart landing in that point.
```
kotp, line 17:
```suggestion
Write a function that given a point in the target (defined by its [cartesian coordinates] `x` and `y`, where `x` and `y` are [real numbers]), returns the correct amount earned by a dart landing in that point.

[cartesian coordinates]: https://www.mathsisfun.com/data/cartesian-coordinates.html
[real numbers]: (https://www.mathsisfun.com/numbers/real-numbers.html)
```

kotp (approved):

angelikatyborska:
test comment

Issue #1977, opened 2022-03-04
Word Count: Apostrophe bug that users' submissions could plausibly include, that current tests won't detect


rneilsen:
The current test suite for the [word-count exercise](https://github.com/exercism/problem-specifications/tree/main/exercises/word-count) tests for words containing apostrophes, but the only examples used in the test cases themselves are words that only have one letter after the apostrophe (specifically, "can't" and "don't").

As a consequence, if a user implements code that doesn't correctly handle words with more than one letter after the apostrophe (like I did, with a regex that looked like `/([a-z]+'[a-z]| ... )/`) the tests will not identify the problem. There aren't a lot of English examples where this might cause miscounts - the only one I could think of was "m'lord"/"m'lady", other than fantasy stuff like "bat'leth"/"bat'leths" - but the thought process that got me there seems pretty plausible as a mistake others might make (I looked at the test cases, saw "don't" and "can't", and thought of "I'm" and "he's", and _didn't_ think of "you're" and "they'll"), so it seems like something worth checking for in the test cases.

Seems like a pretty easy fix to update the "with apostrophes" test case to include a word like "you're" or whatever, I'm happy to submit a PR if it's agreed as worthwhile.

ErikSchierboom:
> Seems like a pretty easy fix to update the "with apostrophes" test case to include a word like "you're" or whatever, I'm happy to submit a PR if it's agreed as worthwhile.

I like the suggestion. My only remark would be that it would have to be a new test case as test cases are immutable. We can discuss whether or not we want the new test case to supersede (reimplements in problem-specifications parlance) the current test case (I think we might).

rneilsen:
Dang it, there's always a catch. Is there a documented procedure for creating a new test to replace an old one? I had a look but couldn't find anything - [this section of CONTRIBUTING.md](https://github.com/exercism/problem-specifications/blob/main/CONTRIBUTING.md#changing-inputs-and-outputs) just says to submit a PR for the JSON file, but it doesn't give any details.

I could probably write up a test, give it a new uuid, and submit a PR with the new test where the old one was, but I don't want to waste anyone else's time if there's a proper procedure to go through that I should be following.

ErikSchierboom:
@rneilsen There is a section in the README that explains how to do this: https://github.com/exercism/problem-specifications#changing-tests

PR #1982, opened 2022-03-07
Add reimplement of "with apostrophes" test

rneilsen:
As per [issue 1977](https://github.com/exercism/problem-specifications/issues/1977), if the user's code was unable to correctly handle words with multiple letters after an apostrophe, e.g. "they've" (a plausible error one might make with a regex, for example), the previous version of the "with apostrophes" test, which only included the word "don't", would not detect the bug (and none of the other tests would detect it, either).

This change is a reimplement of the "with apostrophes" test to now include the word "you're" as part of the input, so if the user's code has a bug and incorrectly handles such words, the new version of the test will fail appropriately, and expose it.

### DIFF
--- a/exercises/word-count/canonical-data.json
+++ b/exercises/word-count/canonical-data.json
@@ -126,6 +126,26 @@
       }
     },
     {
+      "uuid": "4ff6c7d7-fcfc-43ef-b8e7-34ff1837a2d3",
+      "description": "with apostrophes",
+      "comments": ["Added a word with multiple letters after apostrophe"],
+      "reimplements": "4185a902-bdb0-4074-864c-f416e42a0f19",
+      "property": "countWords",
+      "input": {
+        "sentence": "'First: don't laugh. Then: don't cry. You're getting it.'"
+      },
+      "expected": {
+        "first": 1,
+        "don't": 2,
+        "laugh": 1,
+        "then": 1,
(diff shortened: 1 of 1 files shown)


exercises/word-count/canonical-data.json
```diff
@@ -125,6 +125,26 @@
...
+      "input": {
+        "sentence": "'First: don't laugh. Then: don't cry. You're getting it.'"
+      },
+      "expected": {
+        "first": 1,
+        "don't": 2,
+        "laugh": 1,
+        "then": 1,
+        "cry": 1,
+        "you're": 1,
```
kotp, line 143: I notice that the "don't" has a value of 2, while "you're" has a value of 1.  Should it not be a 2 as well?

rneilsen: There's only one instance of "you're" in the input text. Or do you mean we should use an input that includes "you're" multiple times?

kotp: Nope, sorry did not associate it directly with the inputs as a word count occurrence.  Thought perhaps it was a contraction being counted as the two words it involves.

ErikSchierboom (approved):
LGTM

kotp (approved):

BethanyG (approved):

PR #1986, opened 2022-03-13
pangram: remove overly specific description

SaschaMann:
This makes it easier for tracks to add additional test cases that aren't ascii without having to deal with manually syncing the instructions.

I think this is in line with the vast majority of exercises where the inputs aren't described in too much detail.

### DIFF
--- a/exercises/pangram/description.md
+++ b/exercises/pangram/description.md
@@ -5,5 +5,5 @@ Determine if a sentence is a pangram. A pangram (Greek: παν γράμμα, pan
 The best known English pangram is:
 > The quick brown fox jumps over the lazy dog.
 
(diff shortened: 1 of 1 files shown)


SleeplessByte (approved):
Assuming so tracks are allowed to do whatever?

kotp:
Yes, for instance, it states an alphabet from "a to z" but that allows us to use the Italian alphabet for pangram checking, as it is ambiguous enough to allow for that "stretch" exercise, and the work that goes into it.  Even the "a" - "z" in Hawaiian, which is just a colloquialism, since Hawaiian does not even have a "z".

Slightly ambiguous written instructions are a good thing for the exercises.  I use this technique very often for this and similar exercises.

ErikSchierboom (approved):

kotp (approved):

PR #1992, opened 2022-03-29
reorder keys in format

ErikSchierboom:
- Add script to reorder test case keys
- ci: reorder keys as part of /format

I still have some testing to do, to see if this works.

### DIFF
--- a/.github/workflows/action-format.yml
+++ b/.github/workflows/action-format.yml
@@ -60,6 +60,9 @@ jobs:
           git clone $HEAD_REPO .
           git checkout -b "$HEAD_REF" "origin/$HEAD_REF"
 
+      - name: Re-order keys
+        run: ./bin/reorder-keys
+
       - name: Use Node.js LTS (16.x)
         uses: actions/setup-node@9ced9a43a244f3ac94f13bfd896db8c8f30da67a
         with:
--- a/.github/workflows/action-format.yml
+++ b/.github/workflows/action-format.yml
@@ -60,6 +60,7 @@ jobs:
           git clone $HEAD_REPO .
           git checkout -b "$HEAD_REF" "origin/$HEAD_REF"
 
+      # This needs to happen before the actual formatting using Prettier happens
       - name: Re-order keys
         run: ./bin/reorder-keys
 
--- a/bin/reorder-keys
+++ b/bin/reorder-keys
@@ -5,6 +5,8 @@ if [ ! -d exercises ] ; then
     exit 1
 fi
 
+tmp_file='.reordered-canonical-data.json'
+
 for canonical_data_file in exercises/*/canonical-data.json ; do
-    jq 'walk(if (type == "object") and has("uuid") then . as $obj | reduce ["uuid", "reimplements", "description", "comments", "scenarios", "property", "input", "expected"][] as $key ({}; . + (if $obj | has($key) then { ($key): $obj[$key] } else {} end)) else . end)' "$canonical_data_file" > tmp && mv tmp "$canonical_data_file"
+    jq 'walk(if (type == "object") and has("uuid") then . as $obj | reduce ["uuid", "reimplements", "description", "comments", "scenarios", "property", "input", "expected"][] as $key ({}; . + (if $obj | has($key) then { ($key): $obj[$key] } else {} end)) else . end)' "$canonical_data_file" > "$tmp_file" && mv "$tmp_file" "$canonical_data_file"
 done
--- a/bin/reorder-keys
+++ b/bin/reorder-keys
@@ -6,5 +6,5 @@ if [ ! -d exercises ] ; then
 fi
 
 for canonical_data_file in exercises/*/canonical-data.json ; do
-    jq 'walk(if (type == "object") and has("uuid") then . as $obj | reduce ["uuid", "reimplements", "description", "comments", "scenarios", "property", "input", "expected"][] as $key ({}; . + (if $obj | has($key) then { ($key): $obj[$key] } else {} end)) else . end)' $canonical_data_file > tmp && mv tmp $canonical_data_file    
+    jq 'walk(if (type == "object") and has("uuid") then . as $obj | reduce ["uuid", "reimplements", "description", "comments", "scenarios", "property", "input", "expected"][] as $key ({}; . + (if $obj | has($key) then { ($key): $obj[$key] } else {} end)) else . end)' "$canonical_data_file" > tmp && mv tmp "$canonical_data_file"
 done
--- a/bin/reorder-keys
+++ b/bin/reorder-keys
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+
+if [ ! -d exercises ] ; then
+    echo "\"exercises\" directory not found, make sure you're running from the repository root" >&2
+    exit 1
+fi
+
+for canonical_data_file in exercises/*/canonical-data.json ; do
+    jq 'walk(if (type == "object") and has("uuid") then . as $obj | reduce ["uuid", "reimplements", "description", "comments", "scenarios", "property", "input", "expected"][] as $key ({}; . + (if $obj | has($key) then { ($key): $obj[$key] } else {} end)) else . end)' $canonical_data_file > tmp && mv tmp $canonical_data_file    
+done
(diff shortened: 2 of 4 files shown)


.github/workflows/action-format.yml
```diff
@@ -60,6 +60,9 @@ jobs:
           git clone $HEAD_REPO .
           git checkout -b "$HEAD_REF" "origin/$HEAD_REF"
 
+      - name: Re-order keys
+        run: ./bin/reorder-keys
```
ErikSchierboom, line 64: Note: this needs to happen before the actual formatting using Prettier happens

.github/workflows/action-format.yml
```diff
@@ -60,6 +60,9 @@ jobs:
...
+
```
SaschaMann, line 65:
```suggestion
      # This needs to happen before the actual formatting using Prettier happens
      - name: Re-order keys
        run: ./bin/reorder-keys

```

SaschaMann (approved):

bin/reorder-keys
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+if [ ! -d exercises ] ; then
+    echo "\"exercises\" directory not found, make sure you're running from the repository root" >&2
+    exit 1
+fi
+
+for canonical_data_file in exercises/*/canonical-data.json ; do
+    jq 'walk(if (type == "object") and has("uuid") then . as $obj | reduce ["uuid", "reimplements", "description", "comments", "scenarios", "property", "input", "expected"][] as $key ({}; . + (if $obj | has($key) then { ($key): $obj[$key] } else {} end)) else . end)' $canonical_data_file > tmp && mv tmp $canonical_data_file
```
ErikSchierboom, line 9: I'm not great at jq, so if there is a more elegant way to do this, do let me know!

iHiD (approved):

ErikSchierboom:
Small bump for @exercism/reviewers

bin/reorder-keys
ee7, line 9:
```suggestion
    jq 'walk(if (type == "object") and has("uuid") then . as $obj | reduce ["uuid", "reimplements", "description", "comments", "scenarios", "property", "input", "expected"][] as $key ({}; . + (if $obj | has($key) then { ($key): $obj[$key] } else {} end)) else . end)' "$canonical_data_file" > tmp && mv tmp "$canonical_data_file"
```

Removes trailing whitespace and addresses `shellcheck` output.

> [SC2086](https://github.com/koalaman/shellcheck/wiki/Sc2086) (info): Double quote to prevent globbing and word splitting.

(I didn't review this PR properly - I just ran `shellcheck`).

bin/reorder-keys
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+if [ ! -d exercises ] ; then
+    echo "\"exercises\" directory not found, make sure you're running from the repository root" >&2
```
ee7, line 4:
```suggestion
    echo "Error: \"exercises\" directory not found, make sure you're running from the repository root" >&2
```

If you prefer. I know it's already going to stderr, and we're exiting non-zero.

bin/reorder-keys
```diff
@@ -0,0 +1,10 @@
...
+    exit 1
+fi
+
+for canonical_data_file in exercises/*/canonical-data.json ; do
+    jq 'walk(if (type == "object") and has("uuid") then . as $obj | reduce ["uuid", "reimplements", "description", "comments", "scenarios", "property", "input", "expected"][] as $key ({}; . + (if $obj | has($key) then { ($key): $obj[$key] } else {} end)) else . end)' "$canonical_data_file" > tmp && mv tmp "$canonical_data_file"
```
ee7, line 9: Nit: we could consider:
- Splitting this long line.
- Renaming `tmp` to something a bit more unique. In the unlikely case that some user has a file named `tmp` in their prob-specs repo and runs this script, this script will overwrite it.

ee7 (approved):
I'm not enough of a `jq` wizard to suggest improvements to line 9, but it seems to work. It does take a few seconds, so maybe there is a better `jq` incantation.

I've checked that the order in this PR matches that in:
https://github.com/exercism/problem-specifications/blob/7cd0ab13d2f1158633c5d7f551fe2160ba7280c1/bin/check_key_order.rb#L3

The order is currently different than this PR in the README and canonical data schema. In either this PR or a follow-up one, I'd suggest making the key order match in:

https://github.com/exercism/problem-specifications/blob/7cd0ab13d2f1158633c5d7f551fe2160ba7280c1/README.md#L184-L194

https://github.com/exercism/problem-specifications/blob/7cd0ab13d2f1158633c5d7f551fe2160ba7280c1/canonical-data.schema.json#L80-L90

ErikSchierboom:
> The order is currently different than this PR in the README and canonical data schema. In either this PR or a follow-up one, I'd suggest making the key order match in:

Done

bin/reorder-keys
ErikSchierboom: I've renamed `tmp` to something more unique.

As for the splitting of the long line, do you have any good suggestion on how to format it? I've tried several things, but they were all still fairly unwieldy.

bin/reorder-keys
```diff
@@ -0,0 +1,12 @@
...
+if [ ! -d exercises ] ; then
+    echo "Error: \"exercises\" directory not found, make sure you're running from the repository root" >&2
+    exit 1
+fi
+
+tmp_file='.reordered-canonical-data.json'
+
+for canonical_data_file in exercises/*/canonical-data.json ; do
+    jq 'walk(if (type == "object") and has("uuid") then . as $obj | reduce ["uuid", "reimplements", "description", "comments", "scenarios", "property", "input", "expected"][] as $key ({}; . + (if $obj | has($key) then { ($key): $obj[$key] } else {} end)) else . end)' "$canonical_data_file" > "$tmp_file" && mv "$tmp_file" "$canonical_data_file"
+done
```
ee7, line 12: As a follow up on the line splitting: maybe something like the below. But it's fine. PR it if you want.

The default shell for GitHub Actions on non-Windows is bash with `-e`, so I have a slight preference for adding that at the top of the script to get consistent behavior locally. Then we can put the `mv` on a separate line without the `&&`.

```suggestion
#!/usr/bin/env bash

set -e

if [ ! -d exercises ] ; then
    echo "Error: \"exercises\" directory not found, make sure you're running from the repository root" >&2
    exit 1
fi

tmp_file='.reordered-canonical-data.json'

reorder() {
    jq 'walk(if (type == "object") and has("uuid") then . as $obj
        | reduce ["uuid", "reimplements", "description", "comments", "scenarios", "property", "input", "expected"][] as $key(
            {}; . + (if $obj | has($key) then { ($key): $obj[$key] } else {} end)) else . end
        )' "$1" > "$tmp_file"
    mv "$tmp_file" "$1"
}

for canonical_data_file in exercises/*/canonical-data.json ; do
    reorder "$canonical_data_file"
done
```

bin/reorder-keys
ee7: > As for the splitting of the long line, do you have any good suggestion on how to format it?

Maybe something like the below:

```shell
#!/usr/bin/env bash

set -e

if [ ! -d exercises ] ; then
    echo "Error: \"exercises\" directory not found, make sure you're running from the repository root" >&2
    exit 1
fi

tmp_file='.reordered-canonical-data.json'

reorder() {
    jq 'walk(if (type == "object") and has("uuid") then . as $obj
        | reduce ["uuid", "reimplements", "description", "comments", "scenarios", "property", "input", "expected"][] as $key(
            {}; . + (if $obj | has($key) then { ($key): $obj[$key] } else {} end)) else . end
        )' "$1" > "$tmp_file"
    mv "$tmp_file" "$1"
}

for canonical_data_file in exercises/*/canonical-data.json ; do
    reorder "$canonical_data_file"
done
```

But it's fine. PR it if you want.

The default shell for GitHub Actions on non-Windows is bash with the `-e` option (actually `bash --noprofile --norc -eo pipefail`, see [docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell)), so I have a slight preference for adding that at the top of the script to get consistent behavior locally. Then we can put the `mv` on a separate line without the `&&`.

(No strong opinion about the extra function).

SaschaMann: Sorry for the potentially early merge. I saw the `tmp` issue being addressed but missed the refactoring question.

ee7: No problem.

PR #1993, opened 2022-03-30
Add test to Yacht to ensure fives are handled

dstockto:
The current tests for "fives" only include a negative test where the score is 0. This PR adds a test for fives that has a score.


ErikSchierboom (approved):
Good addition!

SaschaMann (approved):

ErikSchierboom:
Thanks @dstockto!

Issue #2002, opened 2022-04-06
Update Instructions For "Twelve Days"


RTurek:

kotp:
In order to create a pull request, one would fork the project,  make the changes on your copy of the repository, and then once pushed, github will respond with text describing where to open a pull request.

RTurek:
Thanks @kotp!

kotp:
Since the PR already exists, I made a comment there, but it should probably have started from the issue, to help inform what the PR would look like once it is started.

RTurek:
Apologies - this is the first time I've contributed to exercism! I'm still learning what is expected 😄

kotp:
Absolutely no problem.  This is a safe place to have first pull requests at all, much less as just a first pull request for the project.

PR #2003, opened 2022-04-07
#2002 Update Instructions For "Twelve Days"

RTurek:
# Overview

Resolves https://github.com/exercism/problem-specifications/issues/2002

# Details

- Improves explanation of what the challenge is asking the student to do
  - Additional language should help non-Christian, non-native-English-speakers to more quickly understand what this exercise is all about when they first read it
- Help clarify the spirit of this challenge and discourage just reading and returning the same text file the test runner is comparing against

### DIFF
--- a/exercises/twelve-days/description.md
+++ b/exercises/twelve-days/description.md
@@ -1,7 +1,12 @@
-# Description
+# Instructions
 
-Output the lyrics to 'The Twelve Days of Christmas'.
+Your task in this exercise is to write a method that returns the lyrics of the song: "The Twelve Days of Christmas."
 
+"The Twelve Days of Christmas" is a common Christmas carol. Each subsequent verse of the song builds on the previous verse.
+
+The lyrics your method returns should _exactly_ match the text shown below.
+
+# The Twelve Days Of Christmas
 ```text
 On the first day of Christmas my true love gave to me: a Partridge in a Pear Tree.
 
@@ -27,3 +32,8 @@ On the eleventh day of Christmas my true love gave to me: eleven Pipers Piping, 
 
 On the twelfth day of Christmas my true love gave to me: twelve Drummers Drumming, eleven Pipers Piping, ten Lords-a-Leaping, nine Ladies Dancing, eight Maids-a-Milking, seven Swans-a-Swimming, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree.
 ```
+
+# Tips
+
+- Each verse of the song has several different elements that are repeated from verse to verse. You should identify these elements and try to re-use them when writing your code.
+- Remember to be "DRY" (Don't Repeat Yourself)!


kotp:
And yet, there is an elegance to the solution that reads the song in, if it is passing the tests to give verses.  Nothing wrong with that solution.  This is a practice exercise as well, and if mentored, it would be natural for the mentor to challenge the author of such a solution to programmatically solve the problem as well, through other algorithms.

The tips are probably belonging in the mentor notes, rather than in the description.

RTurek:
There is - and I would appreciate the cleverness of a student who immediately wrote that solution, but I would definitely encourage them to try writing an algorithm instead. This is me interpreting the spirit of this exercise and trying to explain that to the student.

Come to think of it... is there an exercise that makes more use of importing text files and comparing them or parsing them or something like that? 🤔  seems like a good idea if there isn't one that's specifically for that.

I agree with removing the Tips. It's a bit clunky having them there at the bottom anyway.

I don't know where mentor notes are stored - could you point me in that direction? I wrote those tips trying NOT to lead the student along too much towards any particular solution other than to imply that they shouldn't just do the file import/export method, but that's somewhat opinionated of me.

Help here is much appreciated - this is the first time I've tried to contribute to exercism, so I'm still learning the process @kotp

kotp:
The mentoring notes for the exercise should be found in [exercism/website-copy] in a file called `mentoring.md`.

If we were to start a pull request via the github interface, we might use [this link].

The recommendation I would have is to not have the tips in this file, but perhaps start the mentor notes with that as one of the things to suggest.  You should be able to look at other examples of the `mentoring.md` content from that repository as well to give an idea of the format style, etc.

[exercism/website-copy]: https://github.com/exercism/website-copy
[this-link]: https://github.com/exercism/website-copy/new/main/tracks/ruby/exercises/twelve-days?filename=twelve-days/mentoring.md

exercises/twelve-days/description.md
```diff
@@ -1,6 +1,12 @@
...
+# Instructions
 
-Output the lyrics to 'The Twelve Days of Christmas'.
+Your task in this exercise is to write code that returns the lyrics of the song: "The Twelve Days of Christmas."
+
+"The Twelve Days of Christmas" is a common English Christmas carol. Each subsequent verse of the song builds on the previous verse.
+
+The lyrics your code returns should _exactly_ match the full song text shown below.
+
+## The Twelve Days Of Christmas
```
ErikSchierboom, line 9: I'm wondering whether it might be clearer to have this named:

```suggestion
## Lyrics
```

We're already in the context of Twelve Days of Christmas

exercises/twelve-days/description.md
```diff
@@ -1,6 +1,12 @@
-# Description
+# Instructions
```
ErikSchierboom, line 1: Maybe have this be unchanged to match the other exercises?

```suggestion
# Description
```

ErikSchierboom (commented):
Great that you want to improve this exercise! Two minor comments.

exercises/twelve-days/description.md
RTurek: Agreed!

exercises/twelve-days/description.md
RTurek: Ah, that was unintentional

kotp (approved):

angelikatyborska (approved):

ErikSchierboom (approved):

PR #2014, opened 2022-04-18
Fix the hyperlink to jumpstartlabs

orsenthil:


SaschaMann:
Looking at the [archived version of the old link](https://web.archive.org/web/20210815010331/http://jumpstartlab.com/), these don't appear to be the same websites. Perhaps we should link to an appropriate archive link instead?

PR #2026, opened 2022-04-27
parallel-letter-frequency: align requirements with test

adamwight:


exercises/parallel-letter-frequency/description.md
```diff
@@ -6,3 +6,7 @@ Parallelism is about doing things in parallel that can also be done
 sequentially. A common example is counting the frequency of letters.
 Create a function that returns the total frequency of each letter in a
 list of texts and that employs parallelism.
+
+Uppercase letters count towards their lowercase equivalent.  Support
+at least the extended Latin character set, so that unicode "ü" is
```
SaschaMann, line 11: There are (at least) two ways to write `ü`. Either as a combined character of `u` (U+0075) and `̈` (U+0308) or as `ü` (U+00FC). This difference has tripped me up in the past in another exercise. Does this require support of both those options? Might be good to be quite specific here.

IsaacG:
This doesn't looks like it's happening. I'm not sure there's interest in adding extended letters here.

PR #2042, opened 2022-05-20
Word count: test with quotations for word with apostrophe

dirkkul:
While solving the exercise I noticed that there is no test that contains quotations around a word with apostrophes, eg 'can't'.

Some naive implementations for quotation filtering might not be able to filter these quotations out correctly.

Let me know if I'm missing anything, happy to fix it :)

### DIFF
--- a/exercises/word-count/canonical-data.json
+++ b/exercises/word-count/canonical-data.json
@@ -203,6 +203,18 @@
         "two": 1,
         "three": 1
       }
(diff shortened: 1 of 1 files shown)


jiegillet (approved):

kytrinyx (approved):

glennj (approved):
Interesting test!

PR #2050, opened 2022-06-06
rectangles: Add test for missing sides

hardselius:
Some solutions to this exercise will look for the top corners of the rectangles, making sure they are connected by a horizontal line, then go on to walk down the left and right side to find corners at the same row, in which case a rectangle is detected. These solutions sometimes fail to check that the bottom two corners are connected by a horizontal line. This test case will catch that.

### DIFF
--- a/exercises/rectangles/canonical-data.json
+++ b/exercises/rectangles/canonical-data.json
@@ -174,6 +174,23 @@
         ]
       },
       "expected": 60
+    },
+    {
+      "uuid": "6ef24e0f-d191-46da-b929-4faca24b4cd2",
+      "description": "rectangles must have four sides",
+      "property": "rectangles",
+      "input": {
+        "strings": [
+          "+-+ +-+",
+          "| | | |",
+          "+-+-+-+",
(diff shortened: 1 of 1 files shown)


petertseng (approved):
How likely is it that the equivalent case for a missing right side should also be included?

```
+-+
|
+-+
```

(with other surrounding rectangles to make that an actually-valid input)

hardselius:
@petertseng I'm not sure. I suppose someone would try search columns first and then go from left to right, but it certainly seems less likely than top to bottom. Maybe it could test something like

```
+-+ +-+
| | | |
+-+-+-+
  | |  
+-+-+-+
| | | |
+-+ +-+
```

to cover all scenarios?

hardselius:
I changed this to test all scenarios instead.

petertseng (approved):

jiegillet (approved):

ErikSchierboom (approved):

ErikSchierboom:
Thanks @hardselius! 🎉

PR #2059, opened 2022-06-16
space-age: calculate using correct average calendar year

dsletten:
The expected values are calculated incorrectly based on an average Julian calendar year (365.25 days). They should be computed based on an average Gregorian calendar year (365.2425 days). Fortunately, the only value affected is for Mercury.
https://en.wikipedia.org/wiki/Gregorian_calendar


IsaacG:
See #2060. This is a duplicate.

PR #2060, opened 2022-06-16
Space Age: Change days for earth orbital period

dsletten:
The value 365.25 days/year represents an average Julian calendar year. The correct value for the Gregorian calendar is 365.2425.

### DIFF
--- a/exercises/space-age/description.md
+++ b/exercises/space-age/description.md
@@ -4,7 +4,7 @@ Given an age in seconds, calculate how old someone would be on:
 
 - Mercury: orbital period 0.2408467 Earth years
 - Venus: orbital period 0.61519726 Earth years
-- Earth: orbital period 1.0 Earth years, 365.25 Earth days, or 31557600 seconds
+- Earth: orbital period 1.0 Earth years, 365.2425 Earth days, or 31557600 seconds
 - Mars: orbital period 1.8808158 Earth years
 - Jupiter: orbital period 11.862615 Earth years
 - Saturn: orbital period 29.447498 Earth years


exercises/space-age/description.md
```diff
@@ -4,7 +4,7 @@ Given an age in seconds, calculate how old someone would be on:
 
 - Mercury: orbital period 0.2408467 Earth years
 - Venus: orbital period 0.61519726 Earth years
-- Earth: orbital period 1.0 Earth years, 365.25 Earth days, or 31557600 seconds
+- Earth: orbital period 1.0 Earth years, 365.2425 Earth days, or 31557600 seconds
```
petertseng, line 7: if 365.25 is changed, 31557600 must change as well, because 365.2425 * 24 * 3600 does not equal 31557600

this comment is not to be taken as support for nor opposition against changing 365.25

siebenschlaefer:
@junedev I've had one who complained that it took them very long to figure those constants our by trial and error. I tried to be very tactful when I had to tell them to read the instructions. :-)

rpottsoh:
Seems like an unnecessary change to me.

wolf99:
I am very slightly against this change.
While the pedant in me appreciates the "correctness" of it, the change will result in a sync being required across all tracks that implement this exercise for the sake of only a small fractional difference that does not affect the meaning in the exercise.
That's said, I do not mentor so I have no knowledge of student's experiences with the exercise.

SleeplessByte (changes requested):
Whilst I don't think this change is necessary, or is adding much value, I think one of our base tests: "does this make the exercise unnecessarily harder?" would yield no. It's also not a hassle to update the exercises we already have because the change is inconsequential (as in, we shouldn't mind if tracks don't update), whilst technically make the content of the exercise every so slightly "more correct".

However:
- the seconds must be changed
- any relevant tests much change

A _simpler_ change here may be to note (admonition??) that the "true" orbital year is 365.2425 and that the value of 326.25 is used in x and y and called z.

kytrinyx:
> A simpler change here may be to note (admonition??) that the "true" orbital year is 365.2425 and that the value of 326.25 is used in x and y and called z.

My vote is for this one. @dsletten I realize that it's been four months since this discussion started, and that you might be over it, if you'd like to make the suggested change it looks like there would be enough votes to get this merged.

IsaacG:
@dsletten Would you still like the make a change here?

PR #2063, opened 2022-06-23
pythagorean-triplet: testcase that triggers integer overflow

sudhackar:
While reviewing a C submission for pythagorean triplet exercise, I noticed that larger numbers overflow `int` primitive if not explicitly forced to something like (uint32_t)

I feel this is a good test case to teach integer overflow right from the start.
This was some part of the discussion
```c
            if (a * a + b * b != c * c) {
```
which triggered
```
pythagorean_triplet.c:17:36: runtime error: signed integer overflow: 65529 * 65529 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior pythagorean_triplet.c:17:36 in
```

### DIFF
--- a/exercises/pythagorean-triplet/canonical-data.json
+++ b/exercises/pythagorean-triplet/canonical-data.json
@@ -81,6 +81,19 @@
         [6000, 11250, 12750],
         [7500, 10000, 12500]
       ]
+    },
+    {
+      "uuid": "270b8476-6f59-4621-bef7-70e6a3513923",
+      "description": "triplets to trigger integer overflow",
+      "property": "tripletsWithSum",
+      "input": {
+        "n": 65532
+      },
+      "expected": [
+        [1016, 32250, 32266],
+        [16383, 21844, 27305],
+        [17145, 21156, 27231]
+      ]
     }
   ]
 }


petertseng (approved):
I have verified that these are the correct values for `n = 65532`.

I explicitly have no opinion about these below points. If you are an interested reviewer who cares about these below points, it is up to you to speak up about them, since I will not:

* Note that the O(n) solution will only consider `c` values no greater than 32765 when given n of 65532 (this happens when it is considering `a = 2`), and 32765 * 32765 fits in a signed 32-bit integer. Therefore, if this test case intends to cause integer overflow in the O(n) solution, it must use a larger n.
* The O(n^2) solution should be expected to take perhaps 4x as long for `n = 65532` than for `n = 30000`. What exactly this means in terms of a number of seconds will depend on the language.
* I have no data on how many students submit the O(n) solution versus the O(n^2) solution so I will not be able to assist in evaluating the above statements in terms of the student impact.

sudhackar:
Based on the experience with the C track - the solutions [here](https://exercism.org/tracks/c/exercises/pythagorean-triplet/solutions) I see quite a few "naive" O(n^2) approaches.
Also the given testcases are written with specifically `uint16_t` - students can copy the type for the `sum` for iterating and calculating. Such as this [latest published solution](https://exercism.org/tracks/c/exercises/pythagorean-triplet/solutions/ExercismGhost).

PR #2065, opened 2022-06-29
Update Roman Numeral with secure link and deeper information

apolo49:
Nova Roma's website uses the http connection scheme leaving anyone following that link vulnerable with unencrypted traffic. This enables an easier MITM attack which we want to avoid for the target audience who may not know better.

Additionally the link that has been used with the substitute (Imperivm Romanvm) has their information presented in a wiki format which is easier to consume and more in-depth. Most of the information is illustrated in tables.

Imperivm Romanvm has a converter akin to that which is presented within Nova Roma's website however the one by Imperivm Romanvm presents the answer in an easier to read form and details why the number is like it is. This can be illustrated through imagery which I will present below.

image
image
image
image
image
image
image
image

Nova Roma's website also doesn't represent the current state of the internet and may lead children to be confused by its UI and graphical design, this page in particular shows no signs of updates past 1998 aside from a background change (this could however be very wrong).

For further proof of poor design type in 44444 (forty-four thousand, four-hundred and four).
Imperivm Romanvm presents the number as: ↂↇMↁCDXLIV
Nova Roma presents the number as follows: MMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMCDXLIV

### DIFF
--- a/exercises/roman-numerals/description.md
+++ b/exercises/roman-numerals/description.md
@@ -40,4 +40,4 @@ In Roman numerals 1990 is MCMXC:
 2000=MM
 8=VIII
 
-See also: [http://www.novaroma.org/via_romana/numbers.html](http://www.novaroma.org/via_romana/numbers.html)
+See also: [https://wiki.imperivm-romanvm.com/wiki/Roman_Numerals](https://wiki.imperivm-romanvm.com/wiki/Roman_Numerals)


apolo49:
The links to both websites can be found in the changed file or here:
http://www.novaroma.org/via_romana/numbers.html
https://wiki.imperivm-romanvm.com/wiki/Roman_Numerals

SleeplessByte (approved):
For the sake of having a secure link, I think this change is fine.

BethanyG:
Agree that this makes sense for securing a link.

BethanyG (approved):

junedev (approved):

PR #2067, opened 2022-07-01
word-count: Less ambiguous punctuation rule

adql:
With "ignored", it's unclear if punctuation separates the words or is just removed (relevant for the test case [handles cramped lists](https://github.com/exercism/problem-specifications/blob/16f11cec666076eb1f6f8f4c1fd2fc4d0852dbd6/exercises/word-count/canonical-data.json#L49)).

### DIFF
--- a/exercises/word-count/description.md
+++ b/exercises/word-count/description.md
@@ -12,7 +12,7 @@ When counting words you can assume the following rules:
 
 1. The count is _case insensitive_ (ie "You", "you", and "YOU" are 3 uses of the same word)
 2. The count is _unordered_; the tests will ignore how words and counts are ordered
-3. Other than the apostrophe in a _contraction_ all forms of _punctuation_ are ignored
+3. Other than the apostrophe in a _contraction_ all forms of _punctuation_ are regarded as spaces
 4. The words can be separated by _any_ form of whitespace (ie "\t", "\n", " ")
 
 For example, for the phrase `"That's the password: 'PASSWORD 123'!", cried the Special Agent.\nSo I fled.` the count would be:


exercises/word-count/description.md
```diff
@@ -12,7 +12,7 @@ When counting words you can assume the following rules:
 
 1. The count is _case insensitive_ (ie "You", "you", and "YOU" are 3 uses of the same word)
 2. The count is _unordered_; the tests will ignore how words and counts are ordered
-3. Other than the apostrophe in a _contraction_ all forms of _punctuation_ are ignored
+3. Other than the apostrophe in a _contraction_ all forms of _punctuation_ are regarded as spaces
```
kotp, line 15: What do we do with spaces?  We treat them as part of a word break. It is not identical.  All forms of punction are ignore, and I believe this is correct.

adql: What does "ignored" mean? If it means being skipped and left unread then test case _handles cramped lists_ should be

```json
"input": {
    "sentence": "one,two,three"
},
"expected": {
    "onetwothree": 1
}
```

while actually it is

```json
"input": {
    "sentence": "one,two,three"
},
"expected": {
    "one": 1,
    "two": 1,
    "three": 1
}
```

i.e. `,` is treated as a word break. Perhaps a better formulation would be "regarded as work breaks", but the parsing logic is all the same.

kotp: I am not sure that we can get a canonically accurate word count.  All of the professional word processors get it wrong, or right, even, sometimes.

I was thinking of punctuation as one thing, and non-word things as another.  So punctuation is a word boundary, as is space, vertical tab, while the punctuation that is `-` is not a word boundary, sometimes, and for example.

So my use of "punctuation" in my response was definitely a lazy response.  I guess first we must identify what "all punctuation" means.

petertseng (approved):
I have no preference between "regarded as spaces" or "regarded as word breaks" or "considered to separate words". All of these are more precise (they more unambiguously describe what to do with regard to the test case in question) than "ignored", so we should pick any one to go with.

tejasbubane (approved):

BethanyG (approved):
Like Peter, I think any of the phrasing works and is better than "ignored".  We should just pick one and go with it.

petertseng:
it sounds like "regarded as spaces" has found enough approval.

PR #2068, opened 2022-07-03
Atbash Cipher: Clarify expectation for uppercase letters

dsletten:
The test suite expects uppercase letters to be encoded as lowercase. Neither the instructions nor the examples address this.


glennj:
I suggest adding another example showing the input as upper or mixed case and the output as lower case.

petertseng (approved):
This is an improvement, with or without an additional example.

junedev (approved):

ErikSchierboom (approved):

ErikSchierboom:
Thanks @dsletten!

PR #2071, opened 2022-07-08
matching brackets: reimplements wrong nested bracket tests

kimesf:
Current input value for `paired and wrong nested brackets` (a0205e34-c2ac-49e6-a88a-899508d7d68e) allows for solutions where only the last opened bracket is being traced.

Current input value is `[({*]*})`.
Proposed input value is `[({}*]*)`.
<sub>"*" Used only to highlight the change.<sub>

By wrongly closing the second to last opened bracket (instead of the last one), we best cover wrongly nested brackets, because we ensure users realize they need to track all opened brackets.

I understand the most straighfoward and simple solution is to stack opened brackets and remove then when they're closed, but it's possible to green all tests by tracing brackets count and last opened bracket.

The next code is an example in ruby. It fails if the proposed input is used, but passes with the current one.
<sub>Code was omitted for brevity sake, full code [here](https://github.com/kimesf/exercism_solutions/blob/main/ruby/matching-brackets/matching_brackets_edge_case.rb) | Ruby problem [here](https://exercism.org/tracks/ruby/exercises/matching-brackets) for testing<sub>
```ruby
  def paired?
    brackets.each do |bracket|
      case bracket
      when '{' then count[:braces] += 1
      when '}' then count[:braces] -= 1
      when '(' then count[:parentheses] += 1
      when ')' then count[:parentheses] -= 1
      when '[' then count[:square_brackets] += 1
      when ']' then count[:square_brackets] -= 1
      end

      break if closes_unopened? # any count is negative?
      break if closes_wrong?(bracket) # other bracket being closed other than @last_opened_bracket?

      # By setting this to nil, the code complete ignores previous opened brackets, meaning it does not
      # detect any wrongly closed brackets that were opened before this one
      @last_opened_bracket = nil     if closes_last_opened?(bracket) # closes @last_opened_bracket?
      @last_opened_bracket = bracket if opener?(bracket) # bracket is '{', '(' or '['?
    end

    count.values.all?(&:zero?)
  end
```

I see that it's an unnecessarily overengineered/verbose solution and that one would probably realize that before going through with it, but I feel that when you're learning to code you often fail to see unnecessary complications with your code.

Let me know if you think testing this edge case is a stretch or if there is a better way to do it (maybe with a new test).

This is my first PR here, I've read README, but let me know if I missed anything, `yarn test` was run.

### DIFF
--- a/exercises/matching-brackets/canonical-data.json
+++ b/exercises/matching-brackets/canonical-data.json
@@ -120,8 +120,7 @@
     },
     {
       "uuid": "1d5c093f-fc84-41fb-8c2a-e052f9581602",
-      "reimplements": "a0205e34-c2ac-49e6-a88a-899508d7d68e",
-      "description": "paired and wrong nested brackets",
+      "description": "paired and wrong deep nested brackets",
       "comments": [
         "Ensures last opened bracket is not the only one being traced"
       ],
--- a/exercises/matching-brackets/canonical-data.json
+++ b/exercises/matching-brackets/canonical-data.json
@@ -119,6 +119,19 @@
       "expected": false
     },
     {
+      "uuid": "1d5c093f-fc84-41fb-8c2a-e052f9581602",
+      "reimplements": "a0205e34-c2ac-49e6-a88a-899508d7d68e",
+      "description": "paired and wrong nested brackets",
+      "comments": [
+        "Ensures last opened bracket is not the only one being traced"
+      ],
+      "property": "isPaired",
+      "input": {
+        "value": "[({}])"
+      },
+      "expected": false
+    },
+    {
       "uuid": "ef47c21b-bcfd-4998-844c-7ad5daad90a8",
       "description": "paired and incomplete brackets",
       "property": "isPaired",
--- a/exercises/matching-brackets/canonical-data.json
+++ b/exercises/matching-brackets/canonical-data.json
@@ -120,7 +120,7 @@
     },
     {
       "uuid": "1d5c093f-fc84-41fb-8c2a-e052f9581602",
-      "description": "paired and wrong deep nested brackets",
+      "description": "paired and wrong nested brackets but innermost are correct",
       "comments": [
         "Ensures last opened bracket is not the only one being traced"
       ],


petertseng (approved):
well, never underestimate the creativity of incorrect solutions. note that they also had to ignore the case where `last_opened` was nil to pass all other tests. For the code in the PR description to fully describe the problem it should have inlined the other function calls, since only then does it make apparent it's ignoring a `last_opened` of nil.

petertseng:
EBWODP: I guess my statement doesn't really make sense since ignoring that would only cause it to reject otherwise-valid inputs.

kimesf:
Yes, it would reject valid inputs, if I understood what you meant.

The point I tried to make with the code in the PR description is that it's totally possible to green all tests with an incomplete solution. When closing the last opened bracket by setting `last_opened_bracket` to `nil`, it completely ignores previous opened brackets.

I commented the code in the PR to highlight this.

ErikSchierboom (approved):

exercises/matching-brackets/canonical-data.json
```diff
@@ -118,6 +118,19 @@
       },
       "expected": false
     },
+    {
+      "uuid": "1d5c093f-fc84-41fb-8c2a-e052f9581602",
+      "reimplements": "a0205e34-c2ac-49e6-a88a-899508d7d68e",
```
jiegillet, line 123: The new test is fine, but I don't see why it needs to reimplement another one. Why not have both?

kimesf: I guess my reasoning when chosing to reimplement the test was that having only the new one better conveyed the intent behind the problem, I thought that maybe the first implementating could lure the user to code the wrong answer - or something like that.

That being said, maybe having this redundancy is a better way of asserting a valid implementation and might even help the user to understand the problem in a more didactic way (step by step) - given the user went with a wrong implementation.

Honestly, I didn't give it much thought because it seemed to me that's exactly the kind of dilema better discussed with the people that maintain exercism (you guys). Maybe I should've shared this toughts when creating the PR, instead of just raising the possibility of making a new test for the issue.

I would happily change it to implement a new test if you guys think that's a better approach to it.

petertseng: I don't mind the old test being reimplemented because I didn't see `[({]})` (old test) testing anything that `[({}])` (new test) doesn't already cover. If any other test is warranted in addition to this new test, it would actually be one where the outermost ones are correct, `([{]})`. The implementation that that test is designed for is even more "creatively wrong" though...

```ruby
  {
    # https://github.com/exercism/problem-specifications/pull/2071
    name: 'count only, but also checks last opened',
    should_fail: true,
    f: ->(i, _) {
      open = BracketMatcher::PAIRS.transform_values { 0 } 
      last_opened = nil 
      i['value'].each_char { |c| 
        if open.has_key?(c)
          last_opened = c 
          open[c] += 1
        end 
        if (o = BracketMatcher::CLOSES[c])
          return false if open[o] == 0
          return false if last_opened &.!= o
          last_opened = nil 
          open[o] -= 1
        end 
      }   
      open.values.all?(&:zero?)
    },  
  },  
  {
    # https://github.com/exercism/problem-specifications/pull/2071
    name: 'count only, but also checks first opened',
    should_fail: true,
    f: ->(i, _) {
      open = BracketMatcher::PAIRS.transform_values { 0 } 
      first_opened = nil 
      i['value'].each_char { |c| 
        if open.has_key?(c)
          first_opened ||= c
          open[c] += 1
        end 
        if (o = BracketMatcher::CLOSES[c])
          return false if open[o] == 0
          open[o] -= 1
          if open.values.all?(&:zero?)
            return false if first_opened != BracketMatcher::CLOSES[c]
            first_opened = nil 
          end 
        end 
      }   
      open.values.all?(&:zero?)
    },  
  }, 
```

kimesf: > I don't mind the old test being reimplemented because I didn't see `[({]})` (old test) testing anything that `[({}])` (new test) doesn't already cover. If any other test is warranted in addition to this new test, it would actually be one where the outermost ones are correct, `([{]})`. The implementation that that test is designed for is even more "creatively wrong" though...
> 
> ```ruby
>   {
>     # https://github.com/exercism/problem-specifications/pull/2071
>     name: 'count only, but also checks last opened',
>     should_fail: true,
>     f: ->(i, _) {
>       open = BracketMatcher::PAIRS.transform_values { 0 } 
>       last_opened = nil 
>       i['value'].each_char { |c| 
>         if open.has_key?(c)
>           last_opened = c 
>           open[c] += 1
>         end 
>         if (o = BracketMatcher::CLOSES[c])
>           return false if open[o] == 0
>           return false if last_opened &.!= o
>           last_opened = nil 
>           open[o] -= 1
>         end 
>       }   
>       open.values.all?(&:zero?)
>     },  
>   },  
>   {
>     # https://github.com/exercism/problem-specifications/pull/2071
>     name: 'count only, but also checks first opened',
>     should_fail: true,
>     f: ->(i, _) {
>       open = BracketMatcher::PAIRS.transform_values { 0 } 
>       first_opened = nil 
>       i['value'].each_char { |c| 
>         if open.has_key?(c)
>           first_opened ||= c
>           open[c] += 1
>         end 
>         if (o = BracketMatcher::CLOSES[c])
>           return false if open[o] == 0
>           open[o] -= 1
>           if open.values.all?(&:zero?)
>             return false if first_opened != BracketMatcher::CLOSES[c]
>             first_opened = nil 
>           end 
>         end 
>       }   
>       open.values.all?(&:zero?)
>     },  
>   }, 
> ```

Both cases (and also the original intent of the test) are about phrases where even though all brackets are indeed closed, they are closed in the wrong order. Should we change my reimplementation to solve both cases (with something like `([({}]))` - I'm surrounding the proposed idea with a correct set of outermost brackets), or you would prefer new tests being added? 

One approach would be to (1) change my reimplementation of "paired and wrong nested" to be a new test like "paired and deep wrong nested" (`[({}])`) and (2) then to add another test for "paired and wrong nest within outermost right" (`([{]})`). If we change the reimplementation to solve both edge cases, it might start to get confusing on why the test input is the way it is. It might be better for the user (which is learning) if they were actually separated tests for specific problems.

That being said though, IMO, I would still say to reimplement the one test ("paired and wrong nested") for all edge cases, not only they all refer to the same idea, I tend to believe it's not that big of a deal for the user to realize what they did wrong given they went with a wrong solution. Not only that, but these wrong solutions have to be so creative that most people would probably not understand - and maybe would get confused - on why there is a test for "deep nested" and "wrong within outermost right brackets" - and I don't think they should, these tests are not for them, they are for those who went with these wrong solutions. For explicability, if anyone ever asks, there would be two comments in the canonical JSON, one for each of the edge cases. That's why comments are there for.

What do you think? What do you do in those cases? Reimplements or create new ones?

petertseng: My guess is that it's better for each case to only test one thing because let's say a student submits an implementation that's incorrect in only one way. The student gets a better indication of what went wrong if the test case that failed is focused instead of trying to cover multiple incorrect implementations.

However, it is only a guess. I don't think we've ever had data to answer the question "do students get confused by test cases that are not meant for them, but only for particularly creative wrong solutions?"

If there is doubt about the value of `([{]})`, it doesn't necessarily need to be added until there is an actual sighting of a submitted solution that would get it wrong.

kimesf: Having a better indication of what went wrong is indeed better. I agree.

kimesf:
For better indication of what went wrong in case of wrong implementation, I changed it to add a new test instead of reimplementing the old one.

exercises/matching-brackets/canonical-data.json
```diff
@@ -118,6 +118,18 @@
       },
       "expected": false
     },
+    {
+      "uuid": "1d5c093f-fc84-41fb-8c2a-e052f9581602",
+      "description": "paired and wrong deep nested brackets",
```
petertseng, line 123: I'm concerned that now that this is a new test case instead of reimplementing, the description, when contrasted with the existing "paired and wrong nested brackets", is confusing. The level of nesting is equal to the existing one, so the word "deep" sounds wrong to me.

My preference is that the existing case be reimplemented for reasons I've already discussed previously, but if this must be added as a new case, I would ask for a more indicative name. My instinct is "paired and wrong nested brackets but innermost are correct"

kimesf: I would rather adopt your suggestion, but keep the new test.

For now, I tend to agree with the argument that we want a more indicative test in case of a wrong implementation.

We established that are multiple ways of messing up "paired but wrong order". So instead of one big input value that solves multiple edge cases, having a specific test for each one would be, maybe, more indicative of what is wrong with their implementation.

The new test can be added only when and if a student brings it up here.

I went back and forth with this in my mind, but ended up settling with your previous argument, but now hearing you're team reimplementation, I'm confused hahaha

petertseng (approved):

jiegillet (approved):

Issue #2072, opened 2022-07-08
Triangle: Bug allows faulty solution to be accepted


krion7:
Hello,

I have encountered a bug on the Python exercise of `Triangle`, however it seems to affect other language tracks as well, per my observation at `exercises/triangle/canonical-data.json`. The bug only pertains to the scalene part of the exercise.

This is how the bug goes. (I will be demonstrating it using Python; sidenote: I am quite new to coding and GitHub as well, so please excuse me.) Say, we've got a code like this:
```
def scalene(sides):
    a, b, c = sides
    return valid_triangle(sides) and a != b != c
```
That code is accepted as a valid solution when verified with current test cases. However, this code **shouldn't be a valid solution** because of this:
```
>>> scalene([2, 3, 2])
True
```
Of course, a triangle with lengths of 2, 3, and 2 is **not scalene**, since a scalene triangle has **all sides of different lengths**, but it is recognized as such by that faulty code.

Now, there is no such test case in the current test cases that is identical to `scalene([2, 3, 2])`. That is how faulty codes like the one shown above is incorrectly accepted as valid. So, I recommend adding a new test case akin to `scalene([2, 3, 2])`.

Thanks as always for your continuous work and contributions to Exercism!
(If needed, [this is the issue I opened at the Python Exercism repo](https://github.com/exercism/python/issues/3134#issue-1297628449))

petertseng:
That case should be added.

When adding that case **do not remove** the existing case `c6e15a92-90d9-4fb3-90a2-eef64f8d3e1e "two sides are equal" [4, 4 ,3]`. Instead, change the description of the existing case to be "two long sides are equal". The description of this new added case should be "two short sides are equal". (Or any two descriptions that make clear the difference between these two cases).

SleeplessByte:
@krion7 would you like to (try to) PR this in this repository? We'll help getting the PR to the right state before it's merged. Your request change is very welcome.

PR #2075, opened 2022-07-12
triangle: Add new test case for scalene triangles to solve a bug [#2072]

krion7:
To solve a bug discussed in #2072, I have made this PR. 

In short, I added a new test case to the scalene triangle part of the test cases with the name "two unadjacent sides are equal" of [3, 4, 3], while still keeping the existing "two sides are equal" test case albeit renamed.

### DIFF
--- a/exercises/triangle/canonical-data.json
+++ b/exercises/triangle/canonical-data.json
@@ -192,7 +192,7 @@
           "expected": false
         },
         {
-          "uuid": "",
+          "uuid": "3da23a91-a166-419a-9abf-baf4868fd985",
           "description": "two unadjacent sides are equal",
           "property": "scalene",
           "input": {
--- a/exercises/triangle/canonical-data.json
+++ b/exercises/triangle/canonical-data.json
@@ -184,10 +184,19 @@
         },
         {
           "uuid": "c6e15a92-90d9-4fb3-90a2-eef64f8d3e1e",
-          "description": "two sides are equal",
+          "description": "two adjacent sides are equal",
           "property": "scalene",
           "input": {
             "sides": [4, 4, 3]
+          },
+          "expected": false
+        },
+        {
+          "uuid": "",
+          "description": "two unadjacent sides are equal",
+          "property": "scalene",
+          "input": {
+            "sides": [3, 4, 3]
           },
           "expected": false
         },
--- a/exercises/triangle/canonical-data.json
+++ b/exercises/triangle/canonical-data.json
@@ -184,7 +184,7 @@
         },
         {
           "uuid": "c6e15a92-90d9-4fb3-90a2-eef64f8d3e1e",
-          "description": "two adjacent sides are equal",
+          "description": "first and second sides are equal",
           "property": "scalene",
           "input": {
             "sides": [4, 4, 3]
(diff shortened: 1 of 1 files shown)


glennj:
You appear to be missing a UUID for the new case

exercises/triangle/canonical-data.json
```diff
@@ -184,13 +184,22 @@
...
+          "description": "two adjacent sides are equal",
           "property": "scalene",
           "input": {
             "sides": [4, 4, 3]
           },
           "expected": false
         },
+        {
+          "uuid": "3da23a91-a166-419a-9abf-baf4868fd985",
+          "description": "two unadjacent sides are equal",
```
petertseng, line 196: Calling the sides unadjacent makes me uncomfortable. Since it's a triangle, for any pair of sides, those two sides are adjacent to each other (they share a vertex). This word "adjacent" is therefore used in contravention of usual meaning, because here it was used to mean whether the two numbers are adjacent in the input list.

However, I have recognised this as a situation where I cannot be allowed to let my discomfort stand in the way of progress, which is why I Approved anyway.

petertseng (approved):

exercises/triangle/canonical-data.json
krion7: True, I didn't think about that. Any suggestions on how to better name the description?

petertseng: Referring to the sides as first, second, and third could work, since I don't recall the possibility of these three words being misunderstood to mean something, unlike "adjacent". Of course someone might fairly complain that then a case of the form `[a, b, b]` is missing. Your choice whether to add that `[a, b, b]` case or not.

krion7: Sure, that will work, I think I will just change it to "first and second sides are equal" and "first and third sides are equal".

MatthijsBlom: > Your choice whether to add that `[a, b, b]` case or not.

If it were up to me, this case would be added, just in case (and because it is cheap).

The code that slipped through the tests originally went along the lines of (Python) `a, b, c = sides ; return a != b != c`. A variant that might still slip through is `b != a != c`.

krion7: @MatthijsBlom On a second thought, yep, it's best to add that as well.

exercises/triangle/canonical-data.json
```diff
@@ -200,6 +200,15 @@
           },
           "expected": false
         },
+        {
+          "uuid": "b6a75d98-1fef-4c42-8e9a-9db854ba0a4d",
+          "description": "second and third sides are equal",
+          "property": "scalene",
+          "input": {
+            "sides": [4, 3, 4]
```
MatthijsBlom, line 208: The first instead of the second side is equal to the third.

krion7:
Ah, my apologies for making that mistake.

petertseng (approved):

BethanyG (approved):

ErikSchierboom (approved):

MatthijsBlom:
Congratulations, @krion7 🎉

krion7:
Thanks for the help everyone!

PR #2081, opened 2022-07-21
[Crypto-Square Instructions] Typo and clarification

habere-et-dispertire:
in to => into

### DIFF
--- a/exercises/crypto-square/description.md
+++ b/exercises/crypto-square/description.md
@@ -4,11 +4,10 @@ Implement the classic method for composing secret messages called a square code.
 
 Given an English text, output the encoded version of that text.
 
-First, the input is normalized: the spaces and punctuation are removed
-from the English text and the message is down-cased.
+First, the input is normalized: the spaces and punctuation are removed from the English text and the message is down-cased.
 
-Then, the normalized characters are broken into rows. These rows can be
-regarded as forming a rectangle when printed with intervening newlines.
+Then, the normalized characters are broken into rows.
+These rows can be regarded as forming a rectangle when printed with intervening newlines.
 
 For example, the sentence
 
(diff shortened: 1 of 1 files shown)


exercises/crypto-square/description.md
```diff
@@ -22,13 +22,13 @@ is normalized to:
 "ifmanwasmeanttostayonthegroundgodwouldhavegivenusroots"
 ```
 
-The plaintext should be organized in to a rectangle. The size of the
-rectangle should be decided by the length of the message.
+The plaintext should be organized into a rectangle as square as possible.
```
glennj, line 25: I don't think we need "as square as possible". The criteria for the rectangle follow immediately below.

habere-et-dispertire: I found those criteria confusing without a clear problem statement. It lead me to considering the low-level implementation of meeting these constraints rather than focusing on the nature of the problem more abstractly.

If the idea is that people are to unravel the problem statement as well as solve the problem, then I concur, but it appears to come at the cost of being unhelpful, particularly toward beginners.

glennj: I hear what you're saying.

I suppose saying `c = ceil(sqrt(length))` is a bit too big of a hint.

jiegillet: I like having the requirement explained intuitively before giving the formal criteria 👍

jiegillet (approved):

exercises/crypto-square/description.md
habere-et-dispertire: Thank you @jiegillet for distilling my thoughts. :)

I was striving for an idiomatic raku solution without explicit iterator variables. At least two of the three given constraints can be met [without direct congruence checking](https://exercism.org/tracks/raku/exercises/crypto-square/solutions/habere-et-dispertire).

So I find that to be part of the origin of my concern. The focus on the `r` and `c` criteria felt somewhat like a distracting procedural detail when the problem statement was under-defined for me searching for an orthogonal solution.

kytrinyx (approved):

exercises/crypto-square/description.md
kytrinyx: @glennj are you okay with this phrasing, given the above discussion?

ErikSchierboom: I think the "as square as possible" _could_ help and is unlikely to cause any harm.

ErikSchierboom (approved):

exercises/crypto-square/description.md
kytrinyx: That's where I've landed on this one as well.

PR #2085, opened 2022-08-09
[scale-generator] Fix output description

dlesnoff:
The description is misleading. In the Nim course, we only have to generate 7 notes with the 7 intervals. The last interval is redundant, since the last note it gives is also the first note of the scale.

### DIFF
--- a/exercises/scale-generator/description.md
+++ b/exercises/scale-generator/description.md
@@ -65,7 +65,7 @@ Given a tonic and a set of intervals, generate the musical scale starting with
 the tonic and following the specified interval pattern.
 
 This is similar to generating chromatic scales except that instead of returning
-12 notes, you will return N+1 notes for N intervals.
+12 notes, you will return N notes for N intervals (we omit the last N+1 note in the scale description, since it is the repetition of the tonic).
 The first note is always the given tonic.
 Then, for each interval in the pattern, the next note is determined by starting from the previous note and skipping the number of notes indicated by the interval.
 


petertseng (commented):
My request for requiring only N notes will just be: There needs to be a clear answer for what to do with the last interval. If the test only checks for N notes, then the Nth interval can just be ignored. That was pointed out as a problem in https://github.com/exercism/problem-specifications/issues/1640 and requiring only N notes from N intervals was confusing students, so therefore N+1 notes became required. Is there a better solution to this problem? Some thoughts come to mind.

* Require only N notes, but also require some sort of validation of the Nth interval, namely a check that it would make the `N+1`th note the same as the first. This implies that we must add some tests that test that a set of invalid intervals (intervals that make an `N+1`th note that isn't the same as the first) are rejected.
* Require only N notes, and make explicit mention that the Nth interval can be ignored. That allows students to assume that the intervals are valid as part of the contract of this problem, without having to do the validation. But why would we have an input that's ignored?
* Say "you will return N notes for N-1 intervals", and keep all tests the same. Really that's just rearranging terms so I don't think that changes the situation much.

I realise that my wording implies that I prefer a particular solution over the others, and I apologise for that. I found it difficult to be impartial toward the other solutions. I have no excuse for my behaviour, only an explanation.

petertseng:
EBWODP: I realise that my wording implies that I prefer a particular solution over the others, and I apologise for that. I found it difficult to be impartial toward the other solutions. I have no excuse for my behaviour, only an explanation.

dlesnoff:
There are N intervals for N notes in the tests (at least for the Nim track). So the tests require to return as much notes as intervals. Third option is a no-go.

It makes sense for me, musically, to ignore the last interval.
In modal music, we change the tonic but keep the same intervals of one scale. The last interval is important to define the mode, but not to compute the notes themselves.
I think the second option makes sense due to the context.

In the first option, I do not think that the tests add much to the problem. It is complex enough for non-musical enclined people.

petertseng:
Whatever we do, the description in https://github.com/exercism/problem-specifications/blob/main/exercises/scale-generator/description.md should match the tests in https://github.com/exercism/problem-specifications/blob/main/exercises/scale-generator/canonical-data.json, and the tests have one more note than interval. If a specific track makes a conscious decision to test differently, it can make a conscious decision to modify the description appropriately for their track specifically as well.

So, if the correct decision is "Require only N notes, and make explicit mention that the Nth interval can be ignored", then both the description and the https://github.com/exercism/problem-specifications/blob/main/exercises/scale-generator/canonical-data.json should be modified.

dlesnoff:
Sorry, the Nim's track was out of sync. We fixed it, and thus I do not think this PR is still useful. I did not know about the canonical data. Thanks.

PR #2087, opened 2022-08-14
say: Add more tests

aifrak:
As stated in the [PR #1169 from the Elixir track](https://github.com/exercism/elixir/pull/1169), it is possible that all tests passed but some numbers are still wrong.

Wrong numbers that I found were between 20 and 1000.

### DIFF
--- a/exercises/say/canonical-data.json
+++ b/exercises/say/canonical-data.json
@@ -51,6 +51,24 @@
       "expected": "twenty-two"
     },
     {
+      "uuid": "f010d4ca-12c9-44e9-803a-27789841adb1",
+      "description": "thirty",
+      "property": "say",
+      "input": {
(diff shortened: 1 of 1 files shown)


rvandam:
I came here to point out a similar problem with the lack of tests for 11-19 in higher positions. For instance, 12015 as "twelve thousand fifteen". A naive implementation might end up with something like "ten-two thousand fifteen" and still pass all existing tests because they only have special handling for 11-19 and not for 11-19 thousand or million, etc.

jiegillet (approved):

neenjaw (approved):

SleeplessByte (approved):

SleeplessByte:
Thanks! Merged; tracks can now implement those tests right away.

PR #2094, opened 2022-09-02
issue_2475_roman_numerals_max_3999

dbh:
Adjust Instructions and Unit Tests to support maximum Roman numeral 3999
Issue https://github.com/exercism/go/issues/2475

Adjustment to problem specification, instructions, and unit tests to support the max val 3,999 which can be expressed in Roman numerals (without getting into the "Larger number" scenarios)

### DIFF
--- a/exercises/roman-numerals/description.md
+++ b/exercises/roman-numerals/description.md
@@ -20,7 +20,7 @@ into stone tablets).
  7  => VII
 ```
 
-The maximum number supported by this this notation is 3,999.
+The maximum number supported by this notation is 3,999.
 (The Romans themselves didn't tend to go any higher)
 
 Wikipedia says: Modern Roman numerals ... are written by expressing each
--- a/exercises/roman-numerals/description.md
+++ b/exercises/roman-numerals/description.md
@@ -20,7 +20,7 @@ into stone tablets).
  7  => VII
 ```
 
-There is no need to be able to convert numbers larger than about 3000.
+The maximum number supported by this this notation is 3,999.
 (The Romans themselves didn't tend to go any higher)
 
 Wikipedia says: Modern Roman numerals ... are written by expressing each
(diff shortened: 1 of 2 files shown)


dbh:
Issue write-up is at https://github.com/exercism/problem-specifications/issues/2095 
It is more appropriate in project specifications than in the go language project since the change in spec instructions and unit tests affects more than go.

jiegillet (approved):
Sounds good to me, I wouldn't expect tracks implemented an artificial cutoff at 3000 (although my expectations aren't worth much)

axtens:
Suggestion: have another set of tests that require the implementation of the [Unicode extensions](https://en.wikipedia.org/wiki/Roman_numerals#Unicode) to Roman numerals

IsaacG:
@junedev More reviews are needed here. Mind adding one?

glennj:
> Suggestion: have another set of tests that require the implementation of the [Unicode extensions](https://en.wikipedia.org/wiki/Roman_numerals#Unicode) to Roman numerals

Sounds like that would be a brand new exercise (like resistor-color is a series of exercises)

dbh:
> > Suggestion: have another set of tests that require the implementation of the [Unicode extensions](https://en.wikipedia.org/wiki/Roman_numerals#Unicode) to Roman numerals
> 
> 
> 
> Sounds like that would be a brand new exercise (like resistor-color is a series of exercises)

Agreed. I think Unicode Roman numerals would be a new effort and problem spec. 

This particular PR is to clarify an existing problem spec.

exercises/roman-numerals/description.md
```diff
@@ -20,7 +20,7 @@ into stone tablets).
  7  => VII
 ```
 
-There is no need to be able to convert numbers larger than about 3000.
+The maximum number supported by this this notation is 3,999.
```
petertseng, line 23:
```suggestion
The maximum number supported by this notation is 3,999.
```

junedev (approved):

IsaacG (approved):

PR #2098, opened 2022-09-24
Delete outdated contributing guide

kytrinyx:
The contributing guide was so outdated as to be more wrong than right.

This deletes the guide itself as well as the mention in README pointing to it.

If global documentation is needed for the problem-specifications repository I think we should add documentation to the exercism/docs so that this documentation can be accessed as part of the "Building Exercism" docs here: https://exercism.org/docs/building

We can probably start by opening an issue in the docs repo with a link back to this PR, and bullet points about what we need to document.


junedev (approved):
Good idea deleting this. I am also a fan of keeping docs in the docs repo and then only linking to them where needed.

SleeplessByte (approved):

ErikSchierboom (approved):
Great!

PR #2099, opened 2022-09-24
Fix outdated description for Meetup exercise

kytrinyx:
Back in 2015 we determined that we should not implement fifth, since this exercise is about scheduling regular monthly meetups.

Since not all months have a fifth week, it doesn't make sense to schedule a regular meetup on the fifth Wednesday (or whatever) of every month.

We removed the test cases, but the description was not updated to remove the reference to the 'fifth' scenario.


junedev (approved):
👍

petertseng (approved):

PR #2105, opened 2022-10-01
Simplify and correct the Grep exercise

hkBst:
The Grep exercise description suffers from multiple problems (at least in the Rust variant):
 - it implies that the task is to search for regular expressions, until one enters the editor.
 - it incorrectly states that a single string should be returned (and also gives examples that show that)
 - it leaves out information about the search order of the files

I have simplified the description and fixed these problems.


exercises/grep/description.md
```diff
@@ -1,77 +1,25 @@
-# Description
+# Instructions
```
SleeplessByte, line 1: This really is a description file. The heading is stripped out when imported into a track. It should not be changed to `instructions`.

exercises/grep/description.md
```diff
@@ -1,77 +1,25 @@
...
 
-Search a file for lines matching a regular expression pattern. Return the line
-number and contents of each matching line.
+Search files for lines matching a search string and return all matching lines.
 
-The Unix [`grep`](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/grep.html) command can be used to search for lines in one or more files
-that match a user-provided search query (known as the *pattern*).
+The Unix [`grep`](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/grep.html) command searches for lines in files that match a regular expression. Your task is to implement a simplified `grep` function, which supports searching for literal strings.
```
SleeplessByte, line 5: Can you move the second sentence (`Your task`) to the next line?

As we're updating this file, it should now follow [the Exercism Markdown style](https://exercism.org/docs/building/markdown/markdown#layout).

exercises/grep/description.md
```diff
@@ -1,77 +1,25 @@
...
 
-The `grep` command takes three arguments:
+The `grep` function takes three arguments:
```
SleeplessByte, line 7: It's a command. That in rust (perhaps?) you implement an equally named _function_ is an implementation detail. In various tracks, like JavaScript, you write a working program that has a `grep` command (not a function).

exercises/grep/description.md
```diff
@@ -1,77 +1,25 @@
...
 
-1. The pattern used to match lines in a file.
-2. Zero or more flags to customize the matching behavior.
-3. One or more files in which to search for matching lines.
+1. The string to search for.
+2. Zero or more flags to customize its behavior.
```
SleeplessByte, line 10: In English this `its` refers to `flags`, not `grep`. Repeating "the matching" resolves that. But if you say `matching`, then `pattern` is perhaps a better candidate because of pattern matching.

exercises/grep/description.md
```diff
@@ -1,77 +1,25 @@
...
-```text
-input.txt:hello
-input.txt:hello again
-greeting.txt:hello world
-```
-
-If given just one file, this prefix is not present.
+It then reads the contents of the specified files (in the order specified),
+finds the lines that contain the search string,
+and finally returns those lines (in the order in which they were found).
```
SleeplessByte, line 15: Can you make this a single line?

exercises/grep/description.md
```diff
@@ -1,77 +1,25 @@
...
 
 ## Flags
 
-As said earlier, the `grep` command should also support the following flags:
+The `grep` function should support the following flags:
```
SleeplessByte, line 19:
```suggestion
The `grep` command should support the following flags:
```

exercises/grep/description.md
```diff
@@ -1,77 +1,25 @@
...
 
-- `-n` Prefix each matching line with its line number within its file.
-  When multiple files are present, this prefix goes *after* the filename prefix.
-- `-l` Print only the names of files that contain at least one matching line.
-- `-i` Match line using a case-insensitive comparison.
+- `-n` Prepend the line number to each line in the output.
```
SleeplessByte, line 21: What happened to the "multiple files" note?

SleeplessByte (commented):
Whilst I think simplifying instructions is a good idea, it seems that perhaps Rust has an incorrect implementation of this exercise and should be fixed instead.

I've left some commentary in-line that needs fixing. 

@exercism/reviewers please check thoroughly and compare with your tracks current implementations.

exercises/grep/description.md
hkBst: Done.

exercises/grep/description.md
hkBst: Done.

exercises/grep/description.md
hkBst: Done.

exercises/grep/description.md
hkBst: Is it not true that since "flags" is plural it would need to be referred to with "their", and thus cannot be the target of "its"? Anyway, since it is potentially confusing, I've replaced "its" with "the command's".

exercises/grep/description.md
hkBst: Done.

exercises/grep/description.md
hkBst: Done.

exercises/grep/description.md
hkBst: I've now included this information in line 14 and this flag's description.

exercises/grep/description.md
```diff
@@ -1,77 +1,24 @@
 # Description
 
-Search a file for lines matching a regular expression pattern. Return the line
-number and contents of each matching line.
+Search files for lines matching a search string and return all matching lines.
 
-The Unix [`grep`](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/grep.html) command can be used to search for lines in one or more files
-that match a user-provided search query (known as the *pattern*).
+The Unix [`grep`](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/grep.html) command searches for lines in files that match a regular expression.
```
kytrinyx, line 5: @hkBst Did you dislike the use of "pattern" in the original description?

exercises/grep/description.md
```diff
@@ -1,77 +1,24 @@
...
 ## Flags
 
-As said earlier, the `grep` command should also support the following flags:
+The `grep` command should support the following flags:
 
-- `-n` Prefix each matching line with its line number within its file.
-  When multiple files are present, this prefix goes *after* the filename prefix.
-- `-l` Print only the names of files that contain at least one matching line.
-- `-i` Match line using a case-insensitive comparison.
+- `-n` Prepend the line number and a colon (':') to each line in the output. (Comes after possible file name.)
```
kytrinyx, line 20:
```suggestion
- `-n` Prepend the line number and a colon (':') to each line in the output, placing the number after the filename (if present).
```

exercises/grep/description.md
```diff
@@ -1,77 +1,24 @@
...
+- `-l` Instead of lines, only output the names of the files that contain at least one matching line.
```
kytrinyx, line 21:
```suggestion
- `-l` Output only the names of the files that contain at least one matching line.
```

exercises/grep/description.md
```diff
@@ -1,77 +1,24 @@
...
-```text
-hello
-hello again
-```
-
-The `grep` command should support multiple flags at once.
-
-For example, running `grep -l -v "hello" file1.txt file2.txt` should
-print the names of files that do not contain the string "hello".
+- `-x` Search instead for lines matching the search string entirely, instead of merely containing it.
```
kytrinyx, line 24:
```suggestion
- `-x` Search only for lines where the search string matches the entire line.
```

kytrinyx (commented):
I think this is a very nice simplification.

I've added a few suggested tweaks to the wording in a few places.

exercises/grep/description.md
hkBst: I was not so happy that "pattern" was used ambiguously to refer both to "regular expression" and to "literal string", so I tried to clarify that while the real grep searches for regular expression( pattern)s, in this exercise you only need to implement literal string( pattern) search. The ambiguity of the word "pattern" made me try to avoid its use.

kytrinyx: I think the ambiguity is fine, since below it's clearly stated that this exercise is a simplification and only cares about literal strings. If people are unfamiliar with grep they can look at the linked documentation.

My vote would be to go with `pattern` here.

kotp: And the fact that grep does match patterns, both with literal strings and regular expressions, it covers it (and reiterates what we find to be described in the first(?) line of the documentation.

hkBst: What exactly are you proposing? Do you just want to add the term "pattern"? Or do you want it to replace something else? And to what purpose?

The ambiguity about regular expression vs literal strings was perhaps my main motivation to take a look at improving the prose of this exercise. I am glad that you think this is now very clear, but also a bit afraid to lose this clarification.

hkBst: On further reflection I am happy to replace "regular expression" here with "pattern". Can you propose a change?

kytrinyx: I don't feel strongly about this at all, I was mostly just curious why you preferred regex to pattern when I had the exact opposite reaction :) Usually when this happens I learn something interesting from asking the question.

I have a very weak preference towards:

> The Unix [`grep`](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/grep.html) command searches for lines in files that match a pattern.

But having had this discussion I'm perfectly happy with the current state of the sentence.

kytrinyx (approved):

exercises/grep/description.md
SleeplessByte: Eh. Yeah you're right. I'm that case it would refer to the last singular noun which is not the command so your solution solves all of it. 🙏🏽

exercises/grep/description.md
glennj: This can be read as "files that match a regular expression" instead of "lines that match a regular expression"

Perhaps this (which hand-waves over grep processing its stdin, or handling multiple regexes):

> The grep command searches files for lines matching a regular expression.

I would also prefer "pattern" over "regular expression", mostly because `grep -F` exists to match fixed strings.

> The grep command searches files for lines matching a pattern.

kytrinyx: Ah good catch on that one.

> The grep command searches files for lines matching a pattern.

This is now my favorite.

hkBst: Since fixed strings are a special case of regular expressions, I don't think we gain anything by being non-specific with "pattern". If anyone wonders what a pattern is and they click through, they find out that it is basically a regular expression, although grep supports two kinds of regular expressions and there is special support for easily matching on fixed strings. So, therefore, I wonder why not just say regular expression, when that is basically all grep does?

I do agree that this word reordering is a definite improvement.

hkBst: One more data point: I think the wikipedia page for regular expression is much more relevant here than the one for pattern.

RE: A regular expression is a sequence of [characters](https://en.wikipedia.org/wiki/Character_(computing)) that specifies a [search pattern](https://en.wikipedia.org/wiki/Pattern_matching) in [text](https://en.wikipedia.org/wiki/String_(computer_science)).
Pattern: A pattern is a regularity in the world, in human-made design, or in [abstract](https://en.wikipedia.org/wiki/Abstraction) ideas.

kytrinyx: That works for me.

hkBst:
This has lingered due to not having the 3 required approving reviews. Perhaps I should have re-requested review? Anyway, I solved a merge conflict that came up, and have now tried to re-request review...

SleeplessByte (approved):

neenjaw (approved):
Did a read through, simplifications make sense.

kytrinyx:
Thanks for everybody's persistence in this one. Merging! 🎉 ✨

PR #2111, opened 2022-10-06
Tweak asterisk formatting for plain-text readability

kytrinyx:
In several exercise descriptions we use asterisks for purposes other than markdown text emphasis.

Mostly that is to represent the concept of multiplication, though in one instance we are simply referring to a literal asterisk.

On the website these unescaped asterisks would mess up the formatting, half-emphasized text or invalid HTML.

However escaping the asterisks makes it harder to understand the instructions when working in a local development environment rather than in the online editor.

In other exercises we've chosen to format arithmetic formulae as code rather than escape the asterisks, which works for both environments.

I've therefore chosen to extend this choice to the files that had escaped asterisks.


ErikSchierboom (approved):
I like it a lot! The escaped asterisks have always bugged me.

kotp (approved):

PR #2112, opened 2022-10-06
Delete bonus point sections from descriptions

kytrinyx:
The bonus points sections were a way of encouraging more discussion and exploration in v1.

Discussion:
Now that we have closed mentoring rather than open discussions it makes less sense to encourage discussion in this way. We may find other ways of encouraging discussions, but then it will be a product-level design choice rather than a textual call to action from within the exercise description.

Exploration:
We tend to want people to explore the approach to solving the problem rather than to explore how to expand the scope of the problem. As such I think it's worth deleting the exploration aspect of the bonus points sections.

Also, there are only a handful of exercises that had this sort of section, so it's also neater in terms of consistency.

### DIFF
--- a/exercises/beer-song/description.md
+++ b/exercises/beer-song/description.md
@@ -305,17 +305,3 @@ Take it down and pass it around, no more bottles of beer on the wall.
 No more bottles of beer on the wall, no more bottles of beer.
 Go to the store and buy some more, 99 bottles of beer on the wall.
 ```
-
-## For bonus points
-
-Did you get the tests passing and the code clean? If you want to, these
-are some additional things you could try:
-
-* Remove as much duplication as you possibly can.
-* Optimize for readability, even if it means introducing duplication.
-* If you've removed all the duplication, do you have a lot of
-  conditionals? Try replacing the conditionals with polymorphism, if it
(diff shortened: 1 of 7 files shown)


ErikSchierboom (approved):
The thing that bugged me most about these bonuspoints is that for many tracks, they didn't even apply.

glennj (approved):
Right, without tests these goals are unmeasurable.

junedev (approved):
Totally agree with what was said in the PR description.

siebenschlaefer:
In the "triangle" exercise the bonus section was about degenerate triangles where the sum of the two smaller sides is equal to the third side. The tests didn't cover that case which made sense because the students might have wanted to handle them differently if they "dug deeper".  
But now with that section gone the tests still don't cover them, leaving that case somewhat ambiguous and unspecified. Can we have a test case for a degenerate triangle that makes the expectations clear?

ErikSchierboom:
> But now with that section gone the tests still don't cover them, leaving that case somewhat ambiguous and unspecified. Can we have a test case for a degenerate triangle that makes the expectations clear?

We could. Feel free to PR.

PR #2113, opened 2022-10-06
Use reference links in descriptions

kytrinyx:
Exercism's Markdown Specification says to use reference links instead of inline markdown links.

This goes through all the description.md files for all the exercises and converts inline markdown links to reference links.

There's no pressure for maintainers to sync these changes; rather they're nice to have for future changes, and will make it easier to do other cleanup such as switching to one sentence per line (also preferred in the Exercism markdown specification https://exercism.org/docs/building/markdown/markdown#layout)

### DIFF
--- a/exercises/alphametics/description.md
+++ b/exercises/alphametics/description.md
@@ -2,8 +2,7 @@
 
 Write a function to solve alphametics puzzles.
 
-[Alphametics](https://en.wikipedia.org/wiki/Alphametics) is a puzzle where
-letters in words are replaced with numbers.
+[Alphametics][] is a puzzle where letters in words are replaced with numbers.
 
 For example `SEND + MORE = MONEY`:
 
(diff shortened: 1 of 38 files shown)


ErikSchierboom (approved):

glennj (approved):

exercises/alphametics/description.md
```diff
@@ -2,8 +2,7 @@
 
 Write a function to solve alphametics puzzles.
 
-[Alphametics](https://en.wikipedia.org/wiki/Alphametics) is a puzzle where
-letters in words are replaced with numbers.
+[Alphametics][] is a puzzle where letters in words are replaced with numbers.
```
wolf99, line 5: Should we use this form with the empty brackets?
If it matters, it would be more consistent to use the same form everywhere, i.e. `[Some thing][some-thing]`

kytrinyx: I used the empty form since it was already used in places, but I actually prefer the form that you suggest. I will go ahead and make that change!

PR #2115, opened 2022-10-07
sgf-parsing: Add warning about string literals

petertseng:
The test case added in 3c33c249121c16aa2a7fc3398a50349d8985f752 had to be very careful to specify the difference between the two. Add it to the description as well.

### DIFF
--- a/exercises/sgf-parsing/description.md
+++ b/exercises/sgf-parsing/description.md
@@ -64,6 +64,13 @@ newlines any whitespace character encountered is replaced with a single space
 character irrespective of if the original character is found in its escaped
 from or its unescaped from.
 
+Be careful not to get confused between:
+
+- The string as it is represented in a string literal in the tests
+- The string that is passed to the SGF parser
+
+Escape sequences in the string literals may have already been processed by the programming language's parser before they are passed to the SGF parser.
+
 There are a few more complexities to SGF (and parsing in general), which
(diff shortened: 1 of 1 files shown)


exercises/sgf-parsing/description.md
```diff
@@ -64,6 +64,13 @@ newlines any whitespace character encountered is replaced with a single space
 character irrespective of if the original character is found in its escaped
 from or its unescaped from.
 
+Be careful not to get confused between:
+
+- The string as it is represented in a string literal in the tests
+- The string that is passed to the SGF parser
+
+Escape sequences in the string literals may have already been processed by the programming language's parser before they are passed to the SGF parser.
```
petertseng, line 72: I don't like that I have to be wishy-washy with "may" here, but I can't speak for every language simultaneously because different languages may be able to show the strings differently, such as with some string literal that doesn't have escape sequences (some languages may call this a *raw string*).

It would also be great if I could show an example, but I also doubt I can do one that makes sense universally for all languages.

kytrinyx (approved):

exercises/sgf-parsing/description.md
ErikSchierboom: Agree with the sentiment. The nature or prob-specs means we sometimes have to be somewhat _hand-wavy_, but this is still an improvement!

ErikSchierboom (approved):

PR #2117, opened 2022-10-07
Additional tests for straights

hkBst:
Additional tests for straights with the same tests also for straight flushes. (Originally proposed in the Rust track: https://github.com/exercism/rust/pull/1555)

### DIFF
--- a/exercises/poker/canonical-data.json
+++ b/exercises/poker/canonical-data.json
@@ -169,7 +169,7 @@
       "expected": ["4D AH 3S 2D 5C"]
     },
     {
-      "uuid": "?",
+      "uuid": "e214b7df-dcba-45d3-a2e5-342d8c46c286",
       "description": "aces cannot be in the middle of a straight (Q K A 2 3)",
       "property": "bestHands",
       "input": {
@@ -286,7 +286,7 @@
       "expected": ["5S 7S 8S 9S 6S"]
     },
     {
-      "uuid": "?",
+      "uuid": "d9629e22-c943-460b-a951-2134d1b43346",
       "description": "aces can end a straight flush (10 J Q K A)",
       "property": "bestHands",
       "input": {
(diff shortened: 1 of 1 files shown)


exercises/poker/canonical-data.json
```diff
@@ -269,12 +278,48 @@
     },
     {
       "uuid": "d0927f70-5aec-43db-aed8-1cbd1b6ee9ad",
-      "description": "both hands have straight flush, tie goes to highest-ranked card",
+      "description": "both hands have a straight flush, tie goes to highest-ranked card",
```
kytrinyx, line 281: If you want to fix an exercise description you're going to have to add a new test that reimplements the old one. Check out this bit of the docs: https://github.com/exercism/problem-specifications/#changing-tests

hkBst: That seems to be a bit heavy for such a tiny change.

kytrinyx: Indeed. In that case it will need to stay as it is.

hkBst: I found this in the docs:

> Test cases are immutable, which means that once a test case has been added, it never changes. There are two exceptions:
>  - The comments field can be mutated and thus does not require adding a new test case when changing its value.
>  - **The description field can be mutated and thus does not require adding a new test case when changing its value.**

kytrinyx: Ahhh, I'm so glad you checked. I had misremembered the docs. I thought the exception was the `comments` field and the `scenarios` field.

petertseng (approved):
all added cases are correct

kytrinyx (approved):

glennj (approved):

kytrinyx:
@hkBst Would you like to change the order per @petertseng's suggestion? If not I can take care of it when I get home tonight.

hkBst:
@kytrinyx Thanks for reordering and merging!

PR #2122, opened 2022-10-09
Add titles to the metadata for all exercises

kytrinyx:
Only a handful of exercises had titles.
These were typically exercises where the correct title would be difficult to guess based on the slug (e.g. D&D Characters).

However, it turns out that many exercises have variations on the titles across different tracks.

This hard-codes titles for all the exercises, even if they would technically be fairly straight forward to derive from the slug.

This will make it easier for Configlet to do normalization on the track `config.json`.

### DIFF
--- a/exercises/nucleotide-count/metadata.toml
+++ b/exercises/nucleotide-count/metadata.toml
@@ -1,3 +1,4 @@
+title = "Nucleotide Count"
 blurb = "Given a DNA string, compute how many times each nucleotide occurs in the string."
 source = "The Calculating DNA Nucleotides_problem at Rosalind"
 source_url = "http://rosalind.info/problems/dna/"
--- a/exercises/octal/metadata.toml
+++ b/exercises/octal/metadata.toml
@@ -1,3 +1,4 @@
+title = "Octal"
 blurb = "Convert a octal number, represented as a string (e.g. '1735263'), to its decimal equivalent using first principles (i.e. no, you may not use built-in or external libraries to accomplish the conversion)."
 source = "All of Computer Science"
 source_url = "http://www.wolframalpha.com/input/?i=base+8"
--- a/exercises/two-bucket/metadata.toml
+++ b/exercises/two-bucket/metadata.toml
@@ -1,3 +1,4 @@
+title = "Two Buckets"
 blurb = "Given two buckets of different size, demonstrate how to measure an exact number of liters."
 source = "Water Pouring Problem"
 source_url = "http://demonstrations.wolfram.com/WaterPouringProblem/"
--- a/exercises/two-fer/metadata.toml
+++ b/exercises/two-fer/metadata.toml
@@ -1,3 +1,3 @@
-title = "Two-fer"
+title = "Two Fer"
 blurb = "Create a sentence of the form \"One for X, one for me.\"."
 source_url = "https://github.com/exercism/problem-specifications/issues/757"
--- a/exercises/accumulate/metadata.toml
+++ b/exercises/accumulate/metadata.toml
@@ -1,3 +1,4 @@
+title = "Accumulate"
 blurb = "Implement the `accumulate` operation, which, given a collection and an operation to perform on each element of the collection, returns a new collection containing the result of applying that operation to each element of the input collection."
 source = "Conversation with James Edward Gray II"
 source_url = "https://twitter.com/jeg2"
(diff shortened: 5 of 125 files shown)


junedev (approved):

exercises/nucleotide-count/metadata.toml
```diff
@@ -1,3 +1,4 @@
+title = "nucleotide-count"
```
wolf99, line 1: This one is in the format of the soug rather than title case.

exercises/octal/metadata.toml
```diff
@@ -1,3 +1,4 @@
+title = "octal"
```
wolf99, line 1: This one is not in title case

exercises/nucleotide-count/metadata.toml
kytrinyx: Thank you, good catch!

exercises/octal/metadata.toml
kytrinyx: Thank you for catching that.

wolf99 (approved):

ErikSchierboom (approved):

exercises/two-bucket/metadata.toml
```diff
@@ -1,3 +1,4 @@
+title = "Two Buckets"
```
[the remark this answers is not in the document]
kytrinyx: The plural is not intentional. Thank you for doing this double-checking. In retrospect I realize that I should have written the scripts to check that all exercises had a title, and that all titles matched some of the defined names in the track's config.json files.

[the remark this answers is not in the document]
kytrinyx: I agree; the others look good to me.

Thank you!

kytrinyx:
I will submit PRs to all tracks this week to make them match the title in problem specifications.

Once we know that all tracks are in compliance, I think it would make sense to do 1 & 3.

kytrinyx:
From a tool-maintainer perspective I would also instinctively want to do (a), but from a product point of view I would prefer (c), as the track `config.json` has more to do with how things fit together at a high level and how the track is displayed on the site.

I think I would prefer for `sync` to either to fix the title value in the track-level `config.json` or do nothing.

From the perspective of someone using configlet to help bootstrap a track, I think the most helpful thing would be to have a separate `configlet init` command that takes the `--exercise` argument and assumes that you have not done _anything_ yet to start working on that exercise. It would then

1. Write an entry to the track-level `config.json` with all the required fields, adding the value for `slug` and `title` and adding a new `uuid`.
2. Sync the exercise docs and metadata (without asking for confirmation)
3. Add empty files for all the files specified in the `config.json`

I'm not sure what the ideal behavior would be around the tests. Probably you'd want to either ask for confirmation for each test, but maybe just defer the problem and have people sync the tests separately.

This would remove a lot of the fiddly bits of getting started with a new exercise, even if the track doesn't have a generator, and then we could assume that the `title` is correct for the purposes of the `sync` command, and just error out in the `lint` command if it is wrong for some reason.

kytrinyx:
> Could you please open an issue in the configlet repo?

Yes, I'll do this now.

> And filepaths, right?

Yes, and filepaths.

> Ideally, I think sync should still handle syncing the title. The title could change in prob-specs due to a typo/bad capitalization/style, or because the name of the exercise changed, right? We'd want to propagate that upstream change.

Hm, yes, on reading this my conclusion is that you're right.


So the conclusion so far is:
- we want a separate `init` or `bootstrap` or something command, which I will open an issue for now, and
- `lint` shouldn't error
- `sync` should update the title if there is a mismatch

exercises/two-fer/metadata.toml
```diff
@@ -1,3 +1,3 @@
-title = "Two-fer"
+title = "Two Fer"
```
petertseng, line 1: I feel a little unsure about this once, since it seems to imply that "fer" is a word, but it's difficult for me to really say that it is one.

kytrinyx: It's definitely not a word, but to me `Two Fer` isn't any less wrong than `Two-fer`, and I prefer the former as a title.

kytrinyx: I don't think my having made it up is a good reason, because I didn't put any thought into orthography when I made it.

Given the data about usage outside of Exercism I am leaning more heavily towards the hyphenated version of the title.

kytrinyx: Thank you @ee7. Those are good enough arguments that I don't feel like choosing `Two-Fer` over `Two Fer` is random.

PR #2123, opened 2022-10-10
Add date and datetime scenarios to gigasecond

kytrinyx:
The gigasecond test suite is a bit odd in that there are two different types for the same input.
    
Another oddity of the gigasecond test suite is that most of the tests don't actually add any value. If you get one passing, you get them all passing.
    
This adds a date scenario to the inputs that have only a date, and datetime to the ones that also include a time element.

Am I missing anything important here?

### DIFF
--- a/SCENARIOS.txt
+++ b/SCENARIOS.txt
@@ -1,4 +1,6 @@
 big-integer
+date
+datetime
 floating-point
 immutable
 input-validation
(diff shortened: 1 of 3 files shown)


glennj (approved):
unprompted, LGTM

jiegillet (approved):

PR #2124, opened 2022-10-11
More descriptive summary

gitonthescene:
Slightly more descriptive summary of the purpose of this project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # Problem Specifications
 
-Repository for exercises to be used across tracks.  This includes both problem statements available as markdown and test data described below.
+Repository for practice exercises to be used across tracks.
+This includes both problem statements available as markdown and test data as JSON described below.
 Note that tracks are not _required_ to use these exercises and some may be more or less appropriate to the given language.
 
 See the [new track roadmap](https://exercism.org/docs/building/tracks/new) for full details on setting up a new track.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,9 @@
 # Problem Specifications
 
-Shared metadata for Exercism exercises.
+Repository for exercises to be used across tracks.  This includes both problem statements available as markdown and test data described below.
```
kytrinyx, line 3: I put each sentence on separate lines to match the specification that we adhere to (https://exercism.org/docs/building/markdown/markdown).

Also prefixed exercises with 'practice' since Exercism also has concept exercises.

```suggestion
Repository for practice exercises to be used across tracks.
This includes both problem statements available as markdown and test data described below.
```

kytrinyx (approved):
I think this is an improvement, and as such I'd be happy to see this merged quickly.

README.md
ErikSchierboom: Maybe also mention that the test data is in JSON, similar to how the problem statement is in markdown?

kytrinyx: That's a good call. I updated the suggested change.

kytrinyx (approved):

ErikSchierboom (approved):

glennj (approved):
An question that comes to mind: "what's a 'track'?"

Possibly link to https://exercism.org/docs/building/tracks or https://exercism.org/tracks

gitonthescene:
@glennj - I think you could link that explicitly but a link to setting up a track is in the second paragraph which hopefully serves as breadcrumbs to finding that info.

PR #2125, opened 2022-10-11
custom-set: metadata: add title

ee7:
Commit 6e57ab4490d9 (#2122) intended to add the `title` key to every exercise that lacked one, but custom-set was omitted.


kytrinyx (approved):
Thank you!

glennj (approved):

PR #2126, opened 2022-10-11
two-bucket: metadata: make title singular

ee7:
Follow-up from https://github.com/exercism/problem-specifications/pull/2122#discussion_r992007032

### DIFF
--- a/exercises/two-bucket/metadata.toml
+++ b/exercises/two-bucket/metadata.toml
@@ -1,4 +1,4 @@
-title = "Two Buckets"
+title = "Two Bucket"
 blurb = "Given two buckets of different size, demonstrate how to measure an exact number of liters."
 source = "Water Pouring Problem"
 source_url = "http://demonstrations.wolfram.com/WaterPouringProblem/"


siebenschlaefer:
I'm not a native English speaker, but shouldn't "buckets" be plural because there's two?

kytrinyx:
> I'm not a native English speaker, but shouldn't "buckets" be plural because there's two?

That is why I got it wrong in the first PR. But the slug is `two-bucket`, so it would make sense for the name to be singular.

All the tracks that have implemented this exercise have used `"Two Bucket"` as the name.

(I suppose the name doesn't have to follow the slug exactly, but I feel like we'd need to actually be conscious about that choice).

kytrinyx (approved):

siebenschlaefer:
So this title "Two Bucket" doesn't get displayed anywhere where proper grammar is expected? Ok then, I withdraw my objections.

ErikSchierboom (approved):

kytrinyx:
I'm okay either way. @ErikSchierboom @iHiD What is the current change process if we decide to change the name?

I imagine:
- duplicate the exercise and deprecate the old one
- then potentially change all the tracks to rename `two-bucket` to `two-buckets` and give it a new UUID?
- anything else?

kytrinyx:
Wait, I think I misunderstood the suggestion.

There are two options, as I see it:

1. Change the exercise to be `two-buckets` (name: Two Buckets). That would mean duplicating, deprecating, etc.
2. Just have the name of the `two-bucket` problem to be `Two Buckets`. That wouldn't require any big overhaul.

If (1) then I prefer to merge this as is and do the duplication/deprecation as a second step.
If (2) then I prefer to just close this PR.

kytrinyx:
I've dug into this and I think option 1 is easy enough. I would be happy to do the work to move all tracks to using the new exercise.

As such, I'd like to merge this PR, then I'll open a new one  to deprecate `two-bucket` and implement `two-buckets`.

kytrinyx:
> So we can't just rename the exercise in prob-specs like we did in the past, and we need to add exercises/two-bucket/.deprecated in this repo?

@ee7 Yeah, not since we reopened the repository for changes.

petertseng:
If someone forced me to come up with an argument as to why it should stay two bucket, the argument I'd use is that the https://en.wikipedia.org/wiki/Three-body_problem is called the three-body problem, not the three bodies problem, and therefore in the same vein this problem should be called the two bucket problem instead of the two buckets problem.

Fortunately, I am not being forced to argue for two bucket by anyone or anything, so it's just whatever consensus decides here.

kytrinyx:
@petertseng Had it been "The Two Bucket Problem" I would agree with you. Without "Problem" or "Solution" tacked on, though, I do think "Two Buckets" sounds more correct. The current description doesn't talk about "The Two Bucket Problem", mentioning "two buckets".

Having said all of that... this exercise has existed since 2014 and nobody has ever complained about the title being singular. The only reason this has turned into a discussion is that I was trying to make all titles consistent and accidentally wrote "Two Buckets" when it should have been "Two Bucket".

As I said before, I don't really feel strongly either way about this one, I just want the title to be explicitly defined so that we nudge tracks in the direction of consistency.

ErikSchierboom:
My personal vote would be for option 2: having the title be different from the exercise's slug. I don't really mind that,

iHiD:
To chime in from a website perspective, I'm pretty against deprecating/duplicating the exercise in the **track repos**, as that will mean everyone who's completed it gets a duplicate exercise in their history, and has to redo the same work on a new exercise, which feels like a bad user experience. 

So I think from a track-repo perspective the slug should just be **updated** with the uuid remaining the same. But that sounds like manual work for all the tracks, some of which may be scriptable, but some of which won't be  (e.g. for tracks where directory names are with code (e.g. `TwoFer.foo` must be in a dir called `two_fer`) there's manual code changes).

All of which makes me feel the churn factor here isn't worth changing the slug.

kytrinyx:
> All of which makes me feel the churn factor here isn't worth changing the slug.

Yeah, agreed.

Okay, I'm going to call it:
- We're not going to duplicate and deprecate.
- Whichever way we go, there's potential for someone to get slightly confused, either by the slug being mismatched to the title, or the title not being grammatically correct.
- It's been `Two Bucket` since 2014 with nobody raising it as an issue.

I'm going to go ahead and decide that we will continue to have the slug and title that we've had since 2014: `two-bucket` and `Two Bucket` (singular).

PR #2127, opened 2022-10-11
dominoes: Fix reference to Eulerian graph in comment

dreig:
I believe this was a misprint. A graph that contains such a cycle is called Eulerian. See: https://mathworld.wolfram.com/EulerianGraph.html


kytrinyx (approved):
I'm pretty sure you're right. There is such a thing as a Euclidean Graph, apparently, (this is definitely beyond the scope of my knowledge), but in the context of Dominoes I'm pretty sure we want a Eulerian graph.

petertseng (approved):
my fault

PR #1802, opened 2021-05-07
Tournament: ensure points are sorted numerically not lexically

glennj:
Issue #1801

PR #2135, opened 2022-10-16
Change wording in tournament instructions

kytrinyx:
There was a new test added in #1802 that enforced the behavior of the ordering. A downstream maintainer pointed out that when there is a test enforcing behavior, then the instructions ~should~ must read *must* rather than *should*.

### DIFF
--- a/exercises/tournament/description.md
+++ b/exercises/tournament/description.md
@@ -24,7 +24,7 @@ A win earns a team 3 points.
 A draw earns 1.
 A loss earns 0.
 
-The outcome should be ordered by points, descending.
+The outcome is ordered by points, descending.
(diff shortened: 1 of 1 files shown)


kotp (approved):

kotp:
Also, it might be helpful if there was wording about what happens in the case of the points being the same value.  How are they then sorted? Alphabetically by team name, or in the order that they were received?

kytrinyx:
The next line says that in case of a tie the results are ordered alphabetically by team name.

kotp:
> The next line says that in case of a tie the results are ordered alphabetically by team name.

:facepalm

wolf99 (approved):

PR #2137, opened 2022-10-27
http to https

kytrinyx:
- Convert http links to https in metadata
- Fix http urls in markdown to be https

For all http links where an https equivalent would work,
use the secure link in exercise metadata.

For all JumpstartLab references I switched them to turing.edu
since I was on the team at JumpstartLab at the time, and
we were teaching at what became Turing School of Software and
Design. (JumpstartLab's website doesn't handle https).

There are a few links where I couldn't find an equivalent
that resolved:

- http://butunclebob.com/ArticleS.UncleBob.TheBowlingGameKata
- http://claysnow.co.uk/recycling-tests-in-tdd/"
- http://butunclebob.com/ArticleS.UncleBob.ThePrimeFactorsKata"
- http://hyperphysics.phy-astr.gsu.edu/hbase/Organic/transcription.html
- http://rubykoans.com/

I would appreciate suggestions for what to use for these, as we could then
add link checking to ensure that all links are https.


SaschaMann:
> I would appreciate suggestions for what to use for these, as we could then
add link checking to ensure that all links are https.

Using archive links might work, e.g. https://web.archive.org/web/20220408112140/http://hyperphysics.phy-astr.gsu.edu/hbase/Organic/transcription.html

kytrinyx:
@SaschaMann that is an excellent idea. If I am not mistaken we already have a couple of those already. I will do that.

kotp (approved):

kotp:
Approved, even before the archive links are made.

kytrinyx:
Ok, I think I got all of them.

@ErikSchierboom Do you know if we have a link-checker set up for any of the repositories in the org to ensure we only link to HTTPS links?

ErikSchierboom:
@kytrinyx I don't think  we do

SaschaMann (approved):

wolf99:
We could probably add  the lychee action to the workflows for a link checker and require https.

It is already used in a few repos.

I'll open an issue for this and assign it to myself.

PR #2140, opened 2022-10-28
[DND-Character] Add raku track as a language supporting dice

habere-et-dispertire:
See: https://docs.raku.org/routine/roll

### DIFF
--- a/exercises/dnd-character/description.md
+++ b/exercises/dnd-character/description.md
@@ -26,6 +26,8 @@ Because constitution is 3, the constitution modifier is -4 and the hitpoints are
 
 Most programming languages feature (pseudo-)random generators, but few programming languages are designed to roll dice.
 One such language is [Troll][troll].
+Another is Raku, available as a [track] on Exercism.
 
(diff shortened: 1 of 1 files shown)


jiegillet:
Raku is not "designed to roll dice" [the way Troll is](https://static-curis.ku.dk/portal/files/17117511/Troll-SAC.pdf).
Am I missing something?

habere-et-dispertire:
It does certainly not go as far, but I think it was a design consideration in raku to address a common programming pitfall. Please can you clarify @Util_PM ? I remember some video of yours that expounded on this. Thanks! :-)

IsaacG:
At 22m56s, the presenter is talking about how Perl 6 is written mostly in Perl 6.

Raku has a "roll" function which randomly selects an element from a sequence N times. This is definitely handy when you need to "roll dice" but ... it's just a "select random element" function. I'm not seeing how this is any different from any other language with a "select random element" function.

habere-et-dispertire:
Okay, I don't think I thought this one through enough.
Thank you all for your time and consideration.
:)

PR #2143, opened 2022-10-29
typo: Add missing 'as' to phone number exercise description

jtepe:


jiegillet:
I believe the current version is correct, "so that they can be sent SMS messages" means "so that you can send them SMS messages".

SleeplessByte (changes requested):
The sentence was never meant to indicate the _phone numbers_ themselves would be sent _as_ SMS messages.

jtepe:
Ah, sorry. Then I misread it to be referring to the phone numbers, not the users :smile: 
My bad. Closing.

PR #2146, opened 2022-11-06
Switch schema reference URL to HTTPS

wolf99:
Seems missed from the recent switchover in #2137.

### DIFF
--- a/canonical-data.schema.json
+++ b/canonical-data.schema.json
@@ -22,7 +22,7 @@
     " slightly reducing the amount of manually generated code.   "
   ],
 
-  "$schema": "http://json-schema.org/draft-04/schema#",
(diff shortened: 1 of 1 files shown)


wolf99:
Shoot. Looks like AJV does not support HTTPS schema links, because the schema spec does not either, despite being provided on the web at HTTPS 😕 

https://github.com/ajv-validator/ajv/issues/1450

kytrinyx:
Yeah, I deliberately skipped that link, as it didn't seem like it was supported.

Can we make the link checker ignore json files? I think we only care about markdown and toml.

wolf99:
Happily the link checker already ignores this. I happened on it manually only.

PR #2148, opened 2022-11-11
Add `random` scenario

ErikSchierboom:
This PR adds the `random` scenario to identifying test cases that deal with randomness.

- scenarios: add 'random' scenario
- dnd-character: add 'random' scenario
- diffie-hellman: add 'random' scenario
- simple-cipher: add 'random' scenario

### DIFF
--- a/SCENARIOS.txt
+++ b/SCENARIOS.txt
@@ -6,5 +6,6 @@ immutable
 input-validation
 library-test
 local-scope
+random
 runtime-validation
 unicode
--- a/canonical-data.schema.json
+++ b/canonical-data.schema.json
@@ -127,6 +127,7 @@
         "input-validation",
         "library-test",
         "local-scope",
+        "random",
         "runtime-validation",
         "unicode"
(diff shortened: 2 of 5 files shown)


SCENARIOS.txt
```diff
@@ -6,5 +6,6 @@ immutable
 input-validation
 library-test
 local-scope
+random
```
ErikSchierboom, line 9: I've considered `randomness` as an alternative. I picked random as it seems to best fit the other scenarios, but I could easily be swayed to go a different route.

kytrinyx: I think `random` is fine. I don't feel strongly about the naming and could go either way. If someone has a really strong argument for why it should be something else I'd probably be convinced.

kytrinyx (approved):

jiegillet (approved):

BethanyG (approved):

axtens (approved):

PR #2151, opened 2022-11-16
Add test case

zeltbrennt:
A new test case for certain solutions that would satisfy all previous tests but would still be wrong. 

Resolves Exercism/java#2196

### DIFF
--- a/exercises/dominoes/canonical-data.json
+++ b/exercises/dominoes/canonical-data.json
@@ -170,6 +170,20 @@
         "dominoes": [[1, 2], [2, 3], [3, 1], [4, 5], [5, 6], [6, 4]]
       },
       "expected": false
+    },
+    {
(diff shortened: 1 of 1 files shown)


andrerfcsantos:
Hi @zeltbrennt,

Thanks for opening this. Changed the wording on the PR description from "Resolves <issue>" to "Related <issue>". With "resolves" that would cause the issue on the Java track to close when this PR is merged. We don't necessarily want that because after the PR is approved here we still need one in the Java track to update the Java tests and then you can close the issue there.

I'll let people more familiar with problem specs do the actual review on this one.

IsaacG:
Is there a forum discussion for this issue where the proposed change was discussed and given a go-ahead?

glennj:
> Is there a forum discussion for this issue where the proposed change was discussed and given a go-ahead?

I think this was created before the forum.

PR #2154, opened 2022-11-22
ci: simplify yarn caching

ErikSchierboom:
The `actions/setup-node` action has caching support built-in nowadays: https://github.com/actions/setup-node#caching-global-packages-data
This PR removes the custom yarn caching logic in favor of the built-in caching support.

See https://github.com/exercism/blog/pull/155 for an example where this is used.


kytrinyx (approved):
This is nice :-)

jiegillet (approved):

PR #2155, opened 2022-11-22
Use `ubuntu-22.04` instead of `ubuntu-latest` in GitHub Actions workflows.

ErikSchierboom:
With CI it is good practice to pin your dependencies to a specific
version, as not doing could result in builds suddenly breaking when a
"latest" version is upgraded.


kytrinyx (approved):
Good call

jiegillet (approved):

junedev (approved):
This big fan of version pinning approves. 🙂

PR #2162, opened 2022-12-07
Resistor Color Trio: extend tests, borrowing from awk

IsaacG:
[See here](https://forum.exercism.org/t/canonical-resistor-color-trio-testing-gap-values-over-1000-which-are-not-kiloohms/) for forum discussion

### DIFF
--- a/exercises/resistor-color-trio/canonical-data.json
+++ b/exercises/resistor-color-trio/canonical-data.json
@@ -60,6 +60,78 @@
         "value": 470,
         "unit": "kiloohms"
       }
+    },
+    {
+      "uuid": "ee7a1b79-8848-4b1a-bf50-498c6fab7f27",
+      "description": "Brown and red and red",
+      "property": "label",
+      "input": {
+        "colors": ["brown", "red", "red"]
+      },
+      "expected": {
+        "value": 1200,
+        "unit": "ohms"
+      }
+    },
+    {
+      "uuid": "5f6404a7-5bb3-4283-877d-3d39bcc33854",
+      "description": "Blue and violet and grey",
+      "property": "label",
+      "input": {
+        "colors": ["blue", "violet", "grey"]
+      },
+      "expected": {
+        "value": 6700,
+        "unit": "megaohms"
+      }
+    },
+    {
+      "uuid": "7d3a6ab8-e40e-46c3-98b1-91639fff2344",
+      "description": "Minimum possible value",
+      "property": "label",
+      "input": {
+        "colors": ["black", "black", "black"]
+      },
+      "expected": {
+        "value": 0,
+        "unit": "ohms"
+      }
+    },
+    {
+      "uuid": "ca0aa0ac-3825-42de-9f07-dac68cc580fd",
+      "description": "Maximum possible value",
+      "property": "label",
+      "input": {
+        "colors": ["white", "white", "white"]
+      },
+      "expected": {
+        "value": 99,
+        "unit": "gigaohms"
+      }
+    },
+    {
+      "uuid": "0061a76c-903a-4714-8ce2-f26ce23b0e09",
+      "description": "First two colors make an invalid octal number",
+      "property": "label",
+      "input": {
+        "colors": ["black", "grey", "black"]
+      },
+      "expected": {
+        "value": 8,
+        "unit": "ohms"
+      }
+    },
+    {
+      "uuid": "30872c92-f567-4b69-a105-8455611c10c4",
+      "description": "Ignore extra colors",
+      "property": "label",
+      "input": {
+        "colors": ["blue", "green", "yellow", "orange"]
+      },
+      "expected": {
+        "value": 650,
+        "unit": "kiloohms"
+      }
     }
   ]
 }
--- a/exercises/resistor-color-trio/canonical-data.json
+++ b/exercises/resistor-color-trio/canonical-data.json
@@ -62,26 +62,14 @@
       }
     },
     {
-      "uuid": "ee7a1b79-8848-4b1a-bf50-498c6fab7f27",
-      "description": "Brown and red and red",
-      "property": "label",
-      "input": {
-        "colors": ["brown", "red", "red"]
-      },
-      "expected": {
-        "value": 1200,
-        "unit": "ohms"
-      }
-    },
-    {
       "uuid": "5f6404a7-5bb3-4283-877d-3d39bcc33854",
-      "description": "Blue and violet and grey",
+      "description": "Blue and violet and blue",
       "property": "label",
       "input": {
-        "colors": ["blue", "violet", "grey"]
+        "colors": ["blue", "violet", "blue"]
       },
       "expected": {
-        "value": 6700,
+        "value": 67,
         "unit": "megaohms"
       }
     },
--- a/exercises/resistor-color-trio/description.md
+++ b/exercises/resistor-color-trio/description.md
@@ -46,9 +46,16 @@ So an input of `"orange", "orange", "black"` should return:
 
 > "33 ohms"
 
-When we get more than a thousand ohms, we say "kiloohms".
-That's similar to saying "kilometer" for 1000 meters, and "kilograms" for 1000 grams.
+When we get to larger resistors, a [metric prefix][metric-prefix] is used to indicate a larger magnitude of ohms, such as "kiloohms".
+That is similar to saying "2 kilometers" instead of "2000 meters", or "2 kilograms" for "2000 grams".
 
-So an input of `"orange", "orange", "orange"` should return:
+For example, an input of `"orange", "orange", "orange"` should return:
 
 > "33 kiloohms"
+
+The metric prefixes used in this exercise are "kilo", "mega", and "giga".
+Those give the units of "kiloohms", "megaohms", and "gigaohms" respectively.
+
+Note, the quantity should remain a whole number, so an input of `"orange", "orange", "green"` should return "3300 kiloohms" and not "3.3 megaohms".
+
+[metric-prefix]: https://en.wikipedia.org/wiki/Metric_prefix
--- a/exercises/resistor-color-trio/description.md
+++ b/exercises/resistor-color-trio/description.md
@@ -53,7 +53,4 @@ For example, an input of `"orange", "orange", "orange"` should return:
 
 > "33 kiloohms"
 
-The metric prefixes used in this exercise are "kilo", "mega", and "giga".
-Those give the units of "kiloohms", "megaohms", and "gigaohms" respectively.
-
 [metric-prefix]: https://en.wikipedia.org/wiki/Metric_prefix
--- a/exercises/resistor-color-trio/description.md
+++ b/exercises/resistor-color-trio/description.md
@@ -56,6 +56,4 @@ For example, an input of `"orange", "orange", "orange"` should return:
 The metric prefixes used in this exercise are "kilo", "mega", and "giga".
 Those give the units of "kiloohms", "megaohms", and "gigaohms" respectively.
 
-Note, the quantity should remain a whole number, so an input of `"orange", "orange", "green"` should return "3300 kiloohms" and not "3.3 megaohms".
-
 [metric-prefix]: https://en.wikipedia.org/wiki/Metric_prefix


exercises/resistor-color-trio/canonical-data.json
```diff
@@ -60,6 +60,78 @@
...
+    },
+    {
+      "uuid": "ee7a1b79-8848-4b1a-bf50-498c6fab7f27",
+      "description": "Brown and red and red",
+      "property": "label",
+      "input": {
+        "colors": ["brown", "red", "red"]
+      },
+      "expected": {
+        "value": 1200,
```
meatball133, line 72: You often say 1.2 k ohms instead of 1200 ohms. 
This may seem like a small thing. Although becomes pretty apparent when you come up to resistors with 5 bands.
Writing 11300 ohms just becomes very long.
I am not an expert on how it should be written but Wikipedia uses the style I mentioned.
I know some people who are better (than me) at electrical engineering to double-check this style.

exercises/resistor-color-trio/canonical-data.json
```diff
@@ -60,6 +60,78 @@
...
+    },
+    {
+      "uuid": "7d3a6ab8-e40e-46c3-98b1-91639fff2344",
+      "description": "Minimum possible value",
+      "property": "label",
+      "input": {
+        "colors": ["black", "black", "black"]
+      },
+      "expected": {
+        "value": 0,
```
meatball133, line 96: A bit unsure of this one. 
A lot of the zero ohms resistor is just one black band.
Sure there are some with multiple but the most common ones have just 1 band.

exercises/resistor-color-trio/canonical-data.json
```diff
@@ -60,6 +60,78 @@
...
+    },
+    {
+      "uuid": "5f6404a7-5bb3-4283-877d-3d39bcc33854",
+      "description": "Blue and violet and grey",
+      "property": "label",
+      "input": {
+        "colors": ["blue", "violet", "grey"]
+      },
+      "expected": {
+        "value": 6700,
```
meatball133, line 84: Same as I mentioned earlier.

exercises/resistor-color-trio/canonical-data.json
IsaacG: The goal of this exercise is not to produce common parlance. It is to convert strings to numbers and back to strings.

The units are all integers and is makes sense to remain consistent with the existing tests in this file, as well as with the existing tests in tracks which already use this data.

exercises/resistor-color-trio/canonical-data.json
IsaacG: These resistors, be exercise definition, all have three bands. Other resistors which do not meet the exercise description are irrelevant. This exercise isn't asking students to read all real world resistors.

exercises/resistor-color-trio/canonical-data.json
meatball133: For the first part these changes you are bringing up will brick more or less every tracks solutions.

Meaning it doesnt matter.

For the second part. A lot of people have solved this exercise by using:

If x > 1000
 X / 1000

This goes right in hand with my suggestion.

Your way would mean:

If x % 1000
   X / 1000

And that solution have I not seen in more or less any language I have mentored in ever nor as a community solution.

I am just saying that If this is suppose to be a "realstic" problem with a "realstic" output it should use float values and the way of converting a float to a string in a lot of languages arent different to int.
What is also important to mention is the fact that If there every comes a resistor color quadro. So would that one require the use of float becuase the forth band tells the accuracy.

IsaacG: > For the first part these changes you are bringing up will brick more or less every tracks solutions.
> Meaning it doesnt matter.

I disagree. The problem spec says existing tests cannot be changed. These changes adds additional tests but do not change the behavior of existing tests. Track maintainers can opt to sync these additional tests or not.

> For the second part. A lot of people have solved this exercise by using:
> If x > 1000
> X / 1000

Can you quantify that assertion? Is that 0.5% of the solutions? 50%? I know that I would push back against seeing that in any code I mentor.

> Your way would mean:
> If x % 1000
> X / 1000
> And that solution have I not seen in more or less any language I have mentored in ever nor as a community solution.

In Python, I'd recommend `divmod()`. I've seen modulus used heavily in all the solutions I've read.

> I am just saying that If this is suppose to be a "realstic" problem with a "realstic" output it should use float values and the way of converting a float to a string in a lot of languages arent different to int.

Pros for this approach:
* These changes are consistent with the existing solutions.
* These changes are in line with existing changes with existing tracks.
* These changes are in line with what seems reasonable to me.

Cons for these changes:
* You think some people might not talk this way.

Is there any technical disadvantage or objection to these changes?

> What is also important to mention is the fact that If there ever comes a resistor color quadro. So would that one require the use of float becuase the forth band tells the accuracy.

This exercise is explicitly about three bands. A new exercise could handle the fourth band however it wants. That's not a technical or quantitative reasoning which has any relevancy to this change.

petertseng: Hmm. The verifier for this problem https://github.com/petertseng/exercism-problem-specifications/blob/verify/exercises/resistor-color-trio/verify.rb uses `>=` and `/`, because that's what made the most sense to me personally at the time, writing it. Previously the test cases did not distinguish between these following two rules:

1. If it is possible to express the number as a number of kiloohms greater than 1, express the number in kiloohms. Otherwise, express it in ohms.
2. Expess the number using the smallest SI prefix in {(none), kilo, mega, giga} that results in no fractional part.

So either was acceptable, and I happened to pick the first (which is not to say that the first is better or worse, just that it was a product of circumstance). To clarify, this test case's intention is to make explicit that only the second rule is acceptable, going forward? I will adjust accordingly if I can just get that confirmed, thank you.

petertseng: Hmm. The verifier for this problem https://github.com/petertseng/exercism-problem-specifications/blob/verify/exercises/resistor-color-trio/verify.rb uses `>=` and `/`, because that's what made the most sense to me personally at the time, writing it. Previously the test cases did not distinguish between these following two rules:

1. If it is possible to express the number as a number of kiloohms greater than or equal to 1, express the number in kiloohms. Otherwise, express it in ohms.
2. Expess the number using the largest SI prefix in {(none), kilo, mega, giga} that results in no fractional part (and therefore the smallest magnitude of value).
(a previous version of this message said "smallest", which didn't make sense. it should be "largest")

So either was acceptable, and I happened to pick the first (which is not to say that the first is better or worse, just that it was a product of circumstance). To clarify, this test case's intention is to make explicit that only the second rule is acceptable, going forward? I will adjust accordingly if I can just get that confirmed, thank you.

IsaacG: @petertseng  Yes, please. That would be wonderful and consistent with the tracks were I've seen these extended tests.

petertseng: I **strongly** advise that if the second rule is the correct rule going forward that https://github.com/exercism/problem-specifications/blob/main/exercises/resistor-color-trio/description.md **must** be changed, because it currently says this:

> When we get more than a thousand ohms, we say "kiloohms". That's similar to saying "kilometer" for 1000 meters, and "kilograms" for 1000 grams.

This implies that you **should** use `>= 1000` to decide whether to use kiloohms. Therefore, if this test case is added, it will contradict the description unless the description changes, and we do not want to contradict ourselves. It can't say "more than a thousand ohms" anymore; it needs to mention "a whole number of kiloohms" in order to not violate the rule.

The description will of course also need to be updated to mention mega and giga.

IsaacG: Thanks for noticing that, Peter! I updated the description. I'm hoping this doesn't make it overly verbose.

meatball133: I am leaving a replay to your earlier message Isaac.

> I disagree. The problem spec says existing tests cannot be changed. These changes adds additional tests but do not change the behavior of existing tests. Track maintainers can opt to sync these additional tests or not.

My way requires no current test to be changed.

> Can you quantify that assertion? Is that 0.5% of the solutions? 50%? I know that I would push back against seeing that in any code I mentor.

I will do a check if YOUR solution bricks it for people or for me in the most started solutions for: Javascript, Typescript, Python and Ruby.
Here is the result
Where my solution breaks: 25
Where Isaacs breaks: 36

Analsysis:

Why the numbers are so "low"? That is because a lot of exercises bricked for both of the new solutions. 
I also was fair in my choice of languages since how python and ruby deal with floats will brick my suggestion in all of those exercises more or less.
Even with that said your bricked 11 more solutions than my way.

> In Python, I'd recommend divmod(). I've seen modulus used heavily in all the solutions I've read.

With the first part, it sounds like you try to lecture me?
The exercise has only been out for a few days? None of the most starred solutions uses modulus. Therefore I am not sure what you are talking about. When I searched for the analysis I perhaps search 100+ solutions and I saw perhaps 2-4 use modulus.

> These changes are consistent with the existing solutions.

Also what are you basing this on???

> These changes are in line with existing changes with existing tracks.

I just told you my changes would brick fewer solutions. I did those tests for some of exercism that used languages aswell.

Pros for my approach:

- It is realistic in how you handle metric units.

- Breaks way fewer current solutions (not counting the addition of Giga and etc..)

- Would make it consistent with a further resistor color Quadro if it ever comes.

- Doesn't break any of the current problem spec

Cons for my approach:

- Would brick the solutions for the track awk


> This exercise is explicitly about three bands. A new exercise could handle the fourth band however it wants. That's not a technical or quantitative reasoning which has any relevancy to this change.

I think it does since you can use part of the resistor color dous solution for trio. This would make that impossible.

IsaacG: @iHiD could I impose on you to weigh in, please?

glennj:
A non-exhaustive survey: what would happen with a test for red/red/red => (20 + 2) * 10^2

tracks which produce integral results "2200 ohms"
- awk, bash, jq, tcl (bias alert: all of which I'm involved with)
- typescript

tracks which produce fractional results "2.2 kiloohms"
- groovy
- lua

tracks which produce *wrong* (well, loss of precision) results due to truncation or integer division
- elixir
- javascript
- kotlin
- nim
- python
- ruby
- wren
- csharp (I think, Eric please confirm)

tracks which top out at kiloohms
- csharp, elixir, javascript, lua, python, ruby, wren

popular tracks which do not implement resistor-color-trio
- java, fsharp, go, rust

---

It appears that [the schema](https://github.com/exercism/problem-specifications/blob/main/canonical-data.schema.json) does not disallow the expected value to be "one of a set of possible results", but I don't think that's been done before.

I don't think we'll come to consensus about whether "2200 ohms" or "2.2 kiloohms" is "correct" -- they both represent the exact same value. Should we add these new test cases to the spec, track maintainers can choose to implement them or not, or choose to override the expected value. Nobody is going to jail for ignoring this spec.

From the maintainers perspective, without knowing, I suspect there are many languages where it would be difficult to write tests that can compare the student's result with _either_ of the two possible correct values.

From the student's perspective, if the tests expect 2200 or they expect 2.2, the tests are setting out the requirements that guide the possible solutions, so we may be limiting some creativity there. 

---

That's a fairly "on the fence" opinion. For my tracks I've chosen to go one way. I'd be happy to have this PR approved.

meatball133:
Hi!

You must have done a small typo since you wrote Typescript under integral results.  I guess you meant another language.

I agree with the last part. It is true if some languages like to go another route they are free to do so.

glennj:
No, I mean typescript. It's implemented differently than javascript

https://github.com/exercism/typescript/blob/main/exercises/practice/resistor-color-trio/.meta/proof.ci.ts

meatball133:
I would say what matters is the test files, not the example. The reason for that is 1. The instructions for the typescript exercises are written as fractional results. That the example file goes against that can't I explain. Also secondly most of the typescript solution is made with fractional results in mind.

tltr: Test files are what matters not example.py

glennj:
> I would say what matters is the test files, not the example.

Disagree. From a track maintainers perspective, adding new tests to the file is one thing (may be easy or tedious depending if the track has a test generator). If you don't ensure that the example solution still works though, CI is broken for your track. The two things have to go hand-in-hand.

meatball133:
Yes but the example is often just written by one person and is just an example of how to solve it. It doesn't mean that all maintainers agree that it is the correct one. 

Also, a side, not javascript and typescript shares maintainers, and the languages are the same underneath.
They work should have the same solution because of that.

Also rewriting a solution isn't that much work.

meatball133:
You edited your message with new content so here is response to part 2 of your message. 

In the instructions they refer back to resistor color dou.

> orange-orange-orange would be 33 and 3 zeros, which becomes 33000 ohms.

This way is also incorrect in both my way and yours even if it is mentioned in the docs as an example.
It is for the user to understand what the operation should be. 

Therefore saying  that the docs says:

> orange-orange-red would be 33 and 2 zeros, which becomes 3300 ohms.

And basing that on saying that they want you to write 

3300 ohms is an incorrect assumption. Could be true but could also be false because of the example I mentioned above.

I am basing it on this:

> When we get more than a thousand ohms, we say "kiloohms".

glennj:
One last comment: "2200 ohms" versus "2.2 kiloohms" is currently unspecified by the tests. It doesn't matter how many existing student solutions would pass or fail the new tests. It matters what future students will do with the existing requirements.

I suspect it is the very rare student who goes back to check for new tests for their existing solutions. I bet the vast majority of students solve it then never look at it again.

meatball133:
Yes that is true. Most student don't look back and your inputs are valuable.

I dunno, I have read physics with the metric system. 
There are rules around writing prefixes. 
Personally, I like those to be followed.

I will therefore says this as the last thing:

I recommend changing to using 2.2 k ohms to follow the metric standard.
I know that might require some extra work but I prefer exercises to be as realistic as possible when it is a real problem.
I agree with you about the fact that this exercise would benefit from more tests in general.(I am not against the pr)

I also know that for the track I have done most contributions, python. Would this require a bit of work since how python deals with these kinds of numbers. At the same time, this kind of stuff would be very easy to implement for javascript because how it deals with numbers.

IsaacG:
> tltr: Test files are what matters not example

It's okay and encourages to have opinions and preferences. It's not okay or productive to state your opinions as facts, especially when they are very obviously subjective opinions.

> Personally, I like those to be followed.

That's fine. Opinion recorded. Other maintainers can chime in now :)


> Yes but the example is often just written by one person and is just an example of how to solve it. It doesn't mean that all maintainers agree that it is the correct one.

Yup. Your opinion was noted. Let other maintainers give their input now.

> Also, a side note, javascript and typescript share maintainers, and the languages are the same underneath. They should have the same solution because of that.

Unless I am mistaken, you're not one of those maintainers. It's best practice not to speak for other people. Let them share their own opinions if they want to. Don't assert you know what they want.

> Also rewriting a solution isn't that much work.

That's subjective. Don't assert your opinion is factual. It's not conducive to a productive conversation.

iHiD:
Thanks for opening Isaac, and to Glenn, Carl and Peter for chiming in.

My take on this is that we probably **shouldn't** merge it as it seems there's a strong divergence between tracks on this point. Ideally, I'd like to merge it with a scenario label, but as the description needs to be changed to accommodate the change, that's not currently possible. That means it's not a test that tracks can simply opt in or out of. Merging would therefore mean that all tracks had to adopt this test when they sync their data (else the README would be wrong), and which will then cause work for maintainers in updating their example files etc.

Without strong consensus across tracks (which I don't think we all probably want to try and achieve) I therefore don't think we should merge.

IsaacG:
> Without strong consensus across tracks (which I don't think we all probably want to try and achieve) I therefore don't think we should merge.

Thanks for weighing in, Jeremy. In this of this input, I updated the pull request and removed the test cases which are "ambiguous" i.e. can be represented as a fractional value.

IsaacG:
Can I get reviews on the new version, please?

glennj (approved):

IsaacG:
Thanks @glennj and @kytrinyx !
We just need one more 😄

ErikSchierboom:
> Thanks for weighing in, Jeremy. In light of this input, I updated the pull request and removed the test cases which are "ambiguous" i.e. can be represented as a fractional value.

Sorry to be nitpicky, but the issue Jeremy mentioned still holds with this version:

> Ideally, I'd like to merge it with a scenario label, but as the description needs to be changed to accommodate the change, that's not currently possible. That means it's not a test that tracks can simply opt in or out of. Merging would therefore mean that all tracks had to adopt this test when they sync their data (else the README would be wrong), and which will then cause work for maintainers in updating their example files etc.

It is not just the fractional part, but the fact that the canonical data introduces new metric prefixes _and_ updates the README.

kytrinyx:
Ah, I misread it. I thought the problematic bit was just the fractions.

Yes, I see how the README is problematic for test suites that do not implement the new prefixes.

IsaacG:
I misunderstood that, too :slightly_smiling_face: 

@ErikSchierboom Does it make sense to extend the tests without altering the README? Is there any mechanism for an optional addendum that can be added for those tracks which want to incorporate the additional tests?

ErikSchierboom:
@IsaacG Yes that makes sense! I don't have time to respond now, but I'll do that tomorrow!

kytrinyx:
> Is there any mechanism for an optional addendum that can be added for those tracks which want to incorporate the additional tests?

Yes, tracks can add an optional `.docs/instructions.append.md` file with additional introduction text to append after the existing instructions.

I think it would make sense here to update the README a bit. Something like:

> When we get to larger resistors, a [metric prefix][metric-prefix] is used to indicate a larger magnitude of ohms, such as "kiloohms".
That is similar to saying "2 kilometers" instead of "2000 meters", or "2 kilograms" for "2000 grams".

That's consistent with the old version, while still opening up for other prefixes.

Then we don't have to say anything specific about this exercise having different metric prefixes further down.

Maybe :)

ErikSchierboom:
I really like your suggestion @kytrinyx! With that, I don't even think we need a scenario anymore. What do you think about Katrina's proposal @IsaacG ?

IsaacG:
Hah. I dropped the ball on this one. I like that idea. I'll try to get the PR updated in the next day or two.

IsaacG:
@ErikSchierboom @kytrinyx  How's this look?

kytrinyx (approved):
This looks great to me! I'd say shipit 💙

kotp (approved):

kotp:
I agree.  I like the change in wording in addition to the addition(s) as well.

PR #2174, opened 2022-12-31
[palindrome-products ]: Added test case which test edge case

meatball133:
### Why?

When I played around with palindrome-products I noticed an edge case where you could program in a particular way that would generate incorrect answers but there were no cases testing that.

To be more specific here is the solution that passes all tests for python:

```python
def smallest(min_factor, max_factor):
    """Given a range of numbers, find the smallest palindromes which
    are products of two numbers within that range.
 
    :param min_factor: int with a default value of 0
    :param max_factor: int
    :return: tuple of (palindrome, iterable).
    Iterable should contain both factors of the palindrome in an arbitrary order.
    """
    if min_factor > max_factor:
        raise ValueError("min must be <= max")
    result = 0
    answer = []
    for number_a in range(min_factor, max_factor+1):
        for number_b in range(min_factor, max_factor+1):
            if number_a * number_b <= result or result == 0:
                multipled_number_string = str(number_a * number_b)
                if multipled_number_string == multipled_number_string[::-1]:
                    if number_a * number_b < result:
                        answer = []
                    result = int(multipled_number_string)
                    answer.append([number_a, number_b])
            if result != 0:
                break
        if result != 0:
            break
    if result == 0:
        result = None
    return (result, answer)
```

In the code, you can see that as fast as I found a "result", I can make the code end.
This test case I added will prevent that from happening.

### DIFF
--- a/exercises/palindrome-products/canonical-data.json
+++ b/exercises/palindrome-products/canonical-data.json
@@ -157,6 +157,19 @@
       "expected": {
         "error": "min must be <= max"
       }
+    },
+    {
+      "uuid": "16481711-26c4-42e0-9180-e2e4e8b29c23",
+      "description": "find the actual smallest palindrome",
+      "property": "smallest",
+      "input": {
+        "min": 3215,
+        "max": 4000
+      },
+      "expected": {
+        "value": 10988901,
+        "factors": [[3297, 3333]]
+      }
     }
   ]
 }
--- a/exercises/palindrome-products/canonical-data.json
+++ b/exercises/palindrome-products/canonical-data.json
@@ -160,7 +160,7 @@
     },
     {
       "uuid": "16481711-26c4-42e0-9180-e2e4e8b29c23",
-      "description": "find the actual smallest palindrome",
+      "description": "smallest product does not use the smallest factor",
       "property": "smallest",
       "input": {
         "min": 3215,
--- a/exercises/palindrome-products/canonical-data.json
+++ b/exercises/palindrome-products/canonical-data.json
@@ -161,6 +161,11 @@
     {
       "uuid": "16481711-26c4-42e0-9180-e2e4e8b29c23",
       "description": "smallest product does not use the smallest factor",
+      "comments": [
+        "This test catches code which assumes the smallest product is the product which uses the smallest factor.",
+        "e.g. `for x in (min_factor .. max_factor) for y in (min_factor .. max_factor) if palindrome(x, y) return [x * y, [x, y]]`",
+        "All the other tests pass when using logic like that, i.e. the other tests check for palindrome product but do not always test for the _smallest_ product"
+      ],
       "property": "smallest",
       "input": {
         "min": 3215,


IsaacG:
* Mind explaining what the bug is in that code and what is special about the test case? Is this bug likely to occur in student code?
* The description for the new test case says "actual smallest" but it's not actually the smallest palindrome product. The description is misleading.

meatball133:
The reasson why this test case matters is because you can solve it like this:

```python
def smallest(min_factor, max_factor):
    """Given a range of numbers, find the smallest palindromes which
    are products of two numbers within that range.

    :param min_factor: int with a default value of 0
    :param max_factor: int
    :return: tuple of (palindrome, iterable).
    Iterable should contain both factors of the palindrome in an arbitrary order.
    """
    if min_factor > max_factor:
        raise ValueError("min must be <= max")
    result = 0
    answer2 = []
    for x in range(min_factor, max_factor+1):
        for z in range(min_factor, max_factor+1):
            if x*z >= result:
                k = str(x*z)
                if k == k[::-1]:
                    result = int(k)
                    return (result, [[z,x]])
    return (None, [])
```

Aka finding the first palindrome and return it, without actually checking if it gives an incorrect answer.

IsaacG:
Yes. I see the code. Mind **explaining** what the bug is in that code? Can you explain *why* that code is wrong? What needs to change to make it work? What is incorrect about the value it returns?

> I made my first solution like this, and if you call me a student, so is this something that have occured to a student.

"Likely to occur" doesn't mean "possible" or "happened once". Likely to occur means this is likely to impact a "significant" number of students. Some students might just hard code every test case. We could then "break" that solution by adding one more test case. But that's not the purpose of the problem specs. The specs should have sufficient cases to ensure that most solutions solve the general problem; they should not be tailored to cover very specific bugs which are unlikely to show up in most solutions. See, too, [this discussion](https://forum.exercism.org/t/matching-brackets-tests-dont-cover-more-than-one-balanced-bracket-group/1655/4) for a similar idea.

meatball133:
The "bug" is that if what you do is to setup 2 range loops and return the first palindrome that would get through on the test. Although to find the acutall palindrome you will have to check so there is no other "smaller" palindrome.

Well I found one solution published in the last 15 days that this test would break although that person hardcoded some stuff.
And then I found another person who didn't hardcode which didn't pass my "new" test.

There has been 12 people who have published a solution for the exercise in the last 15 days on python track for context.

IsaacG:
Just to make sure I understand. The issue is: given inputs where there can be multiple palindrome products, if you generate them in a specific order and return the first palindrome product you find, then that value may not be the smallest palindrome product?

meatball133:
Yes

IsaacG:
```
    result = 0
    for x in range(min_factor, max_factor+1):
        for z in range(min_factor, max_factor+1):
            if x*z >= result:
```

Assuming `x, z` are never negative, the `if` condition will always be True and the `result` variable is essentially not used.

This code does no checking to find the smallest value. This example could be simplified to:

```python
def smallest(min_factor, max_factor):
    if min_factor > max_factor:
        raise ValueError("min must be <= max")
    for x in range(min_factor, max_factor + 1):
        for y in range(min_factor, max_factor + 1):
            k = str(x * y)
            if k == k[::-1]:
                return (x * y, [[y, x]])
    return (None, [])
```

It might be helpful to update the description to something that indicates the smallest palindrome product doesn't always have the min factor, eg "smallest product does not use the smallest factor". That can be combined that with a `comments` field to describe what problem this test solves. "actual smallest" doesn't describe the issue very well.

exercises/palindrome-products/canonical-data.json
```diff
@@ -160,7 +160,7 @@
     },
     {
       "uuid": "16481711-26c4-42e0-9180-e2e4e8b29c23",
-      "description": "find the actual smallest palindrome",
+      "description": "smallest product does not use the smallest factor",
```
IsaacG, line 163:
```suggestion
      "description": "smallest product does not use the smallest factor",
      "comments": [
        "This test catches code which assumes the smallest product is the product which uses the smallest factor.",
        "e.g. `for x in (min_factor .. max_factor) for y in (min_factor .. max_factor) if palindrome(x, y) return [x * y, [x, y]]`",
        "All the other tests pass when using logic like that, i.e. the other tests check for palindrome product but do not always test for the _smallest_ product",
      ],
```

IsaacG (approved):

ErikSchierboom (approved):

BethanyG (approved):

PR #2179, opened 2023-01-09
Image alternate text: single sentence per line

kotp:
Noticed image alternative text not being "Single Sentence per Line" for proposed addition to TCL track.

Inspiration: exercism/tcl#267

### DIFF
--- a/exercises/killer-sudoku-helper/description.md
+++ b/exercises/killer-sudoku-helper/description.md
@@ -20,7 +20,17 @@ In a 3-digit cage with a sum of 7, there is only one valid combination: 124.
 - 1 + 2 + 4 = 7
 - Any other combination that adds up to 7, e.g. 232, would violate the rule of not repeating digits within a cage.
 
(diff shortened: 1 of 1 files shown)


glennj:
@SleeplessByte Are there a11y guidelines about such length alt text? I notice the text refers to some numbers being "red".

SleeplessByte:
The red mention is extraneous because there is only one type of highlighting. In general you want alt to be as concise as possible, only mentioning relevant information.

These graphics are incredibly information dense. I think the alt text given (except for the red mention) are very good for allowing me to close my eyes and pretending I can't see or can't see well and make a mental map and model of the grid.

I would say that in this particular case it's a really good alt text.

The alternative would be to build this as a table element with the appropriate th for indices, and highlighting in aria labels, bit that's not possible with Markdown, thus this is the accessible option we have.

SleeplessByte:
A good test for yourself would be:

How would you explain what you see on the screen as quickly as possible to someone who is on the phone?

petertseng (approved):
This is purely a whitespace change; no text has been changed.

(One can argue that the word "followings" should be changed to "following", but then this would no longer purely be a whitespace change, so it should be separately)

kytrinyx:
Does anyone know if the markdown renders correctly when alt text has multiple lines? I have a vague recollection of having had trouble with something similar at some point, but I cannot for the life of me remember the details.

SleeplessByte:
@kytrinyx at least in GFM it is no issue:

long alt text, sentence per newline works in github flavored markdown

BethanyG (approved):

BNAndras:
CI appears to be hung.

IsaacG:
> CI appears to be hung.

I think it might be blocked on an owner approval.

ErikSchierboom (approved):

BNAndras (approved):

IsaacG:
I will try to close and re-open to see if it bumps the checks.

Issue #554, opened 2017-02-13
bank-account: Implement canonical-data.json


kytrinyx:
We want to have a standard set of test inputs and outputs for each exercise to
make it easier to port them to new languages, as well as to help keep the
exercises in sync.

The Bank Account exercise can be found in the [./exercises/bank-account/][exercise-dir] directory.

The step-by-step instructions for how to compile a canonical-data.json file is
described in [this section of the contributing guide][contributing].

See http://exercism.io/contribute/canonical-data/bank-account for the up-to-date list of
language tracks that have an implementation of the Bank Account exercise.

The [tracking issue][tracking-issue] has more context.

[contributing]: https://github.com/exercism/docs/blob/master/you-can-help/improve-exercise-metadata.md#extracting-canonical-test-data
[tracking-issue]: https://github.com/exercism/x-common/issues/552
[exercise-dir]: https://github.com/exercism/x-common/tree/master/exercises/bank-account

petertseng:
I have some specific thoughts about this problem.

Most problems you will see are stateless - write a function, give it some inputs, expect a certain output. All exercises listed in the example section in fact follow that pattern (custom-set is a bit more complex, but fundamentally it still fits into that pattern, just with multiple functoins)

You may find that bank-account doesn't fall into that pattern if it is very stateful - create a bank account, do some things to that bank account, check some state of that bank account.

If it looks like the bank-account tests can't fit well with the examples listed, try looking at what happened in https://github.com/exercism/x-common/pull/488 and https://github.com/exercism/x-common/pull/358 , which were also two very stateful exercises.

(Of course, if you find that bank-account can fit well into the stateless model, then you can ignore everything I've said, no worries!)

PR #2192, opened 2023-02-01
[Bank account]: Add canonical data

meatball133:
The goal with this pr is to bring so a higher percentage of the practice exercises has tests and also to make it easier for new tracks to implement the exercise. 

Some tracks have taken the decision to add a concurrency test, I was a bit unsure how to implement it in the canonical-data.json since it needs for loops for a lot of tracks. And some tracks doesn't it make sense to test for concurrency.

Any way feedback appreciated

resolves: #554


exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,278 @@
...
+    "For this exercise it is possible to implement concurrency tests",
+    "but for language tracks that do not support concurrency, the tests",
+    "are written to be run in a single thread.",
+    "But for language tracks that do support concurrency, so is it possible",
+    "to write tests that run concurrently."
+  ],
+  "cases": [
+    {
+      "uuid": "983a1528-4ceb-45e5-8257-8ce01aceb5ed",
+      "description": "Using pop raises an error if the list is empty",
```
petertseng, line 13: I don't recognise pop as an operation that https://github.com/exercism/problem-specifications/blob/main/exercises/bank-account/description.md should be made on a bank account, nor do I recall the description mentioning lists.

All descriptions in this file similarly reference terms that are unfamiliar to me when thinking about bank-account

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,278 @@
+{
+  "exercise": "bank-account",
+  "comments": [
+    "For this exercise it is possible to implement concurrency tests",
+    "but for language tracks that do not support concurrency, the tests",
+    "are written to be run in a single thread.",
+    "But for language tracks that do support concurrency, so is it possible",
```
petertseng, line 7: "is it possible" seems to imply a question, but the sentence here ends with a period instead of a question mark. Was it meant to be a question?

meatball133: Are we talking about the sentence on line 7-8?

meatball133: Since line 4 has a pretty similar wording

meatball133: I think I improved it, can you check again?

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,278 @@
...
+    "to write tests that run concurrently."
+  ],
+  "cases": [
+    {
+      "uuid": "983a1528-4ceb-45e5-8257-8ce01aceb5ed",
+      "description": "Using pop raises an error if the list is empty",
+      "property": "bankAccont",
```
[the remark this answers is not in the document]
meatball133: Thanks ✨

exercises/bank-account/canonical-data.json
meatball133: Thanks ✨

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,303 @@
...
+          },
+          {
+            "operation": "close"
+          },
+          {
+            "operation": "amount"
+          }
+        ]
+      },
+      "expected": { "error": "account not open" }
```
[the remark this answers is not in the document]
petertseng: It might be a good idea to make this convention clear though - the comments would say something like "if an error is expected, the last operation is the one that caused it; all operations other than the last one should succeed"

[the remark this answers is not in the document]
meatball133: I will see if other reviewers feels like there needs to be a structure change if not I will add that as a comment that it is the last one that should always be the action which will raise error

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,303 @@
+{
+  "exercise": "bank-account",
+  "comments": [
+    "For this exercise it is possible to implement concurrency tests",
+    "but for language tracks that do not support concurrency, the tests",
+    "are written to be run in a single thread.",
+    "But for language tracks that do support concurrency, it is possible",
```
SleeplessByte, line 7: This `But` doesn't work because the previous case also has a `but`. I suggest we change the entire thing to:

```
This exercise is a good candidate to practice concurrency, which
may not be available or possible in the implementing language track.
In this case, all tests can be executed on a single thread. If there
are tests that don't work using a single thread, these should be
marked with a concurrency scenario.
```

We probably also want to add `concurrency` to https://github.com/exercism/problem-specifications/blob/main/SCENARIOS.txt.

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,303 @@
...
+            "operation": "withdraw",
+            "value": 80
+          },
+          {
+            "operation": "withdraw",
+            "value": 20
+          },
+          {
+            "operation": "amount",
+            "expected": 0
```
SleeplessByte, line 122: If we're defining `operations` to have an operand / function call name, and optional argument of `value`, and an optional return expectation of `expected`, I think we should move `error` also into an operation expectation.

This also solves the potential ordering issue @petertseng mentions, and makes it easier to add interesting concurrency tests down the line.

SleeplessByte (commented):
My 2 cents

exercises/bank-account/canonical-data.json
meatball133: Thanks for your feedback, do you think writing it like this makes sense?

```
          {
            "operation": "deposit",
            "value": -50,
            "error": "amount must be greater than 0"
          }
```

SleeplessByte: Perhaps still nest it in `expected` like before? @petertseng what do you think?

petertseng: Is there a reason to prefer one over the other? Since I'm not currently aware of one, it seems we are empowered to make either choice. If a reason presents itself, then we would go with that one. For example, someone might say "all other `error` are currently under `expected`, and it is important for this to remain true because (they give a reason here)". You will notice that since I didn't have something to insert in the blank that I didn't put forth that reason.

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,327 @@
...
+            "value": 100
+          },
+          {
+            "operation": "withdraw",
+            "value": 50
+          },
+          {
+            "operation": "amount",
+            "expected": 50
+          }
```
glennj, line 94: Let's make the withdraw amount different than 50, so people don't get confused about how to calculate the remaining balance
```suggestion
            "operation": "withdraw",
            "value": 75
          },
          {
            "operation": "amount",
            "expected": 25
          }
```

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,327 @@
...
+              "error": "account not open"
+            }
+          }
+        ]
+      },
+      "expected": {}
+    },
+    {
+      "uuid": "c396d233-1c49-4272-98dc-7f502dbb9470",
+      "description": "Close already closed account",
```
glennj, line 198:
```suggestion
      "description": "Close an account that was not opened",
```

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,327 @@
...
+          {
+            "operation": "deposit",
+            "value": 50,
+            "expected": {
+              "error": "account not open"
+            }
+          }
+        ]
+      },
+      "expected": {}
```
glennj, line 171: How about an additional test here:
```
    {
      "uuid": "1234",
      "description": "Deposit into unopened account",
      "property": "bankAccount",
      "input": {
        "operations": [
          {
            "operation": "deposit",
            "value": 50,
            "expected": {
              "error": "account not open"
            }
          }
        ]
      },
      "expected": {}
    }
```

glennj (changes requested):

exercises/bank-account/canonical-data.json
SleeplessByte: Since no one has written any test generator for this format yet, I agree, there is no reason from that point of view right now to pick one or the other.

@meatball133 I'll leave it up to you unless someone else comes up with a good reason.

glennj: I find the empty `expected` sibling to `input` quite confusing.

What if we have some sort of flag that determines which operation produces the expected result?

```json
    {
      "uuid": "983a1528-4ceb-45e5-8257-8ce01aceb5ed",
      "description": "Newly opened account has zero balance",
      "property": "bankAccount",
      "input": {
        "operations": [
          {
            "operation": "open"
          },
          {
            "operation": "amount",
            "result": true
          }
        ]
      },
      "expected": 0
    },
...
    {
      "uuid": "f9facfaa-d824-486e-8381-48832c4bbffd",
      "description": "Checking balance of closed account throws error",
      "property": "bankAccount",
      "input": {
        "operations": [
          {
            "operation": "open"
          },
          {
            "operation": "close"
          },
          {
            "operation": "amount",
            "result": true
          }
        ]
      },
      "expected": {
        "error": "account not open"
      }
    }
```

glennj: Or, just remove the empty outer "expected"

SleeplessByte: I think the outer one should go away indeed.

meatball133: I think the design you suggested Gleen could just make stuff very complicated for generators.
Say you would want to have 2 expect?, then it would be quite hard to implement that with such a system.

And removing expect seems to go against the design scheme.

glennj: Ah, right. I'm now thinking that the input should be an array of operations without any "result" property, and the expected result is the result of the last operation.

Peter's previous comment "If there were ever a case in the future where an operation other than the last one should cause the error": if I were writing unit tests, I wouldn't be performing any more operations following an operation that produces an expected error (unless we're testing the behaviour of the system following an error, in which case we'd explicitly catch the error)

I propose we keep it simple: we'll just expect the results of the last operation.

Would this be obvious to people implementing the test suite for a track? Would we need a comment explicitly stating it?

SleeplessByte: I would always add a comment for a future reader.

petertseng: > unless we're testing the behaviour of the system following an error

This type of test could be quite valuable if one of the desired teaching points of this exercise is to ensure that the system remains in a usable state after an erroneous operation. Ideally that decision wouldn't need to be made now, but I do regret to say that since it might have an effect on how some future tests might need to be written, it seems it's necessary to at least plan for it. If there is no intent at all to have this as a teaching point, then all is well and we can end the discussion here, and reading the rest of this comment is unnecessary.

But if error handling is one of the desired teaching points of this exercise, then just as an example, consider a test that opens an account, deposits a negative amount (which is not allowed), and then checks that the balance is still zero following that erroneous operation:

1. Operations other than the last one could have an `expected`, with the outer `expected` still referring to the very last operation:

```json
    {
      "uuid": "785f57e4-7c32-402e-a118-e3bfd4facf75",
      "description": "Bad deposit",
      "property": "bankAccount",
      "input": {
        "operations": [
          {
            "operation": "open"
          },
          {
            "operation": "deposit",
            "value": -50,
            "expected": {"error": "amount must be greater than 0"}
          },
          {
            "operation": "amount"
          }
        ]
      },
      "expected": 0
    }
```

2. Or, each operation could have its own `expected` and the outer one remains empty.


```json
    {
      "uuid": "785f57e4-7c32-402e-a118-e3bfd4facf75",
      "description": "Bad deposit",
      "property": "bankAccount",
      "input": {
        "operations": [
          {
            "operation": "open"
          },
          {
            "operation": "deposit",
            "value": -50,
            "expected": {"error": "amount must be greater than 0"}
          },
          {
            "operation": "amount"
            "expected": 0
          }
        ]
      },
      "expected": {}
    }
```

I have no reason to prefer one or the other, but if someone does, it would be good to mention the reason for that preference right now.

I also wouldn't consider having such tests a necessary condition to have this pull request be Approved (in the GitHub sense of the word), but I'll phrase it as: It wouldn't be surprising to see an additional pull request from me some time after this pull request is merged if this is indeed a desired teaching point.

glennj: I hear what you're saying about the teaching point.

Of these options, I prefer #1. That indicates that we want to produce *that result* from the *entire sequence* of operations. Contrast to #2 where the actual expected result of the test case is ambiguous.

I think we'd want to give some hint to track implementors that in this situation, the erroring operation has in a non-fatal error or that the error is to be trapped. Perhaps adding a (new) scenario to the case or adding a `"fatal": false` property to the `expected` object for that operation.

meatball133: Sorry for taking a while, I will replay later today. Promise, aslong as I dont get a power outage but that usually doesnt happen.

meatball133: Okay I have read through the feedback. I think we can go with method one. But there are no current test cases which has any other operations in the middle so the fatal flag doesn't need to be applied anywhere. Unless you guys come up with some new tests. But I can convert the tests to suggestion 1. I can also write a comment about how we think about this design.

glennj:
I'm satisfied with the current canonical data. @petertseng @SleeplessByte where do you stand now?

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,317 @@
+{
+  "exercise": "bank-account",
+  "comments": [
+    "This exercise is a good candidate to practice concurrency, which",
+    "may not be available or possible in the implementing language track.",
+    "In this case, all tests can be executed on a single thread. If there",
+    "are tests that don't work using a single thread, these should be",
+    "marked with a concurrency scenario.",
```
petertseng, line 8: I would prefer that the two sentences not be combined into one line, but not so much that it stops me from Approving (in the GitHub sense of that word)

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,317 @@
...
+    "----------------------------------------------------------------",
+    "For tests which requires multiple expected values/errors",
+    "Should put the expected values/errors at each operation",
```
petertseng, line 11:
```suggestion
    "Should put the expected values/errors as a field of each operation",
```

"at" didn't seem clear to me - is this suggestion better?

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,317 @@
...
+    "And then the last expected value/error should be the case expected",
```
petertseng, line 12:
```suggestion
    "And then the last expected value/error should be the overall expected value for the case",
```

"the case expected" seemed to be difficult to understand. How about this?

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,317 @@
...
+    "For errors that are suppose to raise non fatal errors",
+    "Should have an extra field called `fatal` which holds a boolean"
```
petertseng, line 14: I don't currently see a use for this - these two lines could be removed.

Maybe you could help understand why these two lines are here, by answering this question:
What should the tests do differently if an error is fatal, versus if it's not?

petertseng (approved):

exercises/bank-account/canonical-data.json
petertseng: (by that, I mean that I would prefer it if the sentence that starts with "If there" would be on a different line than the sentence that ends with "single thread")

exercises/bank-account/canonical-data.json
meatball133: An error that is not fatal should make the program being able to continued to be runned, I guess

petertseng: Ah, I was thinking single-threadedly.

Here was why I didn't understand at first:
In a single thread, of course any error in an operation other than the very last one **has** to be non-fatal, since the test has to keep running the rest of the operations. Therefore the `fatal` field seemed useless. It always has to be `false`.

Here's why it was a misconception:
But if there are multiple threads, you might need to say some errors are fatal (stops that thread) or not. therefore the field is useful.

Okay, I think you helped me gain an understanding. Thank you.

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,317 @@
...
+  "comments": [
+    "This exercise is a good candidate to practice concurrency, which",
+    "may not be available or possible in the implementing language track.",
+    "In this case, all tests can be executed on a single thread. If there",
+    "are tests that don't work using a single thread, these should be",
+    "marked with a concurrency scenario.",
+    "----------------------------------------------------------------",
+    "For tests which requires multiple expected values/errors",
+    "Should put the expected values/errors as a field of each operation",
+    "And then the last expected value/error should be the overall expected value for the case",
```
petertseng, line 12: I would like to suggest that the overall `expected` value for the case be mentioned first, since it is more important.
It's more important because it's always applicable, regardless of whether any operations also include an `expected` value. Therefore:

```suggestion
    "The overall expected value for the case should be the result of the last operation.",
    "If any operation other than the last one has an expected result,",
    "the expected value/error should be a field of that operation",
```

petertseng:
Please amend the commit message to include the string "`closes https://github.com/exercism/problem-specifications/issues/554`" or any equivalent string. Thank you.

meatball133:
I added so this pr will close that issue

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,317 @@
...
+            "operation": "withdraw",
+            "value": 50
+          }
+        ]
+      },
+      "expected": { "error": "account not open" }
+    },
+    {
+      "uuid": "c396d233-1c49-4272-98dc-7f502dbb9470",
+      "description": "Close an account that was not opened",
```
ErikSchierboom, line 204:
```suggestion
      "description": "Cannot close an account that was not opened",
```

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,317 @@
...
+          {
+            "operation": "close"
+          }
+        ]
+      },
+      "expected": { "error": "account not open" }
+    },
+    {
+      "uuid": "c06f534f-bdc2-4a02-a388-1063400684de",
+      "description": "Open already opened account",
```
ErikSchierboom, line 217:
```suggestion
      "description": "Cannot open an already opened account",
```

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,317 @@
...
+          {
+            "operation": "amount"
+          }
+        ]
+      },
+      "expected": 0
+    },
+    {
+      "uuid": "ec42245f-9361-4341-8231-a22e8d19c52f",
+      "description": "Cannot withdraw more than deposited",
```
ErikSchierboom, line 259: Nice

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,317 @@
...
+          {
+            "operation": "open"
+          }
+        ]
+      },
+      "expected": { "error": "account already open" }
+    },
+    {
+      "uuid": "0722d404-6116-4f92-ba3b-da7f88f1669c",
+      "description": "Reopened account does not retain balance",
```
ErikSchierboom, line 233: This is a new test case AFAICT, correct? This test case and a0a1835d-faae-4ad4-a6f3-1fcc2121380b are the two test cases of which I'm not sure about. Most existing implementations of this exercise don't have these check I think.

ErikSchierboom (commented):
Thanks for working on this!

exercises/bank-account/canonical-data.json
meatball133: Python has both of those tests, I can't speak although for how other tracks have done it.

exercises/bank-account/canonical-data.json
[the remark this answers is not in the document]
meatball133: We don't have a concurrency test among the tests, mostly because I am a little bit unsure how you would write one since most of the time that includes tricking with loops and I don't know how different language deals with that. But perhaps we should.

exercises/bank-account/canonical-data.json
ErikSchierboom: Ah, that's one implementation that I didn't do. Let's keep them then.

exercises/bank-account/canonical-data.json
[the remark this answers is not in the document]
ErikSchierboom: We should definitely have a concurrency test I feel! It's what this exercise is mostly about. How a language implements it is not of any real concern to this canonical data, so we don't really need to consider that. 

BTW: this is how C# implements it https://github.com/exercism/csharp/blob/main/exercises/practice/bank-account/BankAccountTests.cs#L61-L84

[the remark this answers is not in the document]
meatball133: Yeah. I left it out since i was unsure how to add it, and also there are languages that doesn't really have concurrency. But if you have a suggestion on how a concurrency tests data should look like so am I up for suggestions.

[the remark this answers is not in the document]
ErikSchierboom: @meatball133 What do you think about the above suggestion? 

p.s. We have a `concurrent` scenario now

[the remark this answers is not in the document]
meatball133: I think that could work.

meatball133:
@ErikSchierboom did you intend it like this?

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,372 @@
...
+    "may not be available or possible in the implementing language track.",
+    "In that case, all tests can be executed on a single thread. If there",
+    "are tests that don't work using a single thread, these should be",
+    "marked with a concurrency scenario.",
+    "----------------------------------------------------------------",
+    "The overall expected value for the case should be the result of the last operation.",
+    "If any operation other than the last one also has an expected result,",
+    "that individual operation should have its expected value/error as a field.",
+    "For errors that are suppose to raise non fatal errors",
+    "Should have an extra field called `fatal` which holds a boolean"
```
ErikSchierboom, line 14: Is there any value in keeping this comment? I feel like this is more of a discussion than something a track maintainer should actually know about.

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,372 @@
+{
+  "exercise": "bank-account",
+  "comments": [
+    "This exercise is a good candidate to practice concurrency, which",
+    "may not be available or possible in the implementing language track.",
+    "In that case, all tests can be executed on a single thread. If there",
+    "are tests that don't work using a single thread, these should be",
+    "marked with a concurrency scenario.",
```
ErikSchierboom, line 8: I think this comment should, in a slightly different form, be moved to the "concurrent" test case.

exercises/bank-account/canonical-data.json
```diff
@@ -3,15 +3,7 @@
...
-    "In that case, all tests can be executed on a single thread. If there",
-    "are tests that don't work using a single thread, these should be",
-    "marked with a concurrency scenario.",
-    "----------------------------------------------------------------",
-    "The overall expected value for the case should be the result of the last operation.",
-    "If any operation other than the last one also has an expected result,",
-    "that individual operation should have its expected value/error as a field.",
-    "For errors that are suppose to raise non fatal errors",
-    "Should have an extra field called `fatal` which holds a boolean"
+    "It is possible to implement this exercise without concurrency,"
```
ErikSchierboom, line 6: Tiny nit:

```suggestion
    "It is possible to implement this exercise without concurrency."
```

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,370 @@
...
+      "description": "Multiple deposits",
+      "property": "bankAccount",
+      "input": {
+        "operations": [
+          {
+            "operation": "open"
+          },
+          {
+            "operation": "deposit",
+            "value": 100
```
ErikSchierboom, line 56: Maybe we can now use `amount` here as we've dropped it as the operation? This would also apply to the `deposit` actions.

ErikSchierboom (commented):
Just two nits, and then we should be good to go!

exercises/bank-account/canonical-data.json
meatball133: Did I do as you intended in the commit?

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,370 @@
...
+            "operation": "deposit",
+            "amount": 1
+          },
+          {
+            "operation": "withdraw",
+            "amount": 1
+          }
+        ]
+      },
+      "expected": {}
```
ErikSchierboom, line 367: Just noticed, we forgot to add a (non-concurrent) `balance` operation at the  end to verify the state.

ErikSchierboom (commented):
Found just one tiny nit (sorry)

exercises/bank-account/canonical-data.json
meatball133: I separated those operations into 2 groups, is that correct?

ErikSchierboom: Yes! That's perfect

exercises/bank-account/canonical-data.json
ErikSchierboom: You did!

ErikSchierboom (approved):
LGTM!

SleeplessByte (approved):
Let's gooooo

SleeplessByte:
@glennj please have a final look :)

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,375 @@
...
+          {
+            "operation": "withdraw",
+            "amount": 1
+          }
+        ],
+        "operations": [
+          {
+            "operation": "balance"
+          }
+        ]
```
petertseng, line 370: where is the `open` operation for this account?

SleeplessByte: cc @meatball133

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,375 @@
...
+          },
+          {
+            "operation": "withdraw",
+            "amount": 1
+          }
+        ],
+        "operations": [
+          {
+            "operation": "balance"
+          }
```
glennj, line 369: Let's give the current operations some number (number of threads/fibers/etc)
```suggestion
      "input": {
        "initial_operations": [
          {
            "operation": "open"
          }
        ],
        "concurrent_operations": {
          "number": 1000,
          "operations": [
            {
              "operation": "deposit",
              "amount": 1
            },
            {
              "operation": "withdraw",
              "amount": 1
            }
          ]
        },
        "finishing_operations": [
          {
            "operation": "balance"
          }
```

glennj (changes requested):

glennj (approved):

exercises/bank-account/canonical-data.json
glennj: @petertseng can you have a look here please

exercises/bank-account/canonical-data.json
```diff
@@ -0,0 +1,383 @@
...
+      "comments": [
+        "This test is a good candidate to practice concurrency, which",
+        "may not be available or possible in the implementing language track.",
+        "Concurrency testing is often made by running the same operations in",
+        "parallel, and then checking that the final state is as expected."
+      ],
+      "scenarios": ["concurrent"],
+      "property": "bankAccount",
+      "input": {
+        "initial_operations": [
```
petertseng, line 356: if the consensus is that this is the best way to do things then I won't stand in your way. It strikes me as:

* inflexible (since now you can only have concurrent operations in between an initial and final section, instead of arbitrary interleavings)
* violating the schema (`property` now may have `input` of different types: it might either have `operations`, or not) 

What I expected to see was:

```json
{
  "input": {
    "operations": [
      { 
        "operation": "open"
      },
      { 
        "operation": "concurrent",
        "number": 1000,
        "operations": [
          { 
            "operation": "deposit",
            "amount": 1
          },
          { 
            "operation": "withdraw",
            "amount": 1
          } 
        ] 
      },
      { 
        "operation": "balance"
      },
    ]
  },
  "expected": 0
}
```

SleeplessByte: I prefer this too. This also allows us to have multiple _batches_ of concurrent operations instead of always _serial start_ _concurrent_ _serial end_.

meatball133: @glennj and @ErikSchierboom, I think we should have a solution we all agree upon so I want to hear your thoughts

glennj: I'm on board with this suggestion.

ErikSchierboom: Me too. That sounds like a better approach

PR #2199, opened 2023-02-07
Matrix: Ensure whole matrix is correct

homersimpsons:
There are some proposed solutions where the mentee submit solution that would emit extra empty rows. For instance with an off by one error in a for-loop.

By checking that the whole matrix is exactly as expected we can catch such mistakes.

(Re-open of #2195, cc @SleeplessByte)

### DIFF
--- a/exercises/matrix/canonical-data.json
+++ b/exercises/matrix/canonical-data.json
@@ -80,6 +80,134 @@
         "index": 2
       },
       "expected": [1903, 3, 4]
+    },
+    {
+      "uuid": "f5288dbd-2141-453d-9c38-a351d68ed4d9",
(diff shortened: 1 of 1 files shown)


SleeplessByte:
@exercism/reviewers please :)

kotp:
Link back to Forum discussions for continuity:

https://forum.exercism.org/t/pull-requrest-on-exercism-javascript/3440/8

glennj:
The new tests take different inputs, so it seems like they should have different property names.

homersimpsons:
> The new tests take different inputs, so it seems like they should have different property names.

Good catch, should I just remove the `"property"` field completely ? Or should I use a value like `"matrix"` ?

PR #2203, opened 2023-02-17
ci: support instructions.md file

ErikSchierboom:
CI didn't yet account for the fact that an exercise could have an `instructions.md` file instead of a `description.md` file.
Better to fix this now than be bitten by it later :)


iHiD (approved):
Approving from a maintainers-admin perspective, not a code quality one :)

kytrinyx (approved):

PR #2205, opened 2023-02-20
Rework bob documentation

kytrinyx:

We will merge this PR in roughly 72 hours, and then submit PRs to all tracks that have this exercise.

The main change for this exercise is that we've slightly tweaked the introduction to feel more friendly, and structured the instructions to help clarify exactly what responses are in Bob's repertoire.


iHiD (approved):

BethanyG (approved):

ErikSchierboom (approved):

exercises/bob/instructions.md
```diff
@@ -0,0 +1,18 @@
...
+Bob only ever answers one of five things:
+
+- **"Sure."**
+  This is his response if you ask him a question, such as "How are you?"
+- **"Whoa, chill out!"**
+  This is his answer if you YELL AT HIM (in all capitals).
+- **"Calm down, I know what I'm doing!"**
+  This is what he says if you yell a question at him.
+- **"Fine. Be that way!"**
+  This is how he responds if you address him without actually saying anything.
```
vaeng, line 14: As a student this was a frustrating part of trial-and-error. What is the difference between "anything else" and "without actually saying anything"? I had to look into the test cases to get an answer here, and that might not be obvious to new students.

kytrinyx: @vaeng Do you think we should say more about what "not saying anything" means here?

kytrinyx: Hm, rereading your comment, it sounds like probably we need to say more.

Without saying anything means that it’s silence.

In a novel, it would be a conversation like this.
“You did it, didn’t you?”
“…”
“Answer me, when I speak to you!”

In the examples it would be silence. In all the tests, it's represented by all spaces and tabs, I think.

Anything else would be anything that doesn't explicitly fall into the first four categories. So a normal sentence, or an emphatic sentence, or quoting some famous person or whatever.

Is this useful context?

iHiD: How about:
```suggestion
  This is how he responds if you respond with silence.
```

exercises/bob/instructions.md
```diff
@@ -0,0 +1,18 @@
...
+- **"Whatever."**
+  This is what he answers to anything else.
+
+For the purpose of this exercise, you can assume that Bob's conversational partner always follows normal rules regarding sentence punctuation in English.
```
vaeng, line 18: This suggests that input is valid grammar and punctuation. But there are test cases where you have to trim beginning and endings. I think that is okay, but the exercises is not explicit about that.

exercises/bob/instructions.md
vaeng: Now I understand how it was meant, but back then I was really unhappy with the exercise. I have no real suggestion how to put the "white space check" into a non-programmer sentence though.  Elixir and C++ also check for carriage returns and new-lines, which might be confusing if you don't know about them.

It has been a while, but I remember the instructions to be confusing. The intended test-case seems to be "white space" and I think I tested cases like "...", "?" and other not-alphanumeric cases. Look at the [cpp](https://github.com/exercism/cpp/blob/eec57a6b68a39f91a7c2a6f9db55d8a973eb418e/exercises/practice/bob/bob_test.cpp#L86) version for example. That is just a smile and a question mark. That is - in my book - like not saying anything, but would count as a question.

exercises/bob/instructions.md
kytrinyx: Oh, right. Yes. /thinking

exercises/bob/instructions.md
kytrinyx: I don't love _respond with silence_ because it suggests that Bob said something first, which isn't necessarily the case. Maybe a kid at school walks up to him and grabs his shirt and pushes him against the wall and stares him down, without saying anything.

Let me think about this for a bit.

kytrinyx:
@vaeng @iHiD would you two take a look at the changes I made to the instructions and see if this is useful?

vaeng (approved):
Yes, I think that would be a clearer

kytrinyx:
Alright, it seems like we're good on this one. I'm going to go ahead and merge it, and if anyone has any further concerns, please open a PR to suggest changes.

PR #2211, opened 2023-02-28
check_required_files: Only allow legacy or new formats

petertseng:
After an initial implementation in https://github.com/exercism/problem-specifications/pull/2203, ensuing discussion in https://github.com/exercism/problem-specifications/pull/2211 tells us that in fact we only want to allow two possibilities:
    
1. legacy format: Only description.md exists
2. new format: Only instructions.md and introduction.md exist.
    
Previously, no check for introduction was present, so one has to be added.
    
If a an exercise has files that cause it to be a mix of the two formats, that shouldn't be allowed either.

### DIFF
--- a/bin/check_required_files_present
+++ b/bin/check_required_files_present
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 MISSING_FILES=0
+UNNEEDED_FILES=0
 SHOW_PROGRESS='false'
 
 if [ ! -d exercises ] ; then
@@ -21,12 +22,23 @@ function require_file {
 	fi
 }
 
+function reject_file {
+	filename=$1
+	if [ -f "$filename" ]; then
+		echo "unneeded file should be removed: $filename" >&2
+		UNNEEDED_FILES=$((UNNEEDED_FILES + 1))
+	fi
+}
+
 function require_file_or_alternative {
 	filename=$1	
 	alternative_filename=$2
 	if [ ! -f $filename ] && [ ! -f $alternative_filename ]; then		
 		echo "required file missing: $filename or $alternative_filename" >&2
 		let "MISSING_FILES+=1"
+	elif [ -f "$filename" ] && [ -f "$alternative_filename" ]; then
+		echo "one of these two files should be removed: $filename or $alternative_filename" >&2
+		UNNEEDED_FILES=$((UNNEEDED_FILES + 1))
 	fi
 }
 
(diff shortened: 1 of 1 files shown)


petertseng:
Eight possible states (note the two rows that changed as a result of this commit):

description.md exists | instructions.md exists | introduction.md exists | does CI allow?
---|---|---|---
No | No | No | No
No | No | Yes | No
No | Yes | No | No (changed)
No | Yes | Yes | Yes
Yes | No | No | Yes
Yes | No | Yes | Yes
Yes | Yes | No | No (changed)
Yes | Yes | Yes | Yes

If any row does not match its desired behaviour, it would be good to point out which rows don't match, and maybe pointing it out leads to figuring out a suitable solution.

kytrinyx:
I think there are ideally two rows that would be allowed:

Yes-no-no (legacy: only description.md exists)
No-yes-yes (migrated to new format: description is gone, and has been replaced with instructions and introduction)

I would prefer if we don't allow a combination of legacy and new files.

petertseng:
Very well, in that case the code has been updated. To implement the above goal, it was necessary to add a new capability to say that a file that is present should be removed.

ErikSchierboom (approved):
Nice work!

kytrinyx (approved):
Thanks so much, Peter!

bin/check_required_files_present
```diff
@@ -21,20 +22,45 @@ function require_file {
 	fi
 }
 
+function reject_file {
+	filename=$1
+	if [ -f "$filename" ]; then
+		echo "unneeded file should be removed: $filename" >&2
+		let "UNNEEDED_FILES+=1"
```
IsaacG, line 29:
```suggestion
		UNNEEDED_FILES=$(( UNNEEDED_FILES + 1))
```

`let` is very deprecated. `$((` is POSIX sh.

bin/check_required_files_present
```diff
@@ -21,20 +22,45 @@ function require_file {
...
+	fi
+}
+
 function require_file_or_alternative {
 	filename=$1	
 	alternative_filename=$2
 	if [ ! -f $filename ] && [ ! -f $alternative_filename ]; then		
 		echo "required file missing: $filename or $alternate_filename" >&2
 		let "MISSING_FILES+=1"
```
IsaacG, line 38:
```suggestion
		MISSING_FILES=$(( MISSING_FILES +1 ))
```

bin/check_required_files_present
```diff
@@ -21,20 +22,45 @@ function require_file {
...
+	elif [ -f "$filename" ] && [ -f "$alternative_filename" ]; then
+		echo "one of these two files should be removed: $filename or $alternative_filename" >&2
+		let "UNNEEDED_FILES+=1"
```
IsaacG, line 41:
```suggestion
		UNNEEDED_FILES=$(( UNNEEDED_FILES +1 ))
```

bin/check_required_files_present
petertseng: I am unable (as a matter of commit cleanliness) to include this change in this pull request, as it changes code that was not added in this pull request. This change will need to go in its own pull request.

IsaacG (approved):

IsaacG:
I have a pending follow up to tidy up the shell scripts so I won't nitpick those here :)

PR #2212, opened 2023-02-28
check_required_files: Correctly display the name of the missing alt file

petertseng:
there is no such variable alternate_filename. It needs to be alternative_filename instead.


ErikSchierboom (approved):
🤦 Thanks for catching this!

jiegillet (approved):

kytrinyx (approved):

PR #2213, opened 2023-03-01
Rational Numbers: Add error test case to expreal property

natanaelsirqueira:
As suggested by @jiegillet in [gleam!248](https://github.com/exercism/gleam/pull/248#discussion_r1121615055), we would like to add a failure edge case when doing the exponentiation of a real number to a rational number.

### DIFF
--- a/exercises/rational-numbers/canonical-data.json
+++ b/exercises/rational-numbers/canonical-data.json
@@ -400,6 +400,28 @@
             "r": [0, 1]
           },
           "expected": 1.0
+        },
+        {
+          "uuid": "da137f1c-fc94-4035-9813-94f5152a1102",
+          "description": "Raise a negative real number to a rational number with an even numerator",
+          "property": "expreal",
+          "input": {
+            "x": -1,
+            "r": [2, 2]
+          },
+          "expected": 1.0
+        },
+        {
+          "uuid": "f3734f78-91b2-4e59-905e-8c61dbf34f39",
+          "description": "Fails with a negative base if the numerator of the exponent is odd",
+          "property": "expreal",
+          "input": {
+            "x": -1,
+            "r": [1, 2]
+          },
+          "expected": {
+            "error": "impossible operation"
+          }
         }
       ]
     },
--- a/exercises/rational-numbers/canonical-data.json
+++ b/exercises/rational-numbers/canonical-data.json
@@ -413,14 +413,14 @@
         },
         {
           "uuid": "f3734f78-91b2-4e59-905e-8c61dbf34f39",
-          "description": "Fails with a negative base if the numerator of the exponent is odd",
+          "description": "Fails with a negative real number if the numerator of the rational number is odd",
           "property": "expreal",
           "input": {
             "x": -1,
             "r": [1, 2]
           },
           "expected": {
-            "error": "impossible operation"
+            "error": "impossible to raise a negative base to an exponent with an odd numerator"
           }
         }
       ]


exercises/rational-numbers/canonical-data.json
```diff
@@ -400,6 +400,28 @@
...
+        },
+        {
+          "uuid": "da137f1c-fc94-4035-9813-94f5152a1102",
+          "description": "Raise a negative real number to a rational number with an even numerator",
+          "property": "expreal",
+          "input": {
+            "x": -1,
+            "r": [2, 2]
+          },
+          "expected": 1.0
```
jiegillet, line 412: I have some issues with this test
- the expected value should be `-1.0`
- the rational number is not reduced, and when it's reduced to `[1, 1]`, the numerator is odd

exercises/rational-numbers/canonical-data.json
```diff
@@ -400,6 +400,28 @@
...
+        },
+        {
+          "uuid": "f3734f78-91b2-4e59-905e-8c61dbf34f39",
+          "description": "Fails with a negative real number if the numerator of the rational number is odd",
```
jiegillet, line 416: This implies that an odd numerator is the cause of the error, but it's not the case, the case you introduce on top is equivalent to `"x": -1, "r": [1, 1]` and that is fine. Also `"x": -1, "r": [1, 3]` is `-1`.

I think the criterion is that all powers of a negative number will be complex if the **denominator** of the **reduced** fraction is **even**. Like `"r": [1, 2]`, `"r": [5, 6]`...

natanaelsirqueira: @jiegillet I revised the description here and the input number for both test cases (I assumed we want to expect the solution to reduce the number before doing the calculation).

Sorry for all my confusion with the concept here, I should've asked for your assistance in defining the test cases. It also didn't help that I missed reducing the number in my gleam implementation. :(

jiegillet: An implementation may or may not reducing the input, but in a test we want to test only one thing at a time and bbe crystal clear about it, so I suggest that the input should already be reduced. 

You could add another test that specifically checks that giving an un-reduced number also works.

jiegillet (commented):
(sorry I meant to send this days ago, but I forgot to click send, so it stayed pending)

exercises/rational-numbers/canonical-data.json
```diff
@@ -400,6 +400,38 @@
...
           },
           "expected": 1.0
+        },
+        {
+          "uuid": "da137f1c-fc94-4035-9813-94f5152a1102",
+          "description": "Raise a negative real number to a rational number that has been reduced",
+          "property": "expreal",
+          "input": {
+            "x": -1,
+            "r": [1, 1]
```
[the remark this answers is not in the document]
natanaelsirqueira: Hey! I tried using that example in my Gleam implementation but stumbled on [this](https://github.com/gleam-lang/stdlib/blob/v0.27.0/src/gleam/float.gleam#L346). This behaviour is the same in Elixir/Erlang. Any thoughts?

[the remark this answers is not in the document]
natanaelsirqueira: Yeah... I'm up for dropping all of this tbh. It's starting to feel like we are overcomplicating the exercise and not adding much value. Plus, the initial motivation was to add an error case to our Gleam implementation because the function returns a result. Maybe we could just make that same check that `float.power/2` does and have that be our error case?

lpil:
Is this good to go? :)

natanaelsirqueira:
@lpil Thank you for the reminder. In the interest of time, I believe it's reasonable to back out of this change to unblock the PR in the Gleam track. In hindsight, adding an error case to the problem spec solely because the Gleam function returns a Result may not be necessary. I'm open to being challenged on this, but if you agree with this approach, please feel free to leave your review/approval on that PR, and I'll proceed with closing this one.

lpil:
Thank you. What do you think @jiegillet ?

PR #2221, opened 2023-03-05
Fix formatting in two-fer instructions

kytrinyx:
I noticed that CI was unhappy in the javascript track when we synced this exercise.

This reformats the table so that next time we sync we don't undo the fix.

This is not a high-priority fix, so I won't be syncing it to all the tracks; it can come in as part of the next sync that brings in little fixes here and there.

### DIFF
--- a/exercises/two-fer/instructions.md
+++ b/exercises/two-fer/instructions.md
@@ -17,9 +17,9 @@ One for you, one for me.
 
 Here are some examples:
 
-|Name    |Dialogue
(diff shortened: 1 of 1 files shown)


SleeplessByte (approved):
Nice. We run prettier on everything so there is no contention about formatting.

meatball133 (approved):

andrerfcsantos (approved):

ErikSchierboom (approved):

PR #2223, opened 2023-03-07
Revise binary-search instructions

cmcaine:
This commit changes a potentially confusing reference to "number" and "numbers" to the generic "element" and "elements" to match the language used earlier in the description ("item" and "element"). It also specifies what should be done when "we're done" and revises the first step to avoid saying that the list should be divided.

I also use 4 tildes for the admonition to match the style guide or whatever as @SaschaMann commented about that.

### DIFF
--- a/exercises/binary-search/instructions.md
+++ b/exercises/binary-search/instructions.md
@@ -5,17 +5,17 @@ Your task is to implement a binary search algorithm.
 A binary search algorithm finds an item in a list by repeatedly splitting it in half, only keeping the half which contains the item we're looking for.
 It allows us to quickly narrow down the possible locations of our item until we find it, or until we've eliminated all possible locations.
 
-```exercism/caution
+````exercism/caution
 Binary search only works when a list has been sorted.
-```
+````
 
 The algorithm looks like this:
 
-- Divide the sorted list in half and compare the middle element with the item we're looking for.
-- If the middle element is our item, then we're done.
-- If the middle element is greater than our item, we can eliminate that number and all the numbers **after** it.
-- If the middle element is less than our item, we can eliminate that number and all the numbers **before** it.
-- Repeat the process on the part of the list that we kept.
+- Find the middle element of a list and compare it with the item we're looking for.
+- If the middle element is our item, then we're done: return its index.
+- If the middle element is greater than our item, we can eliminate that element and all the elements **after** it.
+- If the middle element is less than our item, we can eliminate that element and all the elements **before** it.
+- Repeat the process on the part of the list that has not been eliminated. If all the list has been eliminated then the item is not in the list.
 
 Here's an example:
 
--- a/exercises/binary-search/instructions.md
+++ b/exercises/binary-search/instructions.md
@@ -5,14 +5,14 @@ Your task is to implement a binary search algorithm.
 A binary search algorithm finds an item in a list by repeatedly splitting it in half, only keeping the half which contains the item we're looking for.
 It allows us to quickly narrow down the possible locations of our item until we find it, or until we've eliminated all possible locations.
 
-````exercism/caution
+~~~~exercism/caution
 Binary search only works when a list has been sorted.
-````
+~~~~
 
 The algorithm looks like this:
 
 - Find the middle element of a sorted list and compare it with the item we're looking for.
-- If the middle element is our item, then we're done: return its index.
+- If the middle element is our item, then we're done!
 - If the middle element is greater than our item, we can eliminate that element and all the elements **after** it.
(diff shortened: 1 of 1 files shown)


exercises/binary-search/instructions.md
```diff
@@ -5,17 +5,17 @@ Your task is to implement a binary search algorithm.
 A binary search algorithm finds an item in a list by repeatedly splitting it in half, only keeping the half which contains the item we're looking for.
 It allows us to quickly narrow down the possible locations of our item until we find it, or until we've eliminated all possible locations.
 
-```exercism/caution
+````exercism/caution
 Binary search only works when a list has been sorted.
-```
+````
```
kytrinyx, line 10: It's either three backticks, or four tildes, according to the styleguide. (See the section called **Special blocks** here). https://exercism.org/docs/building/markdown/markdown

exercises/binary-search/instructions.md
```diff
@@ -5,17 +5,17 @@ Your task is to implement a binary search algorithm.
...
 
 The algorithm looks like this:
 
-- Divide the sorted list in half and compare the middle element with the item we're looking for.
-- If the middle element is our item, then we're done.
-- If the middle element is greater than our item, we can eliminate that number and all the numbers **after** it.
-- If the middle element is less than our item, we can eliminate that number and all the numbers **before** it.
-- Repeat the process on the part of the list that we kept.
+- Find the middle element of a sorted list and compare it with the item we're looking for.
+- If the middle element is our item, then we're done: return its index.
```
kytrinyx, line 15: We need to avoid saying anything about the implementation here, because different tracks will tackle this differently.

kytrinyx (changes requested):
Thanks @cmcaine, I agree that using `element` rather than `number` is better.

exercises/binary-search/instructions.md
```diff
@@ -5,17 +5,17 @@ Your task is to implement a binary search algorithm.
...
+- If the middle element is greater than our item, we can eliminate that element and all the elements **after** it.
+- If the middle element is less than our item, we can eliminate that element and all the elements **before** it.
+- Repeat the process on the part of the list that has not been eliminated. If all the list has been eliminated then the item is not in the list.
```
iHiD, line 18: Split this into two bullet points?

```suggestion
- If list entirely list has been eliminated then the item is not in the list.
- Repeat the process on the part of the list that has not been eliminated. 
```

kytrinyx: Ah, I like that.

exercises/binary-search/instructions.md
SaschaMann: That section says 4 tildes unless in exceptional circumstances, which this doesn't seem to be?

exercises/binary-search/instructions.md
cmcaine: How about this?

```suggestion
- If every element of the list has been eliminated then the item is not in the list.
- Otherwise, repeat the process on the part of the list that has not been eliminated.
```

exercises/binary-search/instructions.md
cmcaine: Makes sense!

```suggestion
- If the middle element is our item, then we're done!
```

exercises/binary-search/instructions.md
cmcaine: Reading is very hard, okay.

cmcaine:
```suggestion
~~~~
```

exercises/binary-search/instructions.md
```diff
@@ -5,17 +5,17 @@ Your task is to implement a binary search algorithm.
 A binary search algorithm finds an item in a list by repeatedly splitting it in half, only keeping the half which contains the item we're looking for.
 It allows us to quickly narrow down the possible locations of our item until we find it, or until we've eliminated all possible locations.
 
-```exercism/caution
+````exercism/caution
```
cmcaine, line 8:
```suggestion
~~~~exercism/caution
```

exercises/binary-search/instructions.md
kytrinyx: @SaschaMann I'm not sure this was clear, but I was agreeing with you 😄 

I've submitted a PR that fixes this in main, so once that is merged we can rebase this branch against that one.

exercises/binary-search/instructions.md
kytrinyx: Yeah, that's nice, @cmcaine. It  flows even better.

ErikSchierboom (approved):

SaschaMann (approved):

kytrinyx (approved):

PR #2225, opened 2023-03-07
Turn off markdown linting rule for code blocks

kytrinyx:
Our style guide says that we want three backticks for code blocks, but to prefer (where possible) four tildes for special blocks like the exercism/caution block or exercism/note block.

The linter we use provides three options for linting:
- only backticks
- only tildes
- be consistent within a file

None of these options work for us, so we're turning off this linting rule.

### DIFF
--- a/.markdownlint.jsonc
+++ b/.markdownlint.jsonc
@@ -3,5 +3,6 @@
 {
     "MD004": "dash", // Prefer hyphens for unordered lists
     "MD013": false,  // Exercism tends to break lines at sentence ends rather than at a column limit
-    "MD024": false   // The format for instructions.md requires  repeated headings, e.g. "Task"
+    "MD024": false,  // The format for instructions.md requires  repeated headings, e.g. "Task"
+    "MD048": false   // We want three backticks for codeblocks, but four tildes for special blocks.
 }


iHiD (approved):

junedev (approved):

kotp (approved):

kotp:
I was going to bring this in, but there is another potential.  I believe we can use tilde, "~", or back tick, "`", for code blocks.  We could document to use 3 of one and 3 of the other for specific semantics.

kytrinyx:
@kotp we _can_ use either tilde or backticks for code blocks, but we would prefer to use backticks for code blocks and tildes for caution blocks.

kotp:
Right on.  I get that.  I was not very clear, is the problem the count of 4 in the one use.  Is it complaining about the use of 3 and 4 marks (seemingly inconsistent use), rather than taking advantage of our preferred use of backtick for one purpose and tilde for another?

kotp:
To clarify, it does have the option of "be consistent" and so if we consistently used 3 instead of 3 and 4, then would that be enough consistency?

kytrinyx:
> To clarify, it does have the option of "be consistent" and so if we consistently used 3 instead of 3 and 4, then would that be enough consistency?

To clarify, @kotp, the "consistent" option is what is currently running on CI. We explicitly don't want consistent, because we want:

- **3 backticks** when the block is a code block
- **4 tildes** when the code is a special block

Currently, the `consistent` option causes CI to fail for us when we consistently use the rules we prefer in a file that has both code blocks and special blocks.

PR #2226, opened 2023-03-08
Tweak wording in binary search introduction

kytrinyx:
As @kotp pointed out in a downstream PR, the word number can refer to a musical number, like what someone would perform on stage, or a mathematical number, like what you would use to count.

In this introduction we talk about songs, confusing the issue about what number refers to.

This adds a parenthetical to clarify which type of numbers we mean.

### DIFF
--- a/exercises/binary-search/introduction.md
+++ b/exercises/binary-search/introduction.md
@@ -1,7 +1,7 @@
 # Introduction
 
 You have stumbled upon a group of mathematicians who are also singer-songwriters.
-They have written a song for each of their favorite numbers, and, as you can imagine, they have a lot of favorite numbers.
+They have written a song for each of their favorite numbers, and, as you can imagine, they have a lot of favorite numbers (like 0 or 73 or _pi_, which starts with 3.14).
 
 You are curious to hear the song for your favorite number, but with so many songs to wade through, finding the right song could take a while.
 Fortunately, they have organized their songs in a playlist sorted by the title — which is simply the number that the song is about.
--- a/exercises/binary-search/introduction.md
+++ b/exercises/binary-search/introduction.md
@@ -1,9 +1,13 @@
 # Introduction
 
 You have stumbled upon a group of mathematicians who are also singer-songwriters.
-They have written a song for each of their favorite numbers, and, as you can imagine, they have a lot of favorite numbers (like 0 or 73 or 6174).
+They have written a song for each of their favorite numbers, and, as you can imagine, they have a lot of favorite numbers (like [0][zero] or [73][seventy-three] or [6174][kaprekars-constant]).
 
 You are curious to hear the song for your favorite number, but with so many songs to wade through, finding the right song could take a while.
 Fortunately, they have organized their songs in a playlist sorted by the title — which is simply the number that the song is about.
 
 You realize that you can use a binary search algorithm to quickly find a song given the title.
+
+[zero]: https://en.wikipedia.org/wiki/0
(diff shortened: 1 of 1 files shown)


exercises/binary-search/introduction.md
```diff
@@ -1,7 +1,7 @@
 # Introduction
 
 You have stumbled upon a group of mathematicians who are also singer-songwriters.
-They have written a song for each of their favorite numbers, and, as you can imagine, they have a lot of favorite numbers.
+They have written a song for each of their favorite numbers, and, as you can imagine, they have a lot of favorite numbers (like 0 or 73 or _pi_, which starts with 3.14).
```
ErikSchierboom, line 4: Wouldn't having pi as a favorite number imply that we expect the number to potentially be a floating-point number?

kytrinyx: Yeah, I was really hesitating about _pi_. Jeremy suggested it in the discussion on the Ruby track, and I like how it's a "famous" number. I also considered _e_ and _square root of -1_. And I don't like how 3.14 isn't pi it's just the start of it.

I picked 73 even though I don't watch Big Bang Theory, because Dr Sheldon is right about it being an awesome number.

> The best number is 73. Why? 73 is the 21st prime number. Its mirror, 37, is the 12th and its mirror, 21, is the product of multiplying 7 and 3.
> 
> In binary 73 is a palindrome, 1001001, which backwards is 1001001.

Jeremy also suggested 42, but that's not a math thing, it's a programmer/geek thing from the Hitchhiker's Guide to the Galaxy (which I tried reading and gave up on, so I'm biased against 42).

Do you have a suggestion for a nice number we can have instead of pi? I'd like a really big one that is interesting for mathematical reasons.

kytrinyx:
```suggestion
They have written a song for each of their favorite numbers, and, as you can imagine, they have a lot of favorite numbers (like 0 or 73 or 6174).
```

ErikSchierboom: I like Kaprekar's number. The speed of light is more physics than maths.

kytrinyx: Yeah, speed of light is a nice one, though it isn't an abstract number (you'd need the unit in order to actually specify it. And also, as you say, physics, not maths).

Ok, cool. I'm going with Keprekar's constant!

junedev (approved):

exercises/binary-search/introduction.md
```diff
@@ -1,7 +1,7 @@
 # Introduction
 
 You have stumbled upon a group of mathematicians who are also singer-songwriters.
-They have written a song for each of their favorite numbers, and, as you can imagine, they have a lot of favorite numbers.
+They have written a song for each of their favorite numbers, and, as you can imagine, they have a lot of favorite numbers (like 0 or 73 or 6174).
```
ErikSchierboom, line 4: One final thought: would it be nice to link to wikipedia for the favorite numbers as 73 and 6174 do have some meaning. Just a thought, feel free to ignore.

ErikSchierboom (approved):

exercises/binary-search/introduction.md
kytrinyx: Yeah I like that.

glennj:
`8675309` is an interesting number, but not mathematically: https://en.wikipedia.org/wiki/867-5309/Jenny

kotp (approved):
Looking good!

kytrinyx:
Okidoki. I think we're sorted here! (No pun intended).

PR #2227, opened 2023-03-09
[Binary Search] Emphasis "sorted" in the algorithm listing

kotp:
Even though there is a caution block, this brings attention to that
requirement where we are describing the solution to the problem.

Reference background discussion in exercism/ruby#1502

### DIFF
--- a/exercises/binary-search/instructions.md
+++ b/exercises/binary-search/instructions.md
@@ -11,7 +11,7 @@ Binary search only works when a list has been sorted.
 
 The algorithm looks like this:
 
-- Find the middle element of a sorted list and compare it with the item we're looking for.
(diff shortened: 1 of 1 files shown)


kotp:
That sounds good to me.  I only noticed it because human.  Did not pay attention the more blatant, but would have noticed the italics, yet still noticed the multiple insistence of "sorted", which is even more subtle.

andrerfcsantos (approved):
I like this, because it is an important part for the implementation and can be easily missed.

ErikSchierboom (approved):

kytrinyx (approved):
Alrighty, this has two votes. I'll add a third so we can merge :)

PR #2231, opened 2023-03-13
Clarify sum-of-multiples description

cmcaine:
I think the current description isn't very clear. It also fails our guidance by not specifying everything the student needs to solve the exercise (doesn't specify that limit and factors are non-negative and isn't super-clear that the multiples must be non-negative).

### DIFF
--- a/exercises/sum-of-multiples/description.md
+++ b/exercises/sum-of-multiples/description.md
@@ -1,6 +1,9 @@
 # Description
 
-Given a number, find the sum of all the unique multiples of particular numbers up to but not including that number.
+Given a limit and a list of factors, find the sum of all unique multiples of the factors that are less than the limit.
+All inputs will be greater than or equal to zero.
+
+## Example
 
 If we list all the natural numbers below 20 that are multiples of 3 or 5, we get 3, 5, 6, 9, 10, 12, 15, and 18.
 


exercises/sum-of-multiples/description.md
```diff
@@ -1,6 +1,9 @@
 # Description
 
-Given a number, find the sum of all the unique multiples of particular numbers up to but not including that number.
+Given a limit and a list of factors, find the sum of all unique multiples of the factors that are less than the limit.
```
kytrinyx, line 3: I completely agree that the original isn't very clear.

I think reversing `limit` and `list of factors` helps with flow, and maybe instead of `find the sum` we can go use the more straight-forward `add up...`.

That gives us:

> Given a list of factors and a limit, add up all the unique multiples of the factors that are less than the limit.

exercises/sum-of-multiples/description.md
```diff
@@ -1,6 +1,9 @@
...
+All inputs will be greater than or equal to zero.
+
+## Example
```
kytrinyx, line 6: I think maybe it would help if we expanded the example a bit as well.

E.g.

Suppose the limit is 20 and the list of factors is [3, 5]. We need to find the sum of all unique multiples of 3 and 5 that are less than 20.

Multiples of 3 less than 20: 3, 6, 9, 12, 15, 18
Multiples of 5 less than 20: 5, 10, 15

The unique multiples are: 3, 5, 6, 9, 10, 12, 15, 18

The sum of the unique multiples is: 3 + 5 + 6 + 9 + 10 + 12 + 15 + 18 = 78

So, the answer is 78.

exercises/sum-of-multiples/description.md
cmcaine: The main thing that's still weird about it is that it's ambiguous whether "that are less than the limit" applies to the factors or the unique multiples. But I couldn't work out a nice way to rewrite it (and in any case the example clears it up).

exercises/sum-of-multiples/description.md
```diff
@@ -1,6 +1,9 @@
...
 
 If we list all the natural numbers below 20 that are multiples of 3 or 5, we get 3, 5, 6, 9, 10, 12, 15, and 18.
```
cmcaine, line 9:
```suggestion
```

exercises/sum-of-multiples/description.md
cmcaine: I agree that that's better :)

cmcaine: I agree that that's better :)

exercises/sum-of-multiples/description.md
kytrinyx: Yeah, the example helps!

exercises/sum-of-multiples/description.md
ErikSchierboom: I like that suggestion!

kytrinyx: Cool. I have made it so!

kytrinyx (approved):
This is a good improvement. Thanks for suggesting it, @cmcaine

ErikSchierboom (approved):

iHiD (approved):
A bazillion times clearer. Thanks.

I'd still like to do some linking around words like "factors" (and maybe also "limit") but I think that can happen as part of the wider set of exercise updates we're doing. This is a big improvement and I think we should just merge it for now.

iHiD:
@kytrinyx This is probably worth propogating across tracks once it's merged if that's easy for you to do.

kytrinyx:
@iHiD Yes, I've got it all scripted up. I'll do that today!

>  I'd still like to do some linking around words like "factors" (and maybe also "limit") 

I actually tried to rewrite to use less "jargon-y" words, but ended up with a mess. Linking might be the way to go. We can definitely look at it as part of our overhaul project. 👍

rainij (commented):
Looks good

PR #2233, opened 2023-03-14
largest-series-product: use double quotes for string

ErikSchierboom:
This PR changes the description to use double quotes instead of single quotes, which seems to be more common across languages.


SaschaMann (approved):

kytrinyx (approved):
That seems perfectly reasonable!

angelikatyborska (approved):

PR #2234, opened 2023-03-14
Remove ambiguity in explanation of empty product

cmcaine:
This commit removes some mathematical jargon (multiplicative identity) and instead describes what should be returned.

The explanation of what the multiplicative identity is has been replaced with a short explanation (in an admonition block) of why the empty product (of numbers) is 1.


ErikSchierboom (approved):
I like it!

kytrinyx (approved):
This is a really nice improvement!

cmcaine:
Great! I'm glad that you both like it :)

iHiD (approved):
I like this.

PR #2237, opened 2023-03-21
Clarify details of the Secret Handshake exercise

safwansamsudeen:
This is the PR in response to [this issue](https://forum.exercism.org/t/error-in-js-exercise/4680). I've corrected the error and explained the concept little more detailedly.

@IsaacG suggested that I shouldn't change the "rightmost five bits" part, but upon checking the metadata, all the tests have numbers under 31. If there are tracks that implement tests with numbers over 31, I'll revert this change.

### DIFF
--- a/exercises/secret-handshake/instructions.md
+++ b/exercises/secret-handshake/instructions.md
@@ -2,8 +2,7 @@
 
 Your task is to convert a number between 1 and 31 to a sequence of actions in the secret handshake.
 
-The sequence of actions is chosen by looking at the rightmost five digits of the number once it's been converted to binary.
-Start at the right-most digit and move left.
+The sequence of actions is chosen by looking at the five digits of the number (which is between 1 and 31, equals 2 ^ 5 - 1) once it's been converted to binary. Start at the right-most digit and move left.
 
 The actions for each number place are:
 
@@ -18,7 +17,7 @@ The actions for each number place are:
 Let's use the number `9` as an example:
 
 - 9 in binary is `1001`.
-- The digit that is farthest to the right is 1, so the first action is `wink`.
+- The digit that is closest to the right is 1, so the first action is `wink`.
 - Going left, the next digit is 0, so there is no double-blink.
 - Going left again, the next digit is 0, so you leave your eyes open.
 - Going left again, the next digit is 1, so you jump.
--- a/exercises/secret-handshake/instructions.md
+++ b/exercises/secret-handshake/instructions.md
@@ -2,7 +2,8 @@
 
 Your task is to convert a number between 1 and 31 to a sequence of actions in the secret handshake.
 
-The sequence of actions is chosen by looking at the five digits of the number (which is between 1 and 31, equals 2 ^ 5 - 1) once it's been converted to binary. Start at the right-most digit and move left.
+The sequence of actions is chosen by looking at the five digits of the number (which is between 1 and 31, equals 2 ^ 5 - 1) once it's been converted to binary. 
+Start at the right-most digit and move left.
 
 The actions for each number place are:
 


exercises/secret-handshake/instructions.md
```diff
@@ -2,8 +2,7 @@
 
 Your task is to convert a number between 1 and 31 to a sequence of actions in the secret handshake.
 
-The sequence of actions is chosen by looking at the rightmost five digits of the number once it's been converted to binary.
-Start at the right-most digit and move left.
+The sequence of actions is chosen by looking at the five digits of the number (which is between 1 and 31, equals 2 ^ 5 - 1) once it's been converted to binary. Start at the right-most digit and move left.
```
IsaacG, line 5: We intentionally use one sentence per line. Please preserve that formatting.

exercises/secret-handshake/instructions.md
```diff
@@ -18,7 +17,7 @@ The actions for each number place are:
 Let's use the number `9` as an example:
 
 - 9 in binary is `1001`.
-- The digit that is farthest to the right is 1, so the first action is `wink`.
+- The digit that is closest to the right is 1, so the first action is `wink`.
```
IsaacG, line 20: Maybe something like:

> - The rightmost digit is 1, so the first action is `wink`.

safwansamsudeen:
> The instructions specify that the largest possible number is 31 (paragraph above).

True, but @IsaacG mentioned in a topic in the forum that other tracks might have numbers larger than 31.

exercises/secret-handshake/instructions.md
safwansamsudeen: OK.

exercises/secret-handshake/instructions.md
safwansamsudeen: As mentioned in the issue, I deliberately changed the wording to provide two different explanations of the same point (rightmost). There shouldn't be a comprehension issue in something this minor, but anyway.

IsaacG: I agree. "farthest to the right" is pretty clear. I'm not entirely sure where the comprehension issue is coming from.

I'm not a fan of this change as it is written, per my comment, so I'm not going to approve as is. But if others think this wording is better, I'll leave it to them :)

kytrinyx:
I would prefer that we check the tracks rather than make a change simply because they might hypothetically have something different.

I manually check all "append" files before merging if track maintainers don't merge the PR themselves.

exercises/secret-handshake/instructions.md
kytrinyx: I would like to understand what the comprehension issue is.

Earlier in the discussion the explanation for this change was that some tracks might have numbers larger than 31 in the test data.

kytrinyx: Ok, so a few things.

We've used `right-most` earlier already. Using `farthest to the right` was a way to vary the wording, in case that helps comprehension for some reason.

_farthest to the right_ is more correct, to my mind, than _closest to the right_, because you have to be closest to a _something_, and _the right_ isn't a something. _farthest to the right_ is a fixed expression that means right-most.

exercises/secret-handshake/instructions.md
```diff
@@ -2,7 +2,7 @@
 
 Your task is to convert a number between 1 and 31 to a sequence of actions in the secret handshake.
 
-The sequence of actions is chosen by looking at the rightmost five digits of the number once it's been converted to binary.
+The sequence of actions is chosen by looking at the five digits of the number (which is between 1 and 31, equals 2 ^ 5 - 1) once it's been converted to binary.
```
kytrinyx, line 5: We had `rightmost` because a lot of people think of a binary number as being 8 bits. So converting 17 to binary would give you: `00010001`.

I don't want to add an equation in the main text. We already have a note saying to go look up how binary works if you don't know. If we think we need to explain more, then we need to explain in the note, not in the main body of the instructions.

We don't need to repeat the fact that the number is between 1 and 31.

safwansamsudeen: OK, I'll revert this, then.

exercises/secret-handshake/instructions.md
safwansamsudeen: > We've used right-most earlier already. Using farthest to the right was a way to vary the wording, in case that helps comprehension for some reason.
That's what I thought, too.

> farthest to the right is more correct, to my mind
I agree with you, I'll change this.

IsaacG: https://www.thefreedictionary.com/Rightmost

> Rightmost: Farthest to the right

Yes. "Farthest to the right" is correct. That's been pointed out several times, both here and on the forum.

safwansamsudeen: Oh. So does this PR have any use, then? We can close it if not.

IsaacG: I don't think this PR has much use, either. Most the comments on the forum when this was raised were along the lines of, "I don't see anything wrong". I'm in favor of closing this out.

kytrinyx: Yeah, I think we can close it.

Thanks for taking the time to think through this.

In the future I would suggest that you work out the details in the forum befor opening the issue.

PR #2242, opened 2023-03-27
Improve book store description

safwansamsudeen:
This is the PR in response to [this topic in the forum](https://forum.exercism.org/t/improve-instructions-of-book-store-exercise/4767). The Javascript version has many more flaws, which I'll correct after this PR is merged.


IsaacG:
Please reach consensus of maintainers on the forum and have explicit input from a maintainer that a PR should be opened before doing so.

PR #2244, opened 2023-03-31
Rework simple-linked-list documentation

kytrinyx:
**This exercise is one of the focus exercises for 12in23.**

We aim to merge this PR in roughly 72 hours, and then submit PRs to all tracks that have this exercise.

The main change is to frame the exercise within the context of story.


ErikSchierboom (approved):

exercises/simple-linked-list/instructions.md
```diff
@@ -0,0 +1,21 @@
+# Instructions
+
+For the prototype, each song will simply be represented by a number.
+Given a range of numbers (the song IDs), create a singly linked list.
+
+Given a singly linked list, you should be able to reverse the list to play the songs in the opposite order.
+
+The prototype also needs to have import/export functionality.
+Given the playlist as a singly linked list, you will need to be able to convert it to the data type used for the external collection.
```
BethanyG, line 9: This feels a bit confusing to me, since (_at least for [Python](https://github.com/exercism/python/blob/main/exercises/practice/simple-linked-list/simple_linked_list_test.py)_) the tests don't have anything that would imply exporting, or returning data generally, except for the `reversed` method.  Are we adding test cases for presentation/export?

kytrinyx: I added it because the original specified going from arrays to linked lists and the opposite. But I didn't want to say array, because not all languages have them.

Definitely open to suggestions!

BethanyG: That's OK.  I can't think of a re-wording that works in this context.  Using `list` or `container` is equally problematic. 🙂 
It it is confusing, we can add or explain more in the [python addendum](https://github.com/exercism/python/blob/main/exercises/practice/simple-linked-list/.docs/instructions.append.md) - which probably needs a re-work --  or also talk about import/export scenarios when we look at constructing some canonical data?  No need to hold this up due to that detail.

BethanyG (approved):

exercises/simple-linked-list/instructions.md
iHiD: Maybe we could say:
> Given the playlist as a singly linked list, you will need to be able to convert it to whichever data type your language is using for the external collection (e.g. an array or list).

Do we maybe want to explicitely say here "Check the tests" too? For languages who aren't as wonderful as Python with their addendums?

BethanyG: That works  🙂  - but we can also leave it as-is. I think (for me) this is just highlighting that the exercise and test cases need a reexamination/rework for Python.

I think we may want to have a long and through discussion around the canonical data for this exercise, when we get to that point (_it doesn't have canonical data right now_). 

For example, we have [this issue in the Forum](https://forum.exercism.org/t/python-simple-linked-list-reversed-tests-do-not-check-the-return-value-is-a-linkedlist/4829), asking for tests to check for/force a particular return type for `reversed` -- so we may want to 'lean in' to the conversion part and have students make a `convert` method, or not as the case may be.

kytrinyx:
I've gone ahead and simplified this a bit further, removing the wording around import/export. I think that if there are questions about this on individuals tracks, maybe we can clarify in the `.append.md` file.

vaeng (approved):

PR #2246, opened 2023-03-31
Rework largest-series-product documentation

kytrinyx:
**This exercise is one of the focus exercises for 12in23.**

We aim to merge this PR in roughly 72 hours, and then submit PRs to all tracks that have this exercise.

The main change is to frame the exercise within the context of a story.

**NOTE:** I removed the bit about the empty product. Talking to Jeremy we concluded that we could simplify the description of this exercise immensely if we remove the one test case for the empty product, and we decided that this is a trade-off that we want to make.

Once this PR is merged, and the downstream PR to all the tracks is submitted, I will go through all those PRs and manually deprecate the test in question in the tests.toml, as well as deleting the test case from the test suite.

### DIFF
--- a/exercises/largest-series-product/instructions.md
+++ b/exercises/largest-series-product/instructions.md
@@ -22,4 +22,5 @@ Let's work through an example, with the input `"63915"`.
   - The product of the series `"639"` is 162 (`6 × 3 × 9 = 162`)
   - The product of the series `"391"` is 27 (`3 × 9 × 1 = 27`)
   - The product of the series `"915"` is 45 (`9 × 1 × 5 = 45`)
-- 162 is bigger than both 27 and 45, so the largest series product of `"63915"` is from the series `"639"`. So the answer is **162**.
+- 162 is bigger than both 27 and 45, so the largest series product of `"63915"` is from the series `"639"`.
+  So the answer is **162**.
--- a/exercises/largest-series-product/instructions.md
+++ b/exercises/largest-series-product/instructions.md
@@ -6,10 +6,10 @@ The technique you're going to use here is called the largest series product.
 
 Let's define a few terms, first.
 
-- **input** - the sequence of digits that you need to analyze
-- **series** - a sequence of adjacent digits (those that are next to each other) that is contained within the input
-- **span** - how many digits long each series is
-- **product** - what you get when you multiply numbers together
+- **input**: the sequence of digits that you need to analyze
+- **series**: a sequence of adjacent digits (those that are next to each other) that is contained within the input
+- **span**: how many digits long each series is
+- **product**: what you get when you multiply numbers together
 
 Let's work through an example, with the input `"63915"`.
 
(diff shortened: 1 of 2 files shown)


exercises/largest-series-product/instructions.md
```diff
@@ -0,0 +1,25 @@
...
+Your task is to look for patterns in the long sequence of digits in the encrypted signal.
+
+The technique you're going to use here is called the largest series product.
+
+Let's define a few terms, first.
+
+- **input** - the sequence of digits that you need to analyze
+- **series** - a sequence of adjacent digits (those that are next to each other) that is contained within the input
+- **span** - how many digits long each series is
+- **product** - what you get when you multiply numbers together
```
IsaacG, line 12:
```suggestion
- **input**: the sequence of digits that you need to analyze
- **series**: a sequence of adjacent digits (those that are next to each other) that is contained within the input
- **span**: how many digits long each series is
- **product**: what you get when you multiply numbers together
```

Definitions traditionally use `term: defintion`

exercises/largest-series-product/instructions.md
```diff
@@ -0,0 +1,25 @@
...
+- To form a series, take adjacent digits in the original input.
+- If you are working with a span of `3`, there will be three possible series:
+  - `"639"`
+  - `"391"`
+  - `"915"`
+- Then we need to calculate the product of each series:
+  - The product of the series `"639"` is 162 (`6 × 3 × 9 = 162`)
+  - The product of the series `"391"` is 27 (`3 × 9 × 1 = 27`)
+  - The product of the series `"915"` is 45 (`9 × 1 × 5 = 45`)
+- 162 is bigger than both 27 and 45, so the largest series product of `"63915"` is from the series `"639"`. So the answer is **162**.
```
IsaacG, line 25: Should this be split into two lines (one sentence per line)?

IsaacG (approved):

exercises/largest-series-product/instructions.md
kytrinyx: Nice catch!

IsaacG (approved):

ErikSchierboom (approved):

kytrinyx:
@BethanyG I saw that there's a discussion about the Python version of this exercise in the forum:
https://forum.exercism.org/t/seeming-discrepancy-between-instructions-and-tests-in-largest-series-product/5050/2

We're planning on removing the test with the zero span, in order to simplify the instructions (removing the entire section about what an empty product means and what the value of it is, and why). If I go ahead and merge this, and sync to the Python track and remove the empty span test, does that seem like it would solve the issue discussed in the forum? Or has the Python exercise diverged a bit from the core spec?

BethanyG:
@kytrinyx  -- I think removing the tests altogether would fix the problem, yes.

The discussion on the forum was in regards to [this Python addendum](https://github.com/exercism/python/blob/main/exercises/practice/largest-series-product/.docs/instructions.append.md) that got super-specific with the error messages, since the Python tests were changed to check for the exact phrasing from problem specs.  But then the phrasing got changed and the addendum became out of sync. 🤦🏽‍♀️ 

Removing the need to check altogether will allow us to omit those messages in the addendum.  😄

BethanyG (approved):

kytrinyx:
Cool! In that case I will merge this now, and then work on updating the tracks over the next week or so.

PR #2254, opened 2023-04-07
largest-series-product: remove description.md

ErikSchierboom:
I think it can be removed, as it has instructions.md and introduction.md files: https://github.com/exercism/problem-specifications/tree/main/exercises/largest-series-product


kytrinyx:
Oh! Yes, I meant to remove it and must have completely forgotten.

Should we make it so CI fails if we have all three md files? We should either have instructions+introduction, or description.

kytrinyx (approved):

ErikSchierboom:
> Should we make it so CI fails if we have all three md files? We should either have instructions+introduction, or description.

I think so.

IsaacG (approved):

Issue #2258, opened 2023-04-10
fetch-configlet: script is outdated


ee7:
I noticed this in #2256.

The version in this repo should be synced with [upstream](https://github.com/exercism/configlet/blob/85745ac34f7c3f557cdc2dc1becaa88c18192b98/scripts/fetch-configlet). The track repos use a good version.

PR #2259, opened 2023-04-10
bin/fetch-configlet: sync with upstream

ee7:
The fetch-configlet script in this repo was outdated. Update it to [the upstream version][1], which has many improvements.

Closes: #2258

[1]: https://github.com/exercism/configlet/blob/85745ac34f7c/scripts/fetch-configlet

### DIFF
--- a/bin/fetch-configlet
+++ b/bin/fetch-configlet
@@ -1,52 +1,90 @@
 #!/usr/bin/env bash
 
+# This file is a copy of the
+# https://github.com/exercism/configlet/blob/main/scripts/fetch-configlet file.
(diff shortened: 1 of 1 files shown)


glennj:
Why do we need configlet in this repo?

IsaacG:
I propose we sync from upstream first. This is a pretty simple and straight forward PR. That unblocks the bash cleanup PR and we can follow up on removing the script as a follow up issue.

glennj (approved):

vaeng (approved):
"The two files are identical"

PR #2281, opened 2023-05-06
Add Game of Life exercise

Steffan153:
https://forum.exercism.org/t/conways-game-of-life-exercise/5509

I don't think this is enough tests, but I'm not quite sure what to test. Should I add more random matrices, or should they be more specific cases?


IsaacG:
Do you want to have the number of turns be an input parameter?

igbanam:
To test for efficiency, the problem statement could be "How many generations are there between populations X and Y?"

…though, this would mean we have a timeout on the test-runner

kytrinyx:
I like the choice of a matrix of ones and zeros to represent the input and output. I would imagine that most languages can use that in their implementations with very little fuss.

kytrinyx:
In order to fully specify the basic rules of the cellular automata, I think we should have at least the following tests:

1. Underpopulation: live cell that dies due to having no live neighbors
2. Underpopulation: live cell that dies due to having only one live neighbor.
3. Survival: A live cell that survives to the next generation due to having two live neighbors.
4. Survival: A live cell that survives to the next generation due to having three live neighbors.
5. Overcrowding: A live cell that dies due to having four live neighbors.
6. Reproduction: A dead cell with exactly three live neighbors becomes a live cell.

In each of these, there would only be one turn.

I would instinctively want to have individual tests for a live cell that dies due to overcrowding with five, six, seven, and eight live neighbors, but I don't know this problem well, so I can't tell if this would be overkill.

Ideally each of these tests would have the smallest possible matrix with the fewest possible changes.

exercises/game-of-life/canonical-data.json
```diff
@@ -0,0 +1,75 @@
...
+    {
+      "uuid": "4ea5ccb7-7b73-4281-954a-bed1b0f139a5",
+      "description": "dead cells with three live neighbors become alive",
+      "property": "gameOfLife",
+      "input": {
+        "matrix": [
+          [1, 1, 0],
+          [0, 0, 0],
+          [1, 0, 0]
+        ]
```
kytrinyx, line 22: I like the choice of a matrix of ones and zeros to represent the input and output. I would imagine that most languages can use that in their implementations with very little fuss.

exercises/game-of-life/canonical-data.json
```diff
@@ -0,0 +1,75 @@
...
+    },
+    {
+      "uuid": "4ea5ccb7-7b73-4281-954a-bed1b0f139a5",
+      "description": "dead cells with three live neighbors become alive",
+      "property": "gameOfLife",
+      "input": {
+        "matrix": [
+          [1, 1, 0],
+          [0, 0, 0],
+          [1, 0, 0]
```
kytrinyx, line 21: I like the choice of a matrix of ones and zeros to represent the input and output. I would imagine that most languages can use that in their implementations with very little fuss.

exercises/game-of-life/canonical-data.json
```diff
@@ -0,0 +1,75 @@
...
+      "expected": [
+        [1, 1, 0, 1, 1, 0, 0, 0],
+        [0, 0, 0, 0, 0, 1, 1, 0],
+        [1, 0, 1, 1, 1, 1, 0, 1],
+        [1, 0, 0, 0, 0, 0, 0, 1],
+        [1, 1, 0, 0, 1, 0, 0, 1],
+        [1, 1, 0, 1, 0, 0, 0, 1],
+        [1, 0, 0, 0, 0, 0, 0, 0],
+        [0, 0, 0, 0, 0, 0, 1, 1]
+      ]
```
kytrinyx, line 72: In order to fully specify the basic rules of the cellular automata, I think we should have at least the following tests:

1. Underpopulation: live cell that dies due to having no live neighbors
2. Underpopulation: live cell that dies due to having only one live neighbor.
3. Survival: A live cell that survives to the next generation due to having two live neighbors.
4. Survival: A live cell that survives to the next generation due to having three live neighbors.
5. Overcrowding: A live cell that dies due to having four live neighbors.
6. Reproduction: A dead cell with exactly three live neighbors becomes a live cell.

In each of these, there would only be one turn.

I would instinctively want to have individual tests for a live cell that dies due to overcrowding with five, six, seven, and eight live neighbors, but I don't know this problem well, so I can't tell if this would be overkill.

Ideally each of these tests would have the smallest possible matrix with the fewest possible changes.

exercises/game-of-life/canonical-data.json
```diff
@@ -0,0 +1,75 @@
+{
+  "exercise": "game-of-life",
+  "cases": [
+    {
+      "uuid": "ae86ea7d-bd07-4357-90b3-ac7d256bd5c5",
+      "description": "empty matrix",
+      "property": "gameOfLife",
```
kytrinyx, line 7: In terms of exercise design, I'd like to have a different name for the property.

If the implementation of the exercise gets created via a generator we'd end up with both the top level namespace and the property under test being _game of life_. I'd like to start with names that are a reasonable default. People can always choose to override this in an individual track, but starting with a good default helps tremendously.

I'm trying to think of names that have to do with the flow of generations, or population... maybe `tick` (the generations tick by)?

IsaacG: tick, round, simulate, advance, next

Any or all of those sounds good to me

IsaacG:
> To test for efficiency, the problem statement could be "How many generations are there between populations X and Y?"
> 
> …though, this would mean we have a timeout on the test-runner

I believe all test runners have time outs. It's part of the test runner infrastructure. 

Another approach is asking how many cells are alive after N rounds. While it might be more "fun" or interesting to have different function signatures, I'm not sure it adds too much to the problem. It does provide a bit more room for flexibility in the approach; if I'm just counting cells or generations until a certain pattern, it would be easier to switch from a 2D array to a collection of points (which is more efficient for sparse data). If I'm doing a single tick, using a sparse collection is harder.

exercises/game-of-life/canonical-data.json
IsaacG: I think this is a very good idea. I think it would be great to see a test for each count of neighbors. Maybe with an empty border so there's no literal edge cases. 

Would it also make sense to have a grid which combines all those individual minimal unit tests into a single test?

MatthijsBlom: > Would it also make sense to have a grid which combines all those individual minimal unit tests into a single test?

That would be a puzzle:

| Alive neighbors | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 |
|--------|-|-|-|-|-|-|-|-|-|
| Cases | 1 | 8 | 28 | 56 | 70 | 56 | 28 | 8 | 1 |

(Multiply by 2 to account for the focus being alive/dead.)

Also, it is impossible to have test cases that _only_ check certain numbers of alive neighbors, for all numbers. (But is is very much possible for _some_.)

IsaacG: It doesn't have to be perfect. 

```
# Zero
000
000
000

# One
100
000
000

# Two
110
000
000

# Three 
101
000
001

# Etc
```

IsaacG: For combining, I meant literally having a larger matrix which has the minimal matrices copy pasted into non overlapping sections.

exercises/game-of-life/canonical-data.json
ErikSchierboom: I like all of those too

Steffan153:
Sorry I took so long! I've added a bunch more tests.

exercises/game-of-life/canonical-data.json
Steffan153: IMO we don't need an Overcrowding test for every number, but the test I added for it naturally has different numbers of neighbors for each cell.

exercises/game-of-life/metadata.toml
```diff
@@ -0,0 +1,3 @@
+title = "Conway's Game of Life"
+blurb = "Implement Conway's Game of Life."
+source_url = "https://en.wikipedia.org/wiki/Conway%27s_Game_of_Life"
```
ErikSchierboom, line 3:
```suggestion
source = "Wikipedia"
source_url = "https://en.wikipedia.org/wiki/Conway%27s_Game_of_Life"
```

exercises/game-of-life/description.md
ErikSchierboom, line 1: Could you split this document into an `introduction.md` and `instructions.md` file? That's the new structure that we are converting exercises to, separating the story/theme/background from the instructions.

Here is an example for `leap`:

- https://github.com/exercism/problem-specifications/blob/main/exercises/leap/introduction.md
- https://github.com/exercism/problem-specifications/blob/main/exercises/leap/instructions.md

ErikSchierboom (commented):
Looks awesome! Just a couple of tiny nits.

Steffan153:
Does this need an icon btw?

IsaacG (approved):

ErikSchierboom:
> Does this need an icon btw?

It does. We'll get it sorted.

exercises/game-of-life/introduction.md
```diff
@@ -0,0 +1,11 @@
+# Introduction
+
+[Conway's Game of Life][game-of-life] is a fascinating cellular automaton created by the British mathematician John Horton Conway in 1970. The game consists of a two-dimensional grid of cells that can either be "alive" or "dead." Each generation, the cells interact with their eight neighbors, which are cells adjacent horizontally, vertically, or diagonally.
+
+On each new generation, the following rules are applied:
+
+- Any live cell with two or three live neighbors lives on.
+- Any dead cell with exactly three live neighbors becomes a live cell.
+- All other cells die or stay dead.
```
ErikSchierboom, line 9: I think this bit should be part of the instructions, as they're defining the rules.

exercises/game-of-life/introduction.md
```diff
@@ -0,0 +1,11 @@
+# Introduction
+
+[Conway's Game of Life][game-of-life] is a fascinating cellular automaton created by the British mathematician John Horton Conway in 1970. The game consists of a two-dimensional grid of cells that can either be "alive" or "dead." Each generation, the cells interact with their eight neighbors, which are cells adjacent horizontally, vertically, or diagonally.
```
ErikSchierboom, line 3: Could you use one line per sentence?

```suggestion
[Conway's Game of Life][game-of-life] is a fascinating cellular automaton created by the British mathematician John Horton Conway in 1970.
The game consists of a two-dimensional grid of cells that can either be "alive" or "dead."
Each generation, the cells interact with their eight neighbors, which are cells adjacent horizontally, vertically, or diagonally.
```

ErikSchierboom: Maybe you briefly explain what a cellular automaton is?

ErikSchierboom: I'd also suggest to copy the last line to the `instructions.md` file and update the text here to something like:

```
Each generation, the cells interact with their eight neighbors via a set of rules, which defined the new generation.
```

This is a bit rough, but you probably get what I'm trying to do here :)

Steffan153: Something like this?

ErikSchierboom (approved):
Great work!

exercises/game-of-life/canonical-data.json
IsaacG: Could we rename this test? "Random" data doesn't make for a good test. Maybe "larger board" or "bigger test"? If it's random, what is it actually testing?

IsaacG (approved):

IsaacG (approved):

MatthijsBlom (commented):
I can't find the right headspace right now to review this thoroughly, sorry.

ErikSchierboom:
Cc @exercism/reviewers

ErikSchierboom:
@glennj or @BethanyG, would one of you have time to review this?

exercises/game-of-life/canonical-data.json
```diff
@@ -0,0 +1,143 @@
...
+          [0, 1, 0],
+          [1, 0, 0],
+          [0, 1, 0]
+        ]
+      },
+      "expected": [
+        [0, 0, 0],
+        [1, 1, 0],
+        [0, 0, 0]
+      ]
```
glennj, line 62: This test jumps ahead to "dead cells with three live neighbors become alive". Would it be better to use this input for testing "live cells with two live neighbors stay alive":
```
      "input": {
        "matrix": [
          [1, 0, 0],
          [1, 0, 0],
          [1, 0, 0]
        ]
      },
      "expected": [
        [0, 0, 0],
        [1, 0, 0],
        [0, 0, 0]
      ]
```
or even
```
      "input": {
        "matrix": [
          [1, 0, 1],
          [1, 0, 1],
          [1, 0, 1]
        ]
      },
      "expected": [
        [0, 0, 0],
        [1, 0, 1],
        [0, 0, 0]
      ]
```

glennj (approved):

exercises/game-of-life/canonical-data.json
Steffan153: On your first test suggestion, there is still a dead cell with three live neighbors. But  the second one is great, will update

BethanyG (approved):

ErikSchierboom:
Thanks @Steffan153 for creating this PR! It's a wonderful addition,

PR #2284, opened 2023-06-05
[queen-attack] SVG chess board

habere-et-dispertire:
I'm unsure how best to use a text fallback in markdown so I reached for `HTML`. I started with `picture` and `source` but that appears to handle alternate image formats and sizes without a final text fallback.

I ended up duplicating it in a `figcaption`. Open to suggestions.


iHiD:
Thanks for this. I see it as an improvement. I'd like to remove the ascii-art altogether though. I don't see much reason to have both.

Files live in https://github.com/exercism/images/tree/main/images/exercises so this would need a PR there.

habere-et-dispertire:
Image uploaded :

https://github.com/exercism/images/pull/12

IsaacG:
Does this (base) image need need attributions/licensing, based on where it came from? Or is this all original work or work with matching licensing?

Note: I thought I already commented, but I don't see my prior comment.

habere-et-dispertire:
My understanding is that screen readers don't do well with ASCII art. We should consider using `aria-hidden="true"` on all ASCII art ( ping @SleeplessByte ) :


~~~html
<pre aria-hidden="true">
  0 1 2 3 4 5 6 7
0 - - * - - * - - 8
1 * - * - * - - - 7
2 - * * * - - - - 6
3 * * W * * * * * 5
4 - * * * - - - - 4
5 * - * - * - - - 3
6 - - * - - B - - 2
7 - - * - - - * - 1
  a b c d e f g h
</pre>
~~~

Exercism does not appear to provide PNG fallback for SVGs -- I imagine over concern with increased bandwidth.

Alternate representations can provide a layered understanding much like different approachs. With it we can "see" that `g1` here is also a possible square to consider while not so in the proposed SVG.

One consideration which avoids a double depiction for sighted users is to put the ASCII art behind a `<detail>` . `lynx` automatically shows me just the ASCII art then :

~~~html
<figure>
    <picture>
        <source srcset="queen-capture.svg" type="image/svg+xml">
        <source srcset="queen-capture.png" type="image/png">
        <img src="queen-capture.svg" alt="">
    </picture>

    <figcaption>A chessboard with two queens. Lines emanating from the queen at <var>c5</var> indicate possible directions of capture along file, rank and diagonal.
        <details>
            <summary>Alternate chessboard as ASCII art</summary>
            <pre aria-hidden="true">
  0 1 2 3 4 5 6 7
0 - - * - - * - - 8
1 * - * - * - - - 7
2 - * * * - - - - 6
3 * * W * * * * * 5
4 - * * * - - - - 4
5 * - * - * - - - 3
6 - - * - - B - - 2
7 - - * - - - * - 1
  a b c d e f g h</pre>
        </details>
    </figcaption>
</figure>
~~~

habere-et-dispertire:
CC @exercism/reviewers

ErikSchierboom (approved):

senekor (approved):

kotp (approved):

PR #2285, opened 2023-06-07
[binary-search-tree] Four SVG graphs

habere-et-dispertire:


exercises/binary-search-tree/description.md
```diff
@@ -19,29 +19,25 @@ All data in the left subtree is less than or equal to the current node's data, a
...
-        /   \
-       /     \
-      2       6
-     / \     / \
-    1   3   5   7
+[A graph with root node 4, two child nodes 2 and 6, and four grandchild nodes 1, 3, 5 and 7.][4-2-6-1-3-5-7]
+
+## Credit
+
+The graphs are made by [habere-et-dispertire](https://exercism.org/profiles/habere-et-dispertire) using [PGF/TikZ](https://en.wikipedia.org/wiki/PGF/TikZ) by Till Tantau.
```
iHiD, line 38:
```suggestion
The images are made by [habere-et-dispertire](https://exercism.org/profiles/habere-et-dispertire) using [PGF/TikZ](https://en.wikipedia.org/wiki/PGF/TikZ) by Till Tantau.
```

SleeplessByte:
Are these image urls auto-signed @iHiD because right now they don't work.

habere-et-dispertire:
I think [ [binary-search-tree] Four SVG graphs #14 ](https://github.com/exercism/images/pull/14) is awaiting consideration and blocking.

habere-et-dispertire:
CC @exercism/reviewers

ErikSchierboom (approved):

MatthijsBlom:
This replaces ASCII art with links. Is this intentional? I expected images instead of links.

habere-et-dispertire:
Well spottted @MatthijsBlom ! I have moved them inline. 👍

GitHub Preview ( and Gmail ) will still not show them due to site-wide SVG scrubbing due to the potential of JavaScript embedding within SVG. A separation of concern between data and executable seems difficult to maintain : see the historical sagas with Microsoft Word Documents, and also PDFs.

An interesting related read on [SVG delineation](https://github.com/w3c/svgwg/issues/266#issuecomment-1336630343) at mime-type.

habere-et-dispertire:
Thanks -- they didn't show up in dark mode as the background is transparent and the image was not inverted with dark mode. ( They didn't work at all in the past. )

MatthijsBlom:
The image not sticking out against dark background sounds like a potential problem for the website as well.

habere-et-dispertire:
Happy to adjust. Are image backgrounds meant to be a solid colour or transparent ? I don't see any documentation for [exercism-images](https://github.com/exercism/images/) .

habere-et-dispertire:
> Is the page CSS inaccessible to the SVG? 

No and that's why I set the background transparent. You only have a single resource then which can be automatically adjusted on light/dark mode preference.

@dem4ron : Does an Exercism CSS class exist for labelling an image as suitable for inversion in dark mode ? It seems they are not all inverted automatically in dark mode since some are not suitable for inversion.

habere-et-dispertire:
So it seems the transparent background is correctly inverted on change of light/dark theme. It is the foreground SVG that is not inverted. If our markdown flavour supports pandoc attribute syntax, we could label the image as fully invertible :

~~~md
![alt text](the.svg){.invertible}
~~~

A compromise approach is to make the stroke on the image a middle gray which should retain enough contrast with the background switching but the foreground not.

ErikSchierboom:
> If our markdown flavour supports pandoc attribute synta

We don't support that IIRC

ErikSchierboom:
I wonder if we should keep some ASCII text. Might be better for accessibility.

iHiD:
Yeah, I don't think we want to lose the ascii text in the local (CLI) instructions. But it would be nice to have them on the website. I don't know how we do that :)

habere-et-dispertire:
Can we put `HTML` into our Markdown ?

I couldn't seem to get an image fallback that is not an image ( ie ASCII Art ).

The closest I could get to a negotiated content/semantic approach was to hide the ASCII Art inside of the [`details` of a `figcaption`](https://github.com/exercism/problem-specifications/pull/2284#issuecomment-1585774018) . This seems to give the shell the ASCII Art, while not cluttering the online version with two representations ( unless you ask for it ).

ErikSchierboom:
Most of the html tags would be escaped, so that'll not work I think

MatthijsBlom:
A few days ago I came across [Typograms](https://google.github.io/typograms/). Might be a cool idea, but perhaps not easy to implement.

habere-et-dispertire:
Interesting @MatthijsBlom -- thanks ! 😀

[Mermaid](https://github.com/mermaid-js/mermaid) leans in a more declarative and semantic way if we're looking towards that approach.

habere-et-dispertire:
The SVG images have landed:

https://github.com/exercism/images/pull/31

ErikSchierboom (approved):

tasxatzial (approved):

BNAndras (approved):

PR #2286, opened 2023-06-08
[Secret Handshake] Fix formatting issue in admonition

safwansamsudeen:
There's a note which is looks incorrect in the UI. I think this change will solve it.
image

This is a basic PR, so I have not discussed it in the forum beforehand.

### DIFF
--- a/exercises/secret-handshake/instructions.md
+++ b/exercises/secret-handshake/instructions.md
@@ -43,6 +43,6 @@ jump, double blink
 
 ~~~~exercism/note
 If you aren't sure what binary is or how it works, check out [this binary tutorial][intro-to-binary].
+~~~~
 
 [intro-to-binary]: https://medium.com/basecs/bits-bytes-building-with-binary-13cb4289aafa
-~~~~


kytrinyx:
Thanks @safwansamsudeen.

It looks correct to me in the browser on the Ruby track:
Screenshot 2023-06-08 at 1 37 37 PM

Which track are you looking at where it is broken?

safwansamsudeen:
Python - is this perhaps track specific?

safwansamsudeen:
@kytrinyx Nope - it turns out [you've](https://github.com/safwansamsudeen/problem-specifications/commit/a9f93a78db8fb77170d84f07a46c6dae88eda68a) corrected this specifically in Ruby.

MatthijsBlom:
This was fixed in #2238; the patch just hasn't been pulled yet by some tracks.

safwansamsudeen:
@kytrinyx Upon looking around, it seems both Python and JS lack the blank line between the link text and the link url.

You've [committed this](https://github.com/safwansamsudeen/problem-specifications/commit/a9f93a78db8fb77170d84f07a46c6dae88eda68a) only on March 27th - perhaps the tracks haven't synced the change yet?

Sorry for the confusion - I originally [opened this in Python](https://github.com/exercism/python/pull/3436), was redirected here, and then failed to see that there was a blank line difference in this version and Python's.

kytrinyx:
@safwansamsudeen this makes sense. The fix would be to run `bin/configlet-fetch` in the Python and JS repos, and then `bin/configlet sync --update --yes --docs` to update any instructions that have fixes from the upstream. `--metadata`

SleeplessByte (approved):
Nice.

MatthijsBlom:
@SleeplessByte I actually expect this PR to _break_ the link. See [this forum thread](http://forum.exercism.org/t/link-references-in-special-blocks-admonitions-are-not-rendered/3803/2).

SleeplessByte (changes requested):
Sorry I reviewed the opposite diff which I thought was moving the link inside the admonition.

There is nothing wrong with this markdown and this PR should be closed.

If tracks have broken admonitions, they should sync with the problem specs.

SleeplessByte:
> @SleeplessByte I actually expect this PR to break the rendering in just another way. See [this forum thread](http://forum.exercism.org/t/link-references-in-special-blocks-admonitions-are-not-rendered/3803/2).

Thank you for that. I just rejected this.

I've ran it myself and the old (current) markdown is correct. This PR is invalid.

safwansamsudeen:
Nice to see this understood, thanks to everybody!

PR #2288, opened 2023-06-11
lychee: do not link check (exercises/house) [papyr.com; it times out from GitHub actions

IsaacG:
Add papyr.com to the .lycheeignore file.

The link is up but when ran from GitHub actions, lychee times out when
retriveing the URL.
Due to that timeout, GH checks fail when everything is, in fact,
working.


BethanyG (approved):

kytrinyx (approved):

exercises/ledger/canonical-data.json
```diff
@@ -0,0 +1,240 @@
+{
```
wolf99, line 1: Not sure if this file is intended to be part of this change?

GitHub shows the whole file as new, but the description of the change means it seems like only the one extra line in the lycheeigbore should be expected?

IsaacG: Thanks! I somehow got that other file/commit on my local/main. It was not meant to be included. It's been dropped.

kotp (approved):

PR #2291, opened 2023-06-15
[New exercise]: Resistor-color master

meatball133:
This exercise is suppose to extend resistor-color trio to make it into a fully flesh converter so it can convert 1,4, or 5 bands resistor.


exercises/resistor-color-master/metadata.toml
```diff
@@ -0,0 +1,3 @@
+title = "Resistor Color Master"
+blurb = "Convert color codes, as used on resistors, to a human-readable label."
+source = "BethanyG, meatball133"
```
SleeplessByte, line 3: There may be an excerpt somewhere from Maud and @ErikSchierboom

BethanyG: I found this in `resistor-color-duo`:

>   "blurb": "Convert color codes, as used on resistors, to a numeric value.",
  "source": "Maud de Vries, Erik Schierboom",
  "source_url": "https://github.com/exercism/problem-specifications/issues/1464"

Which I do think we should add in.  Is that what you had in mind, @SleeplessByte?

IsaacG:
Since I don't see one, I made a [forum post](https://forum.exercism.org/t/new-resistor-color-exercise/6143) to discuss this exercise.

IsaacG:
@BethanyG the ledger exercise probably should have been discussed on the forum. Changes to the problem specifications should be first discussed on the forum. Track maintainers, like myself, should set a good example, not bypass the process. 

Ledger copied existing test data into the spec repo. This PR introduces a new exercise. I feel it should be discussed. I'm happy to open a discussion about ledger, too, if you'd like.

BethanyG:
Had a discussion with @meatball133, and we're going to keep this one track-specific for now.

PR #2295, opened 2023-06-26
High Score: drop implementation details from the problem spec

IsaacG:


vaeng (approved):

kytrinyx (approved):
Thanks! Totally agree that this is out of place.

BethanyG (approved):
Yup.  This has been a source of misery for several tracks, so happy its going.

PR #2297, opened 2023-07-02
Pig Latin: Remove use of "constant clusters"

iHiD:
Context: https://forum.exercism.org/t/pig-latin-ch-in-chair-is-not-consonant-cluster/6285

> maybe try to use those confusing terms as little as possible until fully understand them. the fact that there are different definitions for consonant clusters and consonant digraphs on different websites further proves that there is no reason to continue using them.


IsaacG (approved):

BethanyG (approved):
This reads much better. _Many thanks!_

kotp:
I guess I do not count as a "third" for approval. ;(

BethanyG:
@kotp - I think it's more that even with three maintainers, it takes a `maintainers-admin` before merging can happen.

ErikSchierboom (approved):
Done

PR #2308, opened 2023-08-23
Resistor Color Trio: Typo fix in description.md

BNAndras:
I think the intent here was to use the `resistor-color-duo` slug. Although I'm now wondering if this should just be replaced by the exercise name. Students would probably recognize Resistor Color Duo easier than `resistor-color-duo`.

### DIFF
--- a/exercises/resistor-color-trio/description.md
+++ b/exercises/resistor-color-trio/description.md
@@ -23,7 +23,7 @@ For this exercise, you need to know only three things about them:
 - Grey: 8
 - White: 9
 
-In `resistor-color duo` you decoded the first two colors.
+In `resistor-color-duo` you decoded the first two colors.
 For instance: orange-orange got the main value `33`.
 The third color stands for how many zeros need to be added to the main value.
 The main value plus the zeros gives us a value in ohms.
--- a/exercises/resistor-color-trio/description.md
+++ b/exercises/resistor-color-trio/description.md
@@ -23,7 +23,7 @@ For this exercise, you need to know only three things about them:
 - Grey: 8
 - White: 9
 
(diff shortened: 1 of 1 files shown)


IsaacG (approved):

MatthijsBlom:
I agree that this should be «Resistor Color Duo» rather than «`resistor-color duo`», for the same reason.

BethanyG (approved):

exercises/resistor-color-trio/description.md
```diff
@@ -23,7 +23,7 @@ For this exercise, you need to know only three things about them:
 - Grey: 8
 - White: 9
 
-In `resistor-color duo` you decoded the first two colors.
+In `Resistor Color Duo` you decoded the first two colors.
```
MatthijsBlom, line 26: Is the preservation of the backticks intentional?

BNAndras: It seemed their addition before was intentional (maybe draw attention?) so I left it alone. I don't have any strong feelings on whether to keep the backticks or not.

ErikSchierboom: I think they can be removed here. The previous version was a slug, which we regularly do use backticks on, but not for exercise titles I think.

```suggestion
In Resistor Color Duo you decoded the first two colors.
```

IsaacG: +1. Backticks are used around syntax, code, filenames and other words that are not "regular" prose/English. They should not be used around an exercise title.

IsaacG (approved):

ErikSchierboom (approved):

PR #2313, opened 2023-09-02
Add another test for the last frame

dsardelic:
If the first roll in the last frame is not a strike, then ensure that the next roll (i.e. the second roll in the last frame) pin count is less than or equal to the remaining number of pins in the lane.


IsaacG:
Please start a forum thread to discuss changes to the problem specs.

dsardelic:
https://forum.exercism.org/t/bowling-additional-test-proposal/7323

PR #2314, opened 2023-09-04
Change anagram test case input

olapokon:
The input for the anagram function is supposed to be a list containing single word strings, as in all other test cases. This test case has multiple words in the same string, separated by whitespace.

### DIFF
--- a/exercises/anagram/canonical-data.json
+++ b/exercises/anagram/canonical-data.json
@@ -133,6 +133,17 @@
       "expected": []
     },
     {
+      "uuid": "630abb71-a94e-4715-8395-179ec1df9f91",
+      "reimplements": "7cc195ad-e3c7-44ee-9fd2-d3c344806a2c",
+      "description": "does not detect an anagram if the original word is repeated",
+      "property": "findAnagrams",
+      "input": {
+        "subject": "go",
+        "candidates": ["goGoGO"]
+      },
+      "expected": []
+    },
+    {
       "uuid": "9878a1c9-d6ea-4235-ae51-3ea2befd6842",
       "description": "anagrams must use all letters exactly once",
       "property": "findAnagrams",


IsaacG (approved):

kotp (approved):

ErikSchierboom (changes requested):
This causes the test input to no long match the description, which states that the word is repeated. The new input does not test that, it "merely" checks for case insensitivity, which we already have tests for.

olapokon:
It's true that this change would make this test equivalent to the "words are not anagrams of themselves (case-insensitive)" test. Keeping it as a single string but removing the spaces is another option (the nearest, but not exact, equivalent to this probably is the test case "anagrams must use all letters exactly once").

It could remain as is, but it is a bit weird that this is the only candidate string that has spaces in it, thus making it necessary to account for whitespace in the solution, which doesn't seem to be intended.

ErikSchierboom:
> It could remain as is, but it is a bit weird that this is the only candidate string that has spaces in it, thus making it necessary to account for whitespace in the solution, which doesn't seem to be intended.

I 100% agree with that. Ideally it should be deprecated, but I don't think we yet have a deprecation option. Not sure what the best option would be

olapokon:
Changing it from "go Go GO"  to "goGoGO" would not make the description invalid and there would be no whitespace.  But this PR can be closed without making any changes, if you'd prefer to wait until it is possible to deprecate it.

ErikSchierboom:
> Changing it from "go Go GO" to "goGoGO" would not make the description invalid and there would be no whitespace

That would be perfectly fine with me

olapokon:
I changed it to a single string without whitespace.

ErikSchierboom (approved):

ErikSchierboom:
The final thing left to do is to, instead of mutating the existing test, you create a new test with a `reimplements` field that contains the UUID of the old test.

olapokon:
Thanks, I used configlet for the UUID.

exercises/anagram/canonical-data.json
```diff
@@ -132,6 +132,17 @@
...
+      "property": "findAnagrams",
+      "input": {
+        "subject": "go",
+        "candidates": ["goGoGO"]
+      },
+      "expected": []
+    },
     {
       "uuid": "9878a1c9-d6ea-4235-ae51-3ea2befd6842",
       "description": "anagrams must use all letters exactly once",
```
kotp, line 148: Should we state "all letters of the source word" or something similar?

ErikSchierboom: Isn't that a different test case? I mean, I don't disagree, but it looks like a different test case?

kotp: Does that mean that the description, then, is not right in general?

ErikSchierboom: I think so

olapokon: @kotp Did you mean change "anagrams must use all letters exactly once" to "anagrams must use all letters of the source word exactly once"? As @ErikSchierboom says this is the testcase below. Or did you mean change "does not detect an anagram if the original word is repeated" to "anagrams must use all letters of the source word exactly once"?

If you meant the former, is the description ok as it is now? It currently is "does not detect an anagram if the original word is repeated", which seems to fit the test case even after removing the spaces, since it is the original word being repeated, just without spaces in-between.

kotp: I think your change is fine on lines 135-145, so we are really talking about what you have not changed, and may change, as suggested.

Additionally, if you choose not to fix it while we are here, that is OK too.  It may be changed at another time, in another pull request.

kotp: Well, I must say that your change may be redundant if the following test case is described better.  It is not described well at the moment.  It may be enforcing the same thing.

olapokon: It depends on how one looks at it I guess. The original reason for this change was that the existing test is problematic due to the whitespace. I can change the next test's description as suggested if you prefer, or close the pull request and leave the goGoGO test as  it is until it can be deprecated/removed completely.

ErikSchierboom: Let's merge this as is and do the update of the other test case in a separate PR.

PR #2317, opened 2023-09-06
Phone number: Correct the format

iHiD:
Context: https://forum.exercism.org/t/nanp-phone-numbers/7179/1

This shouldn't change any tests, so I think it's pretty safe to merge.

### DIFF
--- a/exercises/phone-number/description.md
+++ b/exercises/phone-number/description.md
@@ -11,12 +11,14 @@ The first three digits of the local number represent the *exchange code*, follow
 The format is usually represented as
 
 ```text
-(NXX)-NXX-XXXX
+NXX NXX-XXXX
 ```
 
 where `N` is any digit from 2 through 9 and `X` is any digit from 0 through 9.
 
-Your task is to clean up differently formatted telephone numbers by removing punctuation and the country code (1) if present.
+Sometimes they also have the country code (represented as `1` or `+1`) prefixed.
+
+Your task is to clean up differently formatted telephone numbers by removing punctuation and the country code if present.
 
 For example, the inputs
 


IsaacG (approved):

kotp (approved):

ErikSchierboom:
> This shouldn't change any tests, so I think it's pretty safe to merge.

True, but most of the tests use the old format, so that might be a bit weird.

exercises/phone-number/description.md
```diff
@@ -11,12 +11,14 @@ The first three digits of the local number represent the *exchange code*, follow
 The format is usually represented as
 
 ```text
-(NXX)-NXX-XXXX
+NXX NXX-XXXX
```
kotp, line 14: We might show both that are often used... 

The tests are not going to differentiate in a way that errors, but will present it as a cleaned version.

Here we are showing what is "usually" represented, and I might think that 

```text
(NXX) NXX-XXXX
```

And

```text
(NXX)NXX-XXXX
```

are pretty evenly tied in the usual use.

I totally agree with using either with parenthesis or or dash to separate the area code from either or both sides of that, never both at the same time.

kotp (approved):

exercises/phone-number/description.md
IsaacG: We could also change "usually" to "sometimes" and not worry about what is commonly used where :smile:

iHiD:
In response to the comments in general, I think the point is that the tests provide different formats but really the input format is irrelevant - what matters is that it gets normalised. So I think we should make the description actively generic, rather than more specific. If that makes sense?

BethanyG (approved):

ErikSchierboom:
Totally agree

kytrinyx (approved):

PR #2322, opened 2023-09-15
Set up prettier for markdown files

### LINKED ISSUE #2321: Set up prettier for markdown files
On Discord, @siebenschlaefer suggested to format the [table in the acronym instructions](https://github.com/exercism/problem-specifications/blob/main/exercises/acronym/description.md?plain=1#L13-L17) because it's hard to read when you only read the markdown source.

That got me thinking - why do we not have prettier set up here to check formatting of all markdown files? We have it for JSON files.

Are there any unforeseen consequences? I see maybe two:
- A lot of files will get changed when initially run (currently seeing 36 files)
- `~~~~exercism/note` will get changed to ` ```exercism/note` - is that a problem?

I was also surprised not to find any evidence of having discussed this issue before in this repository... did I miss something?

If this is a desired change, I could work on it.

angelikatyborska:
Resolves #2321

Sorry for the notifications spam by pushing single commits, but I needed to see CI results 😇 

I see there's supposed to be a workflow reacting to `/format` comments, but it didn't work for me.

I see no suspicious reformatting happening by prettier's hands. I think this PR should be safe to merge.


exercises/book-store/description.md
```diff
@@ -36,8 +36,8 @@ This would give a total of:
 
 Resulting in:
 
-- 5 × (100% - 25%) * $8 = 5 × $6.00 = $30.00, plus
-- 3 × (100% - 10%) * $8 = 3 × $7.20 = $21.60
+- 5 × (100% - 25%) \* $8 = 5 × $6.00 = $30.00, plus
```
angelikatyborska, line 39: Should those have been changed to `×`?

exercises/circular-buffer/description.md
```diff
@@ -4,39 +4,55 @@ A circular buffer, cyclic buffer or ring buffer is a data structure that uses a
 
 A circular buffer first starts empty and of some predefined length.
 For example, this is a 7-element buffer:
-<!-- prettier-ignore -->
```
angelikatyborska, line 7: Removing the comment and reformatting didn't produce any changes. But just to be safe, I'll wrap those in backticks. A 4 space intent is equivalent to backticks, but it's not as clear in its intention.

.markdownlint.jsonc
```diff
@@ -4,5 +4,6 @@
     "MD004": "dash", // Prefer hyphens for unordered lists
     "MD013": false,  // Exercism tends to break lines at sentence ends rather than at a column limit
     "MD024": false,  // The format for instructions.md requires  repeated headings, e.g. "Task"
-    "MD048": false   // We want three backticks for codeblocks, but four tildes for special blocks.
+    "MD048": false,  // We want three backticks for codeblocks, but four tildes for special blocks.
+    "MD049": false   // Prettier uses underscores, but not all asterisk emphasis can be correctly replaced by underscores.
```
angelikatyborska, line 8: I was battling markdownlint in this PR, and then I realized that there's this problem:
```
- The *n*th number gets multiplied by 10^_(n-1)_.
```

Prettier didn't reformat `*n*th` to `_n_th` and when I tried, it doesn't render correctly in my IDE's markdown renderer. So I am making the assumption that I put in the comment here: it is not possible to have consistent emphasis with underscores because apparently they don't work the same as asterisks?

exercises/book-store/description.md
ErikSchierboom: I think so. Probably just over oversight

exercises/protein-translation/description.md
```diff
@@ -29,16 +29,16 @@ Note the stop codon `"UAA"` terminates the translation and the final methionine
...
+| Codon              | Protein       |
+| :----------------- | :------------ |
+| AUG                | Methionine    |
+| UUU, UUC           | Phenylalanine |
+| UUA, UUG           | Leucine       |
+| UCU, UCC, UCA, UCG | Serine        |
+| UAU, UAC           | Tyrosine      |
+| UGU, UGC           | Cysteine      |
+| UGG                | Tryptophan    |
+| UAA, UAG, UGA      | STOP          |
```
ErikSchierboom, line 41: This is so much better!

.markdownlint.jsonc
wolf99: Guessing that it is not possible to have prettier enforce asterisks only? 
(Which would result in uniformity)

.github/workflows/ci.yml
```diff
@@ -123,6 +123,24 @@ jobs:
           globs: |
             **/*.md
 
+      - name: Setup nodejs
```
wolf99, line 126: Nit: Is there a Prettier action that could be used either? It would remove the need to take care of yarn and node setup and maintenance.

wolf99 (approved):
LGTM, just two small (non-blocking) questions

.markdownlint.jsonc
ErikSchierboom: I don't think the emphasis is _that_ important here to be honest

ErikSchierboom: It could easily be backticks too I think

.github/workflows/ci.yml
angelikatyborska: There seems to be more than once, but I'm not sure if that's a good idea. Historically, prettier upgrades would change the output. So it's important to use exactly the same prettier version to format the files and then later to verify on CI that they have been formatted. If we use a dedicated prettier action, then we make it harder to use exactly the same prettier version in all places. Most likely changing the prettier version would require both changes to `package.json` and to the configuration of such github aciton where a specific prettier version could be passed. I'm not saying that's a blocker, it just makes me unsure about choosing such a solution.

.markdownlint.jsonc
angelikatyborska: Prettier offers almost zero configuration options, be design. It is not possible to change its emphasis preference to asterisks.

I'll remove the unnecessary emphasis on _n_ and remove this new linter rule.

meatball133 (approved):

angelikatyborska:
Yes, it was intended. Customizing prettier is not possible, so keeping tildes would require writing an extra script that undoes prettier's formatting. @ErikSchierboom said it's not necessary and everything will work with triple backticks too.

meatball133:
I think this topic is very scattered across the community, the usage of tildas and backticks really depends on the track. I think there would be the benefit of actually standardizing across tracks. I dont have any direct leaning on which way to go since both have their advantages and disadvantages, but I think some confusion and inconsistency across the tracks could be fixed if we said this is the way to do it instead of having 2 ways which just makes things confusing.

meatball133:
yes I am aware this is only the exercism specific admonitions.

> But the Exercism markdown spec previously said tildes, so tracks should have been using tildes. But I don't know if it ever made a difference.

Yes but various tracks have chosen one over the other, and at least in my view, it creates a weird situation.

meatball133:
> The problem isn't that Exercism didn't pick a side :)

I think there is a divide across the community, I would even like to drag it so far that it is a divide in the exercism team itself.

What has happened is that both methods have been seen as "acceptable" and thereby they have both spread and have both got grounds in their usage areas.

I don't think it may be that simple. And if 39 repos use the "unofficial" way and 88 use the official way those 88 likely mostly come from practice exercises. Does that hint at least to me that there is at least a decent amount of maintainers who use backticks instead of tildas.

iHiD:
A few late night thoughts:
1. If it's in the spec, we should all adhere to it (Erik and I ~often~ sometimes disagree - but we'll always take a "disagree and commit" approach. If it's in the spec. ambiguity is not ok. If it's not in a spec, ambiguity is ok. That's our approach.
2. I quite dislike 4 in 3 backticks (ie I much preferred outer tildes). That said I much prefer Prettier to not having Prettier, so am happy with the change, so consider this now the official Exercism way of doing it.
3. @ee7 Please update the docs - thank you.
4. All other docs not covered by Problem Specs should be updated to match this too. @ErikSchierboom Maybe a bulk PR?
5. @angelikatyborska Thank you for doing the work - its much appreciated by me!

ErikSchierboom:
There is still another option: we could do some post-processing after prettier is done to convert from 3 backticks to 4 tildes.

ErikSchierboom:
If someone is interested in working on this, we don't need to change anything. 

BTW we do something similar when formatting JSON using prettier, where we run this script as post-processing: https://github.com/exercism/problem-specifications/blob/main/bin/format-array.rb

ErikSchierboom:
> I think there is a divide across the community, I would even like to drag it so far that it is a divide in the exercism team itself.

I'd rather we'd not engage in speculation about "divisions" in the Exercism team please.

iHiD:
> There is still another option: we could do some post-processing after prettier is done to convert from 3 backticks to 4 tildes.

Erik and I are discussing this btw. @ee7 hold fire for now on changing the docs.

PR #2294, opened 2023-06-22
Check links in toml and json files

angelikatyborska:
We're adding the same link checker to the Elixir repo to catch dead links ([PR](https://github.com/exercism/elixir/pull/1330)) and I noticed it's failing on some URLs that come from this repository, three URLs from `https://users.csc.calpoly.edu`. How could that be? Well, this repository contains exercise metadata in toml files, which contain source URLs, which do not get checked.

Adding the check for toml and json files increases the number of links checked from 98 to 212.

~This is a draft because I'm not sure yet how to fix those three URLs...~ I will just ignore them.

```
| ✗ [ERR] https://users.csc.calpoly.edu/~jdalbey/103/Projects/ProgrammingPractice.html | Failed: Network error: error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed:ssl/statem/statem_clnt.c:1921: (unable to get local issuer certificate)
```~

PR #2323, opened 2023-09-15
Remove or replace all Twitter links

angelikatyborska:
Currently not logged in users are only allowed to see pages for single tweets. They are not allowed to see profile pages for example. I'm not allowed to access [https://twitter.com/jeg2](https://twitter.com/jeg2). And scripts aren't allow to access anything. `curl -I` on a valid tweet URL always returns 403. Thus Twitter is no longer appropriate to use as sources for anything.

Twitter links are also blocking https://github.com/exercism/problem-specifications/pull/2294

All the `http` instead of `https` are intentional - those are old websites that do not use SSL.

### DIFF
--- a/exercises/clock/metadata.toml
+++ b/exercises/clock/metadata.toml
@@ -1,4 +1,3 @@
 title = "Clock"
 blurb = "Implement a clock that handles times without dates."
 source = "Pairing session with Erin Drummond"
-source_url = "https://twitter.com/ebdrummond"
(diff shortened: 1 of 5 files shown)


exercises/clock/metadata.toml
```diff
@@ -1,4 +1,3 @@
 title = "Clock"
 blurb = "Implement a clock that handles times without dates."
 source = "Pairing session with Erin Drummond"
-source_url = "https://twitter.com/ebdrummond"
```
angelikatyborska, line 4: I could only find Erin Drummond on LinkedIn, but LinkedId does a lot of rate limiting and I suspect it's not going to pass the link checker. It responds with status 999 to curl attempts, something that I learned about today that's possible 🤷

ErikSchierboom (approved):

IsaacG (approved):

meatball133 (approved):

PR #2327, opened 2023-09-20
Format admonitions

ErikSchierboom:
- Add script to format admonitions
- Format admonitions in `format-md` NPM script
- Update admonitions


bin/format-admonitions.rb
```diff
@@ -0,0 +1,8 @@
+BACKTICK_MATCH_REGEX = /(?<start>`{3,})(?<type>exercism\/[a-z]+)(?<admonition>[\s\S]*?)(?<end>\k<start>)/.freeze
```
glennj, line 1: I love your use of named captures.

`[\s\S]` is exactly `.` -- i.e. any character.

bin/format-admonitions.rb
glennj, line 1: The purpose of this is to replace the admonition fences after the document has been formatted, right? Perhaps `format-admonition-boundaries` as a name?

ErikSchierboom: 👍

bin/format-admonitions.rb
iHiD: But I think maybe that was when I was using it in a lookahead.

ErikSchierboom: It doesn't work. I've just double-checked

bin/format-admonitions.rb
ErikSchierboom: I've renamed it to `format-admonition-fences`, which seems to be the proper markdown name: https://spec.commonmark.org/0.30/

bin/format-admonitions.rb
iHiD: Yeah - so this is what I was running into when testing on https://regex101.com/

Screenshot 2023-09-20 at 16 43 20

(Sample regexp if you want to play: 
````
(?<before>[\s\S]*)(?<start>```)(?<type>exercism\/[a-z]+\n(?<admonishion>.*?))(?<end>```)(?<after>[\s\S]*)````

Comment (glennj) [reply bin/format-admonitions.rb]: > Interestingly, this didn't work for me. `.` didn't include newlines.

You need the `m` option for the regex: https://ruby-doc.org/3.2.2/Regexp.html#class-Regexp-label-Options

Comment (glennj) [reply bin/format-admonitions.rb]: > Yeah - so this is what I was running into when testing on https://regex101.com/

Have to be a bit careful with regex testing sites. Are Ruby regexes PCRE?

Comment (iHiD) [reply bin/format-admonitions.rb]: > You need the m option for the regex: https://ruby-doc.org/3.2.2/Regexp.html#class-Regexp-label-Options

Nice! @ErikSchierboom's finished for the day but can update it to this tomorrow :)

> Have to be a bit careful with regex testing sites. Are Ruby regexes PCRE?

🤷 Internet seems to say yes.

Comment (ErikSchierboom) [reply bin/format-admonitions.rb]: Fixed in https://github.com/exercism/problem-specifications/pull/2327/commits/388ed5278588299f30d11b19d5516d0b9def3266

**Comment by ErikSchierboom** [comment]
CC @exercism/reviewers

**Comment by iHiD** [comment]
> So, should tracks that format files with prettier also do this?

Yes. And Erik's going to do a big PR for it I believe.

**Comment by ErikSchierboom** [comment]
> So, should tracks that format files with prettier also do this?

Ideally, yes I think. We could also bake this check into `configlet` maybe?

File: .github/workflows/ci.yml [hunk 2]
```diff
@@ -134,7 +134,7 @@ jobs:
 
       - name: Verify that markdown files are formatted correctly
         run: |
-          yarn format-md
+          yarn format-md && ruby bin/format-admonition-fences.rb
           if ! git diff --quiet --exit-code; then
             echo "please format the files with prettier, or apply the following changes:"
```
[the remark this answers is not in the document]
ErikSchierboom: I've updated

ErikSchierboom:
> That's possible. But I think it'd be better to have an org-wide workflow that runs markdownlint, checks admonitions, and performs any other Exercism-specific markdown linting. Thoughts?

That's a nice idea

ErikSchierboom:
CC @exercism/reviewers Once this PR is merged, I can do the cross-repo PR.

glennj (approved):

kotp (approved):

BethanyG (approved):

PR #2330, opened 2023-09-24
Fix inconsistent 'factor of x' description in raindrops

BNAndras:
All other descriptions referencing factor say 'factor x', but this test's description says 'factor of x' .

### DIFF
--- a/exercises/raindrops/canonical-data.json
+++ b/exercises/raindrops/canonical-data.json
@@ -75,7 +75,7 @@
     },
     {
       "uuid": "37ab74db-fed3-40ff-b7b9-04acdfea8edf",
-      "description": "the sound for 14 is Plong as it has a factor of 7",
+      "description": "the sound for 14 is Plong as it has a factor 7",
       "property": "convert",
       "input": {
         "number": 14


IsaacG (commented):
This should mostly not be visible to the student and changing this can cause churn in downstream track test files. Is there value in changing this? Does that value outweigh the consequences (making generated tests go out of sync)?

BNAndras:
I'm fine leaving it alone then.

ErikSchierboom:
> This should mostly not be visible to the student and changing this can cause churn in downstream track test files. Is there value in changing this? Does that value outweigh the consequences (making generated tests go out of sync)?

I think the value is just consistency, and I'm totally fine with that. The amount of churn this generates is not much and we also don't requeue solutions when they are upgraded but only the tests file has changed, so there is little downside I feel.

BethanyG:
> I think the value is just consistency, and I'm totally fine with that. The amount of churn this generates is not much and we also don't requeue solutions when they are upgraded but only the tests file has changed, so there is little downside I feel.

I am a bit confused here, because when a test file/test case changes, how do we know  that the solution still passes?  I would assume that that is the exact scenario when re-queueing would happen.  Is that because the GUID stays the same, so the file is not technically "different" enough?  Or is it a change that will need to be flagged at the track level to prevent re-queuing?

For the Python test generator, a change in description means a change to the name of the test in the test file, and I would assume a re-queue unless flagged.  For this exercise, that's 17, 133 solutions.  Fine to do that -- but it does seem a bit much considering the change is the removal of the word `of`.

ErikSchierboom:
Doh, I somehow misread 🤦 It _does_ mean requeuing when the tests file change, unless the magic marker is present. Let's close this after all. My bad all!

PR #2341, opened 2023-11-13
Add detailed Instructions to Binary Search Tree

pwadsworth:
Adds instructions to clarify what the exercise expects the student to do by clarifying the expected inputs/outputs and intent of each stub function.


glennj:
Not every track that implements this exercises has "functions" (there may different names for it). Not every track will use the same function names.

Track-specific language does not belong in problem-specifications.

PR #2350, opened 2023-12-09
Knapsack: Fix empty list of items

senekor:


exercises/knapsack/canonical-data.json
```diff
@@ -12,6 +12,17 @@
       },
       "expected": 0
     },
+    {
+      "uuid": "3993a824-c20e-493d-b3c9-ee8a7753ee59",
+      "reimplements": "a4d7d2f0-ad8a-460c-86f3-88ba709d41a7",
```
IsaacG, line 17: Do we need a re-implements when the expected generated test code shouldn't change?

senekor: Well, the CI tests fail if I change the existing test case :shrug:

IsaacG (approved):

wolf99 (approved):

exercises/knapsack/canonical-data.json
ErikSchierboom: It's technically a breaking change, so yeah the reimplements makes sense here I think.

ErikSchierboom (approved):

PR #2357, opened 2023-12-21
Update canonical-data.json

mohamad-shoumar:
Added a check where position of X is not at last position

### DIFF
--- a/exercises/isbn-verifier/canonical-data.json
+++ b/exercises/isbn-verifier/canonical-data.json
@@ -160,6 +160,16 @@
     },
     {
       "uuid": "daad3e58-ce00-4395-8a8e-e3eded1cdc86",
+      "description": "X at any position other than the last one is illegal",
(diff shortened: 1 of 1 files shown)


IsaacG:
See existing test `28025280-2c39-4092-9719-f3234b89c627`

MatthijsBlom:
Test `28025280-2c39-4092-9719-f3234b89c627` proved insufficient to catch the solution that inspired this test.

IsaacG:
Should we move the discussion to the forum? Changes should be discussed there prior to opening PRs.

MatthijsBlom:
I was not planning on championing this proposal, and will not participate in a forum discussion, sorry.

I'm just dropping by here to clear up a confusion that landed in my GH notifications feed.

The only thing I might otherwise have wanted to add in the future: a better test case would be, I think, a _random_ candidate number ending in `X` that is valid, except that it also contains an `X` in another place (making it invalid). The randomness would be there to signal that nothing special is going on, other than the `X`s.

PR #2360, opened 2024-01-14
Update Leap

iHiD:
This:
- Rewords the copy to make the instructions a little less algorithmic in nature
- Brings the copy in line with other recent updates
- Splits the copy into two introduction/instructions

### DIFF
--- a/exercises/leap/instructions.md
+++ b/exercises/leap/instructions.md
@@ -1,3 +1,3 @@
 # Instructions
 
-Your task is to determine if any given year is a leap year.
+Your task is to determine whether a given year is a leap year.
--- a/exercises/leap/instructions.md
+++ b/exercises/leap/instructions.md
@@ -1,0 +1,3 @@
+# Instructions
+
+Your task is to determine if any given year is a leap year.
--- a/exercises/leap/introduction.md
+++ b/exercises/leap/introduction.md
@@ -3,6 +3,7 @@
 A leap year (in the Gregorian calendar) occurs in every year that is evenly divisible by 4,  except if the year is evenly divisible by 100, unless the year is also evenly divisible by 400.
 
 Some examples:
+
 - 1997 was not a leap year as it's not divisible by 4.
 - 1900 was not a leap year as it's not divisible by 400
 - 2000 was a leap year!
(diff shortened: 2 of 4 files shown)


exercises/leap/introduction.md
```diff
@@ -0,0 +1,13 @@
+# Introduction
+
+A leap year (in the Gregorian calendar) occurs in every year that is evenly divisible by 4, except if the year is evenly divisible by 100, unless the year is also evenly divisible by 400.
```
IsaacG, line 3: This sentence is a bit tricky to parse with the except-unless clause. Do those both apply to the base rule? Does the unless apply to except? Does except and unless mean something different?
I'm wondering if it makes sense to split this into multiple sentences and make the exceptions a bit more explicit.

vaeng: Maybe like:

> A leap year (in the Gregorian calendar) occurs for every year evenly divisible by 4, but there are two exceptions.
If the year is evenly divisible by 100 it is not considered a leap year, unless it is also evenly divisible by 400.

iHiD: How about:

```suggestion
A leap year (in the Gregorian calendar) occurs:
- In every year that is evenly divisible by 4
- Unless the year is evenly divisible by 100, in which case it's only a leap year if the year is also evenly divisible by 400.
```

exercises/leap/instructions.md
```diff
@@ -0,0 +1,3 @@
+# Instructions
+
+Your task is to determine if any given year is a leap year.
```
vaeng, line 3: I like that this is now more on track with the "modern" concept exercise style files

vaeng (approved):

exercises/leap/introduction.md
IsaacG: That looks good to me!

BethanyG (approved):

IsaacG (approved):

ErikSchierboom (approved):

kotp (approved):

iHiD:
Thanks everyone! Updating your tracks to the latest version would be greatly appreciated :)

PR #2363, opened 2024-01-17
Update Reverse string to new format

iHiD:


exercises/reverse-string/metadata.toml
```diff
@@ -1,4 +1,4 @@
 title = "Reverse String"
-blurb = "Reverse a string."
+blurb = "Reverse a string without using built-in functionality."
```
iHiD, line 2:
```suggestion
blurb = "Reverse a string, without using built-in functionality."
```

exercises/leap/metadata.toml
```diff
@@ -1,5 +1,5 @@
 title = "Leap"
-blurb = "Determine whether a given year is a leap year."
+blurb = "Determine whether a given year is a leap year, without using built-in functionality."
```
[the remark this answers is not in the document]
ErikSchierboom: Maybe do this in a separate PR?

exercises/reverse-string/instructions.md
```diff
@@ -0,0 +1,11 @@
...
+
+Your task is to reverse a given string.
+
+Some examples:
+
+- Turn `"stressed"` into `"desserts"`.
+- Turn `"strops"` into `"sports"`.
+- Turn `"racecar"` into `"racecar"`.
+
+Many languages provide built-in functions or methods for reversing strings, but for this exercise, presume you have to write it from scratch.
```
IsaacG, line 11:
```suggestion
Many languages provide built-in functions or methods for reversing strings; for this exercise, write your own from scratch.
```

meatball133, line 11: I dont know if this kind of input should be included, for the first so does the majority of solutions use some form of built-in methods to reverse the string. It is also very vividly written and could mean a little bit of everything.

Like, here is "clearly" a method used:

```ruby
text = "ABC"
text.reverse
# => "CBA"
```

But here you could argue you are not using a "reverse" method/function but could still be seen as it:

```python
>>> text = "ABC"
>>> text[::-1]
"CBA"
```

And is an approach like, here I don't use any methods on strings to reverse it but I convert it to another datatype , is that categorized as "built-in" methods, or is the fact that I use reverse on an array seen as I use "built-in" methods?

```ruby
text = "ABC"
characters = text.split
characters.reverse
characters.join
# => "CBA"
```

BethanyG:
Feels like there isn't an easy track-wide answer here.

I _might_ rephrase to say "if you language provides 'methods of connivence' like `text.reverse` or `string.reversed()`, try solving this problem without them". 

But I am with @siebenschlaefer and @meatball133.  Simply saying "don't use [built-in functionality](https://docs.python.org/3/library/functions.html) for reversing strings" is a bit problematic, and will probably lead to non-idiomatic code and/or confusion.

For example, in Python strings are considered a [sequence type](https://docs.python.org/3/library/stdtypes.html#common-sequence-operations).  There is no `str.reverse()`. However, there **is** a generic `reversed()` that applies to all sequence types (_lists, tuples, strings, bytearray, & anything implementing `__reversed__`_).  But it returns a _reverse iterator_ -- so has to be unpacked and converted back into a string, or used in a loop where a string is assembled.  Allowed?  Not allowed?

Foregoing `reversed()` or even `list.reverse()` --  it is still next to impossible to do a string reverse cleanly & idiomatically without walking the sequence backwards (`[::-1]`).  Yes - you could use `enumerate()` + `range()` for that, but that is not what most Python folx would do.  You could argue that bracket notation _isn't_ a built-in method (_although it is considered a `slice()`, and `slice()` is a built-in_).

So maybe it would be better to be more language specific with an addendum?  For Python, I'd say "if you can, avoid methods like `reversed()` and `list.reverse()`", or "Consider what happens when you use `[::-1]`, and think about other ways you might accomplish this, even if it is a bit unidiomatic." 

Or maybe we prompt students to think about the process (and main problems) with reversing strings, and have them think about why the language implements the method(s) they way it does?  Things along the line of:
 
How does your approach change if you need to reverse a Unicode string, or a string as bytes?  What might be the considerations for when you need to reverse a large set of strings? Take a look at your languages implementation of strings, and why it may or may not include a string.reverse function or method.

senekor:
I agree with @siebenschlaefer.

In the real world, strings can almost never be assumed to be ascii and that is the only situation where a home-grown solution can be reasonably simple and correct.

For context, on the Rust track we actually encourage students to use a third-party library to process _grapheme clusters_, since not even unicode code points correspond to what humans think of as a "character". I speak German, writing a function by hand that correctly reverses `Würstchenstand` is no small feat. (As opposed to `Würstchenstand`, which is rather easy. Yes, they are rendered the same.)

I would also add that it's quite valuable to have an exercise that requires students to simply look up the correct function in the standard library. That is a skill that needs to be aquired at first, a fact experienced people can easily overlook.

iHiD:
All good points. Thanks :) I'll remove the bit about not using the built-in functions.

I do think for many tracks that just have `string.reverse()` equivalents, it's worth telling students not to use it. I don't see any real point in there being an exercise where they literally just call a single built-in function. I might consider tell them about that ("there's a really simple way to do this using the standard lib, which is what you should use in production, but let's use the exercise to learn about some deeper aspects of the language and avoid that function" (etc). The same as I don't think calling `Date.leap?` is particularly useful. But let's leave that to the individual tracks, rather than it being in Problem Specs.

jiegillet:
On the Elixir track, we opted not to implement this exercise precisely because it's a built-in function, and there are virtually no alternative methods that are worth spending time on, as a beginner exercise anyway.

We do have `leap` implemented on the track, and even though `Date.leap_year?` exists, the problem is a bit different in the sense that you can easily implement a correct solution, and it's worth learning about. When it comes to strings though, the difficulty is vastly different (of course it's still worth learning about in general).

ErikSchierboom:
> On the Elixir track, we opted not to implement this exercise precisely because it's a built-in function, and there are virtually no alternative methods that are worth spending time on, as a beginner exercise anyway.

I think the key observation here is _there are virtually no alternative methods that are worth spending time on_. Maybe we should recommend tracks forego this exercise when the above holds true. For many tracks though it will still make sense to have this exercise, but maybe for a couple it wouldn't.

exercises/leap/metadata.toml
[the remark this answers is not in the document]
iHiD:
```suggestion
blurb = "Determine whether a given year is a leap year."
```

exercises/reverse-string/instructions.md
iHiD, line 11:
```suggestion
```

exercises/reverse-string/metadata.toml
```diff
@@ -1,4 +1,4 @@
 title = "Reverse String"
-blurb = "Reverse a string."
+blurb = "Reverse a string, without using built-in functionality."
```
iHiD, line 2:
```suggestion
blurb = "Reverse a given string."
```

ErikSchierboom (approved):

ErikSchierboom:
The PR has been updated. Would you all take another look?

iHiD:
@ErikSchierboom @siebenschlaefer @BethanyG @senekor @BNAndras @meatball133 Happy with this now? I'd like to get the PRs out today if possible.

meatball133 (approved):

BethanyG (approved):

ErikSchierboom:
PRs are being created right now.

PR #2371, opened 2024-01-29
Split Raindrops into two files

iHiD:
Prepares Raindrops for #48in24.
- Split into two files
- Splits out modulo into a info block
- Adds remainder as well as modulo
- Change "factors" to "divisible by" (thanks @MatthijsBlom for this suggestion!)


exercises/raindrops/instructions.md
```diff
@@ -0,0 +1,24 @@
+# Instructions
+
+Your task is to convert a number into a string that contains corresponding raindrop sounds, based on the number's factors.
+
+The rules of `raindrops` are that if a given number:
+
+- has 3 as a factor, add 'Pling' to the result.
```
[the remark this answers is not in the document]
iHiD: Nice. Updated :)

exercises/raindrops/instructions.md
```diff
@@ -0,0 +1,24 @@
...
+## Examples
+
+- 28 is divisible by 7, but not 3 or 5, so the result would be `"Plong"`.
+- 30 is divisible by 3 and 5, but not 7, so the result would be `"PlingPlang"`.
+- 34 is not divisible by 3, 5, or 7, so the result would be `"34"`.
+
+~~~~exercism/note
+The simplest way to test if one number is evenly divisible by another is to use the [remainder][remainder] or [modulo][modulo] operator.
+
+[remainder]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Remainder
```
CarlHarrisonNti, line 21: You are reffering to javascript here, even if javascript isnt the only language with a remainder operator.

meatball133, line 21: You are reffering to javascript here, even if javascript isnt the only language with a remainder operator.

iHiD: I am, but I think it's a good doc on what the remainder operator is. I'm welcome to alternative links if they're not overly complex (e.g. wikipedia!)

iHiD: Also, FWIW, most languages have remainder - it's actually modulo that's the rarity.

exercises/raindrops/instructions.md
```diff
@@ -0,0 +1,23 @@
+# Instructions
+
+Your task is to convert a number into its corresponding raindrop sounds.
+
+If a given number:
+
+- is divisible by 3, add 'Pling' to the result.
+- is divisible by 5, add 'Plang' to the result.
+- is divisible by 7, add 'Plong' to the result.
```
ErikSchierboom, line 9: I think most languages use double quotes for strings, so maybe we use them there too?

```suggestion
- is divisible by 3, add "Pling" to the result.
- is divisible by 5, add "Plang" to the result.
- is divisible by 7, add "Plong" to the result.
```

exercises/raindrops/instructions.md
ErikSchierboom: I do feel that linking to JS docs is a bit odd, but the contents is good. I couldn't find a better page, except for maybe the wiki page.

ErikSchierboom: Another option would be to explain the remainder here

iHiD: The best solution will be to have a doc on Exercism that we link to. Let's do that (as it's always part of the plan and this is a low-hanging fruit), but let's link this to the JS doc for now until we have that, then update when we've added that doc.

iHiD: Actually, I'll do it now before we merge.

exercises/raindrops/instructions.md
```diff
@@ -0,0 +1,23 @@
...
+## Examples
+
+- 28 is divisible by 7, but not 3 or 5, so the result would be `"Plong"`.
+- 30 is divisible by 3 and 5, but not 7, so the result would be `"PlingPlang"`.
+- 34 is not divisible by 3, 5, or 7, so the result would be `"34"`.
+
+~~~~exercism/note
+The simplest way to test if one number is evenly divisible by another is to use the [remainder][remainder] or [modulo][modulo] operator.
+
+[remainder]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Remainder
```
iHiD, line 21:
```suggestion
[remainder]: https://exercism.org/docs/programming/operators/remainder
```

exercises/raindrops/instructions.md
iHiD: @ErikSchierboom Updated.

BethanyG:
I will have to instruction append on this because `%`, [`math.remainder()`](https://docs.python.org/3/library/math.html#math.remainder) and [`math.fmod()`](https://docs.python.org/3/library/math.html#math.fmod) do not always return the same thing in Python.
#floatingpointmathisfun

```python
>>> remainder(20, 3)
Out[52]: -1.0

>>> fmod(20, 3)
Out[53]: 2.0

>>> 20 % 3
Out[54]: 2
```

Also, using `remainder` and `fmod` require an import, the operator does not.  🤷🏽‍♀️

iHiD:
@BethanyG Cool, thanks. I guess it's worth noting that `math.remainder` is very much not what an average person would expect and telling them just to use `%`?

The reference to modulo/remainder hasn't really changed in this PR other than we stop telling people that `%` means modulo, which it doesn't in most languages (e.g. Python and JavaScript being the two biggest beginner languages, and two languages where `%` **doesn't** mean modulo)

BethanyG:
> Python and JavaScript being the two biggest beginner languages, and two languages where `%` **doesn't** mean modulo

err.  `%` DOES mean modulo in Python.  But yes, agree that we need to make sure Pythonistas understand the difference between the three.  😄

BethanyG (approved):

iHiD:
> err. % DOES mean modulo in Python. But yes, agree that we need to make sure Pythonistas understand the difference between the three. 😄

ChatGPT lies again! 😁 

Screenshot 2024-01-29 at 16 02 07

iHiD:
I think we can safely say that `%` is ok for raindrops in any language that supports it, regardless of whether it's modulo or remainder?

meatball133:
Yeah, I mean as long as the exercise doesn't deal with signed integers so do all languages I know work the same with the `%` operator.

exercises/raindrops/metadata.toml
```diff
@@ -1,4 +1,4 @@
 title = "Raindrops"
-blurb = "Convert a number to a string, the content of which depends on the number's factors."
+blurb = "Convert a number into its corresponding raindrop sounds - Pling, Plang and Plong"
```
IsaacG, line 2: Should the blurb be a complete sentence with a terminating period? We probably want to keep that consistent across exercises.

IsaacG (approved):

exercises/raindrops/instructions.md
```diff
@@ -0,0 +1,23 @@
+# Instructions
+
+Your task is to convert a number into its corresponding raindrop sounds.
+
+If a given number:
+
+- is divisible by 3, add "Pling" to the result.
+- is divisible by 5, add "Plang" to the result.
+- is divisible by 7, add "Plong" to the result.
+- **is not** divisible by 3, 5, or 7, the result should be the digits of the number.
```
glennj, line 10:
```suggestion
- **is not** divisible by 3, 5, or 7, the result should be the number as a string.
```

"the digits of the number" sounds like some (ordered) collection of the digits. But the result should definitely be a string.

exercises/raindrops/instructions.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Examples
+
+- 28 is divisible by 7, but not 3 or 5, so the result would be `"Plong"`.
+- 30 is divisible by 3 and 5, but not 7, so the result would be `"PlingPlang"`.
+- 34 is not divisible by 3, 5, or 7, so the result would be `"34"`.
+
+~~~~exercism/note
+The simplest way to test if one number is evenly divisible by another is to use the [remainder][remainder] or [modulo][modulo] operator.
```
iHiD, line 19:
```suggestion
A common way to test if one number is evenly divisible by another is to use [the remainder][remainder] or [the modulo operation][modulo].
Most languages provide operators or functions for one or both of these.
```

exercises/raindrops/instructions.md
kotp: I think that we had decided to be vague on the "type" of the return for this, to allow a language to determine what they need to test for.

However, the results are a ordered collection of the digits when a string as the order must be most significant to the left and least significant digit to the right, and in that order.

ErikSchierboom (approved):
Looks good (with the suggested change)

exercises/raindrops/instructions.md
iHiD: > "the digits of the number" sounds like some (ordered) collection of the digits. 

Yeah, I thought this when I was splitting it out, but figured it had been left vague for some reason for languages without strings. But then how do those languages do this exercise at all?

@kotp Do we have any concrete examples of where it's not a string?

exercises/raindrops/instructions.md
```diff
@@ -16,7 +16,8 @@ If a given number:
 - 34 is not divisible by 3, 5, or 7, so the result would be `"34"`.
 
 ~~~~exercism/note
-The simplest way to test if one number is evenly divisible by another is to use the [remainder][remainder] or [modulo][modulo] operator.
+A common way to test if one number is evenly divisible by another is to compare the [remainder][remainder] or [modulus][modulo] to zero.
+Most languages provide operators or functions for one (or both) of these.
```
kotp, line 20: This looks better, I was going to comment about it being "simplest" as, well, it is gone now, and this is better.

exercises/raindrops/instructions.md
kotp: I do not have a concrete example, but remember talking about how even "Hello World" returning a string in some languages is not "simple".  And so leaving the answer to be a collection is well enough.  It has been some years.

kotp: "How do they do this exercise at all?"  They might not.  However, having a collection of some kind can store those numeric values, and may also represent the strings as they see fit.  It is the relationship of the numbers being replaced by some value given some criteria that is important for the exercise, not really the strings values, I mean as far as learning value goes, and that does not diminish the "fun" value in my mind (think ciphergraphic kid games).

iHiD: My sense is that as this exercise is all about strings, we could safely get away with putting "string" in the description for this bit. I think it's much clearer if we can get away with it.

So I think let's fine a concrete counterexample, or go for Glenn's change. And if there's like only one counterexample, maybe that track should just make an exception?

kotp (approved):

exercises/raindrops/instructions.md
kotp: Right, just like all the exercises, the tracks are "inspired" by the problem-specifications  data, not ruled by it.

kotp: If this exercise were a concept exercise, then I would consider that potentially it is all about strings.  In Ruby I usually mentor this as "all about construction and flow control" rather than focusing on the fact that they are strings.  This is what we concentrate on as an Interview question for the simpler version, as well as expansion, when I use this as an interview question.

kotp: > So I think let's fine a concrete counterexample, or go for Glenn's change. And if there's like only one counterexample, maybe that track should just make an exception?

Not against Glenn's change, just the reason for it is not disqualifying for what is there.  (It is not a contradictory thing.)

iHiD: Yeah. OK - I'm going to go with it :) Thanks!

glaxxie:
> > err. % DOES mean modulo in Python. But yes, agree that we need to make sure Pythonistas understand the difference between the three. 😄
> 
> ChatGPT lies again! 😁

In powershell depend on the context and placement, `%` could mean modulus or act as alias for `foreach-object`. 
```powershell
5 % 2  
=> 1

or

@(1,2,3,4,5) | % { $_ % 2 } 
=> @(1,0,1,0,1)
```

Generally I recommend using the fullname to avoid any possible confusion, and easier to read but this practice is still quite prevalent among more experienced user while they do script in terminal.

PR #2382, opened 2024-02-02
Add `eliuds-eggs` exercise

ErikSchierboom:
The `pop-count` exercise should really have been named `eliuds-eggs`.
To help migrate to this new name in a non-breaking way, we're not renaming `pop-count` but instead have both of them temporarily.
Once all tracks have been updated (we'll do a bulk PR), we can remove the obsolete `pop-count` exercise.

### DIFF
--- a/exercises/eliuds-eggs/canonical-data.json
+++ b/exercises/eliuds-eggs/canonical-data.json
@@ -1,0 +1,41 @@
+{
+  "exercise": "eliuds-eggs",
+  "cases": [
+    {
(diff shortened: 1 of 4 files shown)


ErikSchierboom:
I'm gonna force merge this as we need this to continue with out bulk PRs. All the copy and data are the same, so there's nothing really left to check. The schema validation fails due to the uuids being duplicates now, but that is intentional as they'll be replacing the existing ones.

PR #2385, opened 2024-02-14
Add a test for which a greedy algorithm fails

blakelewis:
All of the existing tests can be satisfied with a greedy algorithm along the following lines:

for each coin value in descending order:
   use the maximum number of coins  such that the reduced problem
   omitting this value is still solvable

This test demonstrates that this algorithm is not optimal, since it gives a solution of 11 + 9 * 1, instead of 10 + 10.

This change was discussed in the forum in https://forum.exercism.org/t/a-new-test-for-change/8891.


ErikSchierboom (approved):

BNAndras (approved):

BethanyG (approved):

BNAndras:
I believe you just need to rebase the PR to get the schema validation to pass.

blakelewis:
I think that it is OK now?

Issue #2396, opened 2024-02-25
Exercise asks developer to write a function


kotp:
Having an exercise direct a student to write a function or write a
method or write "some specific structure" can be confusing when the
language they are solving the exercise in does not have that structure.

Where we say things like "Create a function" or "Write a function" we
could be indicating that we need a solution instead.  This will
encourage the student to determine what that solution is,
rather than us telling them to specifically create a function.

Unless the exercise is a concept exercise about functions, of course,
then the student should be told to create a function.

The exercises that mention "create a function" are:

- [ ] exercises/parallel-letter-frequency/description.md

And the exercises that mention "Write a function" are:

- [ ] exercises/sublist/metadata.toml
- [ ] exercises/flatten-array/description.md
- [ ] exercises/nucleotide-codons/metadata.toml
- [ ] exercises/nucleotide-codons/description.md
- [ ] exercises/alphametics/metadata.toml
- [ ] exercises/alphametics/description.md
- [ ] exercises/go-counting/description.md
- [ ] exercises/darts/description.md
- [ ] exercises/darts/metadata.toml

ref: [exercism/ruby#1638-comment](https://github.com/exercism/ruby/pull/1638#discussion_r1501824390)

glaxxie:
I accidentally clicked on the checkbox, I apologize if it triggers anything.

BNAndras:
Sigh. I didn’t mean to close the issue. Stupid thumbs on a small screen.

jiegillet:
Has this been linked to students actually being confused?
The semantic difference seems so small to me, I wouldn't want to spend time and energy on a non-issue.

kotp:
> Has this been linked to students actually being confused? The semantic difference seems so small to me, I wouldn't want to spend time and energy on a non-issue.

Yes, I can see how the meaning shift can be small for someone that is used to hearing it misused as often as I can imagine someone with  your experience as having.  We learn to "you mean this, but said that, I will not mention the difference in meaning" and we can definitely function like that."  That is not really kind, but it is something that happens fairly often.  They will continue to say things that they really do not mean to say, and it will likely not cause a problem.

But when someone is looking at how to make a function in a language that does not have them, how do they find what they can not find?  They have to, a bit clumsily, choose to ignore the meaning of the word and find the closest thing they can to do what was asked of them.

Semantics are all we have at the end of the day, the meaning of words.  So even small differences make a big difference, especially when it is something that is at the very core of certain languages, and non-existent in others.

[...]

PR #2406, opened 2024-03-19
Exercises no longer ask developers to 'write a function' (#2396)

adamazing:
Some proposed wording changes to avoid use of "write a function"/"create a function".

PR for Issue #2396

### DIFF
--- a/exercises/alphametics/description.md
+++ b/exercises/alphametics/description.md
@@ -1,6 +1,6 @@
 # Description
 
-Write a function to solve alphametics puzzles.
+Write a solver for alphametics puzzles.
 
 [Alphametics][alphametics] is a puzzle where letters in words are replaced with numbers.
 
@@ -25,7 +25,5 @@ Replacing these with valid numbers gives:
 This is correct because every letter is replaced by a different number and the words, translated into numbers, then make a valid sum.
 
 Each letter must represent a different digit, and the leading digit of a multi-digit number must not be zero.
-
-Write a function to solve alphametics puzzles.
 
 [alphametics]: https://en.wikipedia.org/wiki/Alphametics
--- a/exercises/alphametics/description.md
+++ b/exercises/alphametics/description.md
@@ -1,6 +1,6 @@
 # Description
 
-Write a solver for alphametics puzzles.
+Write some code to solve alphametics puzzles.
 
 [Alphametics][alphametics] is a puzzle where letters in words are replaced with numbers.
 
--- a/exercises/alphametics/description.md
+++ b/exercises/alphametics/description.md
@@ -1,6 +1,6 @@
 # Description
 
-Write some code to solve alphametics puzzles.
+Given an alphametics puzzle, find the correct solution.
 
 [Alphametics][alphametics] is a puzzle where letters in words are replaced with numbers.
 
--- a/exercises/alphametics/metadata.toml
+++ b/exercises/alphametics/metadata.toml
@@ -1,2 +1,2 @@
 title = "Alphametics"
-blurb = "Write a function to solve alphametics puzzles."
+blurb = "Write a solver for alphametics puzzles."
--- a/exercises/alphametics/metadata.toml
+++ b/exercises/alphametics/metadata.toml
@@ -1,2 +1,2 @@
 title = "Alphametics"
-blurb = "Write a solver for alphametics puzzles."
+blurb = "Write some code to solve alphametics puzzles."
--- a/exercises/alphametics/metadata.toml
+++ b/exercises/alphametics/metadata.toml
@@ -1,2 +1,2 @@
 title = "Alphametics"
-blurb = "Write some code to solve alphametics puzzles."
+blurb = "Given an alphametics puzzle, find the correct solution."
--- a/exercises/darts/description.md
+++ b/exercises/darts/description.md
@@ -16,7 +16,7 @@ In our particular instance of the game, the target rewards 4 different amounts o
 The outer circle has a radius of 10 units (this is equivalent to the total radius for the entire target), the middle circle a radius of 5 units, and the inner circle a radius of 1.
 Of course, they are all centered at the same point — that is, the circles are [concentric][] defined by the coordinates (0, 0).
 
-Given a point in the target (defined by its [Cartesian coordinates][cartesian-coordinates] `x` and `y`, where `x` and `y` are [real][real-numbers]), calculate the correct score earned by a dart landing there.
+Given a point in the target (defined by its [Cartesian coordinates][cartesian-coordinates] `x` and `y`, where `x` and `y` are [real][real-numbers]), calculate the correct score earned by a dart landing at that point.
 
 ## Credit
 
--- a/exercises/darts/description.md
+++ b/exercises/darts/description.md
@@ -1,6 +1,6 @@
 # Description
 
-Write a function that returns the earned points in a single toss of a Darts game.
+Calculate the points scored in a single toss of a Darts game.
 
 [Darts][darts] is a game where players throw darts at a [target][darts-target].
 
@@ -16,7 +16,7 @@ In our particular instance of the game, the target rewards 4 different amounts o
 The outer circle has a radius of 10 units (this is equivalent to the total radius for the entire target), the middle circle a radius of 5 units, and the inner circle a radius of 1.
 Of course, they are all centered at the same point — that is, the circles are [concentric][] defined by the coordinates (0, 0).
 
-Write a function that given a point in the target (defined by its [Cartesian coordinates][cartesian-coordinates] `x` and `y`, where `x` and `y` are [real][real-numbers]), returns the correct amount earned by a dart landing at that point.
+Given a point in the target (defined by its [Cartesian coordinates][cartesian-coordinates] `x` and `y`, where `x` and `y` are [real][real-numbers]), calculate the correct score earned by a dart landing there.
 
 ## Credit
 
--- a/exercises/flatten-array/description.md
+++ b/exercises/flatten-array/description.md
@@ -2,7 +2,7 @@
 
 Take a nested list and return a single flattened list with all values except nil/null.
 
-The challenge is to write a function that accepts an arbitrarily-deep nested list-like structure and returns a flattened structure without any nil/null values.
+The challenge is to take an arbitrarily-deep nested list-like structure and return a flattened structure without any nil/null values.
 
 For example:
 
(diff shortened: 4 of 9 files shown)


exercises/alphametics/description.md
```diff
@@ -1,6 +1,6 @@
 # Description
 
-Write a function to solve alphametics puzzles.
+Write a solver for alphametics puzzles.
```
IsaacG, line 3: What is a "solver"? Should this just ask the student to solve the puzzle? Write code to solve the puzzle?

exercises/darts/description.md
```diff
@@ -16,7 +16,7 @@ In our particular instance of the game, the target rewards 4 different amounts o
 The outer circle has a radius of 10 units (this is equivalent to the total radius for the entire target), the middle circle a radius of 5 units, and the inner circle a radius of 1.
 Of course, they are all centered at the same point — that is, the circles are [concentric][] defined by the coordinates (0, 0).
 
-Write a function that given a point in the target (defined by its [Cartesian coordinates][cartesian-coordinates] `x` and `y`, where `x` and `y` are [real][real-numbers]), returns the correct amount earned by a dart landing at that point.
+Given a point in the target (defined by its [Cartesian coordinates][cartesian-coordinates] `x` and `y`, where `x` and `y` are [real][real-numbers]), calculate the correct score earned by a dart landing there.
```
IsaacG, line 19: Avoid words like "there" which are imprecise references.

exercises/flatten-array/description.md
```diff
@@ -2,7 +2,7 @@
 
 Take a nested list and return a single flattened list with all values except nil/null.
 
-The challenge is to write a function that accepts an arbitrarily-deep nested list-like structure and returns a flattened structure without any nil/null values.
+The challenge is to take an arbitrarily-deep nested list-like structure and return a flattened structure without any nil/null values.
```
IsaacG, line 5: "Return" is function-specific terminology.

exercises/go-counting/description.md
```diff
@@ -5,10 +5,10 @@ Count the scored points on a Go board.
 In the game of go (also known as baduk, igo, cờ vây and wéiqí) points are gained by completely encircling empty intersections with your stones.
 The encircled intersections of a player are known as its territory.
 
-Write a function that determines the territory of each player.
+Your task is to calculate the territory of each player.
```
IsaacG, line 8:
```suggestion
Calculate the territory of each player.
```

exercises/go-counting/description.md
```diff
@@ -5,10 +5,10 @@ Count the scored points on a Go board.
...
 You may assume that any stones that have been stranded in enemy territory have already been taken off the board.
 
-Write a function that determines the territory which includes a specified coordinate.
+Given a board, determine the territory which includes a specified coordinate.
```
IsaacG, line 11:
```suggestion
Determine the territory which includes a specified coordinate.
```

exercises/nucleotide-codons/description.md
```diff
@@ -1,6 +1,6 @@
 # Description
 
-Write a function that returns the name of an amino acid a particular codon, possibly using shorthand, encodes for.
+Return the name of an amino acid encoded by a particular codon, possibly using shorthand.
```
IsaacG, line 3: "Return" is function specific terminology.

IsaacG (changes requested):

adamazing:
Thanks for your feedback @IsaacG! I've applied your specific suggestions and made changes based on the feedback about use of the word `return` and preferring specific/direct language to vague. Lmk if there's anything that's still not quite right! 🙇

MatthijsBlom:
For what it's worth:

> "Return" is function specific terminology.

I partly disagree. Rather, 'return' has a function-specific meaning _in contexts that includes functions_. In contexts that do not include functions, such as the programming language tracks that this PR aims to serve, it does not have this specific meaning.

Words do not inherently have meanings. Instead, their meanings always depend on context. Case in point: [berries](https://blog.plover.com/lang/jargon-failures.html). Another example: what was a _compiler_ prior to 1900, what did it mean to the person who coined the term in computer context, and what does it mean to most programmers today (and which word took on the previous meaning)?

I do not think it makes sense to navigate around using 'return' when it makes the overall sentence worse.

exercises/alphametics/metadata.toml
```diff
@@ -1,2 +1,2 @@
 title = "Alphametics"
-blurb = "Write a function to solve alphametics puzzles."
+blurb = "Write a solver for alphametics puzzles."
```
MatthijsBlom, line 2: It seems this use of 'solver' was overlooked.

( I do not mind it myself. )

exercises/nucleotide-codons/description.md
```diff
@@ -1,6 +1,6 @@
 # Description
 
-Write a function that returns the name of an amino acid a particular codon, possibly using shorthand, encodes for.
+Write some code which will output the name of an amino acid encoded by a particular codon, possibly using shorthand.
```
MatthijsBlom, line 3: Perhaps

```suggestion
Produce the name of an amino acid encoded by a particular codon, possibly using shorthand.
```

or, more radically,

```suggestion
Translate a given codon into the name of the amino acid that it encodes, possibly using shorthand.
```

exercises/nucleotide-codons/metadata.toml
```diff
@@ -1,2 +1,2 @@
 title = "Nucleotide Codons"
-blurb = "Write a function that returns the name of an amino acid a particular codon, possibly using shorthand, encodes for."
+blurb = "Return the name of an amino acid encoded by a particular codon, possibly using shorthand."
```
MatthijsBlom, line 2: It looks like this blurb was overlooked as well.

( Not that I mind &ndash; I actually prefer this one &ndash; but others might. )

exercises/parallel-letter-frequency/description.md
```diff
@@ -4,4 +4,4 @@ Count the frequency of letters in texts using parallel computation.
 
 Parallelism is about doing things in parallel that can also be done sequentially.
 A common example is counting the frequency of letters.
-Create a function that returns the total frequency of each letter in a list of texts and that employs parallelism.
+Employ parallelism to return the total frequency of each letter in a list of texts.
```
MatthijsBlom, line 7:
```suggestion
Employ parallelism to calculate the total frequency of each letter in a list of texts.
```

Not because I dislike 'return', but because I feel 'calculate' fits better.

IsaacG:
> I partly disagree. Rather, 'return' has a function-specific meaning _in contexts that include functions_. In contexts that do not include functions, such as the programming language tracks that this PR aims to serve, it does not have this specific meaning.
> ...
> I do not think it makes sense to navigate around using 'return' when it makes the overall sentence worse.

If the goal of this issue/PR is to have the text "work" for all tracks, "return" doesn't work well for tracks like awk or bash that asks you to print a value to STDOUT. awk and  bash _do_ have functions but the test runners operate by executing a process and capturing STDOUT. Without getting into semantic definitions, I don't think changing "function" to "return" makes the text apply better. The sqlite tests require writing code that updates a column in a table. "Return" isn't a great word to use there, either.

exercises/parallel-letter-frequency/description.md
IsaacG: There is an ongoing discussion about this exercise and whether solutions must use parallelism vs concurrency vs allowing non interleaved solutions. It might be worth holding off on any changes here for a bit to reduce churn. On the other hand, this said parallelism before and that word isn't changing.

exercises/nucleotide-codons/description.md
adamazing: I'm going to Bravely Fix it to be like your second suggestion. That reworking had occurred to me too but I was being cautious and trying to use as light a touch as possible. 😅

exercises/parallel-letter-frequency/description.md
adamazing: @IsaacG, I made a call to keep the change with @MatthijsBlom's `s/return/calculate/` suggestion since -- as you say -- the word parallelism is there as it was before (and still in the exercise name/slug 😅) but LMK if you feel strongly that it should be removed from this PR. 

My feeling is that this way, the function-specific language issue is resolved, whereas it may be forgotten/overlooked amid the parallelism/concurrency debate? 

Thanks for your time and your feedback 🙇

exercises/nucleotide-codons/description.md
```diff
@@ -1,6 +1,6 @@
 # Description
 
-Write a function that returns the name of an amino acid a particular codon, possibly using shorthand, encodes for.
+Translate a given codon into the name of the amino acid that it encodes, possibly using shorthand.
```
IsaacG, line 3: If we're rewording this, the exercise asks you to translate a sequence of codons into a sequence of amino acids. There's no shorthands in any of the tracks that I've seen.

exercises/alphametics/description.md
```diff
@@ -1,6 +1,6 @@
 # Description
 
-Write a function to solve alphametics puzzles.
+Write some code to solve alphametics puzzles.
```
IsaacG, line 3: I'm not a big fan of "some code". Can we reword/rewrite this sentence?

exercises/nucleotide-codons/description.md
adamazing: I think that's explained in the description. DNA has the 4 bases, Adenine, Cytine, Thymine, Guanine. The shorthand it talks about is to do with the fact that multiple codons can map to the same amino acid, e.g. AAA and AAG both map to Lysine. So the shorthand they came up with is AAR (R means A|G). I think the "may use shorthand" language is due to the fact that the input might be "AAA", "AAG", or "AAR" and the expected output would be "Lysine"? But I haven't done the exercise yet.

adamazing: Sorry, wrote that from phone while horizontal and in pain 😅. Had a quick chance to search, and -- because the exercise is deprecated? -- I've only found it (locked) on the Rust track. Because of that and the fact that there's no canonical-data.json file I can't confirm my assumption above is correct. 

Since the exercise is deprecated, is it worth further effort to refine the text?

ErikSchierboom: > Since the exercise is deprecated, is it worth further effort to refine the text?

Not really

ErikSchierboom (approved):

IsaacG (approved):

kotp (approved):

PR #2408, opened 2024-03-21
Update two-fer instructions.md

iHiD:
See https://forum.exercism.org/t/clarify-two-fer-instructions/10411

### DIFF
--- a/exercises/two-fer/introduction.md
+++ b/exercises/two-fer/introduction.md
@@ -5,4 +5,4 @@ Two-for-one is a way of saying that if you buy one, you also get one for free.
 So the phrase "two-fer" often implies a two-for-one offer.
 
 Imagine a bakery that has a holiday offer where you can buy two cookies for the price of one ("two-fer one!").
-You go for the offer and (very generously) decide to give the extra cookie to a friend.
+You take the offer and (very generously) decide to give the extra cookie to someone else in the queue.
--- a/exercises/two-fer/instructions.md
+++ b/exercises/two-fer/instructions.md
@@ -2,14 +2,13 @@
 
 Your task is to determine what you will say as you give away the extra cookie.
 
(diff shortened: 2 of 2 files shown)


BNAndras (approved):

glennj (approved):

kotp (commented):
This still implies that we need to have a "knowledge check".  The knowledge being requested before was knowing if they like cookies, the new information is finding out if we know the persons name.

Prior, this was addressed by the simple language of "if no name is given".  So the knowledge was built in to the action of giving or not giving a name.

exercises/two-fer/introduction.md
```diff
@@ -5,4 +5,4 @@ Two-for-one is a way of saying that if you buy one, you also get one for free.
 So the phrase "two-fer" often implies a two-for-one offer.
 
 Imagine a bakery that has a holiday offer where you can buy two cookies for the price of one ("two-fer one!").
-You go for the offer and (very generously) decide to give the extra cookie to a friend.
+You take the offer and (very generously) decide to give the extra cookie to someone else in the queue.
```
IsaacG, line 8: Should this be, "You accept the offer"?

IsaacG (approved):

exercises/two-fer/introduction.md
kotp: It could be, colloquially it means the same thing.

IsaacG: Given that we have people for whom English is not a first language, should we try to avoid colloquialisms?

kotp: Colloquially does not necessarily mean hard to understand, it only means informal, and we use informal language here.  It is conversational.  If it were slang, then it would be less appropriate.

We do not currently avoid colloquialisms.

ErikSchierboom (approved):

PR #2413, opened 2024-03-27
Refine wording in Saddle Points

MatthijsBlom:
As per @iHiD's request [here](http://forum.exercism.org/t/dubious-wording-in-saddle-points-exercise/6877/7).

### DIFF
--- a/exercises/saddle-points/instructions.md
+++ b/exercises/saddle-points/instructions.md
@@ -5,7 +5,7 @@ Your task is to find the potential trees where you could build your tree house.
 The data company provides the data as grids that show the heights of the trees.
 The rows of the grid represent the east-west direction, and the columns represent the north-south direction.
 
-An acceptable tree will be the largest in its row, while being the smallest in its column.
+An acceptable tree will have the greatest height in its row, while having the least height in its column.
 
 A grid might not have any good trees at all.
 Or it might have one, or even several.
--- a/exercises/saddle-points/introduction.md
+++ b/exercises/saddle-points/introduction.md
@@ -7,5 +7,5 @@ You need to analyze each grid on the map to find good trees for your tree house.
 
 A good tree is both:
 
-- taller than every tree to the east and west, so that you have the best possible view of the sunrises and sunsets.
-- shorter than every tree to the north and south, to minimize the amount of tree climbing.
+- no shorter than any tree to the east and west, so that you have the best possible view of the sunrises and sunsets.
+- no taller than any tree to the north and south, to minimize the amount of tree climbing.


exercises/saddle-points/instructions.md
```diff
@@ -5,7 +5,7 @@ Your task is to find the potential trees where you could build your tree house.
 The data company provides the data as grids that show the heights of the trees.
 The rows of the grid represent the east-west direction, and the columns represent the north-south direction.
 
-An acceptable tree will be the largest in its row, while being the smallest in its column.
+An acceptable tree will have the greatest height in its row, while having the least height in its column.
```
MatthijsBlom, line 8: I agree with @jordancurve that there is a risk here of the very same confusion that we are trying to avoid.

exercises/saddle-points/introduction.md
```diff
@@ -7,5 +7,5 @@ You need to analyze each grid on the map to find good trees for your tree house.
 
 A good tree is both:
 
-- taller than every tree to the east and west, so that you have the best possible view of the sunrises and sunsets.
-- shorter than every tree to the north and south, to minimize the amount of tree climbing.
+- no shorter than any tree to the east and west, so that you have the best possible view of the sunrises and sunsets.
+- no taller than any tree to the north and south, to minimize the amount of tree climbing.
```
ErikSchierboom, line 11: I'm not sure this is an improvement to be honest. Negatives are harder to read than positives in general.

MatthijsBlom: I agree. Sadly though there is no single word synonym of «no shorter». The previous «larger» has confused people by being false.

ErikSchierboom: Was that because it could also be of the same length?

MatthijsBlom: Yes.

«At least as tall as all» might be the next most succinct.

ErikSchierboom: I like that better!

IsaacG: Does it need to be succinct?

"The tree must be equal in height to the tallest tree to the north and south. No other trees can be taller to the north and south. This minimizes climbing."

ErikSchierboom: It just needs to be clear, so I'm happy to have a less succinct description.

> The tree must be equal in height to the tallest tree to the north and south

To me this reads as if it couldn't be taller.

MatthijsBlom: «the tallest tree» also suggests (again!) that there is only one.

ErikSchierboom: What about:

"The tree must either be:
- taller than the other trees to the north and south, or
- as tall as the tallest tree to the north and south.
 
No other trees can be taller to the north and south. This minimizes climbing."

Not sure about the formatting, but I'd like the fact that we're explicit about the two variants.

MatthijsBlom: The «at least as tall as all» variant suggested earlier:

```suggestion
- at least as tall as all trees to the east and west, so that you have the best possible view of the sunrises and sunsets.
- at most as tall as all tree to the north and south, to minimize the amount of tree climbing.
```

The more verbose variant suggested most recently:

```suggestion
- either
  - taller than the other trees to the east and west, or
  - as tall as the tallest trees to the east and west.

  No other trees can be taller to the east and west.
  This produces the best possible view of the sunrises and sunsets.
- either
  - shorter than the other trees to the north and south, or
  - as tall as the tallest trees to the north and south.

  No other trees can be shorter to the north and south.
  This minimizes climbing.
```

A note about this last one: it tripped up both Erik (the original is wrong) and me (I was insecure correcting and re-correcting it). The above is only a naive rendering; there might well be a better one.

---

For the moment I prefer one of «no shorter than any» and «at least as tall as all». Comparison:

- Pro «no shorter than any»: people like to think in terms of violated rules. This formulation focuses on how a tree might violate the requirements: by being too short.
- Contra «no shorter than any»: it contains a negation.
- Pro «at least as tall as all»: it does not contain a negation.
- Contra «at least as tall as all»: less immediately clear how this requirement could be violated.

ErikSchierboom (approved):

exercises/saddle-points/instructions.md
kotp: Isn't the largest or greatest height "tallest tree" and the least height being the "shortest tree", given the problem we are solving?

kotp, line 8:
```suggestion
An acceptable tree will the tallestt in its row, while being the shortest  n its column.
```

kotp (commented):
I provided a "commit suggestion" block to address the tallest and shortest tree.

exercises/saddle-points/instructions.md
MatthijsBlom: The problem is that «the tallest tree» might not be unique. People have been frustrated by test cases with several tallest/shortest trees per row/column, because based on the instructions they expected there to be only one.

ErikSchierboom: @kotp "will the tallest" should that be "will be the tallest"?

kotp: The tests should enforce the detail though, right?  So any ambiguity is clarified by expressed expectation, one would think.

IsaacG: The whole point of this PR is to clarify this point. It is okay fort the prose to omit a detail and for the test to enforce that detail. It is less okay for the prose to imply one thing and the test to enforce something that contradicts that thing.

kotp: The tests do seem to have covered multiple "shortest tree" at least in the single column test.  Or to express it in the way the test explains it "saddle points" meaning there can be more than one.
I wholly agree that the prose should not contradict, while omission is OK (not great, necessarily, but sometimes purposefully done).

I think that the detail that there may be more than one saddle point is expressed in the tests even by name.

We may see that here:

```json
    {
      "uuid": "ee99ccd2-a1f1-4283-ad39-f8c70f0cf594",
      "description": "Can identify that saddle points in a single column matrix are those with the minimum value",
      "property": "saddlePoints",
      "input": {
        "matrix": [
          [2],
          [1],
          [4],
          [1]
        ]
      },
      "expected": [
        {
          "row": 2,
          "column": 1
        },
        {
          "row": 4,
          "column": 1
        }
      ]
    },
    {
      "uuid": "63abf709-a84b-407f-a1b3-456638689713",
      "description": "Can identify that saddle points in a single row matrix are those with the maximum value",
      "property": "saddlePoints",
      "input": {
        "matrix": [
          [2, 5, 3, 5]
        ]
      },
      "expected": [
        {
          "row": 1,
          "column": 2
        },
        {
          "row": 1,
          "column": 4
        }
```

ErikSchierboom: @MatthijsBlom What about the above suggestion?

ErikSchierboom: Couldn't we use the phrasing from the introduction?

MatthijsBlom: We could. I figured the difference between `introduction.md` and `instructions.md` might be deliberate, so I kept them different.

ErikSchierboom: I'm fine with unifying them, especially when it turns out that the terminology is so hard to get right.

MatthijsBlom: If we use the same wording in both cases, we could also de-duplicate.

One potential advantage of using two different descriptions: these provide different perspectives on the same problem.

exercises/saddle-points/introduction.md
```diff
@@ -7,5 +7,5 @@ You need to analyze each grid on the map to find good trees for your tree house.
 
 A good tree is both:
 
-- taller than every tree to the east and west, so that you have the best possible view of the sunrises and sunsets.
-- shorter than every tree to the north and south, to minimize the amount of tree climbing.
+- at least as tall as all trees to the east and west, so that you have the best possible view of the sunrises and sunsets.
+- at most as tall as all tree to the north and south, to minimize the amount of tree climbing.
```
ErikSchierboom, line 11:
```suggestion
- at most as tall as all trees to the north and south, to minimize the amount of tree climbing.
```

exercises/saddle-points/instructions.md
MatthijsBlom: I did not agree with applying this suggestion, but it was applied anyway.

The point of my change was to fix an inaccuracy. This now applied suggestion effectively reverts my change and _reintroduces_ the original inaccuracy. This inaccuracy has previously irritated people, hence this PR. See the forum thread, and the other threads here.

ErikSchierboom (approved):

ErikSchierboom (commented):
What I think might help is to have an additional example, besides the

> Here is a grid that has exactly one candidate tree.

exercises/saddle-points/introduction.md
```diff
@@ -7,5 +7,5 @@ You need to analyze each grid on the map to find good trees for your tree house.
 
 A good tree is both:
 
-- taller than every tree to the east and west, so that you have the best possible view of the sunrises and sunsets.
-- shorter than every tree to the north and south, to minimize the amount of tree climbing.
+- at least as tall as all trees to the east and west, so that you have the best possible view of the sunrises and sunsets.
+- at most as tall as all trees to the north and south, to minimize the amount of tree climbing.
```
IsaacG, line 11:
```suggestion
- at least as tall as all the trees to the east and west, so that you have the best possible view of the sunrises and sunsets.
- at most as tall as all the trees to the north and south, to minimize the amount of tree climbing.
```

PR #2436, opened 2024-05-09
custom-set: add test for removing duplicates

senekor:


senekor:
[forum post](https://forum.exercism.org/t/custom-set-new-test-against-deduplication/11130)

IsaacG (approved):

BNAndras (approved):

exercises/custom-set/canonical-data.json
```diff
@@ -396,6 +396,16 @@
...
+        },
+        {
+          "uuid": "20d0a38f-7bb7-4c4a-ac15-90c7392ecf2b",
+          "description": "difference removes all duplicates in the first set",
+          "property": "difference",
+          "input": {
+            "set1": [1, 1],
+            "set2": [1]
+          },
+          "expected": [0]
```
glennj, line 408: Why is the expected set `[0]`? Where does element `0` come from? Shouldn't I expect an empty set as the difference?

ErikSchierboom: Yeah, I think that might be a mistake?

senekor: Yes, that was a mistake. thanks!

ErikSchierboom (approved):

PR #2448, opened 2024-05-25
Queen Attack - Fix test names and errors for non-negative position requirements

ryanplusplus:
The test names and error strings for negative rows and columns state that row/columns must be positive, but 0 is a valid value and 0 is not positive. This change adjusts the test names and error strings so that they correctly describe the requirement.


ryanplusplus:
Should I deprecate the old cases and replace them with new cases?

PR #2465, opened 2024-07-31
Update description for Luhn algorithm

masters3d:
The use of `_` was ambiguous to me. Attempting to make this clearer.


masters3d:
Updated the branch but since its closed it doesn't update the PR. 
Here is the discussion: https://forum.exercism.org/t/update-description-for-luhn-algorithm/12334

PR #2466, opened 2024-08-02
[Luhn algorithm] Replace _ with arrows

masters3d:
[Luhn algorithm] Replace _ with arrows

### DIFF
--- a/exercises/luhn/description.md
+++ b/exercises/luhn/description.md
@@ -23,7 +23,7 @@ We will be doubling
 
 ```text
 4539 3195 0343 6467
-↑ ↑  ↑ ↑  ↑ ↑  ↑ ↑  (doubling)
+↑ ↑  ↑ ↑  ↑ ↑  ↑ ↑  (double these)
 ```
 
 If doubling the number results in a number greater than 9 then subtract 9 from the product.
--- a/exercises/luhn/description.md
+++ b/exercises/luhn/description.md
@@ -22,7 +22,8 @@ The first step of the Luhn algorithm is to double every second digit, starting f
 We will be doubling
 
 ```text
-4_3_ 3_9_ 0_4_ 6_6_
+4539 3195 0343 6467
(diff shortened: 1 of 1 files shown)


BethanyG (approved):

exercises/luhn/description.md
```diff
@@ -22,7 +22,8 @@ The first step of the Luhn algorithm is to double every second digit, starting f
 We will be doubling
 
 ```text
-4_3_ 3_9_ 0_4_ 6_6_
+4539 3195 0343 6467
```
IsaacG, line 25: Could you please implement @kotp 's suggest of using an odd number of digits?

IsaacG:
Forum discussion [here](https://forum.exercism.org/t/update-description-for-luhn-algorithm/12334)

exercises/luhn/description.md
iHiD: I think @kotp's suggestion makes it less clear. The bit everyone misses is "starting from the right" not "every second digit". If we make it odd number of digits, then we double the second one from the left and from the right, which I think makes it confusing again.

IsaacG (approved):

exercises/luhn/description.md
kotp: It was exactly what I was indicating with the "starting from the right" as that is what I see people miss as well.

So having a second example that shows this, with an odd number of digits, makes it apparent that it is not "start from the right".

It shows them side by side (or vertically) together, and visually shows the difference.

I agree, if we make it odd number of digits, and that alone, it would definitely be confusing, but along with the example of the even number of digits, this helps (to me) clarify visually, that there is something different going on on the left hand side.

kotp: Well, it is already in.  So it does not matter, if this matters it will eventually come in.  So  do not sweat it.  I did not recognize that it was already merged and so the point is moot.

PR #2469, opened 2024-08-13
Clarify Series errors in README

iHiD:
Not set on the wording, but putting this here so it can be riffed on.

Forum discussion: http://forum.exercism.org/t/series-problem-description-unclear/12477

### DIFF
--- a/exercises/series/description.md
+++ b/exercises/series/description.md
@@ -13,7 +13,7 @@ And the following 4-digit series:
 - "4914"
 - "9142"
 
-And if you ask for a 6-digit series from a 5-digit string, you deserve whatever you get.
+Trying to extract a 6-digit series from a 5-digit string should result in an error.
 
 Note that these series are only required to occupy _adjacent positions_ in the input;
 the digits need not be _numerically consecutive_.


ErikSchierboom (approved):

glennj:
Perhaps
```
The series length `n` cannot be longer than the string of digits.
Trying to extract a 6-digit series from a 5-digit string should result in an error.
```

exercises/series/description.md
```diff
@@ -13,7 +13,7 @@ And the following 4-digit series:
 - "4914"
 - "9142"
 
-And if you ask for a 6-digit series from a 5-digit string, you deserve whatever you get.
+Trying to extract a 6-digit series from a 5-digit string should result in an error.
```
IsaacG, line 16: Is "should result in an error" something that should be track specific behavior?

tasxatzial:
Old PR. Should we move it forward?

Simply by looking at the canonical data, it appears that there are even more cases where an error is expected:

* slice length cannot be zero
* slice length cannot be negative
* empty series is invalid

So, it's best to either mention all cases explicitly or remove any reference to errors from the description altogether, which seems more reasonable to me.

PR #2474, opened 2024-08-24
New test - Update canonical-data.json

karginoleg:
Add test to check if two unordered lists are separated properly


IsaacG:
Ongoing forum discussion: https://forum.exercism.org/t/markdown-add-new-test/12652/7

Please do not reopen PR until there is maintainer agreement that we want to proceed here.

PR #2477, opened 2024-09-20
Forth: additional tests for "combined arithmetic" category

glennj:
This will ensure that students are popping numbers off the correct end of the stack (for languages without a builtin Stack type).

From: http://forum.exercism.org/t/new-tests-for-forth-exercise-canonical-data/13045

### DIFF
--- a/exercises/forth/canonical-data.json
+++ b/exercises/forth/canonical-data.json
@@ -215,7 +215,7 @@
           "expected": [2]
         },
         {
-          "uuid": "",
+          "uuid": "f749b540-53aa-458e-87ec-a70797eddbcb",
           "description": "multiplication and addition",
           "property": "evaluate",
           "input": {
@@ -224,7 +224,7 @@
           "expected": [13]
         },
         {
-          "uuid": "",
+          "uuid": "c8e5a4c2-f9bf-4805-9a35-3c3314e4989a",
           "description": "addition and multiplication",
           "property": "evaluate",
           "input": {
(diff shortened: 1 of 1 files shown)


exercises/forth/canonical-data.json
```diff
@@ -213,6 +213,24 @@
...
           "expected": [2]
+        },
+        {
+          "uuid": "f749b540-53aa-458e-87ec-a70797eddbcb",
+          "description": "multiplication and addition",
+          "property": "evaluate",
+          "input": {
+            "instructions": ["1 3 4 * +"]
+          },
+          "expected": [13]
```
IsaacG, line 224: Would it make sense to simplify this and only do one operation?
`["1 3 4 *] => [1, 12]`

glennj: Yes. In that case, I should do one for each of `+`, `-`, `*` and `/`. These tests are in the "combined arithmetic". I'll rework this.

IsaacG: Do the combined arithmetic tests add anything that isn't already captured by prior tests? If we test all the operators when there are both 2 and 3 values on the stack, the combined tests shouldn't be adding any additional coverage, I think?

glennj: I think they are enhancing the "combined arithmetic" tests, in that it's OK to load the stack with numbers first and then perform multiple operations to consume/transform the stack.

IsaacG (approved):

glennj:
Added some reviewers to get some traction on the request.

ErikSchierboom (approved):

BNAndras (approved):

BethanyG (approved):

glennj:
rebased and force pushed to pull in the updates to link checker.

PR #2495, opened 2024-11-05
Update square-root description.md

xanni:
Moving generally applicable information from the Python specific addendum as per https://forum.exercism.org/t/moving-generic-instructions-from-python-addendum-to-overall-problem-specification/13678/1 and https://forum.exercism.org/t/square-root-exercise-instructions-could-be-clearer/13664/11

See also https://github.com/exercism/python/pull/3815/files

### DIFF
--- a/exercises/square-root/description.md
+++ b/exercises/square-root/description.md
@@ -9,5 +9,7 @@ Check out the Wikipedia pages on [square root][square-root] and [methods of comp
 
 Recall also that natural numbers are positive real whole numbers (i.e. 1, 2, 3 and up).
 
+It is possible to compute the square root of any natural number using only natural numbers in the computation.
+
 [square-root]: https://en.wikipedia.org/wiki/Square_root
 [computing-square-roots]: https://en.wikipedia.org/wiki/Methods_of_computing_square_roots
--- a/exercises/square-root/description.md
+++ b/exercises/square-root/description.md
@@ -1,17 +1,24 @@
 # Description
 
-Given a natural radicand, return its square root.
+We are launching a deep space exploration rocket and we need a way to make sure the navigation system stays on target.
 
-Note that the term "radicand" refers to the number for which the root is to be determined.
-That is, it is the number under the root symbol.
+The first step in our coordinate calculation involves taking a number as input and finding another number which multiplied by itself will equal our input number, or in other words, finding the square root of our input number.
 
-Check out the Wikipedia pages on [square root][square-root] and [methods of computing square roots][computing-square-roots].
+Because the journey will be very long, we had to make our rocket’s onboard computer very power efficient so it wouldn’t drain the batteries. Unfortunately that means that we can’t rely on fancy math libraries and functions, as they would take more power. Instead we want to calculate or approximate the square root directly.
 
-Recall also that natural numbers are positive real whole numbers (i.e. 1, 2, 3 and up).
+Your task is to implement a function that calculates or approximates the square root of a given number.
 
-Avoid using built-in functions or libraries that directly compute the square root.
-Instead, create your own approach to approximate or calculate the square root.
-It is possible to calculate the square root of any natural number using only natural numbers in the calculation.
+* Try to avoid using the pre-existing math libraries of your language.
+* As input you’ll be given a positive whole number, i.e. 1, 2, 3, 4…
+* The function also only needs to output positive whole numbers.
 
-[square-root]: https://en.wikipedia.org/wiki/Square_root
+Examples of some approaches you could consider are:
+
+* Linear or binary search for a number that gives the input number when squared
+* Successive approximation using Newton's or Heron's method
+* Calculating one digit at a time or one bit at a time
+
+You can check out the Wikipedia pages on [integer square root][integer-square-root] and [methods of computing square roots][computing-square-roots] to help with choosing a method of calculation.
+
+[integer-square-root]: https://en.wikipedia.org/wiki/Integer_square_root
 [computing-square-roots]: https://en.wikipedia.org/wiki/Methods_of_computing_square_roots
--- a/exercises/square-root/description.md
+++ b/exercises/square-root/description.md
@@ -9,7 +9,9 @@ Check out the Wikipedia pages on [square root][square-root] and [methods of comp
 
 Recall also that natural numbers are positive real whole numbers (i.e. 1, 2, 3 and up).
 
-It is possible to compute the square root of any natural number using only natural numbers in the computation.
+Avoid using built-in functions or libraries that directly compute the square root.
+Instead, create your own approach to approximate or calculate the square root.
+It is possible to calculate the square root of any natural number using only natural numbers in the calculation.
 
 [square-root]: https://en.wikipedia.org/wiki/Square_root
 [computing-square-roots]: https://en.wikipedia.org/wiki/Methods_of_computing_square_roots
--- a/exercises/square-root/description.md
+++ b/exercises/square-root/description.md
@@ -4,13 +4,15 @@ We are launching a deep space exploration rocket and we need a way to make sure 
 
 The first step in our coordinate calculation involves taking a number as input and finding another number which multiplied by itself will equal our input number, or in other words, finding the square root of our input number.
 
-Because the journey will be very long, we had to make our rocket’s onboard computer very power efficient so it wouldn’t drain the batteries. Unfortunately that means that we can’t rely on fancy math libraries and functions, as they would take more power. Instead we want to calculate or approximate the square root directly.
+Because the journey will be very long, we had to make our rocket’s onboard computer very power efficient so it wouldn’t drain the batteries.
+Unfortunately that means that we can’t rely on fancy math libraries and functions, as they would take more power.
+Instead we want to calculate or approximate the square root directly.
 
-Your task is to implement a function that calculates or approximates the square root of a given number.
+Your task is to implement a calculation of the square root of a given number.
 
 - Try to avoid using the pre-existing math libraries of your language.
 - As input you’ll be given a positive whole number, i.e. 1, 2, 3, 4…
-- The function also only needs to output positive whole numbers.
+- You also only need to output a positive whole number.
 
 Examples of some approaches you could consider are:
 
--- a/exercises/square-root/description.md
+++ b/exercises/square-root/description.md
@@ -8,15 +8,15 @@ Because the journey will be very long, we had to make our rocket’s onboard com
 
 Your task is to implement a function that calculates or approximates the square root of a given number.
 
-* Try to avoid using the pre-existing math libraries of your language.
-* As input you’ll be given a positive whole number, i.e. 1, 2, 3, 4…
-* The function also only needs to output positive whole numbers.
+- Try to avoid using the pre-existing math libraries of your language.
+- As input you’ll be given a positive whole number, i.e. 1, 2, 3, 4…
+- The function also only needs to output positive whole numbers.
 
 Examples of some approaches you could consider are:
 
-* Linear or binary search for a number that gives the input number when squared
-* Successive approximation using Newton's or Heron's method
-* Calculating one digit at a time or one bit at a time
+- Linear or binary search for a number that gives the input number when squared
+- Successive approximation using Newton's or Heron's method
+- Calculating one digit at a time or one bit at a time
 
 You can check out the Wikipedia pages on [integer square root][integer-square-root] and [methods of computing square roots][computing-square-roots] to help with choosing a method of calculation.
 
--- a/exercises/square-root/instructions.md
+++ b/exercises/square-root/instructions.md
@@ -4,7 +4,7 @@ Your task is to calculate the square root of a given number.
 
 - Try to avoid using the pre-existing math libraries of your language.
 - As input you'll be given a positive whole number, i.e. 1, 2, 3, 4…
-- You also only need to return a positive whole number.
+- You are only required to handle cases where the result is a positive whole number.
 
 Some potential approaches:
 
--- a/exercises/square-root/instructions.md
+++ b/exercises/square-root/instructions.md
@@ -4,7 +4,7 @@ Your task is to calculate the square root of a given number.
 
 - Try to avoid using the pre-existing math libraries of your language.
 - As input you’ll be given a positive whole number, i.e. 1, 2, 3, 4…
-- You also only need to output a positive whole number.
+- You also only need to return a positive whole number.
 
 Some potential approaches:
 
--- a/exercises/square-root/instructions.md
+++ b/exercises/square-root/instructions.md
@@ -3,7 +3,7 @@
 Your task is to calculate the square root of a given number.
 
 - Try to avoid using the pre-existing math libraries of your language.
-- As input you’ll be given a positive whole number, i.e. 1, 2, 3, 4…
+- As input you'll be given a positive whole number, i.e. 1, 2, 3, 4…
 - You also only need to return a positive whole number.
 
 Some potential approaches:
--- a/exercises/square-root/instructions.md
+++ b/exercises/square-root/instructions.md
@@ -1,12 +1,4 @@
-# Description
-
-We are launching a deep space exploration rocket and we need a way to make sure the navigation system stays on target.
-
-The first step in our coordinate calculation involves taking a number as input and finding another number which multiplied by itself will equal our input number, or in other words, finding the square root of our input number.
-
-Because the journey will be very long, we had to make our rocket’s onboard computer very power efficient so it wouldn’t drain the batteries.
-Unfortunately that means that we can’t rely on fancy math libraries and functions, as they would take more power.
-Instead we want to calculate or approximate the square root directly.
+# Instructions
 
 Your task is to implement a calculation of the square root of a given number.
 
--- a/exercises/square-root/instructions.md
+++ b/exercises/square-root/instructions.md
@@ -11,3 +11,8 @@ Some potential approaches:
 - Linear or binary search for a number that gives the input number when squared.
 - Successive approximation using Newton's or Heron's method.
 - Calculating one digit at a time or one bit at a time.
+
+You can check out the Wikipedia pages on [integer square root][integer-square-root] and [methods of computing square roots][computing-square-roots] to help with choosing a method of calculation.
+
+[integer-square-root]: https://en.wikipedia.org/wiki/Integer_square_root
+[computing-square-roots]: https://en.wikipedia.org/wiki/Methods_of_computing_square_roots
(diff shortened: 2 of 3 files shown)


exercises/square-root/description.md
IsaacG, line 1: We should split this up into an introduction and instruction file. See here, for example. https://github.com/exercism/problem-specifications/tree/main/exercises/eliuds-eggs

exercises/square-root/description.md
```diff
@@ -1,13 +1,24 @@
...
 
-Given a natural radicand, return its square root.
+We are launching a deep space exploration rocket and we need a way to make sure the navigation system stays on target.
 
-Note that the term "radicand" refers to the number for which the root is to be determined.
-That is, it is the number under the root symbol.
+The first step in our coordinate calculation involves taking a number as input and finding another number which multiplied by itself will equal our input number, or in other words, finding the square root of our input number.
 
-Check out the Wikipedia pages on [square root][square-root] and [methods of computing square roots][computing-square-roots].
+Because the journey will be very long, we had to make our rocket’s onboard computer very power efficient so it wouldn’t drain the batteries. Unfortunately that means that we can’t rely on fancy math libraries and functions, as they would take more power. Instead we want to calculate or approximate the square root directly.
```
IsaacG, line 7: We use one line per sentence, one sentence per line.

exercises/square-root/description.md
```diff
@@ -1,13 +1,24 @@
...
 
-Recall also that natural numbers are positive real whole numbers (i.e. 1, 2, 3 and up).
+Your task is to implement a function that calculates or approximates the square root of a given number.
```
IsaacG, line 9: We avoid the word "function". Not all tracks have functions.

kotp (approved):
I think it looks good.

exercises/square-root/introduction.md
```diff
@@ -0,0 +1,9 @@
+# Introduction
+
+We are launching a deep space exploration rocket and we need a way to make sure the navigation system stays on target.
+
+The first step in our coordinate calculation involves taking a number as input and finding another number which multiplied by itself will equal our input number, or in other words, finding the square root of our input number.
```
IsaacG, line 5: This sentence is hard for me to follow, even as a native English speaker. It might be worth trying to break this into smaller pieces.

exercises/square-root/introduction.md
```diff
@@ -0,0 +1,9 @@
...
+
+Because the journey will be very long, we had to make our rocket’s onboard computer very power efficient so it wouldn’t drain the batteries.
```
IsaacG, line 7:
```suggestion
The journey will be very long.
To avoid draining the battery, ee had to make our rocket’s onboard computer very power efficient.
```

exercises/square-root/introduction.md
```diff
@@ -0,0 +1,9 @@
...
+Unfortunately that means that we can’t rely on fancy math libraries and functions, as they would take more power.
```
IsaacG, line 8:
```suggestion
Unfortunately that means that we can’t rely on fancy math libraries and functions, as they use more power.
```

exercises/square-root/instructions.md
```diff
@@ -0,0 +1,18 @@
+# Instructions
+
+Your task is to implement a calculation of the square root of a given number.
```
IsaacG, line 3:
```suggestion
Your task is to calculate the square root of a given number.
```

Cool-Katt (approved):

exercises/square-root/instructions.md
```diff
@@ -0,0 +1,18 @@
...
+
+- Try to avoid using the pre-existing math libraries of your language.
+- As input you’ll be given a positive whole number, i.e. 1, 2, 3, 4…
+- You also only need to output a positive whole number.
```
IsaacG, line 7:
```suggestion
- You are only required to handle cases where the output is a positive whole number.
```

exercises/square-root/instructions.md
```diff
@@ -0,0 +1,18 @@
...
+
+Examples of some approaches you could consider are:
```
IsaacG, line 9:
```suggestion
Here are some approaches you could use.
```

exercises/square-root/instructions.md
```diff
@@ -0,0 +1,18 @@
...
+
+- Linear or binary search for a number that gives the input number when squared
```
IsaacG, line 11: These sentences should have a period at the end

exercises/square-root/instructions.md
xanni: The suggested change is misleading because it seems to preclude approximating the output to a positive whole number even when it is not the exact result, which most implementations of integer square root will do.  It biases the student to implementations that only handle perfect squares, which overly narrows the exercise.

kotp: I think this is a good suggestion, we can see what the requirements are from the tests.  What is allowed will pass the tests, even (and partly because of) the tests not being exhaustive.

The tests guide what the requirement is, and potentially directly what is not allowed.

The reading can be a bit ambiguous, but may  not contradict.

**What about leaving that line out?**  We can see what the tests require, and the tests are going to communicate, and we also encourage the students to read them.

Also, "output" does not happen in the sense of "to  a screen or file" but what results is a positive whole number.

The tests are also a source of bias, though, we may not be able to avoid that.

exercises/square-root/instructions.md
```diff
@@ -1,16 +1,16 @@
...
 
-Your task is to implement a calculation of the square root of a given number.
+Your task is to calculate the square root of a given number.
 
 - Try to avoid using the pre-existing math libraries of your language.
 - As input you’ll be given a positive whole number, i.e. 1, 2, 3, 4…
 - You also only need to output a positive whole number.
 
-Examples of some approaches you could consider are:
+Here are some approaches you could use.
```
kotp, line 9:
```suggestion
Some potential approaches:
```

Since what follows is a bullet list.

BethanyG (approved):
This look good to me in its current state.

exercises/square-root/introduction.md
```diff
@@ -4,6 +4,7 @@ We are launching a deep space exploration rocket and we need a way to make sure
 
 The first step in our coordinate calculation involves taking a number as input and finding another number which multiplied by itself will equal our input number, or in other words, finding the square root of our input number.
```
kotp, line 5:
```suggestion
As the first step in our calculation, we take a number and find its square root—that is, the number that, when multiplied by itself, equals the given number.
```

exercises/square-root/instructions.md
```diff
@@ -0,0 +1,18 @@
...
+
+Here are some approaches you could use.
+
+- Linear or binary search for a number that gives the input number when squared.
+- Successive approximation using Newton's or Heron's method.
+- Calculating one digit at a time or one bit at a time.
+
+You can check out the Wikipedia pages on [integer square root][integer-square-root] and [methods of computing square roots][computing-square-roots] to help with choosing a method of calculation.
+
+[integer-square-root]: https://en.wikipedia.org/wiki/Integer_square_root
```
tasxatzial, line 17:
```suggestion
```

The integer square root differs from the usual square root, and the existing wiki link already provides enough information. It might lead students down an unintended path and cause extra confusion.

tasxatzial (changes requested):

exercises/square-root/instructions.md
```diff
@@ -0,0 +1,18 @@
...
+- As input you’ll be given a positive whole number, i.e. 1, 2, 3, 4…
+- You also only need to output a positive whole number.
+
+Here are some approaches you could use.
+
+- Linear or binary search for a number that gives the input number when squared.
+- Successive approximation using Newton's or Heron's method.
+- Calculating one digit at a time or one bit at a time.
+
+You can check out the Wikipedia pages on [integer square root][integer-square-root] and [methods of computing square roots][computing-square-roots] to help with choosing a method of calculation.
```
tasxatzial, line 15:
```suggestion
You can check out the Wikipedia pages on [methods of computing square roots][computing-square-roots] to help with choosing a method of calculation.
```

tasxatzial:
I'd like to point out that both wiki links provide code implementations, with the second link offering even more examples. Someone reading these could unintentionally come across the solution.

@maintainers Is this ok?

It also seems that all suggested methods refer to the integer square root wiki, so it should be sufficient to keep only that link.

xanni:
> I'd like to point out that both wiki links provide code implementations, with the second link offering even more examples. Someone reading these could unintentionally come across the solution.
> 
> @maintainers Is this ok? If the intention is to read the wiki pages, then we are basically asking them to read the solution.

My understanding is that (particularly for exercises marked as "easy"!) we are happy for students to implement existing solutions in the language of their track and not necessarily expecting them to develop mathematical algorithms from first principles.

exercises/square-root/introduction.md
```diff
@@ -0,0 +1,10 @@
+# Introduction
+
+We are launching a deep space exploration rocket and we need a way to make sure the navigation system stays on target.
+
+As the first step in our calculation, we take a number and find its square root—that is, the number that, when multiplied by itself, equals the given number.
```
iHiD, line 5:
```suggestion
As the first step in our calculation, we take to add functionality that can take a target number and find its square root — (i.e. the number that, when multiplied by itself, equals the target number).
```

I got a bit lost in the definition of the square root with the use of "number" to mean different numbers. It's correct, but a little hard to parse. So I think labelling the "target number" (better name welcome) makes it clearer. I'd also rather put the definition in brackets as, while I'm very :+1: on having it, I think most people know what a square root is, and it's potentially a little confusing having to read it.

exercises/square-root/introduction.md
```diff
@@ -0,0 +1,10 @@
...
+
+The journey will be very long.
+To make the batteries last as long as possible, we had to make our rocket’s onboard computer very power efficient.
+Unfortunately that means that we can’t rely on fancy math libraries and functions, as they use more power.
+Instead we want to calculate or approximate the square root directly.
```
iHiD, line 10: Are we calculating or approximating? If we're approximating I think we need that to be clearer in the statement above.

exercises/square-root/instructions.md
```diff
@@ -0,0 +1,18 @@
+# Instructions
+
+Your task is to calculate the square root of a given number.
+
+- Try to avoid using the pre-existing math libraries of your language.
+- As input you’ll be given a positive whole number, i.e. 1, 2, 3, 4…
+- You also only need to output a positive whole number.
```
iHiD, line 7: Is the point here that we're getting the nearest whole number? (ie approximating/rounding?)

IsaacG:
> My understanding is that (particularly for exercises marked as "easy"!) we are happy for students to implement existing solutions in the language of their track and not necessarily expecting them to develop mathematical algorithms from first principles.

Note, the problem specs do not have a difficulty rating associated with the exercises. Those are applied at the track level. 

That said, I'm personally not concerned about linking to a Wiki page that may have pseudo code which shows an algorithm.

exercises/square-root/introduction.md
kotp: Is "take to add"  correct?  Seems to be a missing word in there.

exercises/square-root/instructions.md
xanni: I thought the idea was to have students specifically implement integer square root, not the full floating point square root, as a way of narrowing down the scope of the task.  The backstory was chosen partly to support this.

exercises/square-root/instructions.md
xanni: Yes.  This was chosen to narrow the scope of the exercise and hopefully make it both easier and more interesting.  It also opens the possibility of more alternative implementations such as linear and binary search.

exercises/square-root/instructions.md
xanni: Additional reasons why I intentionally want to explicitly require the integer square root in the problem specification:

* Algorithms that compute the non-integer square root run forever on inputs that are not a perfect square.  The integer square root algorithms do not run forever.
* The existing submissions, at least in the Python track, have taken their cues from the tests expecting integer results and the students have typically truncated the results to an integer even when using the built in operators, for example many submissions are of the form int(n^0.5).
* The integer square root has a wider range of alternative implementations as suggested in these new instructions, which I think makes the exercise more interesting.

Therefore if any Wikipedia link is to be removed I would prefer it be the one to the "Computing square roots" page, not the "Integer square roots" page.  The inverse of your requested change.

exercises/square-root/instructions.md
xanni: OK, I've replaced "output" with "return".

exercises/square-root/instructions.md
Cool-Katt: I'm with @xanni on this one. Judging from the original implementation, the author meant this exercise as an exercise on perfect squares.
I also thin that both links can stay. Not that many people will bother following up two links before solving the exercise. If they get stuck, they loo to community solutions instead.

kotp: Not to mention the risk of them doing their own research and being exposed to things.

tasxatzial:
The wiki link includes actual implementations in C but doesn’t appear to have code in any other languages. Are we still okay with that?

xanni:
> The wiki link includes actual implementations in C but doesn’t appear to have code in any other languages. Are we still okay with that?

Well, we could edit the Wiki page and replace the C with pseudocode...

mostly kidding.

Cool-Katt:
I'm feeling 

> The wiki link includes actual implementations in C but doesn’t appear to have code in any other languages. Are we still okay with that?

I'm feeling okay with that. As I said before, it's much more likely that students would copy existing community solutions, than read the actual wiki page.

tasxatzial:
I’m more concerned about students who will spend time reading the wiki. If we direct them to exactly what they’re trying to implement, it takes away their opportunity to develop this themselves.

xanni:
> I’m more concerned about students who will spend time reading the wiki. If we direct them to exactly what they’re trying to implement, it takes away their opportunity to develop this themselves.

That's a reasonable concern, but of course students can also copy-and-paste from Stack Overflow, or ask an AI to write it for them... or to take us right back to what started this whole proposal, many of them are just calling sqrt() or n^0.5.

Cool-Katt:
> I’m more concerned about students who will spend time reading the wiki. If we direct them to exactly what they’re trying to implement, it takes away their opportunity to develop this themselves.

If all the effort makes students *really* read the wiki page and try to understand, then i feel it is time well spent. It's like when you are  preparing a cheat sheet for a test, you're revising the material so you and up remembering more.

xanni:
I'm actually OK with deleting all the Wikipedia links (which are full of lots of math notation) and letting enterprising students find that themselves if they want.  The implementation hints should be sufficient to provide some handy search terms.

tasxatzial:
Copying from Stack Overflow or asking AI implies an intentional choice by students—they know they’re not putting in the effort. That’s very different from directing them to read a page that contains the solution. We expect them to do that in order to learn how the algorithm works. But they have no way of knowing in advance that the wiki includes solutions.

Cool-Katt:
Well, if they end up looking up the wiki page themselves, it's the same result so it's not less intentional

exercises/square-root/instructions.md
xanni: Wikipedia links removed.

tasxatzial:
@xanni Please hold on a sec. Let's give everyone the chance to catch up.

IsaacG:
I'm not a fan of removing the Wiki links. If we're asking students to implement an algorithm, we should be providing them with the algorithm details.

IsaacG:
Could you please put the Wiki links back for now. If you'd like to discuss removing them, please start a new thread on the forum. Let's please restrict the scope of this PR to what was agreed upon in the forum and not expand the scope.

Cool-Katt:
> Could you please put the Wiki links back for now. If you'd like to discuss removing them, please start a new thread on the forum. Let's please restrict the scope of this PR to what was agreed upon in the forum and not expand the scope.

100% agree with Isaac. Let's tackle the links in a separate thread and PR.

exercises/square-root/instructions.md
tasxatzial: I second IsaacG's suggestion. It emphasizes that we are limited to inputs that yield a specific type of result. The current change just focuses on what the result is. The difference is subtle but important.

exercises/square-root/instructions.md
```diff
@@ -0,0 +1,18 @@
+# Instructions
+
+Your task is to calculate the square root of a given number.
+
+- Try to avoid using the pre-existing math libraries of your language.
+- As input you’ll be given a positive whole number, i.e. 1, 2, 3, 4…
```
tasxatzial, line 6:
```suggestion
- As input you'll be given a positive whole number, i.e. 1, 2, 3, 4…
```

exercises/square-root/introduction.md
```diff
@@ -0,0 +1,10 @@
+# Introduction
+
+We are launching a deep space exploration rocket and we need a way to make sure the navigation system stays on target.
+
+As the first step in our calculation, we take a target number and find its square root (that is, the number that when multiplied by itself equals the target number).
+
+The journey will be very long.
+To make the batteries last as long as possible, we had to make our rocket’s onboard computer very power efficient.
+Unfortunately that means that we can’t rely on fancy math libraries and functions, as they use more power.
```
tasxatzial, line 9:
```suggestion
To make the batteries last as long as possible, we had to make our rocket's onboard computer very power efficient.
Unfortunately that means that we can't rely on fancy math libraries and functions, as they use more power.
```

tasxatzial (commented):
Few minor changes for consistency.

tasxatzial (approved):

exercises/square-root/introduction.md
```diff
@@ -0,0 +1,10 @@
+# Introduction
+
+We are launching a deep space exploration rocket and we need a way to make sure the navigation system stays on target.
+
+As the first step in our calculation, we take a target number and find its square root (that is, the number that when multiplied by itself equals the target number).
+
+The journey will be very long.
+To make the batteries last as long as possible, we had to make our rocket's onboard computer very power efficient.
+Unfortunately that means that we can't rely on fancy math libraries and functions, as they use more power.
+Instead we want to calculate the square root directly.
```
IsaacG, line 10: Note, "directly" suggests that the library functions do something less direct than the suggested algorithms. Using the library functions is just as direct implementing this yourself. 

We may want to consider replacing "directly" with something more accurate, like suggesting that we want to build a minimal math library that has only the one function we actually need: square root. 

Or we can leave it as is for now.

IsaacG (approved):

IsaacG (approved):

kotp (approved):

ErikSchierboom (approved):

PR #2496, opened 2024-11-12
Update the description for the complex-numbers exercise

jagdish-15:
### Pull Request Description

This pull request updates the description file for the **Complex Number** exercise, as discussed in the [forum thread](https://forum.exercism.org/t/guidance-needed-for-adding-new-test-cases-for-complex-number-exercise/13649/13). The changes reflect the updated description with no objections raised during the discussion.

### DIFF
--- a/exercises/complex-numbers/description.md
+++ b/exercises/complex-numbers/description.md
@@ -37,7 +37,7 @@ The square of the absolute value is computed as the product of `z` and its conju
 The sum of two complex numbers `z1 = a + b * i` and `z2 = c + d * i` is computed by adding their real and imaginary parts separately:
 
 ```text
-z1 + z2 = (a + b * i) + (c + d * i) 
+z1 + z2 = (a + b * i) + (c + d * i)
         = (a + c) + (b + d) * i
 ```
 
@@ -46,7 +46,7 @@ z1 + z2 = (a + b * i) + (c + d * i)
 The difference of two complex numbers is obtained by subtracting their respective parts:
 
 ```text
-z1 - z2 = (a + b * i) - (c + d * i) 
+z1 - z2 = (a + b * i) - (c + d * i)
         = (a - c) + (b - d) * i
 ```
 
@@ -55,7 +55,7 @@ z1 - z2 = (a + b * i) - (c + d * i)
 The product of two complex numbers is defined as:
 
 ```text
-z1 * z2 = (a + b * i) * (c + d * i) 
+z1 * z2 = (a + b * i) * (c + d * i)
         = (a * c - b * d) + (b * c + a * d) * i
 ```
 
@@ -64,7 +64,7 @@ z1 * z2 = (a + b * i) * (c + d * i)
 The reciprocal of a non-zero complex number is given by:
 
 ```text
-1 / z = 1 / (a + b * i) 
+1 / z = 1 / (a + b * i)
       = a / (a^2 + b^2) - b / (a^2 + b^2) * i
 ```
 
@@ -73,8 +73,8 @@ The reciprocal of a non-zero complex number is given by:
 The division of one complex number by another is given by:
 
 ```text
-z1 / z2 = z1 * (1 / z2) 
-        = (a + b * i) / (c + d * i) 
+z1 / z2 = z1 * (1 / z2)
+        = (a + b * i) / (c + d * i)
         = (a * c + b * d) / (c^2 + d^2) + (b * c - a * d) / (c^2 + d^2) * i
 ```
 
@@ -83,7 +83,7 @@ z1 / z2 = z1 * (1 / z2)
 Raising _e_ (the base of the natural logarithm) to a complex exponent can be expressed using Euler's formula:
 
 ```text
-e^(a + b * i) = e^a * e^(b * i) 
+e^(a + b * i) = e^a * e^(b * i)
               = e^a * (cos(b) + i * sin(b))
 ```
 
--- a/exercises/complex-numbers/description.md
+++ b/exercises/complex-numbers/description.md
@@ -86,10 +86,10 @@ e^(a + b * i) = e^a * e^(b * i) = e^a * (cos(b) + i * sin(b))
 
 Given that you should not use built-in support for complex numbers, implement the following operations:
 
-- **Addition** of two complex numbers.
-- **Subtraction** of two complex numbers.
-- **Multiplication** of two complex numbers.
-- **Division** of two complex numbers.
-- Calculation of the **conjugate** of a complex number.
-- Calculation of the **absolute value** of a complex number.
-- Calculation of the **exponent** of a given complex number.
+- **addition** of two complex numbers
+- **subtraction** of two complex numbers
+- **multiplication** of two complex numbers
+- **division** of two complex numbers
+- **conjugate** of a complex number
+- **absolute value** of a complex number
+- **exponent** of a complex number
--- a/exercises/complex-numbers/description.md
+++ b/exercises/complex-numbers/description.md
@@ -6,7 +6,7 @@ A **complex number** is expressed in the form `z = a + b * i`, where:
 
 - `b` is the **imaginary part** (also a real number), and
 
-- `i` is the imaginary unit satisfying `i^2 = -1`.
+- `i` is the **imaginary unit** satisfying `i^2 = -1`.
 
 ## Key Properties
 
--- a/exercises/complex-numbers/description.md
+++ b/exercises/complex-numbers/description.md
@@ -56,7 +56,7 @@ The difference of two complex numbers is obtained by subtracting their respectiv
 z₁ - z₂ = (a - c) + (b - d) * i
 ```
 
-### Multiplication 
+### Multiplication
 
 The product of two complex numbers is defined as:
 
--- a/exercises/complex-numbers/description.md
+++ b/exercises/complex-numbers/description.md
@@ -26,6 +26,8 @@ The absolute value (or modulus) of `z` is defined as:
 |z| = sqrt(a^2 + b^2)
 ```
 
+### Sqaure
+
 The square of the absolute value, `|z|²`, can be computed as the product of `z` and its conjugate:
 
 ```plaintext
@@ -58,12 +60,20 @@ The product of two complex numbers is defined as:
 z₁ * z₂ = (a + b * i) * (c + d * i) = (a * c - b * d) + (b * c + a * d) * i
 ```
 
+### Reciprocal
+
+The reciprocal of a non-zero complex number is given by:
+
+```math
+1 / z = 1 / (a + b * i) = a / (a² + b²) - b / (a² + b²) * i
+```
+
 ### Division
 
 The division of one complex number by another is given by:
 
 ```plaintext
-z₁ / z₂ = (a + b * i) / (c + d * i) = (a * c + b * d) / (c² + d²) + (b * c - a * d) / (c² + d²) * i
+z₁ / z₂ = z₁ * (1 / z₂) = (a + b * i) / (c + d * i) = (a * c + b * d) / (c² + d²) + (b * c - a * d) / (c² + d²) * i
 ```
 
 ### Exponentiation
--- a/exercises/complex-numbers/description.md
+++ b/exercises/complex-numbers/description.md
@@ -92,4 +92,4 @@ Given that you should not use built-in support for complex numbers, implement th
 - **division** of two complex numbers
 - **conjugate** of a complex number
 - **absolute value** of a complex number
-- **exponentiation** of the base _e_ (the natural logarithm) to a complex number
+- **exponentiation** of _e_ (the base of the natural logarithm) to a complex number
--- a/exercises/complex-numbers/description.md
+++ b/exercises/complex-numbers/description.md
@@ -64,7 +64,7 @@ z₁ * z₂ = (a + b * i) * (c + d * i) = (a * c - b * d) + (b * c + a * d) * i
 
 The reciprocal of a non-zero complex number is given by:
 
-```math
+```plaintext
 1 / z = 1 / (a + b * i) = a / (a² + b²) - b / (a² + b²) * i
 ```
 
--- a/exercises/complex-numbers/description.md
+++ b/exercises/complex-numbers/description.md
@@ -1,29 +1,85 @@
 # Description
 
-A complex number is a number in the form `a + b * i` where `a` and `b` are real and `i` satisfies `i^2 = -1`.
+A **complex number** is expressed in the form:
 
-`a` is called the real part and `b` is called the imaginary part of `z`.
-The conjugate of the number `a + b * i` is the number `a - b * i`.
-The absolute value of a complex number `z = a + b * i` is a real number `|z| = sqrt(a^2 + b^2)`. The square of the absolute value `|z|^2` is the result of multiplication of `z` by its complex conjugate.
+```
+z = a + b * i
+```
 
-The sum/difference of two complex numbers involves adding/subtracting their real and imaginary parts separately:
-`(a + i * b) + (c + i * d) = (a + c) + (b + d) * i`,
-`(a + i * b) - (c + i * d) = (a - c) + (b - d) * i`.
+where:
 
-Multiplication result is by definition
-`(a + i * b) * (c + i * d) = (a * c - b * d) + (b * c + a * d) * i`.
+- `a` is the **real part** (a real number),
 
-The reciprocal of a non-zero complex number is
-`1 / (a + i * b) = a/(a^2 + b^2) - b/(a^2 + b^2) * i`.
+- `b` is the **imaginary part** (also a real number), and
 
-Dividing a complex number `a + i * b` by another `c + i * d` gives:
-`(a + i * b) / (c + i * d) = (a * c + b * d)/(c^2 + d^2) + (b * c - a * d)/(c^2 + d^2) * i`.
+- `i` is the imaginary unit satisfying `i^2 = -1`.
 
-Raising e to a complex exponent can be expressed as `e^(a + i * b) = e^a * e^(i * b)`, the last term of which is given by Euler's formula `e^(i * b) = cos(b) + i * sin(b)`.
+### Key Properties
 
-Implement the following operations:
+1. **Conjugate**: The conjugate of the complex number `z = a + b * i` is given by:
 
-- addition, subtraction, multiplication and division of two complex numbers,
-- conjugate, absolute value, exponent of a given complex number.
+```
+z̅ = a - b * i
+```
 
-Assume the programming language you are using does not have an implementation of complex numbers.
+2. **Absolute Value**: The absolute value (or modulus) of `z` is defined as:
+
+```
+|z| = sqrt(a^2 + b^2)
+```
+
+The square of the absolute value, `|z|²`, can be computed as the product of `z` and its conjugate:
+
+```
+|z|² = z * z̅ = a² + b²
+```
+
+### Operations on Complex Numbers
+
+1. **Addition**: The sum of two complex numbers `z₁ = a + b * i` and `z₂ = c + d * i` is computed by adding their real and imaginary parts separately:
+
+```
+z₁ + z₂ = (a + c) + (b + d) * i
+```
+
+2. **Subtraction**: The difference of two complex numbers is obtained by subtracting their respective parts:
+
+```
+z₁ - z₂ = (a - c) + (b - d) * i
+```
+
+3. **Multiplication**: The product of two complex numbers is defined as:
+
+```
+z₁ * z₂ = (a + b * i) * (c + d * i) = (a * c - b * d) + (b * c + a * d) * i
+```
+
+4. **Division**: The division of one complex number by another is given by:
+
+```
+z₁ / z₂ = (a + b * i) / (c + d * i) = (a * c + b * d) / (c² + d²) + (b * c - a * d) / (c² + d²) * i
+```
+
+5. **Reciprocal**: The reciprocal of a non-zero complex number is given by:
+
+```
+1 / z = 1 / (a + b * i) = a / (a² + b²) - b / (a² + b²) * i
+```
+
+6. **Exponentiation**: Raising _e_ (the base of the natural logarithm) to a complex exponent can be expressed using Euler's formula:
+
+```
+e^(a + b * i) = e^a * e^(b * i) = e^a * (cos(b) + i * sin(b))
+```
+
+### Implementation Requirements
+
+Given that you should not use built-in support for complex numbers, implement the following operations:
+
+- **Addition** of two complex numbers.
+- **Subtraction** of two complex numbers.
+- **Multiplication** of two complex numbers.
+- **Division** of two complex numbers.
+- Calculation of the **conjugate** of a complex number.
+- Calculation of the **absolute value** of a complex number.
+- Calculation of the **exponent** of a given complex number.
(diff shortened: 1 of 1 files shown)


Cool-Katt:
@jagdish-15 Looks like you might have forgotten to run the linter on this.

jagdish-15:
@Cool-Katt, apologies for the initial oversight!

I’ve updated the file so it now passes all tests. I’m a bit uncertain about the ordered list formatting—I've prefixed each item with “1.” as required by markdownlint, but if there’s a better approach, I'd appreciate any guidance. Thanks!

exercises/complex-numbers/description.md
```diff
@@ -1,29 +1,85 @@
...
 
-Dividing a complex number `a + i * b` by another `c + i * d` gives:
-`(a + i * b) / (c + i * d) = (a * c + b * d)/(c^2 + d^2) + (b * c - a * d)/(c^2 + d^2) * i`.
+- `i` is the imaginary unit satisfying `i^2 = -1`.
 
-Raising e to a complex exponent can be expressed as `e^(a + i * b) = e^a * e^(i * b)`, the last term of which is given by Euler's formula `e^(i * b) = cos(b) + i * sin(b)`.
+## Key Properties
 
-Implement the following operations:
+1. **Conjugate**: The conjugate of the complex number `z = a + b * i` is given by:
```
SleeplessByte, line 19: Instead of making this a list, try:

```markdown
### Conjugate

[...]

```

For example below this line:


-------------------------------

## Key Properties

### Conjugate

The conjugate of the complex number `z = a + b * i` is given by:

```math
z̅ = a - b * i
```

### Absolute Value

The absolute value (or modulus) of `z` is defined as:

```math
|z| = sqrt(a^2 + b^2)
```

tasxatzial:
@jagdish-15 Please replace the `1.` with just `-`. This is an unordered list.

SleeplessByte (commented):
For those who want to see the rich text [in the latest form](https://github.com/jagdish-15/problem-specifications/blob/chnage-description-complex/exercises/complex-numbers/description.md) when reviewing. And [here is the split raw view without whitespace changes](https://github.com/exercism/problem-specifications/pull/2496/files?diff=split&w=1#diff-e6831cd784873c50d20d539bd9fc87ee4efa9d03439180ffc54bbe2b5fbcd10c)

exercises/complex-numbers/description.md
tasxatzial: @jagdish-15 Go with SleeplessByte's suggestion, disregard mine.

SleeplessByte (approved):
Tentatively approved if you fix the linter.

I think this is better than what it was, so yah.

jagdish-15:
Thank you so much for the feedback, @SleeplessByte! I'll definitely address the linter issues. If you notice any other areas for improvement or have any suggestions for refining this version further, I’d love to hear them!

exercises/complex-numbers/description.md
```diff
@@ -1,29 +1,101 @@
...
-The conjugate of the number `a + b * i` is the number `a - b * i`.
-The absolute value of a complex number `z = a + b * i` is a real number `|z| = sqrt(a^2 + b^2)`. The square of the absolute value `|z|^2` is the result of multiplication of `z` by its complex conjugate.
+```math
+z = a + b * i
+```
 
-The sum/difference of two complex numbers involves adding/subtracting their real and imaginary parts separately:
-`(a + i * b) + (c + i * d) = (a + c) + (b + d) * i`,
-`(a + i * b) - (c + i * d) = (a - c) + (b - d) * i`.
+where:
```
tasxatzial, line 9: Doesn't let me apply a suggestion, so I'm leaving it as a comment:

Consider replacing the selected text with:

A **complex number** is expressed in the form `z = a + b * i`, where:

jagdish-15:
I’d also love to know if this change will automatically update in instructions.md across all tracks, or if it’ll require a manual update.

BNAndras:
> I’d also love to know if this change will automatically update in instructions.md across all tracks, or if it’ll require a manual update.

Each track will need to sync it manually. Some tracks do have tooling to alert the maintainer when there are unsynced changes in the problem specifications repo, but the actual syncing is still done by a human.

tasxatzial:
It is possible to trigger a PR that updates the description across all tracks with a script. However, it's more likely that track maintainers will perform a batch sync once a month, for example, to pull in all changes from this repo.

Cool-Katt (approved):

jagdish-15:
Thank you so much, everyone!

Before wrapping up, I’d love to know if there’s anything I should keep in mind for next time, or if you’d like any additional changes to the current file.

exercises/complex-numbers/description.md
```diff
@@ -1,29 +1,95 @@
...
+- `a` is the **real part** (a real number),
 
-The sum/difference of two complex numbers involves adding/subtracting their real and imaginary parts separately:
-`(a + i * b) + (c + i * d) = (a + c) + (b + d) * i`,
-`(a + i * b) - (c + i * d) = (a - c) + (b - d) * i`.
+- `b` is the **imaginary part** (also a real number), and
 
-Multiplication result is by definition
-`(a + i * b) * (c + i * d) = (a * c - b * d) + (b * c + a * d) * i`.
+- `i` is the imaginary unit satisfying `i^2 = -1`.
```
tasxatzial, line 9:
```suggestion
- `i` is the **imaginary unit** satisfying `i^2 = -1`.
```

tasxatzial (commented):
One small change for consistency

jagdish-15:
Thank you so much, @tasxatzial! I really appreciate all your help. If there’s anything else that might need fixing, I’d love to hear your suggestions.

exercises/complex-numbers/description.md
```diff
@@ -1,29 +1,95 @@
...
+
+Given that you should not use built-in support for complex numbers, implement the following operations:
+
+- **addition** of two complex numbers
+- **subtraction** of two complex numbers
+- **multiplication** of two complex numbers
+- **division** of two complex numbers
+- **conjugate** of a complex number
+- **absolute value** of a complex number
+- **exponent** of a complex number
```
tasxatzial, line 95:
```suggestion
- **exponentiation** of the base _e_ (the natural logarithm) to a complex number
```

tasxatzial (commented):
Hopefully the final suggestion. I just realized that the "exponent of a complex" number isn't even proper math. Let's fix that.

exercises/complex-numbers/description.md
jagdish-15: Makes sense!!

jagdish-15:
Done!

tasxatzial (approved):

SleeplessByte:
Now we just need @exercism/maintainers-admin to sign-off. Great work @jagdish-15.

tasxatzial (commented):
@jagdish-15 Final suggestion, this is also for consistency with the previous part that introduces exponentiation

BNAndras (approved):

exercises/complex-numbers/description.md
```diff
@@ -1,29 +1,95 @@
...
 
-Dividing a complex number `a + i * b` by another `c + i * d` gives:
-`(a + i * b) / (c + i * d) = (a * c + b * d)/(c^2 + d^2) + (b * c - a * d)/(c^2 + d^2) * i`.
+### Conjugate
 
-Raising e to a complex exponent can be expressed as `e^(a + i * b) = e^a * e^(i * b)`, the last term of which is given by Euler's formula `e^(i * b) = cos(b) + i * sin(b)`.
+The conjugate of the complex number `z = a + b * i` is given by:
 
-Implement the following operations:
+```math
```
IsaacG, line 17: Does Exercism's website render `math` blocks?

exercises/complex-numbers/description.md
```diff
@@ -1,29 +1,95 @@
...
+### Absolute Value
 
-Assume the programming language you are using does not have an implementation of complex numbers.
+The absolute value (or modulus) of `z` is defined as:
+
+```math
+|z| = sqrt(a^2 + b^2)
+```
+
+The square of the absolute value, `|z|²`, can be computed as the product of `z` and its conjugate:
```
IsaacG, line 29: Square isn't part of the tests. Should it be here? If yes, should it have a header?

IsaacG (approved):

exercises/complex-numbers/description.md
ErikSchierboom: I don't think we do

exercises/complex-numbers/description.md
jagdish-15: Thank you for pointing this out! Since the square of the absolute value isn't part of the tests or required to solve the exercise, I think it might be better to leave it out altogether. Including it could potentially confuse learners by making them think it's something they need to implement.

On a related note, I noticed that the reciprocal isn't tested either. I included it because it was in the previous version of the file, but if you feel it’s unnecessary, I can remove that as well.

Let me know your thoughts, and I’ll make the changes accordingly. 😊

exercises/complex-numbers/description.md
jagdish-15: @IsaacG, yes, @ErikSchierboom is right! Exercism doesn't render math blocks (as far as I know, since I haven't seen them anywhere), and they don't work in the forum. However, the tests and checks on GitHub require me to specify the block type. I initially thought it would make sense to specify it as a math block, but my original intention was to keep it as plaintext. I didn't specify plaintext initially, though.

Now, I'm wondering whether I should change the block type to plaintext or leave it as is (since Exercism doesn't support math blocks, so I assume it will render it as plain text instead of throwing an error). The reasoning is that if Exercism ever starts supporting math blocks in the future, it might be useful to leave it as is so it renders appropriately.

What do you think?

tasxatzial: It might be better to change it to 'plaintext' or 'text'. When I preview the file in vscode, I'm getting rendering errors due to the `math` block.

exercises/complex-numbers/description.md
tasxatzial: The square of a complex number could be used in a division operation (the square of the second number is the denominator in both fractions), so it might be preferable to leave it for those who wish to explore that approach:

```(a + i * b) / (c + i * d) = (a * c + b * d)/(c^2 + d^2) + (b * c - a * d)/(c^2 + d^2) * i```

The reciprocal section doesn't seem to be useful at all and can likely be safely removed.

Comment (jagdish-15) [reply exercises/complex-numbers/description.md:29]: You're right that the reciprocal doesn't have a direct use in the given context. It might be the case that the author included it initially as a way to offer an alternative approach or to keep the option open for students who want to explore multiplying by the reciprocal, as it essentially performs division. But in terms of the immediate solution, it does seem unnecessary.

File: exercises/complex-numbers/description.md (hunk 5 above)
Comment (jagdish-15) [reply exercises/complex-numbers/description.md:17]: Thanks, @tasxatzial for the clarification! I agree with your reasoning that switching to plain text seems like the better option, especially given Exercism’s current limitations with not rendering math blocks. Aesthetically, it makes sense to avoid rendering errors and ensure everything displays cleanly. You’re right that even if math blocks are supported in the future, they may not necessarily be labelled as such. Changing it to plaintext would likely be the safest choice for now. I’ll push the changes ASAP.

File: exercises/complex-numbers/description.md (hunk 6 above)
Comment (tasxatzial) [reply exercises/complex-numbers/description.md:29]: True, dividing by a number is the same as multiplying by its reciprocal, just like with real numbers. However, this might not be immediately obvious, especially when working with complex numbers. Looking through community solutions, I noticed that some people use the square method, while none seem to use the reciprocal.

Also, I noticed that in the new instructions, the reciprocal is introduced after the division. This might cause people to focus more on the division formula and ignore the reciprocal as an option. In the original instructions, the reciprocal was mentioned right after multiplication, though it didn't seem to make any difference overall. If we still want people to consider the reciprocal route, we might want to explicitly indicate its connection with the division operation, like so:

`z₁ / z₂ = z₁ * (1 / z₂) = <formula goes here>`

I don't have a strong opinion on this, but since I'm biased due to my background, I'll stay out of the decision. Some might argue that the instructions are too technical and prefer removing both the square and the reciprocal altogether.

Comment (jagdish-15) [reply exercises/complex-numbers/description.md:29]: I agree that removing the reciprocal concept could simplify things and make the instructions more straightforward, especially for those who might find it confusing when dealing with complex numbers. Simplifying it would likely make the learning process smoother, so I think it could be a great move. Thanks for sharing your thoughts!

**Comment by jagdish-15** [comment]
I believe this addresses both of the issues mentioned above. @tasxatzial, @ErikSchierboom, @IsaacG, @SleeplessByte, @Cool-Katt, and @BNAndras—please let me know if there’s anything else that needs attention.

File: exercises/complex-numbers/description.md (hunk 6 above)
Comment (tasxatzial) [reply exercises/complex-numbers/description.md:29]: Keep in mind that there are already 5 approvals, so it might be better to leave the discussion on whether to remove the square and reciprocal for another thread. As it stands, I would still suggest moving the reciprocal before the division (as it was in the original instructions) and just adding the `z₁ * (1 / z₂)` to give everyone the chance to use the reciprocal if they want.

Comment (jagdish-15) [reply exercises/complex-numbers/description.md:31]: Sure, I've changed it accordingly!

Comment (IsaacG) [reply exercises/complex-numbers/description.md:31]: 1. There's a typo in the header.
2. I think the Square section ought to be dropped. It's not relevant to this exercise. Reciprocal is useful for division, so it can be useful and bumped to a later discussion. I'm not sure why Square would be kept.

File: exercises/complex-numbers/description.md [hunk 7]
```diff
@@ -1,29 +1,97 @@
...
-`(a + i * b) - (c + i * d) = (a - c) + (b - d) * i`.
+- `b` is the **imaginary part** (also a real number), and
 
-Multiplication result is by definition
-`(a + i * b) * (c + i * d) = (a * c - b * d) + (b * c + a * d) * i`.
+- `i` is the **imaginary unit** satisfying `i^2 = -1`.
 
-The reciprocal of a non-zero complex number is
-`1 / (a + i * b) = a/(a^2 + b^2) - b/(a^2 + b^2) * i`.
+## Key Properties
```
IsaacG, line 11: What makes the absolute value a property and not an operation?

exercises/complex-numbers/description.md
```diff
@@ -1,29 +1,97 @@
...
 
-Dividing a complex number `a + i * b` by another `c + i * d` gives:
-`(a + i * b) / (c + i * d) = (a * c + b * d)/(c^2 + d^2) + (b * c - a * d)/(c^2 + d^2) * i`.
+### Conjugate
 
-Raising e to a complex exponent can be expressed as `e^(a + i * b) = e^a * e^(i * b)`, the last term of which is given by Euler's formula `e^(i * b) = cos(b) + i * sin(b)`.
+The conjugate of the complex number `z = a + b * i` is given by:
 
-Implement the following operations:
+```plaintext
```
IsaacG, line 17: This repo uses `text` (117 times) and not `plaintext` (4 times).

jagdish-15: Sure, I've changed it to text!

exercises/complex-numbers/description.md
jagdish-15: Upon reflecting further, I realized that the header isn't necessary at all, as the property in question is the square of the absolute value of the complex number, not the complex number itself. Regarding its significance, this property is widely used in various operations, from division to finding the reciprocal of a complex number. If you examine the formulas, you'll notice it appears in the denominators of both the real and imaginary components.

tasxatzial: That's correct, all denominators are the square of either the first or the complex number.

Here's a solution that defines it on line 7 and uses it for division on line 29

https://exercism.org/tracks/clojure/exercises/complex-numbers/solutions/fergalbyrne

exercises/complex-numbers/description.md
tasxatzial: I believe the concept of "key property" isn't very well defined and can be interpreted in different ways. Technically, neither the conjugate nor the absolute value is regarded as a property of complex numbers. Properties refer to characteristics of the numbers, like the associative rule `(a + b) + c = a + (b + c)` for real numbers.

Personally, I'm not too picky about this, given that we're not really teaching math here, and what we already have should be sufficient for this exercise. However, if we want to be accurate, both the conjugate and the absolute value (or square) are simply operations.

IsaacG: It sounds like you're defining "property" as a function which operates on one complex number and defining "operation" as a function which operates on two complex numbers. In that case, wouldn't "exponent" be a property? Or would it make more sense to not apply this distinction and just call them all "operations"?

exercises/complex-numbers/description.md
```diff
@@ -1,29 +1,95 @@
...
 
-- addition, subtraction, multiplication and division of two complex numbers,
-- conjugate, absolute value, exponent of a given complex number.
+### Absolute Value
 
-Assume the programming language you are using does not have an implementation of complex numbers.
+The absolute value (or modulus) of `z` is defined as:
+
+```text
+|z| = sqrt(a^2 + b^2)
```
IsaacG, line 26: Should these use superscript 2 and not the `^` symbol, in like with the other codeblocks?

tasxatzial: The exponentiation still uses the `^` symbol. It might be better to avoid superscripts and subscripts altogether, as they can be a bit too small for some people to read.

kotp: Consistency over "too small to read" because I can read them with a reader, or make the font size bigger, but being inconsistent makes things "jitter" a bit in what has to be understood.

tasxatzial: Completely agree. Removing the superscripts and subscripts means that we revert back to how the instructions currently are. So a^2, b^2, z1, z2, e^a etc.

jagdish-15: Sure, I am happy to change it to superscript!

exercises/complex-numbers/description.md
jagdish-15: The key point I want to highlight is that operations like conjugate, absolute value, exponentiation, and reciprocal are unary operators, while addition, subtraction, multiplication, and division are binary operators for complex numbers. This raises an important question: should unary and binary operators be categorized separately, or should they all be grouped together under a unified "Operations" section?  

My reasoning for placing **absolute value** and **conjugate** in the *Key Properties* section, separate from other operations, is that these two play a foundational role in complex number arithmetic. For example:  
- **Absolute value** is central to defining magnitude, normalization, and division.  
- **Conjugate** is essential for computing reciprocals and simplifying division.  

I acknowledge, however, that both involve computation and are technically unary operations, not inherent "properties" in the strict mathematical sense. As @tasxatzial rightly pointed out, properties typically refer to inherent characteristics like associativity or commutativity.  

That said, my intention in categorizing them as "key properties" was to emphasize their foundational role in complex number operations. If unifying all unary and binary operators under "Operations" better suits the context of this repository, I’m happy to align with that approach.  

Looking forward to your thoughts on whether separating these foundational elements adds clarity or if a unified categorization would be more consistent.

exercises/complex-numbers/description.md
tasxatzial: Superscripts and subscripts aren't easy for everyone to read. Some people may have low-quality monitors, while others may have impaired eyesight. They pose accessibility challenges, as they are small compared to normal text and disrupt visual alignment, making it harder to follow the content. This creates a similar 'jitter' effect that Kotp mentioned. If Exercism were to implement proper math blocks in the future, subscript and superscript notation would be a great option. For now, they are just a workaround due to the lack of proper math rendering.

Please revert everything to how it was before, as it wasn't an issue previously.

jagdish-15: @tasxatzial I haven’t made any changes yet; I’m still waiting for a conclusion to be drawn!

tasxatzial: There are two options. Either switch to how it was before, or use superscripts/subscripts. Both of them have no consistency issues. And since there are issues with superscripts/subscripts, it's probably best if we avoid them altogether.

Also, take a look at the complex conjugate `z̅ = a - b * i`. What is the first symbol? A `z`? The line above z is almost touching the letter, making it illegible. The current instructions don't even use a special notation for the conjugate.

All those issues wouldn't exist if Exercism properly supported math.

jagdish-15: We can also represent the conjugate as `z*` if the bar notation causes issues, or we can avoid denoting it altogether. However, I believe this representation should work fine.  

I’m also happy to revert to the original notation and avoid using subscript or superscript if no one has any concerns with that.

tasxatzial: > However, I believe the `z*` representation should be fine.

This now looks like the multiplication symbol. I'd just assign the conjugate to a variable and use it:

The conjugate `zc` of the complex number `z = a + b * i` is given by...

exercises/complex-numbers/description.md
tasxatzial: Since the exercise requires implementing 'operations', which include both the conjugate and the absolute value, it might be better to move them to the operations section for consistency.

exercises/complex-numbers/description.md
IsaacG: It sounds like people are leaning towards using the `^` symbol. I don't have a strong opinion here. Using `^` sounds like the most consistent option.

jagdish-15: Sure, we’ll use the `^` symbol and avoid using notation for the conjugate altogether to prevent issues with bar rendering and the potential confusion of `z*` with multiplication.

exercises/complex-numbers/description.md
```diff
@@ -1,29 +1,93 @@
...
+```text
+z1 - z2 = (a - c) + (b - d) * i
+```
+
+### Multiplication
+
+The product of two complex numbers is defined as:
+
+```text
+z1 * z2 = (a + b * i) * (c + d * i) = (a * c - b * d) + (b * c + a * d) * i
```
IsaacG, line 56: I think that intermediate step is helpful to show. It would be nice to include it in the addition and subtraction formula, too!

IsaacG (approved):

exercises/complex-numbers/description.md
jagdish-15: I've made these changes, please check them out!

tasxatzial: I feel that the equations are becoming a bit too long as one-liners. Perhaps consider breaking the following into multiple lines?

```text
z1 + z2 = (a + b * i) + (c + d * i)
        = (a + c) + (b + d) * i

z1 * z2 = (a + b * i) * (c + d * i)
        = (a * c - b * d) + (b * c + a * d) * i

1 / z = 1 / (a + b * i)
      = a / (a^2 + b^2) - b / (a^2 + b^2) * i

z1 / z2 = z1 * (1 / z2)
        = (a + b * i) / (c + d * i)
        = (a * c + b * d) / (c^2 + d^2) + (b * c - a * d) / (c^2 + d^2) * i

e^(a + b * i) = e^a * e^(b * i)
              = e^a * (cos(b) + i * sin(b))
```

The absolute value does not need to be broken as it's already short and readable.

jagdish-15: Sure

jagdish-15:
Shouldn't we mark this issue as resolved since we have verified that the tests for the Yacht exercise are already in sync, and updates for the other exercises are currently in progress?

jagdish-15:
With this commit, I believe the formatting issues are resolved, and this should be ready to merge. Please let me know if I’ve missed anything.

A big thank you to @ErikSchierboom, @kotp, @Cool-Katt, @BNAndras, and @SleeplessByte, and especially to @tasxatzial and @IsaacG for reviewing this and providing your valuable feedback!

IsaacG (approved):
Looks good to me!

kotp (approved):

ErikSchierboom (approved):

PR #2497, opened 2024-11-12
eliuds-eggs: Reword exercise restrictions

tasxatzial:
Forum thread

https://forum.exercism.org/t/eliuds-eggs-instructions-restrictions-are-too-specific/13672

### DIFF
--- a/exercises/eliuds-eggs/instructions.md
+++ b/exercises/eliuds-eggs/instructions.md
@@ -4,5 +4,5 @@ Your task is to count the number of 1 bits in the binary representation of a num
 
 ## Restrictions
 
-Keep your hands off that bit-count functionality provided by your standard library!
-Solve this one yourself using other basic tools instead.
+Avoid using any library functions that count the number of bits in a number.
+Instead, try solving this yourself using more basic tools.


BNAndras (approved):

exercises/eliuds-eggs/instructions.md
```diff
@@ -4,5 +4,5 @@ Your task is to count the number of 1 bits in the binary representation of a num
 
 ## Restrictions
 
-Keep your hands off that bit-count functionality provided by your standard library!
-Solve this one yourself using other basic tools instead.
+Avoid using any library functions that count the number of bits in a number.
+Instead, try solving this yourself using more basic tools.
```
IsaacG, line 8: How does one measure _how_ basic something is? Would bitshifts and loops be more basic or less basic? I think sticking to "other" might be better here.

tasxatzial: > How does one measure _how_ basic something is? Would bitshifts and loops be more basic or less basic? I think sticking to "other" might be better here.

Good point. We can't. I didn't really think much about that for two reasons:

* The restriction specifies what should be avoided but doesn't clearly define what is permitted.
* Using 'other basic' might suggest a comparison, implying that the restricted tools and potential alternatives are similarly complex.

So, if we want to avoid any kind of comparison —which might be preferable— we could simply say 'other tools.' If a comparison is still desired, I’m personally indifferent.

Thoughts, maintainers?

Cool-Katt: I feel that other tools/approaches fits better in this case.

IsaacG: "other tools" sounds good to me!

kotp:
On Sun, Nov 17, 2024 at 09:30:28AM -0800, Isaac Good wrote:
> @IsaacG commented on this pull request.
> 
> 
> 
> > @@ -4,5 +4,5 @@ Your task is to count the number of 1 bits in the binary representation of a num
>  
>  ## Restrictions
>  
> -Keep your hands off that bit-count functionality provided by your standard library!
> -Solve this one yourself using other basic tools instead.
> +Avoid using any library functions that count the number of bits in a number.

That seems excessive, especially if you choose to write a library
function not using the built in bit-count functionality provided by your
standard library.

> +Instead, try solving this yourself using more basic tools.
> 
> "other tools" sounds good to me!

I agree with the "other tools" approach, for sure.

exercises/eliuds-eggs/instructions.md
kotp: On Sun, Nov 17, 2024 at 09:30:28AM -0800, Isaac Good wrote:
> @IsaacG commented on this pull request.
> 
> 
> 
> > @@ -4,5 +4,5 @@ Your task is to count the number of 1 bits in the binary representation of a num
>  
>  ## Restrictions
>  
> -Keep your hands off that bit-count functionality provided by your standard library!
> -Solve this one yourself using other basic tools instead.
> +Avoid using any library functions that count the number of bits in a number.

That seems excessive, especially if you choose to write a library
function not using the built in bit-count functionality provided by your
standard library.

> +Instead, try solving this yourself using more basic tools.
> 
> "other tools" sounds good to me!

I agree with the "other tools" approach, for sure.

kotp: For the purpose of the student learning, some of this interpretation should be the responsibility of the student based on their level.  If we are not testing and restricting specific tooling in the language, then this is a suggestion, rather than a requirement.

habere-et-dispertire:
I made an [image for these instructions](https://forum.exercism.org/t/new-exercise-suggestion-pop-count/7632/17) ( forum.exercism.org ) .
Do we want to include this ?

ac5cfac4e01dedf7ae74e2621b5cd258afadea1e

exercises/eliuds-eggs/instructions.md
tasxatzial: Correct. I intentionally reworded it to sound like a suggestion because I expected most tracks to treat it that way. I'm not even sure why a track would go to all that trouble to restrict specific functions. If someone chooses not to follow a suggestion, that's entirely up to them.

Could we just remove the header? The description is only two sentences anyway.

kotp: We could just leave it as it is.  It does not seem to have caused problems historically in other tracks.  Also not all languages have functions, and sometimes we do restrict specific functions or methods or procedures, depending on the nature of the exercise (not at the Problem Specification level though.

tasxatzial: How do you propose we solve the issue with the word "functions" so that it applies to all tracks? Here are two alternative sentences that come to mind:

`Avoid library functionality that directly counts the bits in a number.`

`Avoid library-provided features that directly count the bits in a number.`

I also prefer using 'directly,' as I originally proposed in the forum thread, to make it perfectly clear what we are looking for, and I removed 'any' to avoid potential confusion.

kotp: > How do you propose we solve the issue with the word "functions" 

A third choice:

"Keep your hands off that bit-count functionality"

tasxatzial: > "Keep your hands off that bit-count functionality"

This is identical to what we already have.

IsaacG: Possibly because people discussed how to avoid using the word function before 😂

I think "library functions" here isn't too bad. We don't want to ask students to "write a function" because not all tracks ask students to implement a function. However, I think "avoid built in library functions" is fine in any track. SQLite asks you to update a table, but still has functions. awk asks you to generate output, but still has library functions. bash doesn't really have libraries, but the intent is clear -- avoid any pre built solutions. Functions here can be understood as "some way to run code" and not strictly a function call.

habere-et-dispertire:
example-1-coop-min

example-1-binary-min

example-2-coop-min

example-2-binary-min

[Pull request](https://github.com/exercism/images/pull/30) with four SVG images.

exercises/eliuds-eggs/instructions.md
tasxatzial: May I suggest that we start wrapping up this PR and move the discussion to the forum? The problem with how restrictions are stated isn't limited to this exercise. Don't we often repeat, to ourselves and others, that using built-ins is perfectly fine and serves a purpose, such as exploring the library? So why not simply add a comment in the canonical data stating these restrictions as suggestions, leaving it to track maintainers to decide how to handle them? On one hand, we seem to care a lot about restrictions; on the other, we prefer when those restrictions aren't there. I feel this is a topic worth discussing in the forum, if there isn't already a thread about it.

Given the discussion so far, here are some of alternative options

`Avoid pre-built solutions that directly count the bits in a number. Implement your own approach instead.`

`Avoid pre-existing solutions that directly count the bits in a number. Implement your own approach instead.`

`Avoid pre-built functionality that directly counts the bits in a number. Implement your own approach instead.`

`Avoid pre-existing functionality that directly counts the bits in a number. Implement your own approach instead.`

kotp: Keeping what we have on Problem Specifications is also an option.  The wording is a suggestion, a request, rather than a restriction, the restriction would be in the tests, if those restrictions exist.

Is a comment needed for that in general?  Based on the solutions I have seen that ignore the suggestion, it seems that it is generally interpreted as a suggestion rather than a restriction.

IsaacG: If we don't have a consensus here, it might be best to bump this back to the forum to discuss until we can agree on something.

tasxatzial: > Based on the solutions I have seen that ignore the suggestion, it seems that it is generally interpreted as a suggestion rather than a restriction.

I don't know how it is interpreted. All I see is a clear restriction stated in a very aggressive manner.

`Restrictions: Keep your hands off that bit-count functionality provided by your standard library!`

tasxatzial: > If we don't have a consensus here, it might be best to bump this back to the forum to discuss until we can agree on something.

Well, folks, I don’t know. The only thing I'll say is that those instructions didn't achieve what they were supposed to. My reasoning for this is outlined in my initial forum post.

Regarding the wording, while I agree that it matters, for this particular exercise, once someone reads the phrase "Avoid ... that directly counts the number of bits" they'll immediately understand the restriction because it clearly conveys the intent. In contrast, terms like "bit-count functionality" can easily be misinterpreted, as "functionality" implies a broad range of things. In fact, it's even worse than using the word function.

That said, none of the terms — function, functions, functionality, or more basic tools—seems likely to make a significant difference here. As I mentioned earlier, the purpose of the restriction is to specify what cannot be used, not what can.

Additionally, the second sentence — Instead, try solving this yourself using other tools — feels superfluous at this point.

kotp: > As I mentioned earlier, the purpose of the restriction is to specify what cannot be used, not what can.

This basically outlines why I think that an append file would be the better solution, as it can be directed to the specific language track, and that restriction is very much dependent on the language, in terms of what should or should not be used in that track.  Those specifics can be addressed more locally than Problem Specifications can.

I get the wording "Restrictions" is pretty strong, yet no matter how we communicate to the student, it is up to them to determine where they get the value of "learning" from, even if it contradicts the restriction and the learn something by using something that is restricted.  In terms of "absolutes" there are ways to more directly restrict something programmatically, but also ways that a clever student can bypass those restrictions, and doing so may also provide learning opportunities to them.

It is not an easy problem.

The "confused me" problem is something that we likely deal with (OK, I deal with this) every day, and the solution to this is very often solving the problem (or not) but then submitting it for review and asking, which likely will give an opportunity for clarity.

> Additionally, the second sentence — Instead, try solving this yourself using other tools — feels superfluous at this point.

As I mentioned in the forum, I would interpret this as "provide this functionality yourself", regardless of the "other tools" since solving this myself will provide that "other tool" at that point.  Is it superfluous?  By definition it definitely is as it is extra confirmation that the solution is "something other than a built in solution".

habere-et-dispertire:
Images added to introduction -- not these instructions.

[ Added 4 SVG images #2501 ]( https://github.com/exercism/problem-specifications/pull/2501 )

tasxatzial:
Alright, i'm closing this. Back to the forum.

PR #2524, opened 2025-01-31
grains exercise: Split into introduction and instructions

tasxatzial:
Forum discussion
https://forum.exercism.org/t/grains-exercise-split-into-introduction-and-instructions/15361

Also:

* Minor revision to the instructions, nothing substantial


exercises/grains/instructions.md
```diff
@@ -0,0 +1,11 @@
+# Instructions
+
+Calculate the number of grains of wheat on a chessboard.
+
+A chessboard has 64 squares.
+Square 1 has one grain, square 2 has two grains, square 3 has four grains, and so on, doubling each time.
+
+Write code that calculates:
```
BNAndras, line 8: On at least one track (Batch), we only test the number of grains on specific squares so referring to two separate calculations might be confusing. We could simply add an append that Batch doesn't test the total number of grains, but dropping lines 8 - 11 would also be consistent with the documentation for Difference of Squares. There, most test suites expect three functions, but we don't describe each one in the docs.

tasxatzial: Sure, but this isn't something that can be addressed in this PR at this stage. I feel that people are going to have different opinions on this topic. So it requires a forum discussion first and is also outside the scope of this PR.

IsaacG (approved):

exercises/grains/instructions.md
BNAndras: I’m not sure it’s out of scope though. I recall past #48in24 PRs to split exercise documentation where the material itself was edited.

IsaacG: This PR doesn't significantly alter the text; it splits the file into two. The text might be ripe for updates, but it might be best to discuss that separately from the split. I don't think the split needs to also update the text at the same time; those are separate purposed.

tasxatzial: I was under the impression that changes involving major modifications should be discussed in the forum first. I'm open to adjusting the PR to accommodate changes. However, your suggestion is quite interesting, and since it involves other exercises and tracks, it would be helpful to have a clearer perspective before making any changes here. A forum discussion seems more suitable and would also make me feel more comfortable, especially since this is a topic I'm not very familiar with.

tasxatzial: Okay, let's keep this PR focused, and any kind of forum decision can go in a subsequent PR.

Cool-Katt (approved):

ErikSchierboom (approved):

PR #2528, opened 2025-02-09
Label alphabets as English rather than Latin

ellnix:
See the forum post for more details:
https://forum.exercism.org/t/rotational-cipher-atbash-cipher-affine-cipher-the-english-alphabet-is-not-the-latin-roman-alphabet/15618


IsaacG:
See ongoing discussion on the forum. Please use the forum for any discussion until a consensus is reached.

tasxatzial (dismissed):

ellnix:
The discussion is not closed, but the change requested by this PR does not seem to be popular. Probably the change will be different, if there will be a change at all.

PR #2537, opened 2025-03-02
feat: add pathfinding exercise

vaeng:
This is an idea I proposed on [Discord ](https://discord.com/channels/854117591135027261/1344967617391231020/1344970101618905148) as a reaction to @ErikSchierboom's call for exercises.

### DIFF
--- a/exercises/relative-distance/instructions.md
+++ b/exercises/relative-distance/instructions.md
@@ -14,7 +14,7 @@ Given the following family tree:
 
 ```text
       ┌──────────┐            ┌──────────┐ ┌───────────┐
-      │  Helena  │            │  Helena  │ │  Shusaku  │
+      │  Helena  │            │  Erdős   │ │  Shusaku  │
       └───┬───┬──┘            └─────┬────┘ └──────┬────┘
       ┌───┘   └───────┐             └──────┬──────┘
       ▼               ▼                    ▼
--- a/exercises/relative-distance/instructions.md
+++ b/exercises/relative-distance/instructions.md
@@ -13,20 +13,22 @@ Your task is to determine the degree of separation between two individuals in a 
 Given the following family tree:
 
 ```text
-      ┌──────────┐            ┌──────────┐ ┌──────────┐
-      │  Helena  │            │  Helena  │ │   Umar   │
-      └───┬───┬──┘            └─────┬────┘ └─────┬────┘
-      ┌───┘   └───────┐             └──────┬─────┘
+      ┌──────────┐            ┌──────────┐ ┌───────────┐
+      │  Helena  │            │  Helena  │ │  Shusaku  │
+      └───┬───┬──┘            └─────┬────┘ └──────┬────┘
+      ┌───┘   └───────┐             └──────┬──────┘
       ▼               ▼                    ▼
 ┌──────────┐     ┌────────┐          ┌──────────┐
-│   Isla   │     │  Tariq │          │   Kaito  │
+│   Isla   │     │  Tariq │          │   Kevin  │
 └────┬─────┘     └────┬───┘          └──────────┘
      ▼                ▼
-┌─────────┐       ┌───────┐
-│   Uma   │       │ Aditi │
-└─────────┘       └───────┘
+┌─────────┐      ┌────────┐
+│   Uma   │      │ Morphy │
+└─────────┘      └────────┘
 ```
 
 The degree of separation between Tariq and Uma is 3 (Tariq → Helena → Isla → Uma).
-The degree of separation between Isla and Kaito is not defined, as there is no visible relationship in the given data.
+The degree of separation between Isla and [Kevin][six-bacons] is not defined, as there is no visible relationship in the given data.
 The degree of separation between Uma and Isla is 1.
+
+[six-bacons]: https://en.m.wikipedia.org/wiki/Six_Degrees_of_Kevin_Bacon
--- a/exercises/relative-distance/instructions.md
+++ b/exercises/relative-distance/instructions.md
@@ -1,0 +1,32 @@
+# Instructions
+
+Your task is to determine the degree of separation between two individuals in a family tree.
+
+- You will be given an input, with all parent names and their children.
+- Each name is unique, a child _can_ have one or more parents.
+- The degree of separation is defined as the shortest number of connections from one person to another.
+- If two individuals are not connected, return a value that represents "no relation."
+  Please see the test cases for the actual implementation.
+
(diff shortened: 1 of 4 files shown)


IsaacG (approved):

BNAndras:
Degrees of separation make me think of https://en.m.wikipedia.org/wiki/Six_Degrees_of_Kevin_Bacon. Maybe we could add a Kevin as a homage?

exercises/relative-distance/instructions.md
```diff
@@ -0,0 +1,34 @@
...
+
+```text
+      ┌──────────┐            ┌──────────┐ ┌───────────┐
+      │  Helena  │            │  Erdős   │ │  Shusaku  │
+      └───┬───┬──┘            └─────┬────┘ └──────┬────┘
+      ┌───┘   └───────┐             └──────┬──────┘
+      ▼               ▼                    ▼
+┌──────────┐     ┌────────┐          ┌──────────┐
+│   Isla   │     │  Tariq │          │   Kevin  │
+└────┬─────┘     └────┬───┘          └──────────┘
```
kotp, line 23: Nice touch, that must be Kevin Bacon...

vaeng: It is. Thanks to @BNAndras for the trip down memory lane when we did this with other film nerds in my "sneak preview" circle.

BethanyG (approved):

exercises/relative-distance/introduction.md
```diff
@@ -0,0 +1,11 @@
+# Introduction
+
+You've been hired to develop **Noble Knots**, the hottest new dating app for nobility!
+With centuries of royal intermarriage, things have gotten… _complicated_.
+To avoid any _oops-we're-twins_ situations, your job is to build a system that checks how closely two people are related.
+
+Inspired by Iceland’s famous "[Islendinga-App][islendiga-app]" (which prevents accidental dating of a close relative), your algorithm will determine the **degree of separation** between two individuals in the royal family tree.
```
tasxatzial, line 7:
```suggestion
Inspired by Iceland's famous "[Islendinga-App][islendiga-app]" (which prevents accidental dating of a close relative), your algorithm will determine the **degree of separation** between two individuals in the royal family tree.
```

exercises/relative-distance/instructions.md
```diff
@@ -0,0 +1,34 @@
...
+     ▼                ▼
+┌─────────┐      ┌────────┐
+│   Uma   │      │ Morphy │
+└─────────┘      └────────┘
+```
+
+The degree of separation between Tariq and Uma is 3 (Tariq → Helena → Isla → Uma).
+The degree of separation between Isla and [Kevin][six-bacons] is not defined, as there is no visible relationship in the given data.
```
tasxatzial, line 31: If the "not defined" here is the same as "no relation" on line 8, then it might be better to standardize the terminology.

exercises/relative-distance/introduction.md
BNAndras, line 7: The app itself isn’t famous nor is it even available any more. The database with the genealogical information should be the focus. I’d also remove the accidental dating part because that’s a thing  the app added. The database itself is a standard genealogical database so it doesn’t stop accidental dating any more than FamilySearch.

exercises/relative-distance/instructions.md
kotp: "No known relationship" as it probably can not claim absolute knowledge.

exercises/relative-distance/introduction.md
ErikSchierboom: I second this comment.

exercises/relative-distance/instructions.md
```diff
@@ -0,0 +1,34 @@
+# Instructions
+
+Your task is to determine the degree of separation between two individuals in a family tree.
+
+- You will be given an input, with all parent names and their children.
+- Each name is unique, a child _can_ have one or more parents.
```
ErikSchierboom, line 6: Maybe:

```suggestion
- Each name is unique, a child _can_ have one or two parents.
```

ErikSchierboom (commented):
Lovely!

exercises/relative-distance/introduction.md
vaeng: Oh, I had no idea the app was discontinued. When it came out, it was quite the talk in my area, so I assumed it was well-known around the globe.

```suggestion
Noble Knots is inspired by Iceland’s "[Islendinga-App][islendiga-app]," which is backed up by a database that traces all known family connections between Icelanders from the time of the settlement of Iceland.
Your algorithm will determine the **degree of separation** between two individuals in the royal family tree.
```

vaeng: Thanks, I put that into the other suggestions, as they overlap.

exercises/relative-distance/instructions.md
tasxatzial: @vaeng Could we then be consistent across the exercise and change "no relation" to "no known relationship"?

vaeng: Good idea, I just pushed that.

jiegillet:
I've been trying to implement this one in Elixir, but I'm having trouble solving it, I think some cases may have issues.
1. "Direct parent-child relation" => degree 1, no problem
2. "Sibling relationship" => expected degree between siblings is 1, but from the description I expected it to be 2, with the path being: left child => parent => right child. Bypassing the parent for siblings is fine as a rule, but it should be explicit.
3. "Two degrees of separation, grandchild" => degree 2, no problem
4.  "Unrelated individuals" => no problem
5. "Complex graph, cousins" => degree 4, but my solution gave degree 9. I did it by hand and I also get 9 (with the rule of siblings having a degree of separation 1, otherwise you would get 10)
6. "Complex graph, no shortcut, far removed nephew" => degree 15, but my solution fives me 14 (it could be 15 without the siblings rule potentially, I did not do this one by hand)
7. "Complex graph, some shortcuts, cross-down and cross-up, cousins three times removed" => degree 8, but my solution gives me no connection, and I confirmed by hand. Sofia doesn't have a parent, and she starts a second family with Qi in it.

@vaeng could you help me here?

Also in general, I like having large inputs, but only at the end when all base case have been covered. The cousin case, for example could be much smaller, it took me a long time to verify by hand.
I also noticed that the Elixir exercise generator does not preserve the ordering of the family tree entries, but with random order I never would have been able to check my solutions by hand, so maybe it's worth adding a comment saying that preserving the order would be a good idea.

PR #2539, opened 2025-03-06
Add `baffling-birthdays` exercise

ErikSchierboom:
This exercise combines two subjects that I feel we are still lacking exercises for:

1. Dates
2. Randomness

I've created a sample C# implementation for reference: https://github.com/exercism/csharp/pull/2402

### DIFF
--- a/exercises/baffling-birthdays/canonical-data.json
+++ b/exercises/baffling-birthdays/canonical-data.json
@@ -20,7 +20,6 @@
           },
           "expected": false
         },
-        ,
         {
           "uuid": "f7b3eb26-bcfc-4c1e-a2de-af07afc33f45",
           "description": "two birthdates with same year, month, and day",
--- a/exercises/baffling-birthdays/canonical-data.json
+++ b/exercises/baffling-birthdays/canonical-data.json
@@ -160,7 +160,7 @@
       "input": {
         "groupSize": 23
       },
-      "expected": 50.7
+      "expected": 50.729723
     },
     {
       "uuid": "e43e6b9d-d77b-4f6c-a960-0fc0129a0bc5",
--- a/exercises/baffling-birthdays/canonical-data.json
+++ b/exercises/baffling-birthdays/canonical-data.json
@@ -9,156 +9,176 @@
   ],
   "cases": [
     {
-      "uuid": "f7b3eb26-bcfc-4c1e-a2de-af07afc33f45",
-      "description": "matching birthday for same date",
-      "property": "matchingBirthday",
-      "input": {
-        "birthdate1": "2000-01-01",
-        "birthdate2": "2000-01-01"
-      },
-      "expected": true
-    },
-    {
-      "uuid": "7193409a-6e16-4bcb-b4cc-9ffe55f79b25",
-      "description": "matching birthday for same year and month but different day",
-      "property": "matchingBirthday",
-      "input": {
-        "birthdate1": "2012-05-09",
-        "birthdate2": "2012-05-17"
-      },
-      "expected": false
-    },
-    {
-      "uuid": "d04db648-121b-4b72-93e8-d7d2dced4495",
-      "description": "matching birthday for same month and day but different year",
-      "property": "matchingBirthday",
-      "input": {
-        "birthdate1": "1999-10-23",
-        "birthdate2": "1988-10-23"
-      },
-      "expected": true
-    },
-    {
-      "uuid": "3c8bd0f0-14c6-4d4c-975a-4c636bfdc233",
-      "description": "matching birthday for same year but different month and day",
-      "property": "matchingBirthday",
-      "input": {
-        "birthdate1": "2007-12-19",
-        "birthdate2": "2007-04-27"
-      },
-      "expected": false
-    },
-    {
-      "uuid": "df5daba6-0879-4480-883c-e855c99cdaa3",
-      "description": "matching birthday for different year, month and day",
-      "property": "matchingBirthday",
-      "input": {
-        "birthdate1": "1997-08-04",
-        "birthdate2": "1963-11-23"
-      },
-      "expected": false
-    },
-    {
-      "uuid": "70b38cea-d234-4697-b146-7d130cd4ee12",
-      "description": "random birthdates generates requested number of birthdays",
-      "scenarios": ["random"],
-      "property": "randomBirthdates",
-      "input": {},
-      "expected": "length == groupsize"
-    },
-    {
-      "uuid": "d9d5b7d3-5fea-4752-b9c1-3fcd176d1b03",
-      "description": "random birthdates have non-leap year",
-      "scenarios": ["random"],
-      "property": "randomBirthdates",
-      "input": {},
-      "expected": {
-        "years": {
-          "leapYear": false
+      "description": "matching birthday",
+      "cases": [
+        {
+          "uuid": "f7b3eb26-bcfc-4c1e-a2de-af07afc33f45",
+          "description": "same year, month, and day",
+          "property": "matchingBirthday",
+          "input": {
+            "birthdate1": "2000-01-01",
+            "birthdate2": "2000-01-01"
+          },
+          "expected": true
+        },
+        {
+          "uuid": "7193409a-6e16-4bcb-b4cc-9ffe55f79b25",
+          "description": "same year and month, but different day",
+          "property": "matchingBirthday",
+          "input": {
+            "birthdate1": "2012-05-09",
+            "birthdate2": "2012-05-17"
+          },
+          "expected": false
+        },
+        {
+          "uuid": "d04db648-121b-4b72-93e8-d7d2dced4495",
+          "description": "same month and day, but different year",
+          "property": "matchingBirthday",
+          "input": {
+            "birthdate1": "1999-10-23",
+            "birthdate2": "1988-10-23"
+          },
+          "expected": true
+        },
+        {
+          "uuid": "3c8bd0f0-14c6-4d4c-975a-4c636bfdc233",
+          "description": "same year, but different month and day",
+          "property": "matchingBirthday",
+          "input": {
+            "birthdate1": "2007-12-19",
+            "birthdate2": "2007-04-27"
+          },
+          "expected": false
+        },
+        {
+          "uuid": "df5daba6-0879-4480-883c-e855c99cdaa3",
+          "description": "different year, month, and day",
+          "property": "matchingBirthday",
+          "input": {
+            "birthdate1": "1997-08-04",
+            "birthdate2": "1963-11-23"
+          },
+          "expected": false
         }
-      }
+      ]
     },
     {
-      "uuid": "d1074327-f68c-4c8a-b0ff-e3730d0f0521",
-      "description": "random birthdates have random month",
-      "scenarios": ["random"],
-      "property": "randomBirthdates",
-      "input": {},
-      "expected": {
-        "months": {
-          "random": true
+      "description": "random birthdates",
+      "cases": [
+        {
+          "uuid": "70b38cea-d234-4697-b146-7d130cd4ee12",
+          "description": "generate requested number of birthdates",
+          "scenarios": ["random"],
+          "property": "randomBirthdates",
+          "input": {},
+          "expected": "length == groupsize"
+        },
+        {
+          "uuid": "d9d5b7d3-5fea-4752-b9c1-3fcd176d1b03",
+          "description": "years are not leap years",
+          "scenarios": ["random"],
+          "property": "randomBirthdates",
+          "input": {},
+          "expected": {
+            "years": {
+              "leapYear": false
+            }
+          }
+        },
+        {
+          "uuid": "d1074327-f68c-4c8a-b0ff-e3730d0f0521",
+          "description": "months are random",
+          "scenarios": ["random"],
+          "property": "randomBirthdates",
+          "input": {},
+          "expected": {
+            "months": {
+              "random": true
+            }
+          }
+        },
+        {
+          "uuid": "7df706b3-c3f5-471d-9563-23a4d0577940",
+          "description": "days are random",
+          "scenarios": ["random"],
+          "property": "randomBirthdates",
+          "input": {},
+          "expected": {
+            "days": {
+              "random": true
+            }
+          }
         }
-      }
+      ]
     },
     {
-      "uuid": "7df706b3-c3f5-471d-9563-23a4d0577940",
-      "description": "random birthdates have random day",
-      "scenarios": ["random"],
-      "property": "randomBirthdates",
-      "input": {},
-      "expected": {
-        "days": {
-          "random": true
+      "description": "has shared birthdays",
+      "cases": [
+        {
+          "uuid": "ade37c87-f41d-4929-962a-286b4d1d048a",
+          "description": "with match",
+          "property": "hasSharedBirthday",
+          "input": {
+            "birthdates": ["1970-01-19", "1975-06-03", "2003-01-19"]
+          },
+          "expected": true
+        },
+        {
+          "uuid": "1d155b33-c6e1-46b9-81fa-09123ae378ea",
+          "description": "without match",
+          "property": "hasSharedBirthday",
+          "input": {
+            "birthdates": ["1984-04-05", "2000-09-17", "1966-10-12"]
+          },
+          "expected": false
         }
-      }
+      ]
     },
     {
-      "uuid": "ade37c87-f41d-4929-962a-286b4d1d048a",
-      "description": "has matching birthdays with matching birthday pair",
-      "property": "hasMatchingBirthdays",
-      "input": {
-        "birthdates": ["1970-01-19", "1975-06-03", "2003-01-19"]
-      },
-      "expected": true
-    },
-    {
-      "uuid": "1d155b33-c6e1-46b9-81fa-09123ae378ea",
-      "description": "has matching birthdays without matching birthday pair",
-      "property": "hasMatchingBirthdays",
-      "input": {
-        "birthdates": ["1984-04-05", "2000-09-17", "1966-10-12"]
-      },
-      "expected": false
-    },
-    {
-      "uuid": "89a462a4-4265-4912-9506-fb027913f221",
-      "description": "probability for matching birthday pair for one person",
-      "scenarios": ["random"],
-      "property": "probabilityForMatchingBirthdays",
-      "input": {
-        "groupSize": 1
-      },
-      "expected": 0.0
-    },
-    {
-      "uuid": "ec31c787-0ebb-4548-970c-5dcb4eadfb5f",
-      "description": "probability for matching birthday pair for ten people",
-      "scenarios": ["random"],
-      "property": "probabilityForMatchingBirthdays",
-      "input": {
-        "groupSize": 10
-      },
-      "expected": 11.694818
-    },
-    {
-      "uuid": "b548afac-a451-46a3-9bb0-cb1f60c48e2f",
-      "description": "probability for matching birthday pair for twenty-three people",
-      "scenarios": ["random"],
-      "property": "probabilityForMatchingBirthdays",
-      "input": {
-        "groupSize": 23
-      },
-      "expected": 50.729723
-    },
-    {
-      "uuid": "e43e6b9d-d77b-4f6c-a960-0fc0129a0bc5",
-      "description": "probability for matching birthday pair for seventy people",
-      "scenarios": ["random"],
-      "property": "probabilityForMatchingBirthdays",
-      "input": {
-        "groupSize": 70
-      },
-      "expected": 99.915958
+      "description": "probability of at least one shared birthday",
+      "cases": [
+        {
+          "uuid": "89a462a4-4265-4912-9506-fb027913f221",
+          "description": "for one person",
+          "scenarios": ["random"],
+          "property": "probabilityForSharedBirthday",
+          "input": {
+            "groupSize": 1
+          },
+          "expected": 0.0
+        },
+        {
+          "uuid": "ec31c787-0ebb-4548-970c-5dcb4eadfb5f",
+          "description": "among ten people",
+          "scenarios": ["random"],
+          "property": "probabilityForSharedBirthday",
+          "input": {
+            "groupSize": 10
+          },
+          "expected": 11.694818
+        },
+        {
+          "uuid": "b548afac-a451-46a3-9bb0-cb1f60c48e2f",
+          "description": "among twenty-three people",
+          "scenarios": ["random"],
+          "property": "probabilityForSharedBirthday",
+          "input": {
+            "groupSize": 23
+          },
+          "expected": 50.729723
+        },
+        {
+          "uuid": "e43e6b9d-d77b-4f6c-a960-0fc0129a0bc5",
+          "description": "among seventy people",
+          "scenarios": ["random"],
+          "property": "probabilityForSharedBirthday",
+          "input": {
+            "groupSize": 70
+          },
+          "expected": 99.915958
+        }
+      ]
     }
   ]
 }
--- a/exercises/baffling-birthdays/canonical-data.json
+++ b/exercises/baffling-birthdays/canonical-data.json
@@ -150,7 +150,7 @@
       "input": {
         "groupSize": 10
       },
-      "expected": 11.7
+      "expected": 11.694818
     },
     {
       "uuid": "b548afac-a451-46a3-9bb0-cb1f60c48e2f",
--- a/exercises/baffling-birthdays/canonical-data.json
+++ b/exercises/baffling-birthdays/canonical-data.json
@@ -50,7 +50,7 @@
     },
     {
       "uuid": "df5daba6-0879-4480-883c-e855c99cdaa3",
-      "description": "matching birthday for different year month and day",
+      "description": "matching birthday for different year, month and day",
       "property": "matchingBirthday",
       "input": {
         "birthdate1": "1997-08-04",
@@ -122,7 +122,7 @@
     },
     {
       "uuid": "89a462a4-4265-4912-9506-fb027913f221",
-      "description": "probability for matching birthday for one person",
+      "description": "probability for matching birthday pair for one person",
       "scenarios": ["random"],
       "property": "probabilityForMatchingBirthdays",
       "input": {
@@ -132,7 +132,7 @@
     },
     {
       "uuid": "ec31c787-0ebb-4548-970c-5dcb4eadfb5f",
-      "description": "probability for matching birthday for ten people",
+      "description": "probability for matching birthday pair for ten people",
       "scenarios": ["random"],
       "property": "probabilityForMatchingBirthdays",
       "input": {
@@ -142,7 +142,7 @@
     },
     {
       "uuid": "b548afac-a451-46a3-9bb0-cb1f60c48e2f",
-      "description": "probability for matching birthday for twenty-three people",
+      "description": "probability for matching birthday pair for twenty-three people",
       "scenarios": ["random"],
       "property": "probabilityForMatchingBirthdays",
       "input": {
@@ -152,7 +152,7 @@
     },
     {
       "uuid": "e43e6b9d-d77b-4f6c-a960-0fc0129a0bc5",
-      "description": "probability for matching birthday for seventy people",
+      "description": "probability for matching birthday pair for seventy people",
       "scenarios": ["random"],
       "property": "probabilityForMatchingBirthdays",
       "input": {
--- a/exercises/baffling-birthdays/canonical-data.json
+++ b/exercises/baffling-birthdays/canonical-data.json
@@ -12,54 +12,102 @@
       "description": "shared birthday",
       "cases": [
         {
-          "uuid": "f7b3eb26-bcfc-4c1e-a2de-af07afc33f45",
-          "description": "same year, month, and day",
+          "uuid": "716dcc2b-8fe4-4fc9-8c48-cbe70d8e6b67",
+          "description": "one birthdate",
           "property": "sharedBirthday",
           "input": {
-            "birthdate1": "2000-01-01",
-            "birthdate2": "2000-01-01"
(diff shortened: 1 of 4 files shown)


exercises/baffling-birthdays/introduction.md
```diff
@@ -0,0 +1,25 @@
...
+Over 70 people have shown up, including your mildly eccentric Uncle Ted.
+
+In one of his usual antics, he bets you £100 that at least two people in the room share the same birthday.
+That sounds ridiculous—there are 365 possible birthdays, so you confidently accept.
+
+To your astonishment, after collecting just 32 birthdays, you've already found a match.
+Magnanimous, you hand Uncle Ted his £100, but something feels off.
+
+The next day, curiosity gets the better of you.
+A quick web search leads you to the [Birthday Problem][birthday-problem], which reveals that with just 23 people, the probability of a shared birthday exceeds 50%.
```
BNAndras, line 13:
```suggestion
A quick web search leads you to the [birthday paradox][birthday-problem], which reveals that with just 23 people, the probability of a shared birthday exceeds 50%.
```

We should clearly refer to this as the birthday paradox. That'll align with how the Wikipedia article distinguishes the birthday problem and the paradox. Later on we also refer to "other paradoxes" without introducing a paradox so that's awkwardly written without this change.

exercises/baffling-birthdays/introduction.md
```diff
@@ -0,0 +1,25 @@
...
+
+Ah. So _that's_ why Uncle Ted was so confident.
+
+Determined to turn the tables, you start looking up other paradoxes—next time, _you'll_ be the one making the bets.
+
+~~~~exercism/note
+The birthday paradox is a veridical paradox: even though it feels wrong, it is actually true.
```
BNAndras, line 20:
```suggestion
The birthday paradox is a [veridical paradox][veridical-paradox]: even though it feels wrong, it is actually true.
```

exercises/baffling-birthdays/instructions.md
```diff
@@ -0,0 +1,15 @@
+# Instructions
+
+Your task is to implement a solution that verifies the Birthday Paradox's probabilities.
+
+To do this, you need to:
+
+- Determine if two birthdates match (same month and day).
```
tasxatzial, line 7: This is slightly misleading. Two birthdates match if they have the same month, day, year. Same month and day isn't necessarily a match.

exercises/baffling-birthdays/canonical-data.json
```diff
@@ -0,0 +1,176 @@
...
+    "",
+    "The expected probability values should be compared using some tolerance to allow for small deviations."
+  ],
+  "cases": [
+    {
+      "uuid": "f7b3eb26-bcfc-4c1e-a2de-af07afc33f45",
+      "description": "matching birthday for same date",
+      "property": "matchingBirthday",
+      "input": {
+        "birthday1": "2000-01-01",
```
tasxatzial, line 16: birthdate1 might be a better name if the full date is provided

tasxatzial (commented):
I feel there may be some confusion between 'birthday' and 'birthdate' among non-native speakers. It might be helpful to clearly explain the difference in the instructions.

exercises/baffling-birthdays/instructions.md
BNAndras:
```suggestion
- Determine if two birthdates have the same month and day.
```

exercises/baffling-birthdays/canonical-data.json
BNAndras: I strongly agree. The test case description also references birthdates that happen to have matching birthdays so the input names should match.

exercises/baffling-birthdays/introduction.md
```diff
@@ -0,0 +1,25 @@
+# Introduction
+
+Fresh out of graduation, you're throwing a huge party to celebrate with friends and family.
+Over 70 people have shown up, including your mildly eccentric Uncle Ted.
+
+In one of his usual antics, he bets you £100 that at least two people in the room share the same birthday.
+That sounds ridiculous—there are 365 possible birthdays, so you confidently accept.
```
BNAndras, line 7:
```suggestion
That sounds ridiculous—there are a lot of possible birthdays, so you confidently accept.
```
Leap years have 366 possible birthdays so I don't think we should specify a number. Referring to the count as some sort of large number still conveys the point that it's surprising the person found a match so quickly.

exercises/baffling-birthdays/instructions.md
```diff
@@ -0,0 +1,15 @@
+# Instructions
+
+Your task is to implement a solution that verifies the Birthday Paradox's probabilities.
```
IsaacG, line 3:
```suggestion
Your task is to verify the Birthday Paradox's probabilities.
```

exercises/baffling-birthdays/introduction.md
```diff
@@ -0,0 +1,25 @@
+# Introduction
+
+Fresh out of graduation, you're throwing a huge party to celebrate with friends and family.
```
IsaacG, line 3: out of college/after graduation

exercises/baffling-birthdays/introduction.md
IsaacG, line 7:
```suggestion
That sounds ridiculous — there are 365 possible birthdays, so you confidently accept.
```

exercises/baffling-birthdays/introduction.md
```diff
@@ -0,0 +1,25 @@
...
+Over 70 people have shown up, including your mildly eccentric Uncle Ted.
+
+In one of his usual antics, he bets you £100 that at least two people in the room share the same birthday.
+That sounds ridiculous—there are 365 possible birthdays, so you confidently accept.
+
+To your astonishment, after collecting just 32 birthdays, you've already found a match.
+Magnanimous, you hand Uncle Ted his £100, but something feels off.
```
IsaacG, line 10: Magnanimous or Magnanimously?

exercises/baffling-birthdays/introduction.md
```diff
@@ -0,0 +1,25 @@
...
+
+The next day, curiosity gets the better of you.
+A quick web search leads you to the [Birthday Paradox][birthday-problem], which reveals that with just 23 people, the probability of a shared birthday exceeds 50%.
+
+Ah. So _that's_ why Uncle Ted was so confident.
+
+Determined to turn the tables, you start looking up other paradoxes—next time, _you'll_ be the one making the bets.
```
IsaacG, line 17:
```suggestion
Determined to turn the tables, you start looking up other paradoxes; next time, _you'll_ be the one making the bets.
```

exercises/baffling-birthdays/introduction.md
BNAndras: I think the wording is right since magnanimous as an adjective is describing the subject of the sentence (you) rather than the verb (hand). We might want to drop it though to avoid taking the reader out of the story.

exercises/baffling-birthdays/introduction.md
BNAndras: Is that a hyphen or an em dash? I’m on my phone currently so I can’t tell, but generally I’d expect an em dash (the long dash) to be used to conjoin independent clauses. I think spaces or no spaces around the em dash depends on the style guide you use. I haven’t had to write a term paper in over a decade so I’m not sure which style guide would be appropriate if we had to pick one.

exercises/baffling-birthdays/introduction.md
BNAndras: Would `fresh out of school` be confusing to folks? That’s how I would say it vernacularly.

exercises/baffling-birthdays/instructions.md
BNAndras, line 3:
```suggestion
Your task is to implement a solution that verifies the birthday paradox's probabilities.
```

This isn’t a proper noun as far as I can tell, and Wikipedia doesn’t capitalize it either.

exercises/baffling-birthdays/instructions.md
```diff
@@ -0,0 +1,15 @@
...
+To do this, you need to:
+
+- Determine if two birthdates have the same month and day.
+- Generate random birthdates.
+- Check if a set of randomly generated birthdates contains at least one matching pair.
+- Calculate the probability of at least one match for different group sizes.
+
+~~~~exercism/caution
+The Birthday Paradox assumes that birthdays are uniformly distributed within a single year.
+While the year should vary between different random birthdate generation calls, all birthdates within a single call must share the same year.
```
jiegillet, line 14: We might also want to address leap years. The probability to be born on Feb 29 is 4(ish) time smaller than other dates, which changes the probability ever so slightly. I'd suggest adding that we can also assume that a year is 365 days.

exercises/baffling-birthdays/instructions.md
```diff
@@ -0,0 +1,15 @@
+# Instructions
+
+Your task is to implement a solution that verifies the Birthday Paradox's probabilities.
+
+To do this, you need to:
+
+- Determine if two birthdates have the same month and day.
+- Generate random birthdates.
+- Check if a set of randomly generated birthdates contains at least one matching pair.
+- Calculate the probability of at least one match for different group sizes.
```
jiegillet, line 10: "Calculate" sounds like "Find the an analytical solution", which can easily be done for this problem. I think you could use "Estimate" instead if you want .

(and just for fun, let me see if I can find the solution)
The probability of at least one match in N people is 1 minus the probability of no match at all.
The probability to pick a different date N times in a pool of D (365) days is:
$1 * (D-1)/D * (D-2)/D * ... * (D-N+1)/D = D!/N!/D^N$
(first pick is guaranteed, second pick only has D-1 choices, etc)
that's if N <= D, if N > D, the probability is 0.

ErikSchierboom: Fair point! I'll update

exercises/baffling-birthdays/instructions.md
ErikSchierboom: Good point.

exercises/baffling-birthdays/introduction.md
tasxatzial: It looks like an em dash. Spaces around it make it clearer that it's not a hyphen.

exercises/baffling-birthdays/introduction.md
tasxatzial: If the sentence is kept, i'd rather drop the word and use something that's easier to understand
```suggestion
Accepting your loss, you hand Uncle Ted his £100, but something feels off.
```

exercises/baffling-birthdays/instructions.md
ErikSchierboom: I've updated. Could you check if it is okay now?

exercises/baffling-birthdays/introduction.md
ErikSchierboom: I've rephrased

exercises/baffling-birthdays/canonical-data.json
```diff
@@ -0,0 +1,176 @@
...
+    "Dates are formatted as 'YYYY-MM-DD'.",
+    "",
+    "To increase the likelihood of consistent result, the tests should run the code returning random results many times.",
+    "",
+    "The expected probability values should be compared using some tolerance to allow for small deviations."
+  ],
+  "cases": [
+    {
+      "uuid": "f7b3eb26-bcfc-4c1e-a2de-af07afc33f45",
+      "description": "matching birthday for same date",
```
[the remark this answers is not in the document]
ErikSchierboom: Okay, but the suggestion does not mention anything about birthdays, and I'd like to keeo that in.

exercises/baffling-birthdays/canonical-data.json
```diff
@@ -0,0 +1,176 @@
...
+    },
+    {
+      "uuid": "ec31c787-0ebb-4548-970c-5dcb4eadfb5f",
+      "description": "probability for matching birthday for ten people",
+      "scenarios": ["random"],
+      "property": "probabilityForMatchingBirthdays",
+      "input": {
+        "groupSize": 10
+      },
+      "expected": 11.7
```
tasxatzial, line 153: If these values are to be compared using a tolerance, they should be as accurate as possible rather than approximations with only one decimal point. I believe 6-8 decimal places should suffice, allowing each track to decide how closely they want to test against those values.

tasxatzial, line 153: Rounded to 8 decimal places
```suggestion
      "expected": 11.694818
```

exercises/baffling-birthdays/canonical-data.json
```diff
@@ -0,0 +1,176 @@
...
+    },
+    {
+      "uuid": "b548afac-a451-46a3-9bb0-cb1f60c48e2f",
+      "description": "probability for matching birthday for twenty-three people",
+      "scenarios": ["random"],
+      "property": "probabilityForMatchingBirthdays",
+      "input": {
+        "groupSize": 23
+      },
+      "expected": 50.7
```
tasxatzial, line 163:
```suggestion
      "expected": 50.729723
```

exercises/baffling-birthdays/canonical-data.json
```diff
@@ -0,0 +1,176 @@
...
+    },
+    {
+      "uuid": "e43e6b9d-d77b-4f6c-a960-0fc0129a0bc5",
+      "description": "probability for matching birthday for seventy people",
+      "scenarios": ["random"],
+      "property": "probabilityForMatchingBirthdays",
+      "input": {
+        "groupSize": 70
+      },
+      "expected": 99.9
```
tasxatzial, line 173:
```suggestion
      "expected": 99.915958
```

exercises/baffling-birthdays/instructions.md
jiegillet: The "no leap year" line works, but I don't understand the point of "While the year should vary between different random birthdate generation calls, all birthdates within a single call _must_ share the same year.". Is this to further avoid leap year issues? If so, maybe it would work better to do filter out any Feb29s, or ask not generate them at all. Or we could leave it to the students to figure out how to avoid leap years.

ErikSchierboom:
I've tried to update the instructions a bit to be more clear on what rules the student is expected to follow. I've also updated the canonical data. I hope it is now clearer.

exercises/baffling-birthdays/instructions.md
```diff
@@ -0,0 +1,19 @@
+# Instructions
+
+Your task is to estimate the birthday paradox's probabilities.
+
+To do this, you need to:
+
+- Determine if two birthdates have the same month and day.
+- Generate random birthdates.
+- Check if a collection of randomly generated birthdates contains at least one matching pair.
```
tasxatzial, line 9: This still talks about "matching birthdays", but we haven't defined what matching means. How about something similar to L7
```suggestion
- Check if a collection of randomly generated birthdates contains any two with the same month and day.
```

jiegillet (approved):

exercises/baffling-birthdays/instructions.md
```diff
@@ -0,0 +1,19 @@
+# Instructions
+
+Your task is to estimate the birthday paradox's probabilities.
+
+To do this, you need to:
+
+- Determine if two birthdates have the same month and day.
+- Generate random birthdates.
+- Check if a collection of randomly generated birthdates contains any two with the same month and day.
+- Estimate the probability of at least one match for different group sizes.
```
tasxatzial, line 10: This is asking for the `probabilityForMatchingBirthdays`, right? I think can be rephrased so that it's more straightforward and clear.
```suggestion
- Estimate the probability that at least two people in a group share the same birthday (month and day) for different group sizes.
```

exercises/baffling-birthdays/instructions.md
```diff
@@ -0,0 +1,19 @@
+# Instructions
+
+Your task is to estimate the birthday paradox's probabilities.
+
+To do this, you need to:
+
+- Determine if two birthdates have the same month and day.
+- Generate random birthdates.
+- Check if a collection of randomly generated birthdates contains any two with the same month and day.
+- Estimate the probability that at least two people in a group share the same birthday (month and day) for different group sizes.
```
ErikSchierboom, line 10: I feel that it is confusing _not_ to mention birthday, as it _is_ the birthday paradox after all. Maybe we can either explain it in parentheses:

```suggestion
- Determine if two birthdates (day/month/year) have the birthday (day/month).
- Generate random birthdates.
- Check if a collection of randomly generated birthdates contains any two with the same birthday.
- Estimate the probability that at least two people in a group share the same birthday for different group sizes.
```

or  else maybe use superscripts:

```suggestion
- Determine if two birthdates¹ have the birthday².
- Generate random birthdates.
- Check if a collection of randomly generated birthdates contains any two with the same birthday.
- Estimate the probability that at least two people in a group share the same birthday for different group sizes.

¹ A birthdate is the combination of day, month, and year
² A birthday is the combination of day and month
```

or something similar.

IsaacG: Rather than making these definitions footnotes, can we explicitly define, at the top of the instructions, how we intend these terms to be used/understood?

tasxatzial: I think the instructions are fine as they are. The only thing i would add is a note (maybe an exercism note) right after the bullets.

`Note: A birthdate includes the full date of birth (year, month, and day), whereas a birthday refers only to the month and day, which repeat each year. Two birthdates with the same month and day correspond to the same birthday.`

ErikSchierboom: I've updated

ErikSchierboom:
I'm wondering if the properties should maybe be renamed at bit:

- `matchingBirthday` -> `matchingBirthdayPair`
- `hasMatchingBirthdays` -> `hasMatchingBirthdayPair`
- `probabilityForMatchingBirthdays` -> `probabilityForMatchingBirthdayPair`

exercises/baffling-birthdays/canonical-data.json
```diff
@@ -0,0 +1,164 @@
...
+          "random": true
+        }
+      }
+    },
+    {
+      "uuid": "ade37c87-f41d-4929-962a-286b4d1d048a",
+      "description": "has matching birthdays with matching birthday pair",
+      "property": "hasMatchingBirthdays",
+      "input": {
+        "birthdays": ["1970-01-19", "1975-06-03", "2003-01-19"]
```
tasxatzial, line 110: This looks like it should have been birthdates
```suggestion
        "birthdates": ["1970-01-19", "1975-06-03", "2003-01-19"]
```

exercises/baffling-birthdays/canonical-data.json
```diff
@@ -0,0 +1,164 @@
...
+      },
+      "expected": true
+    },
+    {
+      "uuid": "1d155b33-c6e1-46b9-81fa-09123ae378ea",
+      "description": "has matching birthdays without matching birthday pair",
+      "property": "hasMatchingBirthdays",
+      "input": {
+        "birthdays": ["1984-04-05", "2000-09-17", "1966-10-12"]
```
tasxatzial, line 119:
```suggestion
        "birthdates": ["1984-04-05", "2000-09-17", "1966-10-12"]
```

tasxatzial:
> I'm wondering if the properties should maybe be renamed at bit:
> 
> * `matchingBirthday` -> `matchingBirthdayPair`
> * `hasMatchingBirthdays` -> `hasMatchingBirthdayPair`
> * `probabilityForMatchingBirthdays` -> `probabilityForMatchingBirthdayPair`

Tough call. I'll share my thoughts.

* The tests with `matchingBirthday` prop seem to have consistent description and props, and it's only 2 birthdates as input, so i don't find any reason for adding `pair` anywhere.

* The tests with `hasMatchingBirthdays` have odd descriptions:

  `has matching birthdays with matching birthday pair`
  `has matching birthdays without matching birthday pair`

  I think something like this makes more sense:

  `has matching birthday pair`
  `does not have matching birthday pair`

  Given that the input is an array of birthdates, the word `pair` might suggest a pair from the array, so birthdates pair. So to avoid any kind of confusion, I would just leave the word `pair` out and adjust the descriptions to:

  `has matching birthdays`
  `does not have matching birthdays`

  or something similar.

* The `probabilityForMatchingBirthdays` also have odd descriptions:

  `probability for matching birthday pair for ten people`

  This is a bit misleading because it suggests the probability of exactly one matching pair. A better way to phrase this so that it aligns with the paradox would be:

  `probability of at least one shared birthday among ten people`

  and i would leave anything related to `pair` out, for sure.

tasxatzial:
@ErikSchierboom Have you also considered nesting (->), or is it not appropriate for these tests?

ErikSchierboom:
In general we try to not use nesting, but in this case it might be appropriate due to several completely different properties being tested

ErikSchierboom:
I've used nesting and tweaked the properties again. Once again, we're proving that naming things is hard :)

ErikSchierboom:
There are now two variants (of which we need to choose one, or an alternative).

People with the same birthday are referred to as:

- sharing their birthday
- having a matching birthday

Which one do you prefer? Or should it be "same birthday?"

tasxatzial:
> Or should it be "same birthday?"

I think "have the same birthday" sounds more natural. "Sharing their birthday" sounds like they are celebrating together. "Matching birthday" does sound a bit awkward.

You'd have to do a ton of renaming though. Properties, descriptions, instructions...

ErikSchierboom:
I don't mind as I'm agreeing with you so I think it is a good idea

tasxatzial:
For the record, the Wikipedia article uses a mix of "share the same birthday" and "have the same birthday"

ErikSchierboom:
> For the record, the Wikipedia article uses a mix of "share the same birthday" and "have the same birthday"

I feel like "have" is less ambiguous that "share" (for the reason you mentioned earlier)

ErikSchierboom:
Okay, another iteration :) I've gone for "share" after all, as it seems to be used quite a lot in online material.

BNAndras (approved):

exercises/baffling-birthdays/introduction.md
```diff
@@ -0,0 +1,25 @@
+# Introduction
+
+Fresh out of college, you're throwing a huge party to celebrate with friends and family.
+Over 70 people have shown up, including your mildly eccentric Uncle Ted.
+
+In one of his usual antics, he bets you £100 that at least two people in the room share their birthdays.
```
tasxatzial, line 6: Consistency
```suggestion
In one of his usual antics, he bets you £100 that at least two people in the room share the same birthday.
```

exercises/baffling-birthdays/canonical-data.json
```diff
@@ -0,0 +1,184 @@
...
+          "expected": {
+            "days": {
+              "random": true
+            }
+          }
+        }
+      ]
+    },
+    {
+      "description": "contains shared birthdays",
```
tasxatzial, line 116: With this description the full description for the second case will become `contains shared birthdays -> without shared birthday`. This does not read well. What do you think about the following changes in the parent/child descriptions
```suggestion
      "description": "shared birthdays",
```

exercises/baffling-birthdays/canonical-data.json
```diff
@@ -0,0 +1,184 @@
...
+      "cases": [
+        {
+          "uuid": "ade37c87-f41d-4929-962a-286b4d1d048a",
+          "description": "with shared birthday",
```
tasxatzial, line 120:
```suggestion
          "description": "contains shared birthday",
```

exercises/baffling-birthdays/canonical-data.json
```diff
@@ -0,0 +1,184 @@
...
+          "property": "containsSharedBirthday",
+          "input": {
+            "birthdates": ["1970-01-19", "1975-06-03", "2003-01-19"]
+          },
+          "expected": true
+        },
+        {
+          "uuid": "1d155b33-c6e1-46b9-81fa-09123ae378ea",
+          "description": "without shared birthday",
```
tasxatzial, line 129:
```suggestion
          "description": "does not contain shared birthday",
```

exercises/baffling-birthdays/canonical-data.json
```diff
@@ -0,0 +1,184 @@
...
+        }
+      ]
+    },
+    {
+      "description": "contains shared birthdays",
+      "cases": [
+        {
+          "uuid": "ade37c87-f41d-4929-962a-286b4d1d048a",
+          "description": "with shared birthday",
+          "property": "containsSharedBirthday",
```
tasxatzial, line 121: A more appropriate name would be `containsSharedBirthdays` since the array may contain more than one pair of shared birthdays.

exercises/baffling-birthdays/canonical-data.json
```diff
@@ -0,0 +1,184 @@
...
+          "input": {
+            "birthdates": ["1970-01-19", "1975-06-03", "2003-01-19"]
+          },
+          "expected": true
+        },
+        {
+          "uuid": "1d155b33-c6e1-46b9-81fa-09123ae378ea",
+          "description": "without shared birthday",
+          "property": "containsSharedBirthday",
```
tasxatzial, line 130:
```suggestion
          "property": "containsSharedBirthdays",
```

exercises/baffling-birthdays/canonical-data.json
tasxatzial, line 121: A more appropriate name would be containsSharedBirthdays since the array may contain more than one pair of shared birthdays.
```suggestion
          "property": "containsSharedBirthdays",
```

tasxatzial:
I don't know if I'm missing something, but I've noticed that the props `sharedBirthday` and `containsSharedBirthday` have very similar tests. The first one is about checking two dates, the other is about checking an array of dates. Would it make sense to remove the `sharedBirthday` and move those tests to the `containsSharedBirthday` ?

ErikSchierboom:
As suggested, I've merged the `sharedBirthday` and `sharedBirthdays` properties.

exercises/baffling-birthdays/canonical-data.json
```diff
@@ -0,0 +1,209 @@
...
+        {
+          "uuid": "716dcc2b-8fe4-4fc9-8c48-cbe70d8e6b67",
+          "description": "one birthdate",
+          "property": "sharedBirthday",
+          "input": {
+            "birthdates": ["2000-01-01"]
+          },
+          "expected": false
+        },
+        ,
```
tasxatzial, line 23:
```suggestion
```

tasxatzial (commented):
I think it looks good. I only have one remaining suggestion from a previous review, and the schema error in this one.

ErikSchierboom:
Fixed both

exercises/baffling-birthdays/introduction.md
```diff
@@ -0,0 +1,25 @@
...
+
+Determined to turn the tables, you start looking up other paradoxes; next time, _you'll_ be the one making the bets.
+
+~~~~exercism/note
+The birthday paradox is a [veridical paradox][veridical-paradox]: even though it feels wrong, it is actually true.
+
+[veridical-paradox]: https://en.wikipedia.org/wiki/Paradox#Quine's_classification
+~~~~
+
+[birthday-problem]: https://en.wikipedia.org/wiki/Birthday_problem
```
tasxatzial, line 25:
```suggestion
[birthday-problem]: https://en.wikipedia.org/wiki/Birthday_problem
[veridical-paradox]: https://en.wikipedia.org/wiki/Paradox#Quine's_classification
```

exercises/baffling-birthdays/introduction.md
```diff
@@ -0,0 +1,25 @@
...
+A quick web search leads you to the [birthday paradox][birthday-problem], which reveals that with just 23 people, the probability of a shared birthday exceeds 50%.
+
+Ah. So _that's_ why Uncle Ted was so confident.
+
+Determined to turn the tables, you start looking up other paradoxes; next time, _you'll_ be the one making the bets.
+
+~~~~exercism/note
+The birthday paradox is a [veridical paradox][veridical-paradox]: even though it feels wrong, it is actually true.
+
+[veridical-paradox]: https://en.wikipedia.org/wiki/Paradox#Quine's_classification
```
tasxatzial, line 22: i believe this should no be inside the note?
```suggestion
```

exercises/baffling-birthdays/instructions.md
```diff
@@ -0,0 +1,24 @@
+# Instructions
+
+Your task is to estimate the birthday paradox's probabilities.
+
+To do this, you need to:
+
+- Determine if two birthdates have the same month and day.
```
tasxatzial, line 7:
```suggestion
```

exercises/baffling-birthdays/introduction.md
ErikSchierboom: Which it is?

tasxatzial: Apologies for the two separate comments.

exercises/baffling-birthdays/instructions.md
tasxatzial: I think it's fine to leave this one in.

exercises/baffling-birthdays/introduction.md
tasxatzial: > Which it is?

Both suggestions are valid. Remove from the note, and move at the end of the file

exercises/baffling-birthdays/introduction.md
tasxatzial, line 25: This was the correct suggestion. Ignore if the link renders correctly when inside the codeblock.
```suggestion
~~~~

[birthday-problem]: https://en.wikipedia.org/wiki/Birthday_problem
[veridical-paradox]: https://en.wikipedia.org/wiki/Paradox#Quine's_classification
```

tasxatzial (approved):

exercises/baffling-birthdays/introduction.md
ErikSchierboom: It does not only render correctly when it is in the note, but it has to be inside the note (for reasons)

tasxatzial: Thanks for letting me know, I wasn't aware of this.

PR #2565, opened 2025-05-15
Rewrite the Simple Cipher description.

IsaacG:
See [forum discussion](https://forum.exercism.org/t/simple-cipher-change-proposal-rewrite-the-description/17380)

Key improvements:

* Remove mentions of "Steps" and "Extension".
* Replace "Caeser cipher" with "Vigenère cipher" to accurately match
  what is being asked.
* Use a format in line with the other cipher exercises.

### DIFF
--- a/exercises/simple-cipher/description.md
+++ b/exercises/simple-cipher/description.md
@@ -1,66 +1,38 @@
 # Description
 
-Implement a simple shift cipher like Caesar and a more secure substitution cipher.
+Create an implementation the [Vigenère cipher][wiki].
+The Vigenère cipher is simple substitution cipher.
 
-## Step 1
+## Cipher terminology
 
-"If he had anything confidential to say, he wrote it in cipher, that is, by so changing the order of the letters of the alphabet, that not a word could be made out.
-If anyone wishes to decipher these, and get at their meaning, he must substitute the fourth letter of the alphabet, namely D, for A, and so with the others."
-—Suetonius, Life of Julius Caesar
+A cipher is an algorithm used to encrypt, or encode, a string.
+The unencrypted string is called the _plaintext_ and the encrypted string is called the _ciphertext_.
+Converting plaintext to ciphertext is called _encoding_ while the reverse is called _decoding_.
 
-Ciphers are very straight-forward algorithms that allow us to render text less readable while still allowing easy deciphering.
-They are vulnerable to many forms of cryptanalysis, but Caesar was lucky that his enemies were not cryptanalysts.
+In a _substitution cipher_, each letter is replaced with a different letter which is computed with the help of a _key_.
 
-The Caesar cipher was used for some messages from Julius Caesar that were sent afield.
-Now Caesar knew that the cipher wasn't very good, but he had one ally in that respect: almost nobody could read well.
-So even being a couple letters off was sufficient so that people couldn't recognize the few words that they did know.
+## Encoding details
 
-Your task is to create a simple shift cipher like the Caesar cipher.
-This image is a great example of the Caesar cipher:
+In this cipher, the key is a series of lowercase letters, such as `"abcd"`.
+Each letter of the plaintext is _shifted_ or _rotated_ by a distance based on a corresponding letter in the key.
+An `"a"` in the key means a shift of 0 (that is, no shift).
+A `"b"` in the key means a shift of 1.
+A `"c"` in the key means a shift of 2, and so on.
 
-![Caesar cipher][img-caesar-cipher]
+The first letter of the plaintext uses the first letter of the key, the second letter of the plaintext uses the second letter of the key and so on.
+If you run out of letters in the key before you run out of letters in the plaintext, start over from the start of the key again.
 
-For example:
+If the key only contains one letter, such as `"dddddd"`, then all letters of the plaintext are shifted by the same amount (three in this example), which would make this the same as a rotational cipher or shift cipher (sometimes called a Caesar cipher).
+For example, the plaintext `"iamapandabear"` would become `"ldpdsdqgdehdu"`.
 
-Giving "iamapandabear" as input to the encode function returns the cipher "ldpdsdqgdehdu".
-Obscure enough to keep our message secret in transit.
+If the key only contains the letter `"a"` (one or more times), the shift distance is zero and the ciphertext is the same as the plaintext.
 
-When "ldpdsdqgdehdu" is put into the decode function it would return the original "iamapandabear" letting your friend read your original message.
+Usually the key is more complicated than that, though!
+If the key is `"abcd"` then letters of the plaintext would be shifted by a distance of 0, 1, 2 then 3.
+Using that key, `"hello"` would become `"hfnoo"`.
 
-## Step 2
+## Random keys
 
-Shift ciphers quickly cease to be useful when the opposition commander figures them out.
-So instead, let's try using a substitution cipher.
-Try amending the code to allow us to specify a key and use that for the shift distance.
+If no key is provided, generate a key which consists of at least 100 random lowercase letters from the Latin alphabet.
 
-Here's an example:
-
-Given the key "aaaaaaaaaaaaaaaaaa", encoding the string "iamapandabear"
-would return the original "iamapandabear".
-
-Given the key "ddddddddddddddddd", encoding our string "iamapandabear"
-would return the obscured "ldpdsdqgdehdu"
-
-In the example above, we've set a = 0 for the key value.
-So when the plaintext is added to the key, we end up with the same message coming out.
-So "aaaa" is not an ideal key.
-But if we set the key to "dddd", we would get the same thing as the Caesar cipher.
-
-## Step 3
-
-The weakest link in any cipher is the human being.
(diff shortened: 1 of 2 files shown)


exercises/simple-cipher/description.md
```diff
@@ -1,66 +1,38 @@
...
+If the key is `"abcd"` then letters of the plaintext would be shifted by a distance of 0, 1, 2 then 3.
+Using that key, `"hello"` would become `"hfnoo"`.
 
-## Step 2
+## Random keys
 
-Shift ciphers quickly cease to be useful when the opposition commander figures them out.
-So instead, let's try using a substitution cipher.
-Try amending the code to allow us to specify a key and use that for the shift distance.
+If no key is provided, generate a key which consists of at least 100 random lowercase letters from the Latin alphabet.
```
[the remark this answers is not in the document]
IsaacG: It's an arbitrary requirement. There's no reason it needs to be "at least 100" vs "exactly 75" or "50 to 100".

The prior description included "at least 100" so I carried that over.

I don't see anything in the test data that enforces this, so I think it would be fine to change this requirement ... but I would suggest making that part of a separate discussion/PR. This PR rewrites the description without altering the requirements.

exercises/simple-cipher/description.md
```diff
@@ -1,66 +1,38 @@
...
+In a _substitution cipher_, each letter is replaced with a different letter which is computed with the help of a _key_.
 
-The Caesar cipher was used for some messages from Julius Caesar that were sent afield.
-Now Caesar knew that the cipher wasn't very good, but he had one ally in that respect: almost nobody could read well.
-So even being a couple letters off was sufficient so that people couldn't recognize the few words that they did know.
+## Encoding details
 
-Your task is to create a simple shift cipher like the Caesar cipher.
-This image is a great example of the Caesar cipher:
+In this cipher, the key is a series of lowercase letters, such as `"abcd"`.
```
tasxatzial, line 16: Should it mention that letters are from the Latin alphabet, so that it's consistent with the random keys?
```suggestion
In this cipher, the key is a series of lowercase letters from the Latin alphabet, such as `"abcd"`.
```

ErikSchierboom (approved):

exercises/simple-cipher/description.md
```diff
@@ -1,66 +1,38 @@
...
-"If he had anything confidential to say, he wrote it in cipher, that is, by so changing the order of the letters of the alphabet, that not a word could be made out.
-If anyone wishes to decipher these, and get at their meaning, he must substitute the fourth letter of the alphabet, namely D, for A, and so with the others."
-—Suetonius, Life of Julius Caesar
+A cipher is an algorithm used to encrypt, or encode, a string.
+The unencrypted string is called the _plaintext_ and the encrypted string is called the _ciphertext_.
+Converting plaintext to ciphertext is called _encoding_ while the reverse is called _decoding_.
 
-Ciphers are very straight-forward algorithms that allow us to render text less readable while still allowing easy deciphering.
-They are vulnerable to many forms of cryptanalysis, but Caesar was lucky that his enemies were not cryptanalysts.
+In a _substitution cipher_, each letter is replaced with a different letter which is computed with the help of a _key_.
```
tasxatzial, line 12: Based on the explanation below, the new letter can also be the same letter, right?
```suggestion
In a _substitution cipher_, each letter is replaced with another letter, computed with the help of a _key_.
```

exercises/simple-cipher/description.md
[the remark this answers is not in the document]
BNAndras: Okay, thanks. I was just curious.

exercises/simple-cipher/description.md
IsaacG: We could, but ... I'm not sure it's needed. I think it's obvious enough from the context and we already mention the Latin alphabet. Do you feel the description needs to mention it to be more clear/less confusing

exercises/simple-cipher/description.md
IsaacG: Correct. The mapping doesn't always change the value. I'm not sure that "another" is much more clear than "different". If this sentence is confusing, we could call it out explicitly.

```suggestion
In a _substitution cipher_, each plaintext letter is replaced with a ciphertext letter which is computed with the help of a _key_. (Note, it is possible for replacement letter to be the same as the original letter! As mentioned earlier, an `"a"` in the key means there is no shift/change.)
```

exercises/simple-cipher/description.md
tasxatzial: Once someone reads the section about random keys, they'll already know what the alphabet is. So it's clear enough as it is.

The suggestion is purely for consistency, and I don't even have a strong opinion on it. It can also stay as it is, I guess :)

tasxatzial:
metadata.toml also needs to be updated

IsaacG:
> metadata.toml also needs to be updated

Good catch! I completely missed that. Fixed.

exercises/simple-cipher/description.md
```diff
@@ -1,66 +1,40 @@
...
-—Suetonius, Life of Julius Caesar
+A cipher is an algorithm used to encrypt, or encode, a string.
+The unencrypted string is called the _plaintext_ and the encrypted string is called the _ciphertext_.
+Converting plaintext to ciphertext is called _encoding_ while the reverse is called _decoding_.
 
-Ciphers are very straight-forward algorithms that allow us to render text less readable while still allowing easy deciphering.
-They are vulnerable to many forms of cryptanalysis, but Caesar was lucky that his enemies were not cryptanalysts.
+In a _substitution cipher_, each plaintext letter is replaced with a ciphertext letter which is computed with the help of a _key_.
+(Note, it is possible for replacement letter to be the same as the original letter!
+As mentioned earlier, an `"a"` in the key means there is no shift/change.)
```
tasxatzial, line 14: This is a detail mentioned later rather than earlier. I think it's safe to remove this sentence, since the explanation comes a couple of sentences later.

exercises/simple-cipher/description.md
```diff
@@ -1,66 +1,39 @@
...
+For example, the plaintext `"iamapandabear"` would become `"ldpdsdqgdehdu"`.
 
-Giving "iamapandabear" as input to the encode function returns the cipher "ldpdsdqgdehdu".
-Obscure enough to keep our message secret in transit.
+If the key only contains the letter `"a"` (one or more times), the shift distance is zero and the ciphertext is the same as the plaintext.
 
-When "ldpdsdqgdehdu" is put into the decode function it would return the original "iamapandabear" letting your friend read your original message.
+Usually the key is more complicated than that, though!
+If the key is `"abcd"` then letters of the plaintext would be shifted by a distance of 0, 1, 2 then 3.
+Using that key, `"hello"` would become `"hfnoo"`.
```
tasxatzial, line 33: Would it make sense to put this example in its own subsection and show the process in more detail? For instance: "e" is shifted by 1, so it becomes the next letter, "f". "l" is shifted by 2, so it becomes the letter two positions after "l", and so on. For "o", we've run out of key letters, so we start over with "a", and the shift is 0. Basically, just more explanation about how the instructions are applied to the example.

IsaacG: :shrug: 
Either way makes sense. I think this is detailed enough to get students started. Do you think it needs to be more detailed for students to understand the algorithm and solve this exercise?

IsaacG: FWIW no one on the forum asked for more detail here, implicitly suggesting they feel this is sufficiently detailed as is.

tasxatzial: > I think this is detailed enough to get students started. Do you think it needs to be more detailed for students to understand the algorithm and solve this exercise?

I agree, this is enough to get started. It's enough for me since I already understand how the algorithm works, and it's probably enough for someone who figures it out from the input and output. So if the goal is for students to work out the algorithm based on the description and tests, then the given instructions are probably sufficient. It really depends on what you have in mind. Personally, I would prefer to see the algorithm in action to make things clearer. Even a simple rephrasing would help:

```If the plaintext is "hello" and the key is "abcd", the letters `h, e, l, l, o` are shifted by `0, 1, 2, 3, 0` positions respectively, resulting in "hfnoo".```

Comment (IsaacG) [reply exercises/simple-cipher/description.md:33]: So ... to clarify what you're saying, you specifically would like the example which calls out/shows the part about how the key is repeated?

**Review by keiravillekode — APPROVED** [review]

File: exercises/simple-cipher/description.md (hunk 5 above)
Comment (tasxatzial) [reply exercises/simple-cipher/description.md:33]: Both the key repetition and how individual letters are shifted. Basically what the rephrased version shows. It's what you already have, just explained in a little more detail.

Comment (IsaacG) [reply exercises/simple-cipher/description.md:33]: How about this?
```
If the key is `"abcd"` then letters of the plaintext would be shifted by a distance of 0, 1, 2, and 3.
If the plaintext is `"hello"`, we need 5 shifts so the key would wrap around, giving shift distances of 0, 1, 2, 3, and 0.
Applying those shifts to the letters of `"hello"` we get `"hfnoo"`.
```

tasxatzial: That should be fine 👍

tasxatzial (approved):

PR #2566, opened 2025-05-18
Relative distance fix

mk-mxp:
See [Incorrect degrees of separation in Relative Distance instructions](https://forum.exercism.org/t/incorrect-degrees-of-separation-in-relative-distance-instructions/17245) forum thread.


IsaacG (approved):

exercises/relative-distance/instructions.md
```diff
@@ -1,6 +1,7 @@
 # Instructions
 
 Your task is to determine the degree of separation between two individuals in a family tree.
+This is similar to the pop culture idea that everyone is [within six degrees of Kevin Bacon][six-bacons].
```
jiegillet, line 4: Isn't that every actor rather than everyone?
I mean, it might hold true for everyone, but the pop culture idea is about actors.

ErikSchierboom: Yeah, I think making it about just actors is a good idea

ErikSchierboom (approved):
Nice improvement!

exercises/relative-distance/instructions.md
BNAndras: I’d be in favor as well. The underlying concept is everyone should be up to six degrees of separation from one another, but the Kevin Bacon game is specifically about actors in Hollywood.

kotp (approved):

jiegillet (approved):

PR #2575, opened 2025-06-18
[ Flower Field ] Deprecate minesweeper for flower-field

habere-et-dispertire:
See:
https://forum.exercism.org/t/suggestion-deprecate-minesweeper-for-flower-field/17967

### DIFF
--- a/exercises/flower-field/canonical-data.json
+++ b/exercises/flower-field/canonical-data.json
@@ -30,36 +30,68 @@
       "description": "no flowers",
       "property": "annotate",
       "input": {
-        "garden": ["   ", "   ", "   "]
+        "garden": [
+          "   ",
+          "   ",
+          "   "
+        ]
       },
-      "expected": ["   ", "   ", "   "]
+      "expected": [
+        "   ",
+        "   ",
+        "   "
+      ]
     },
     {
       "uuid": "225176a0-725e-43cd-aa13-9dced501f16e",
       "description": "garden full of flowers",
       "property": "annotate",
       "input": {
-        "garden": ["***", "***", "***"]
+        "garden": [
+          "***",
+          "***",
+          "***"
+        ]
       },
-      "expected": ["***", "***", "***"]
+      "expected": [
+        "***",
+        "***",
+        "***"
+      ]
     },
     {
       "uuid": "3f345495-f1a5-4132-8411-74bd7ca08c49",
       "description": "flower surrounded by spaces",
       "property": "annotate",
       "input": {
-        "garden": ["   ", " * ", "   "]
+        "garden": [
+          "   ",
+          " * ",
+          "   "
+        ]
       },
-      "expected": ["111", "1*1", "111"]
+      "expected": [
+        "111",
+        "1*1",
+        "111"
+      ]
     },
     {
       "uuid": "6cb04070-4199-4ef7-a6fa-92f68c660fca",
       "description": "space surrounded by flowers",
       "property": "annotate",
       "input": {
-        "garden": ["***", "* *", "***"]
+        "garden": [
+          "***",
+          "* *",
+          "***"
+        ]
       },
-      "expected": ["***", "*8*", "***"]
+      "expected": [
+        "***",
+        "*8*",
+        "***"
+      ]
     },
     {
       "uuid": "272d2306-9f62-44fe-8ab5-6b0f43a26338",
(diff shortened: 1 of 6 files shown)


exercises/flower-field/canonical-data.json
```diff
@@ -0,0 +1,201 @@
+{
+  "exercise": "flower-field",
+  "comments": [
+    " The expected outputs are represented as arrays of strings to   ",
+    " improve readability in this JSON file.                         ",
+    " Your track may choose whether to present the input as a single ",
+    " string (concatenating all the lines) or as the list.           "
```
IsaacG:
```suggestion
    "The expected outputs are represented as arrays of strings to",
    "improve readability in this JSON file.",
    "Your track may choose whether to present the input as a single",
    "string (concatenating all the lines) or as the list."
```

exercises/flower-field/metadata.toml
```diff
@@ -0,0 +1,4 @@
+title = "Flower Field"
+blurb = "Help avoid stepping on the flowers in a flower garden."
```
IsaacG: There is a single reference in this commit to a cutting room floor in the intro file, a single reference to a garden in the intro, and a single reference stepping on flowers in the metadata. The rest of the docs don't mention cutting rooms, floors or stepping on things. There's mention of counting flowers in fields and boards, though.

Do we want to make things more consistent about the location and goal?

exercises/flower-field/canonical-data.json
kotp: The list reference can be the same as the string reference.  It is a single list, so perhaps:

```suggestion
    " string (concatenating all the lines) or as a list."
```

exercises/flower-field/introduction.md
```diff
@@ -0,0 +1,7 @@
+# Introduction
+
+[Flower Field][Wikipedia] is a compassionate reimagining of the popular game Minesweeper.
```
kotp: If we are doing away with triggers, then we should really not mention the honored occupation, as well.

exercises/flower-field/introduction.md
```diff
@@ -0,0 +1,7 @@
...
+The object of the game is to find all the flowers in the garden using numeric hints that indicate how many flowers are directly adjacent (horizontally, vertically, diagonally) to a square.
+"Flower Field" shipped in regional versions of Microsoft Windows in Italy, Germany, South Korea, Japan and Taiwan.
+
+[Wikipedia]: https://en.wikipedia.org/wiki/Minesweeper_(video_game)#History
```
kotp: I we are avoiding triggers, then we should warn here as well, very likely.

exercises/flower-field/introduction.md
SleeplessByte:
```suggestion
[Flower Field (trigger warning)][history] is a compassionate reimagining of a popular game.
```

and then

```
[history]: https://web.archive.org/web/20020409051321fw_/http://rcm.usr.dsi.unimi.it/rcmweb/fnm/
```

No need to link to wiki ;)

exercises/flower-field/metadata.toml
habere-et-dispertire: The web reference to the cutting room floor website is not in the intro file since I switched it to using wikipedia as the auto-linter was getting an http response code of forbidden 403.

exercises/flower-field/introduction.md
habere-et-dispertire: FWIW I don't think we need to shy away from detailing the history as it is a minor success story of manifest change in an orthogonal sphere. If keeping a warning, then I'd go with a "sensitivity disclosure" as opposed to a "trigger warning". This helps _combat_ the militaristic metaphor.

exercises/flower-field/metadata.toml
habere-et-dispertire: Switched per @kotp to using webarchive for the link which doesn't then 403.

IsaacG: The blurb should still not contain additional details which aren't in the intro/instructions ;)

habere-et-dispertire: Adjusted blurb

exercises/flower-field/canonical-data.json
```diff
@@ -0,0 +1,119 @@
...
+        "garden": [""]
+      },
+      "expected": [""]
+    },
+    {
+      "uuid": "d774d054-bbad-4867-88ae-069cbd1c4f92",
+      "description": "no flowers",
+      "property": "annotate",
+      "input": {
+        "garden": ["   ", "   ", "   "]
```
kotp: While the linting change can make some sense, having it as it was can make it easier to see that the space lengths are identical (apparently) from the alignment of the quotes.  If they are not spaces and some other unprinted white space, then that can be harder to immediately see, but when thinking of four legs, hooves and manes, we think "horse" rather than "zebra" (unless of course, you are more accustomed to seeing zebras).

So I understand the linting change, but miss having the nice easily seen "all the same" from having the individual lines.  Plus if one element is changed, it is also cleaner in diffs.

So there are benefits to having these the other way.

exercises/minesweeper/.deprecated
```diff
@@ -0,0 +1,3 @@
+**NOTE: This exercise has been deprecated**
+Your track should implement "flower-field" instead.
```
kotp, line 2: Should or may?

exercises/flower-field/canonical-data.json
habere-et-dispertire: Adjusted the ruby linting rule -- hope that's okay.
Multi-line arrays for this specific exercise are back to being on multiple lines now.

exercises/minesweeper/.deprecated
IsaacG: I believe should is correct. The problem spec should generally be followed. There's always exceptions but all a rule tracks should follow the spec.

habere-et-dispertire: Of the eleven other deprecated exercises, the language all reads "should" (in .deprecated) so maybe stick with that for consistency.

habere-et-dispertire:
[Greener exercise icon](https://github.com/exercism/website-icons/pull/198)

flower-field cleaned

kotp (approved):

exercises/flower-field/introduction.md
kotp: Also, hopefully it keeps the issues alive, as there are still a lot of mines out there that need to be swept and we get more people trained to care for this problem.  So not unnecessarily exposing those traumatized by it should not be enough to throw the problem under the rug where no one sees that it is something that still needs to be addressed.

habere-et-dispertire: > Also, hopefully it keeps the issues alive, as there are still a lot of mines out there that need to be swept and we get more people trained to care for this problem.

... and we get more people _and [rats](https://en.wikipedia.org/wiki/Demining#Rats
)_ trained to care for this problem.

[APOPO](https://en.wikipedia.org/wiki/APOPO) credits the rats with clearing more than 100,000 mines.

SleeplessByte (approved):

BNAndras (approved):

ErikSchierboom (approved):

PR #2589, opened 2025-09-09
two-bucket: test buckets with a big difference in size

nix-gh:
Those tests have been discussed here : https://forum.exercism.org/t/adding-tests-to-the-two-bucket-exercise/19163

There goal is to verify the solution is able to do several pour without refilling or emptying a bucket. This should catch some common pitfall when implementing it with Diophantine equations. 

I added some randomly generated  v4 uuid. I'm not sure if it's ok or if it should come from something else.


exercises/two-bucket/canonical-data.json
```diff
@@ -142,4 +180,4 @@
       }
     }
   ]
-}
+}
```
IsaacG: The linter is complaining about the missing newline here.

nix-gh: My formatter removed it automatically :/ Would it be possible to add a pre-commit hook for this? It would have avoided me to do several commits. I guess it should be discussed on the forum. If yes I'll open a discussion.

IsaacG: A forum thread is a good idea. Go for it!

glennj (approved):

BethanyG:
I have yet to test these for my track, but am fine with merging them.  I can always omit them if they create issues.

nix-gh:
> I have yet to test these for my track, but am fine with merging them. I can always omit them if they create issues.

Please take the time to do it, I was a bit surprised there is no automatic way to test the tests data. I guess it's not easy to put in place. So better double checking before merging in main. I paid attention but I'm human, a copy/past error can happen.

PR #2601, opened 2025-10-16
[ISBN Verifier] Error in tests

Keftcha:
Check if the `X` character is only the verification digit

PR according to [this](https://forum.exercism.org/t/isbn-verifier-error-in-tests-that-check-if-the-x-character-is-only-the-verification-digit/19639) topic.

Resolve exercism/python#4014

### DIFF
--- a/exercises/isbn-verifier/canonical-data.json
+++ b/exercises/isbn-verifier/canonical-data.json
@@ -69,6 +69,15 @@
       "expected": false
     },
     {
+      "uuid": "fdb14c99-4cf8-43c5-b06d-eb1638eff343",
+      "description": "X is not substituted by the value 10",
+      "property": "isValid",
+      "input": {
+        "isbn": "3-598-2X507-5"
+      },
+      "expected": false
+    },
+    {
       "uuid": "f6294e61-7e79-46b3-977b-f48789a4945b",
       "description": "valid isbn without separating dashes",
       "property": "isValid",


exercises/isbn-verifier/canonical-data.json
```diff
@@ -61,13 +61,31 @@
     },
     {
       "uuid": "28025280-2c39-4092-9719-f3234b89c627",
-      "description": "X is only valid as a check digit",
```
Cool-Katt, line 64: Tests are immutable, so you shouldn't touch the already existing test cases. 

Take a look at the [guide](https://github.com/exercism/problem-specifications?tab=readme-ov-file#changing-tests) for more info.

Cool-Katt (changes requested):

exercises/isbn-verifier/canonical-data.json
```diff
@@ -68,6 +68,37 @@
       },
       "expected": false
     },
+    {
+      "uuid": "0153f517-3008-4832-a168-8155c2ec0d94",
+      "reimplements": "28025280-2c39-4092-9719-f3234b89c627",
+      "description": "X is only valid as a check digit - X character is substituted for 10 and multiplied",
```
Cool-Katt: Do we need to keep this test at all of its superseded by the other one that also catches the edge cases? 

If yes, then I'd vote to change the names to something more distinct for the new test, so we can avoid the reimplementation. (I don't have a good name suggestion at the moment, sry)

exercises/isbn-verifier/canonical-data.json
```diff
@@ -68,6 +68,37 @@
...
+      ],
+      "property": "isValid",
+      "input": {
+        "isbn": "3-598-2X507-9"
+      },
+      "expected": false
+    },
+    {
+      "uuid": "fdb14c99-4cf8-43c5-b06d-eb1638eff343",
+      "description": "X is only valid as a check digit - X character is substituted for 10 and not multiplied",
```
IsaacG: I'm not convinced this test is needed.

exercises/isbn-verifier/canonical-data.json
```diff
@@ -68,6 +68,37 @@
...
+      "property": "isValid",
+      "input": {
+        "isbn": "3-598-2X507-5"
+      },
+      "expected": false
+    },
+    {
+      "uuid": "47fa54cb-507e-45d3-901a-096be447dfd9",
+      "description": "X is only valid as a check digit - X character is ignored altogether",
```
IsaacG: I'm not convinced this test is needed.

exercises/isbn-verifier/canonical-data.json
IsaacG: If we're not adding the other tests, this description doesn't need to change.

exercises/isbn-verifier/canonical-data.json
```diff
@@ -68,6 +68,37 @@
       },
       "expected": false
     },
+    {
+      "uuid": "0153f517-3008-4832-a168-8155c2ec0d94",
+      "reimplements": "28025280-2c39-4092-9719-f3234b89c627",
+      "description": "X is only valid as a check digit - X character is substituted for 10 and multiplied",
+      "comments": [
+        "Rename this test due to multiple tests on X that should be only valid as a check digit"
```
IsaacG: This comment looks like a commit comment. Looking at the test data and not the changes, this comment isn't useful so it should not be here.

exercises/isbn-verifier/canonical-data.json
oxe-i: > Do we need to keep this test at all of its superseded by the other one that also catches the edge cases?

I've commented in the forum topic, but one does not supersede the other. The current one checks when X is substituted for 10 and multiplied by the position (as the other digits are). The proposed one checks when X is substituted for 10 without being multiplied by the position, but fails to assert when it is multiplied by the position. So, they are complementary tests.

exercises/isbn-verifier/canonical-data.json
oxe-i: > I'm not convinced this test is needed.

This is the test proposed in the topic. The other one is the current test already present in the specs, only renamed.

exercises/isbn-verifier/canonical-data.json
Keftcha: Is something like *«Add a more precise description»* better ?

IsaacG: For a changelog? Sure. For the canonical data? There's no need for the canonical data file to contain change history. That's what the git log is for. The comments in the file should be useful to explain the data in the file. The comments in the file should _not_ be used to explain the history of the file.

Keftcha: I don't really know what to put in comments as I just rename the test. The [guide](https://github.com/exercism/problem-specifications?tab=readme-ov-file#changing-tests) specifically tell that I should add a comment.

> Re-implemented test cases must use the comments field to explain why a test case was re-implemented

This is what I try to do by explaining why I've renamed the test.  
Do you have an idea of what a correct comment could be for this case ?

Cool-Katt: In that case, I suggest not touching the original test and instead just come up with a better name to the new one. Seems like the simplest option to me, even if it's not the most clear one.

IsaacG (approved):
+cc @ErikSchierboom  for write approval

Keftcha:
Thanks you !

PR #2645, opened 2026-04-15
[simple-linked-list] Add missing initialValues field

senekor:
Templating languages that distinguish between undefined and empty list
will crash out on this missing field, or force the template author to
handle the field's absence explicitly.

### DIFF
--- a/exercises/simple-linked-list/canonical-data.json
+++ b/exercises/simple-linked-list/canonical-data.json
@@ -186,6 +186,22 @@
           "expected": {}
         },
         {
+          "uuid": "f3197f0a-1fea-45a5-939f-4a5ea60387ec",
+          "reimplements": "70d747a1-2e84-4ebc-bc3f-dcbee6a05f6b",
+          "description": "Can push to an empty list",
+          "property": "list",
+          "input": {
+            "initialValues": [],
+            "operations": [
+              {
+                "operation": "push",
+                "value": 1
+              }
+            ]
+          },
+          "expected": {}
+        },
+        {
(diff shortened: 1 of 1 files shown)


IsaacG (approved):

ErikSchierboom (approved):

kotp (approved):

IsaacG:
This actually causes the Go generator to swap `[]int(nil)` (a `null` slice) for `[]int{}` (an empty slice) which is less idiomatic :D Now I'm debating trying to add something to the test to force things back to `nil`.

senekor:
There are several other test cases with an empty `initialValues`, so the problem should've existed already?

Maybe off topic here, but I am curious why `nil` is more idiomatic :) I assume that `[]int{}` is optimized not to cause an allocation. When I was using Go a couple years ago, "empty" and "nil" objects being different things was an unpleasant source of bugs.

IsaacG:
> There are several other test cases with an empty `initialValues`, so the problem should've existed already?

Yup. Though it's not a problem so much as not following accepted style.

> Maybe off topic here, but I am curious why `nil` is more idiomatic :) I assume that `[]int{}` is optimized not to cause an allocation. When I was using Go a couple years ago, "empty" and "nil" objects being different things was an unpleasant source of bugs.

Go encourages the "zero value" to be useable as is. You should be able to use the `nil` slice interchangably with the empty slice for most things in Go.

See <https://google.github.io/styleguide/go/decisions#nil-slices> and <https://github.com/pthethanh/effective-go#declaring-empty-slices>

PR #2652, opened 2026-05-02
Add `save-the-cow`, deprecating `hangman`

BNAndras:
I'm proposing to deprecate Hangman due to its violent subject matter and replace it with a gentler Save the Cow exercise.
See http://forum.exercism.org/t/replace-hangman-with-non-violent-content/34156.

### DIFF
--- a/exercises/hangman/.deprecated
+++ b/exercises/hangman/.deprecated
@@ -1,0 +1,3 @@
+**NOTE: This exercise has been deprecated**
+Your track should implement "save-the-cow" instead.
+https://forum.exercism.org/t/replace-hangman-with-non-violent-content/34156/
--- a/exercises/save-the-cow/canonical-data.json
+++ b/exercises/save-the-cow/canonical-data.json
@@ -1,0 +1,146 @@
+{
+  "exercise": "save-the-cow",
+  "comments": [
+    "The expected values may be returned from the guess function, or returned from a separate game state call.",
+    "They may also be accessed from a different call for each value, based on what is most idiomatic for the language.",
+    "To explore objects, each guessed letter should be a separate call."
+  ],
+  "cases": [
+    {
+      "uuid": "71d340f9-fc29-4826-872e-ad7d0b83dd98",
+      "description": "Initially 9 failures are allowed and no letters are guessed",
+      "property": "guess",
+      "input": {
+        "word": "loot",
+        "guesses": []
+      },
+      "expected": {
+        "state": "Ongoing",
+        "maskedWord": "____",
+        "remainingFailures": 9
+      }
+    },
+    {
+      "uuid": "76759c24-8f1a-4fc8-9ffd-d6a1ff0cf03b",
+      "description": "After 10 failures the game is over",
+      "property": "guess",
+      "input": {
+        "word": "loot",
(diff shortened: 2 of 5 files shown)


jiegillet:
I'm good with the name change, but `hangman` was [explicitly asking](https://github.com/BNAndras/exercism-problem-specifications/blob/main/exercises/hangman/description.md) to solve the exercise using reactive functional programming. Are we dropping that aspect of the exercise on purpose?

To be fair, FRP is a complex thing, I tried implementing `hangman` in Elixir, but FRP is not really a thing in the language, so I couldn't find a clean way to do it. On the other hand, I realized that it was a perfect fit for Elm, since the Elm Architecture is a descendant of FRP, and it ended up being the first practice exercise for the `web-applications-sandbox` concept. It never would have happened without the exercise asking for it.

BNAndras:
I removed the mention thinking tracks could implement it without FRP and then not need an instructions append to clarify FRP was in fact not needed. However, I can put it back in since the exercise is intended to replace Hangman, and the forum discussion approving the PR to be made didn’t touch upon the content of the instructions.

IsaacG:
Would it make sense for exercises to come with optional appends that tracks could add that would include suggestions about intents/approaches/concepts? Multiple tracks implement hangman without FRP. But it would be nice not to drop it completely.

IsaacG (approved):

ErikSchierboom:
That would mean the backend code would need to take those extra files into account, and I don't see Jeremy having time to do that anytime soon. Maybe instead of having the mention of FRP in the instructions, we put it in the top level comments in the canonical data?

IsaacG:
> That would mean the backend code would need to take those extra files into account

I was suggesting additional md files in the problem spec that maintainers could opt to rename to `instructiond.append.md` if they want to use those instructions. I don't think this would require any backend changes.

ErikSchierboom:
Ah right. Yeah that is an option

keiravillekode (approved):

ErikSchierboom (approved):

oxe-i (approved):

kotp (approved):

BNAndras:
> Would it make sense for exercises to come with optional appends that tracks could add that would include suggestions about intents/approaches/concepts? Multiple tracks implement hangman without FRP. But it would be nice not to drop it completely.

Would it be appropriate for a comment to be inserted in the canonical data JSON to suggest a FRP approach if applicable? Then it's not dropped completely and the maintainers can choose if they want to use FRP or not.

IsaacG:
> Would it be appropriate for a comment to be inserted in the canonical data JSON to suggest a FRP approach if applicable? Then it's not dropped completely and the maintainers can choose if they want to use FRP or not.

I think it's always safe to dump comments into the canonical data file :) I think those are easier to miss and aren't markdown so it might not be quite as nice, but that works for me.

mk-mxp (approved):
I'd also add the FRP bit as a comment to the JSON.

IsaacG:
Looks good to me

jiegillet:
Realistically, the added comment won't make any difference, I'm guessing after reading it, hardly any track will go out of their way to implement the exercise with FRP.
However, it's clear that we collectively don't care about keeping the FRP angle very much, so I'll say that it's good enough.

oxe-i (approved):
The comment looks good to me

PR #2657, opened 2026-06-01
Update `anagram` blurb

BNAndras:
The blurb assumes someone knows what an anagram is and what correct means in this context.
https://forum.exercism.org/t/tweak-metadata-blurb-for-anagram/54255

### DIFF
--- a/exercises/anagram/metadata.toml
+++ b/exercises/anagram/metadata.toml
@@ -1,4 +1,4 @@
 title = "Anagram"
-blurb = "Find the words that use the letters of another word."
+blurb = "Find the words that use the same letters as another word."
 source = "Inspired by the Extreme Startup game"
 source_url = "https://github.com/rchatley/extreme_startup"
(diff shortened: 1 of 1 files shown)


exercises/anagram/metadata.toml
```diff
@@ -1,4 +1,4 @@
 title = "Anagram"
-blurb = "Given a word and a list of possible anagrams, select the correct sublist."
+blurb = "Find the words that use the letters of another word."
```
ErikSchierboom: Not sure if this is grammatically correct, but I would think having them match the exact same letters is important.

```suggestion
blurb = "Find the words that use the same letters as another word."
```

IsaacG (approved):

ErikSchierboom (approved):

mk-mxp (approved):

tasxatzial (commented):
As a non-native speaker, I find "same letters" a bit ambiguous. Does it ignore letter counts, or does it require an exact match of all letters and their frequencies?

But maybe that's okay  given it's a blurb.

IsaacG:
I think some ambiguity in the blurb is fine, so long as it's fairly simple and clear.

PR #2659, opened 2026-06-01
Update `isogram` blurb

BNAndras:
See https://forum.exercism.org/t/tweak-metadata-blurb-for-isogram/54253. The current blurb assumes the reader knows what an isogram is.

### DIFF
--- a/exercises/isogram/metadata.toml
+++ b/exercises/isogram/metadata.toml
@@ -1,4 +1,4 @@
 title = "Isogram"
-blurb = "Determine whether a phrase has any repeated letters."
+blurb = "Determine whether a phrase is an isogram, a word with no repeated letters."
 source = "Wikipedia"
 source_url = "https://en.wikipedia.org/wiki/Isogram"
--- a/exercises/isogram/metadata.toml
+++ b/exercises/isogram/metadata.toml
@@ -1,4 +1,4 @@
 title = "Isogram"
-blurb = "Determine if a word or phrase is an isogram."
+blurb = "Determine whether a phrase has any repeated letters."
 source = "Wikipedia"
 source_url = "https://en.wikipedia.org/wiki/Isogram"


exercises/isogram/metadata.toml
```diff
@@ -1,4 +1,4 @@
 title = "Isogram"
-blurb = "Determine if a word or phrase is an isogram."
+blurb = "Determine whether a phrase has any repeated letters."
```
ErikSchierboom: Isn't isogram more focused on _not_ having any repeated letters? It's a subtle difference.

BNAndras: It is. If you determine there aren’t repeated letters, it’s an isogram so a negative means an affirmative in practice. It’s a little clumsy so we can change it so an affirmative to the question here means it is an isogram.

IsaacG: Negative logic is harder to process. Affirmative logic is simpler. While the function's output is the inverse of this, the same logic applies to either phrasing (modulo a `not` operator). The affirmative wording is simpler and blurbs are meant to be short and simple.

This blurb _is_ the inverse of the exercise, but it's simpler and better conveys what this exercise is about. Is the improved simplicity worth the technical inaccuracy?

BNAndras: `"Determine whether a phrase is an isogram, a word with no repeated letters."` is a bit longer but it lines up the affirmative response with what we're expecting later in the instructions and test suite.

ErikSchierboom: I like that

BethanyG: I like that as well, and would happily approve if it was changed to that.

IsaacG (approved):

ErikSchierboom (approved):

oxe-i (approved):

BethanyG (approved):

PR #2661, opened 2026-06-16
Bob: Add test case for shouting statement with question mark not at end.

jack-garratt:
The previous test cases would consider any shouting statement with a question mark to be a shouting question. This is not the case as a question must **end** with a question mark. This test case covers this scenario.

### DIFF
--- a/exercises/bob/canonical-data.json
+++ b/exercises/bob/canonical-data.json
@@ -30,7 +30,7 @@
     },
     {
       "uuid": "3c954328-86fb-4c71-8961-e18d6a5e2517",
-      "description": "shouting statement containing question mark",
+      "description": "shouting a statement containing a question mark",
       "property": "response",
       "input": {
         "heyBob": "DO LIONS EAT PEOPLE? AHHHHH."
--- a/exercises/bob/canonical-data.json
+++ b/exercises/bob/canonical-data.json
@@ -29,6 +29,15 @@
       "expected": "Whoa, chill out!"
     },
     {
+      "uuid": "3c954328-86fb-4c71-8961-e18d6a5e2517",
+      "description": "shouting statement containing question mark",
+      "property": "response",
+      "input": {
+        "heyBob": "DO LIONS EAT PEOPLE? AHHHHH."
+      },
+      "expected": "Whoa, chill out!"
+    },
+    {
       "uuid": "8a2e771d-d6f1-4e3f-b6c6-b41495556e37",
       "description": "asking a question",
       "property": "response",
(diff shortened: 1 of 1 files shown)


IsaacG:
Relates to https://forum.exercism.org/t/bob-request-to-add-test-case-for-shouting-statement-with-question-mark-not-at-end/59368/

exercises/bob/canonical-data.json
```diff
@@ -236,6 +236,15 @@
         "heyBob": "\nDoes this cryogenic chamber make\n me look fat?"
       },
       "expected": "Sure."
+    },
+    {
+      "uuid": "3c954328-86fb-4c71-8961-e18d6a5e2517",
+      "description": "Shouting statement containing question mark",
```
IsaacG: The other descriptions are all lowercase. Could we follow suit here?


```suggestion
      "description": "shouting statement containing question mark",
```

glennj: "shouted"?

jack-garratt:
I have updated the description to fully lowercase. 
The failure seems to be unrelated to the changes made so not sure why that's failing?

oxe-i (approved):

IsaacG:
Ignore that CI failure. It's unrelated.

IsaacG (approved):

kotp (approved):

exercises/bob/canonical-data.json
```diff
@@ -236,6 +236,15 @@
         "heyBob": "\nDoes this cryogenic chamber make\n me look fat?"
       },
       "expected": "Sure."
+    },
+    {
+      "uuid": "3c954328-86fb-4c71-8961-e18d6a5e2517",
+      "description": "shouting statement containing question mark",
```
BNAndras:
```suggestion
      "description": "shouted statement containing question mark",
```

Glenn noted this earlier, but I agree. A statement itself can't be shouting but it can be shouted. The test descriptions are often student-facing so we should be unambiguous here.

jack-garratt: All other test cases refer to the type of message as shouting rather than shouted
For example:
"shouting",
"shouting gibberish"
"shouting with special characters"
"shouting numbers"
"shouting with no exclamation mark"
If this needs to be changes then I will do it but it does not seem necessary given all the other test cases follow the "shouting" format.

IsaacG: We could definitely change that, but that might be better in a follow up PR. For this PR, keeping it consistent is probably the best option. 

@BNAndras any objections to that plan?

Though it might be good to reorder this one test to be closer to the other shouting tests.

jack-garratt: We could move it closer to the other ones but they are already quite spread out so it would only be a rough area, possibly line 22 or 94?

IsaacG: Adjacent to any other shouting sounds good to me :) When we change "shouting" to "shouted" we could consider re-sorting them all.

IsaacG (approved):

exercises/bob/canonical-data.json
glennj: The other cases use "shouting" as a verb. If we want to stick with shouting, we just need to adjust the description slightly: `shouting a statement containing a question mark`

jack-garratt: Thats a fair point, I can update that now.

BNAndras (approved):

PR #2662, opened 2026-06-26
[bob] Sort test cases

IsaacG:
Sort test cases based on the response and their order in the instructions. First test the simple versions of each response. Then test the more complex versions, ordered by response.

This is a follow up to #2661

### DIFF
--- a/exercises/bob/canonical-data.json
+++ b/exercises/bob/canonical-data.json
@@ -2,13 +2,13 @@
   "exercise": "bob",
   "cases": [
     {
-      "uuid": "e162fead-606f-437a-a166-d051915cea8e",
-      "description": "stating something",
+      "uuid": "8a2e771d-d6f1-4e3f-b6c6-b41495556e37",
+      "description": "asking a question",
       "property": "response",
       "input": {
-        "heyBob": "Tom-ay-to, tom-aaaah-to."
+        "heyBob": "Does this cryogenic chamber make me look fat?"
       },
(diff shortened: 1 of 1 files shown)


kotp (approved):

keiravillekode (approved):

BNAndras (approved):

exercises/bob/canonical-data.json
```diff
@@ -65,186 +65,186 @@
...
-        "heyBob": "Okay if like my  spacebar  quite a bit?   "
+        "heyBob": "Ending with ? means a question."
       },
-      "expected": "Sure."
+      "expected": "Whatever."
     },
     {
-      "uuid": "72bd5ad3-9b2f-4931-a988-dce1f5771de2",
-      "description": "other whitespace",
+      "uuid": "66953780-165b-4e7e-8ce3-4bcb80b6385a",
```
keiravillekode, line 223: This should have appeared earlier, line 103, before the test case that re-implements it.

BNAndras: I don’t think there’s a hard rule about that. For example, list-ops has tests that appear several slots after the ones they reimplement and svg-parsing has a test that implements the following test.

A test generator should be following the uuids not the test order to determine which test is reimplemented where.

IsaacG: What @BNAndras said. Nothing here should rely on the order of the cases.

PR #2666, opened 2026-07-03
Check reimplemented case prior to new implementation

keiravillekode:
See discussion: https://forum.exercism.org/t/enforcing-an-order-in-canonical-data-between-deprecated-test-case-and-its-reimplementation-s/63361

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,6 +90,31 @@ jobs:
 
           exit "$fail"
 
+      - name: Verify that reimplemented test cases appear before the case that reimplements them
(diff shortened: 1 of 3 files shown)


BNAndras:
This CI passes if the reimplementation follows the original test even if there are other tests in between. For example, https://github.com/exercism/problem-specifications/blob/77d50e4a40e93b90bf45fa610a2329087e4ca3d1/exercises/list-ops/canonical-data.json#L206 occurs three tests after the one it’s replacing. The reimplementation test should probably occur immediately after for readability’s sake.

PR #2679, opened 2026-08-07
Update `nucleotide-count` source field

BNAndras:
See https://forum.exercism.org/t/remove-stray-underscore-from-nucleotide-count-metadata/71051.

### DIFF
--- a/exercises/nucleotide-count/metadata.toml
+++ b/exercises/nucleotide-count/metadata.toml
@@ -1,4 +1,4 @@
 title = "Nucleotide Count"
 blurb = "Given a DNA string, compute how many times each nucleotide occurs in the string."
-source = "The Calculating DNA Nucleotides_problem at Rosalind"
+source = "The Calculating DNA Nucleotides problem at Rosalind"
 source_url = "https://rosalind.info/problems/dna/"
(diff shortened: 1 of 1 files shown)


exercises/nucleotide-count/metadata.toml
```diff
@@ -1,4 +1,4 @@
 title = "Nucleotide Count"
 blurb = "Given a DNA string, compute how many times each nucleotide occurs in the string."
-source = "The Calculating DNA Nucleotides_problem at Rosalind"
+source = "The Calculating DNA Nucleotides problem at Rosalind"
```
blakelewis: The original problem is called Counting DNA Nucleotides, which seems more accurate.

IsaacG (approved):

ErikSchierboom (approved):

exercises/nucleotide-count/metadata.toml
BNAndras: Good catch. I wonder if Rosalind changed it or someone here put the wrong name by accident. Either way, we should probably change the name here so it matches Rosalind. Otherwise, that might be confusing for people clicking through.

blakelewis (approved):

keiravillekode (approved):

